PR #16, opened 2018-01-20
Add airflowlite example

andscoop:
@schnie Was discussed on side, but wanted to share full picture of the code changes. If still out of scope just decline and I will resubmit with examples git clone removed.

### DIFF
--- a/examples/airflow/start
+++ b/examples/airflow/start
@@ -1,12 +1,29 @@
 #!/usr/bin/env bash
 
 WORKERS_COUNT=2
+COMPOSEFILE='./docker-compose.yml'
+SCALEWORKER='--scale worker='
+AIRFLOWLITE=false
+# Check flags
+while getopts "l" OPTION
+do
+	case $OPTION in
+		l)
+			echo "Starting airflow-lite..."
+			AIRFLOWLITE=true
+			COMPOSEFILE='./docker-compose-lite.yml'
+			SCALEWORKER=''
+			WORKERS_COUNT=''
+			;;
+	esac
+done
 
 if [ ! -z $1 ] && [ -d $1 ]; then
 	BASE_IMAGE_NAME="astronomerinc/ap-airflow"
 	REQUIREMENTS="$1/requirements.txt"
 	PACKAGES="$1/packages.txt"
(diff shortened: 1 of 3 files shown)


examples/airflow/.env
```diff
@@ -1,3 +1,4 @@
 AIRFLOW_IMAGE=astronomerinc/ap-airflow
 AIRFLOW_IMAGE_TAG=latest
 AIRFLOW_HOME=/tmp/astronomer
+AIRFLOW__CORE__LOAD_EXAMPLES=False
```
schnie: @andscoop what happens if we set this to `True`?

examples/airflow/start
```diff
@@ -7,7 +7,7 @@ if [ ! -z $1 ] && [ -d $1 ]; then
 	REQUIREMENTS="$1/requirements.txt"
 	PACKAGES="$1/packages.txt"
 	DOCKERFILE="$1/Dockerfile"
-
+
```
schnie: @andscoop can you kill the extra whitespace?

andscoop:
@schnie Could I get a clarification on what is in-scope for this PR?

Goal was to implement airflow-lite as an example.

Two outstanding questions are

Do we need to have airflow and and airflow-lite consolidated into a single example? Best way I see this being done is to have `./start` call multiple `docker-compose-up` cmds, but I will need to see how well that solution works.

Is pulling in examples via the `git clone` cmd out of scope or in scope?

schnie:
@andscoop was chatting with @vparekh94 about this on Friday. He's down to hack on https://github.com/astronomerio/open-example-dags and help make it part of the documentation, so you guys should collaborate.

Idea was to keep cloning out of this repo and just include docs on how to get started by cloning the examples and point to it with start cmd.

What do you think about adding a flag to start script to determine which compose file to start or something. Nothing too fancy.

schnie:
Nice!

schnie:
@andscoop the `kill` and `stop` scripts are actually still using the normal `docker-compose.yml` file, even if you `./start -l` since it just looks for the default filename. This causes some extra logging about not being able to delete things that don't exist. Want to take a crack at patching that up?

PR #20, opened 2018-02-01
Use portable shebangs

tedmiston:
Similar to https://github.com/astronomerio/astronomer-ee/pull/12

### DIFF
--- a/docker/platform/airflow/include/clean-airflow-logs
+++ b/docker/platform/airflow/include/clean-airflow-logs
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 DIRECTORY=${AIRFLOW_HOME:-/usr/local/airflow}
 RETENTION=${ASTRONOMER__AIRFLOW__WORKER_LOG_RETENTION_DAYS:-15}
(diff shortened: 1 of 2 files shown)


tedmiston:
@schnie Does this one look good to merge to you?

schnie:
@tedmiston `bash` is cool for anything that runs outside of the container but for everything that runs inside an alpine container, I'd rather they stay `sh`. Typically we wouldn't want to even install bash in our containers - airflow is the special case here. Keeping them `sh` will actually make them more portable within our ecosystem.

tedmiston:
Makes sense.  Do you think it'd be preferred to use sh over bash in the docker entrypoint too, or is there something about it that depends on bash instead?

https://github.com/astronomerio/astronomer/blob/9527b7a219765e72b63a9c36d3c9aa843e99c195/docker/platform/airflow/include/entrypoint#L1

schnie:
Something in the entrypoint or wait-for-command scripts needs bash but could probably be patched to work with sh. I just left it so users could use the bash operator from airflow.

tedmiston:
Makes sense - closing this PR.

PR #39, opened 2018-03-17
Update Airflow examples docs

tedmiston:
Add instructions to start Airflow examples and make comparison clear between Lite vs Enterprise examples.

Here's how it looks now:

screen shot 2018-03-16 at 7 42 44 pm

Side note: I think we should consider a clearer naming convention for the examples.  (Lite sounds like it's not full featured Airflow; Enterprise makes sense as an approximation of EE but maybe another name could convey that better.)

Some suggestions for Airflow Lite from discussion with Andy & Ben are:

- Airflow Core
- Airflow Base
- Airflow (Simple)

For Airflow Enterprise:

- Airflow Suite
- Airflow Plus Plus
- Airflow (Fancy)

@schnie @andscoop @ryw Thoughts?

And also - does it make sense to add the CLI to the comparison table for EE?


ryw:
I'll let @schnie decide

* i like "Airflow Core" and "Airflow w/ Monitoring"
* cli isn't part of open, so shouldn't be part of this chart, if we're simply comparing the examples, and i don't think we should add marketing for astronomer enterprise in the open docs

andscoop:
+1 for Airflow Core.

My preference is Airflow Suite, Airflow Bundle or Airflow Stack for "Enterprise" bundle.

andscoop:
We should also consider the branding Astronomer Core, Astronomer [Bundle, Suite, Stack].

schnie:
Core seems fine for the slim version. I kind of like keeping enterprise though.

tedmiston:
Cool - I just pushed an update renaming all references of Airflow Lite to Airflow Core; no change to Airflow Enterprise name for now.

Are we good to merge this?

tedmiston:
Andy talked with a customer today who was confused that the Airflow Enterprise example is not Astronomer Enterprise (https://github.com/astronomerio/astronomer/issues/44).  It's not the same reason we made this issue/PR but it is the same topic we've been discussing here.

@schnie Any more thoughts on what we should do here?  Andy has another suggestion in the linked issue as well.

andscoop:
LGTM - this PR adds some clarity that didn't exist before. We can revisit naming issue if it continues to be a problem

PR #47, opened 2018-04-02
Hotfix/update maintainer label

andscoop:
Updates repo to remove deprecated MAINTAINER attr in favor of LABEL maintainer="" attr 

https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

### DIFF
--- a/docker/platform/airflow/Dockerfile
+++ b/docker/platform/airflow/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 FROM astronomerinc/ap-base
-MAINTAINER Astronomer <humans@astronomer.io>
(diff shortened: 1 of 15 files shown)


schnie:
@andscoop can you split this out into two PRs? One for the setuptools fix and one for the `MAINTAINER` stuff? What do you think about keeping the maintainer labels on the same line as they were orignally? Kind of separate from the rest of the labels.

schnie:
Actually, now I see the separate one for the setuptools. Can you just update this one then?

andscoop:
@schnie I had previously branched off of the other feature branch so this one contained both commits. I have removed the 'setup tools fix' from this branch and made requested changes on the maintainer stuff and forced pushed to this branch. 

Should be good to go

PR #49, opened 2018-04-03
Pin setuptools for deterministic build

tedmiston:
Recently we merged upgrading to the latest `setuptools` to support another dependency.  This builds on that to PR #46 to pin `setuptools` like we do other Python packages so we can ensure that our builds are deterministic when future releases of `setuptools` are published.

### DIFF
--- a/docker/platform/airflow/Dockerfile
+++ b/docker/platform/airflow/Dockerfile
@@ -49,7 +49,7 @@ RUN apk add --no-cache --virtual .build-deps \
 		postgresql \
 		python3 \
 		tini \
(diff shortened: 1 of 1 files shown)


schnie:
@tedmiston were you able to successfully run a `docker build` with this?

schnie:
Worked for me.

tedmiston:
Thanks.

I tried to run the build like 5 times from the office yesterday afternoon/evening with the storm, but it kept hanging on downloading random dependencies.  I think it must have been some weird network issue though.

PR #50, opened 2018-04-04
Upgrade + pin pip to 9.0.3

tedmiston:
We were using a slightly outdated pip (9.0.1).  This upgrades to the latest and pins it similarly to #49.


tedmiston:
@schnie rebased off master - should be good now

PR #51, opened 2018-04-09
Reduce cache from 30 min to 5 min

tedmiston:


Makefile
```diff
@@ -69,4 +69,7 @@ build-docs:
 
 .PHONY: push-docs
 push-docs: build-docs
-	gsutil -m rsync -a public-read -d -r ${DOCS_DEST} ${DOCS_BUCKET}
+	gsutil \
+		-h "Content-Type:text/html" \
```
schnie: @tedmiston not every file that gets synced is html. There's also css and js, etc. Maybe just leave this line off?

tedmiston: fixed

tedmiston:
@schnie should be good now

PR #56, opened 2018-04-12
Pin Grafana to 5.0.4

tedmiston:
Note: Untested

Pinning this to be consistent for deterministic builds like with the other images.

Note: This tag is the latest stable but it is from 15 days ago while master was updated today.

### DIFF
--- a/docker/vendor/grafana/Dockerfile
+++ b/docker/vendor/grafana/Dockerfile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


schnie:
@tedmiston the only reason this was on master was because at the time of building it was the only way to get the feature where you can preload dashboards. I believe that was slated to be officially released in v5, so I'm guessing our dashboards are still automatically being loaded?

tedmiston:
No clue but we can test it.  If nothing else, we've got that written down for future reference.

PR #74, opened 2018-04-28
Change link to HTTPS to fix live

jlaneve:
https://open.astronomer.io currently doesn't load jQuery because the jQuery link isn't `https://...`


tedmiston:
Thanks for this - I tested with http earlier to avoid the 5-min cache on https.  Deployed.

I have made and deployed the same fix on EE as well.

PR #122, opened 2018-06-21
Adding start/stop/pause scripts back. Updated docs

cwurtz:
Readded the start, stop, and pause scripts as they are the best way to run the examples when custom python and/or alpine packages are needed.


schnie:
My only concern with this is that now we have functionality in two/three places. Things need to be duplicated in two scripts here, as well as the CLI. Seems like we'll just get more and more out of sync as we push forward.

schnie:
@cwurtz should we close this PR after our chat yesterday?

PR #124, opened 2018-06-28
Disable pickling by default

tedmiston:
see https://github.com/astronomerio/astronomer/issues/123

### DIFF
--- a/docker/platform/airflow/include/airflow.cfg
+++ b/docker/platform/airflow/include/airflow.cfg
@@ -78,7 +78,7 @@ plugins_folder = /usr/local/airflow/plugins
 fernet_key = d6Vefz3G9U_ynXB3cr7y_Ak35tAHkEGAVxuz_B-jzWw=
 
 # Whether to disable pickling dags
-donot_pickle = False
+donot_pickle = True
 
 # How long before timing out a python file import while filling the DagBag
 dagbag_import_timeout = 30


tedmiston:
Are we cool with having an opinionated default like this?  If not, I'm thinking I could instead copy this over as an example in our docs.

What do you guys think? @schnie @andscoop @cwurtz

schnie:
@tedmiston isn't pickling only required for the mesos executor? Does anything else use this setting? I'm assuming we aren't currently utilizing pickles, but want to make sure.

tedmiston:
The error @vparekh94 got without this setting was in our current Astro Open setup trying to use VirtualenvOperator (haven't tested this myself).

My understanding is that any executor besides LocalExecutor and SequentialExecutor at least pulls it and potentially uses it.

https://github.com/apache/incubator-airflow/blob/b0061f1369636c86df87829b93d5ece582591e2a/airflow/jobs.py#L2501-L2508

https://github.com/apache/incubator-airflow/blob/fa6c35743a5129057d870752554b0827dd7d462f/airflow/models.py#L4068

vparekh94:
@tedmiston so doing some digging, it looks like the PythonVirtualEnvOperator does pickle the DAG if arguments are passed to it:

https://github.com/apache/incubator-airflow/blob/master/airflow/operators/python_operator.py#L184

https://github.com/apache/incubator-airflow/blob/master/airflow/operators/python_operator.py#L311

I was able to get an example to run without overriding the pickling setting after I took out passing context and `kwargs` to the operator.

tedmiston:
Cool - sounds like it's relevant when the op has context / kwargs / args passed to it.

I'm ready to merge this if someone can give a 👍.

schnie:
So it looks like the issue @vparekh94 hit, may be related to pickling, but I'm not sure `donot_pickle` is actually used by the `PythonVirtualEnvOperator`. Seems like that's a different issue. @tedmiston have you tested this with the CeleryExecutor to be sure?

tedmiston:
I haven't tested it with that operator personally yet.  Since this is more maintenance, I think this is lower priority than the sprint —  cool if we revisit this post-sprint?

schnie:
Yep

ryw:
lets just merge this, shouldn't hurt anything right?

schnie:
Let's merge it after we test it. @tedmiston can you test this quick? We should start testing PRs _before_ sending them to prevent them from lingering around like this.

PR #125, opened 2018-07-09
Add ability to build and push release candidate builds

schnie:
This adds the ability for us to maintain a list of components to be built from their respective github `master` branches, rather than the tagged version in the actual `Dockerfile`.


schnie:
Codeclimate is showing issues on the `${BUILD_ARGS}` in the `docker build` commands. Adding the quotations breaks it at runtime.

andscoop:
LGTM

PR #132, opened 2018-07-24
Upgrades to dashboard for new kube environment

schnie:
Minor changes to regex matching and label formatting. Prometheus config in kubernetes passes release name in as part of `{{component_name}}`. Tested in docker-compose and in helm charts on kubernetes.


tedmiston:
Everything looks good here too.

I want to do more testing with this one too once I have my GKE environment spun where I can test with auth.

PR #134, opened 2018-07-25
Rearrange COPY commands

schnie:
This creates several layers for caching and ensures that we don't `apk add` and `pip install` every time a dag/plugin file changes.

- If `packages.txt` changes, we essentially rebuild from our airflow base.
- If `requirements.txt` changes, we only rebuild starting with `pip install`, and skip `apk add`.
- If dag/plugin/anything else changes, we skip both `apk add` and `pip install`.

Since we aren't creating new layers each time, `astro airflow deploy` (`docker push`) is also a lot faster since it determines that the previously pushed layers already exist in the remote registry. 

This should speed up development and deployment.

Initial pushes will still take a little extra time since it has to push all the layers for the first time. Wondering if we could "warm the cache" in the registry with an init container or something that pre-loads the registry with the base image and all it's layers.

### DIFF
--- a/docker/platform/airflow/onbuild/Dockerfile
+++ b/docker/platform/airflow/onbuild/Dockerfile
@@ -20,12 +20,9 @@ ARG BUILD_NUMBER=-1
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
 
-# Update repositories
-ONBUILD RUN apk update
-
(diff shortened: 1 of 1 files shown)


tedmiston:
Overall I'm in support of this but one thing to keep in mind is if a user's requirements.txt looks like:

```
requests
```

(Without version tags) and a new version of the package comes out, I think the user would still get the previous version unless they do an uncached build.

PR #149, opened 2018-08-15
Airflow versioning

schnie:
This PR breaks the build scripts for the platform and airflow out and tags images with both the airflow version and platform version.

It also adds an initial image for 1.10.0 that builds from master, until we have an official release on the apache side. It is currently omitted from official builds via `make`, as it will change as we start moving towards using it in production for the 0.3.4 release.


cwurtz:
Do you think it would make sense to continue building them off of `ap-base` rather than directly off `alpine`?  Or at least a common airflow image that has all the common deps?

schnie:
Yea, I went in with that same thought but kind of ran into some awkward situations with sharing `ap-base`. Then started with a separate `ap-airflow-base`, but that kinda introduces some potentially awkward scenarios down the line too as we have more versions to manage. 

I figured the most maintainable/safest way would be to just accept some duplication through the different versions. It also seems to be the way a lot of big library images are maintained on dockerhub. For example, mariadb - https://github.com/docker-library/mariadb. They duplicate the entrypoint script as well as the entire `Dockerfile` in each version. They do have a script and template to automate some of it, but that might be harder with Airflow and it's rapid dependency and config changes.

We might be able to introduce some automation to our process as the project matures.

schnie:
Forgot about changes to the `onbuild` images, will have to fix those and some bash linting issues.

@cwurtz @tedmiston @andscoop what do you guys think about removing `airflow.cfg` from our images? If left out, airflow generates the default one on startup if it doesn't exist. It'd be easier to maintain these images (less files), and produces a more default/vanilla image. Anything we want to enforce at runtime will get configured with environment variables anyway.

tedmiston:
I support removing it and letting Airflow generate the default.

PR #151, opened 2018-08-18
Add cli installer image to platform

schnie:
This image is the same one that powers install.astronomer.io. In the charts, we override the script to template in a matching version number to download by default.


tedmiston:
lgtm

PR #163, opened 2018-11-29
Ee logging doc

rhwang10:
prelim doc, mostly fluentD config info


schnie:
@rhwang10 looks great. Thanks for taking the time do this!

PR #171, opened 2019-01-02
ee configuration docs [skip ci]

vparekh94:


vparekh94:
I made a bunch of changes and pushed up another version, so we can remove this pull request.

PR #172, opened 2019-01-06
Added notes on setup

vparekh94:
Added information on where to set up SMTP and minimum cluster size.


docs/ee-installation-config.md
```diff
@@ -26,9 +28,22 @@ global:
...
+
+You'll need to set up SMTP to use email invites with Astronomer.
+
+In your Helm config, nested under `astronomer.smtp.uri` add something like:
+
+```yaml
+astronomer:
+  smtp:
+    uri: "smtp://user:pass@email-smtp.us-east-1.amazonaws.com/?requireTLS=true"
+    uri: "smtps://user:pass@smtp.mailgun.org/?pool=true"
```
schnie: @vparekh94 we really only need one here. I just combined two examples that I had laying around. Maybe just use `smtps://user:pass@host/options` or something generic here. We could mention that we are using `nodemailer` under the hood and that any options supported by it can be passed.

https://nodemailer.com/smtp/

vparekh94: Got it - I'm making a few more changes today and I'll rope this into one of them

PR #262, opened 2019-02-20
Init operator-sdk Dockerfile

andriisoldatenko:
**Why?**
We need docker image of opeator-sdk to run inside CI.

Related ->
https://github.com/operator-framework/operator-sdk/issues/1115

### DIFF
--- a/docker/platform/operator-sdk/Dockerfile
+++ b/docker/platform/operator-sdk/Dockerfile
@@ -1,0 +1,9 @@
+FROM alpine:3.6
+
+ARG operator_sdk_version=0.5.0
+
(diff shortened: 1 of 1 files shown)


schnie:
@andriisoldatenko this repo is really only used to build and tag images that end up on https://hub.docker.com/u/astronomerinc and are part of the platform install.

In addition to CI, will this be the same image we will reference from the operator helm chart?

schnie:
@andriisoldatenko should we close this for now? The plan was to reopen the PR with the actual docker image for the full operator right? If we need this CI only image, we can just put it in our `astronomerio` repo, the same place our existing `ap-build` Dockerfile lives.

andriisoldatenko:
yes. let's close it for now.

PR #297, opened 2019-04-30
bump Airflow version to 1.10.3

andriisoldatenko:


docker/airflow/1.10.3/onbuild/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM astronomerinc/ap-airflow:0.9.0-alpha.2-1.10.2
+FROM astronomerinc/ap-airflow:0.9.0-alpha.1-1.10.3
```
andriisoldatenko: what's the next release of platform? `0.10.0`?

docker/airflow/1.10.3/Dockerfile
```diff
@@ -21,10 +21,9 @@ LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.module="airflow"
 LABEL io.astronomer.docker.component="airflow"
-LABEL io.astronomer.docker.airflow.version="1.10.2"
+LABEL io.astronomer.docker.airflow.version="1.10.3"
 
 ARG ORG="astronomer"
-# ARG VERSION="1.10.2-1"
 ARG VERSION="feature/kube-labels"
```
schnie: @andriisoldatenko also need to change it here to point at v1-10-3 (the new branch).

docker/airflow/1.10.3/onbuild/Dockerfile
schnie: @andriisoldatenko you can revert this change. We will automatically update this tag as part of the release process (https://github.com/astronomer/astronomer/blob/master/Makefile#L144-L148).

docker/airflow/1.10.3/Dockerfile
andriisoldatenko: @schnie fixed.

Issue #303, opened 2019-05-09
Fix Grafana dashboards for KubeExecutor


schnie:
The current "Airflow Deployment Overview" dashboard shows a list of pods/containers at the bottom. We should find a cleaner way to show this chart, maybe excluding the task pods, or show them in some other chart or some other way.

This also results in the "Deployment Status" going to unhealthy as pods transition states rapidly. We need to skip task pods as part of that calculation.

PR #305, opened 2019-05-15
Add filter only pods without dag_id label to exclude from health

andriisoldatenko:
fix https://github.com/astronomer/astronomer/issues/303


docker/vendor/grafana/include/airflow-overview.json
```diff
@@ -2055,6 +2055,149 @@
       "title": "Container Status",
```
andriisoldatenko: @schnie 
do you want me to change title?

PR #373, opened 2019-07-18
Update FAB Security Manager to v1.1.2

ashb:


ashb:
I updated both versions numbers this time :)

Issue #416, opened 2019-08-13
Relabel Airflow StatsD metrics in Prometheus 


kaxil:
It might useful to re-label the statsd/airflow metrics back in to metrics with labels - like `dag.<dag_id>.<task_id>.duration` which appears as `airflow_dag_mydagid_taskid_duration` in Prometheus would be nice if it was` airflow_dag_duration{dag_id=X, task_id=Y}`

PR #463, opened 2019-09-03
[#416] Add mappings for new Statsd metrics

kaxil:
**Issue**: https://github.com/astronomer/astronomer/issues/416

This PR adds mappings to new metrics so that we no longer have metrics as follows:

image

Instead stuff like dag_id, task_id would become labels.


kaxil:
I haven't tested this yet but the mappings should be straight-forward

schnie:
Looks good! We can adjust if we need to.

PR #464, opened 2019-09-03
sync tag names for build and scan

andriisoldatenko:
Fix this:
>Error response from daemon: manifest for astronomerinc/ap-commander:0.0.0 not found

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ push-airflow-ref:
 scan-platform: check-env
 	for component in ${PLATFORM_COMPONENTS} ; do \
 		PUSH_IMAGE=${REPOSITORY}/ap-$${component} \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -112,7 +112,7 @@ scan-platform: check-env
 	done;
 	for component in ${VENDOR_COMPONENTS} ; do \
 		PUSH_IMAGE=${REPOSITORY}/ap-$${component} \
-		PUSH_TAGS="${ASTRONOMER_VERSION}" \
+		PUSH_TAGS="${ASTRONOMER_REF}" \
```
schnie: @andriisoldatenko should we update the `${ASTRONOMER_VERSION}` in the platform components to be `${ASTRONOMER_REF}` too?

andriisoldatenko: @schnie done.

PR #465, opened 2019-09-04
Use pre-packaged Alpine packages from edge repo for faster builds

ashb:
Alpine has added lots of new python packages in their edge repos that we
can use to not have to build the big ones (numpy, panda, grpcio) form
source. This change took the image build time down to 12 minutes for me
on my laptop.

### DIFF
--- a/docker/airflow/1.10.4/Dockerfile
+++ b/docker/airflow/1.10.4/Dockerfile
@@ -42,6 +42,9 @@ ENV ASTRONOMER_GROUP=${ASTRONOMER_GROUP}
 RUN addgroup -S ${ASTRONOMER_GROUP} \
 	&& adduser -S -G ${ASTRONOMER_GROUP} ${ASTRONOMER_USER}
 
+RUN echo @edge-community http://dl-cdn.alpinelinux.org/alpine/edge/community >> /etc/apk/repositories && \
+    echo @edge-testing http://dl-cdn.alpinelinux.org/alpine/edge/testing >> /etc/apk/repositories && \
+    echo @edge http://dl-cdn.alpinelinux.org/alpine/edge/main >> /etc/apk/repositories
 # Install packages
 RUN apk update \
 	&& apk add --no-cache --virtual .build-deps \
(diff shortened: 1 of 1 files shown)


docker/airflow/1.10.4/Dockerfile
```diff
@@ -55,39 +58,43 @@ RUN apk update \
...
-	# https://pynacl.readthedocs.io/en/stable/install/
-	&& SODIUM_INSTALL=system pip3 install --no-cache-dir pynacl==1.3.0 \
 	&& pip3 install --no-cache-dir "${AIRFLOW_MODULE}" \
 	&& pip3 install --no-cache-dir "https://github.com/astronomer/astronomer-fab-securitymanager/releases/download/v1.2.1/astronomer_fab_security_manager-1.2.1-py3-none-any.whl" \
 	&& cd /usr/lib/python3.7/site-packages/airflow/www_rbac \
 	&& npm install \
 	&& npm run build \
 	&& rm -rf node_modules \
-	&& apk del .build-deps \
+	&& apk del .build-deps py3-numpy-dev \
```
schnie: @ashb should we add `py3-numpy-dev` to the upper half of this command as part of `.build-deps`?

ashb: Ideally yes, but you can't put "masked" deps (the `@edge` edge is a mask) in to a virtual package

ashb:
A bit smaller, but not much:

```
astro-alpine-airflow                                                 latest                        bfad4761dd0c        About an hour ago   590MB
astronomerinc/ap-airflow                                             0.10.0-1.10.4-onbuild         1390012010a0        12 days ago         721MB
```

schnie:
LGTM, pushed a small cosmetic change. Testing locally.

schnie:
@ashb any idea where we lost some weight with this image? All seems good to me, just hoping we didn't loose something important :thinking:

ashb:
Yeah, I think we should test it out a bit more before putting this in front of customers.

PR #466, opened 2019-09-05
[#416] Fix Statsd mapping yaml

kaxil:
should solve the issue in https://github.com/astronomer/biz-cloud/issues/95#issuecomment-528381191


ryw:
Is this a situation where the config needs to be different depending on Airflow version?

schnie:
Shouldn't be. This is just a filter so old deployments wont be emitting them and that's fine. We just have to account for it downstream in any consumers though.

PR #493, opened 2019-09-18
Add Grafana Dashboard for Velero

kaxil:
Add Grafana Dashboard for Velero

### DIFF
--- a/docker/vendor/grafana/Dockerfile
+++ b/docker/vendor/grafana/Dockerfile
@@ -32,3 +32,4 @@ COPY include/elasticsearch.json /var/lib/grafana/dashboards/elasticsearch.json
 COPY include/prometheus.json /var/lib/grafana/dashboards/prometheus.json
 COPY include/docker-registry.json /var/lib/grafana/dashboards/docker-registry.json
 COPY include/platform-overview.json /var/lib/grafana/dashboards/platform-overview.json
+COPY include/velero.json /var/lib/grafana/dashboards/velero.json
--- a/docker/vendor/grafana/include/velero.json
+++ b/docker/vendor/grafana/include/velero.json
@@ -1,0 +1,1040 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [
+    {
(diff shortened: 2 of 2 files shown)


schnie:
Awesome @kaxil! This is super useful. My only question is if it should exist in this repo or not. These dashboards are shippped to every install, including installations where we don't even use our terraform (openshift, home-grown, etc). In some of these cases, Velero won't be present and I'm guessing that this dashboard will essentially be empty.

I'm worried about users logging into Grafana and interpreting that as if something is wrong. We recently had this happen with some charts that were designed for Celery workers, but are empty when using KubernetesExecutor.

Just thinking out loud here.. maybe we could include this dashboard in our helm charts and mount it into the same directory, but wrap that mount block with something like `{{ if .Values.global.velero.enabled }}` or something. It could be disabled by default and our cloud terraform could set the `enabled` flag here to include.

I've wanted to actually move all the dashboards to our helm charts to make modifications easier, but we actually use these charts in our `examples` directory in `astronomer/astronomer`, for usage with our "demo" `docker-compose` version. This is probably out of date and not many people even know about it. Not sure if its worth us keeping around or not.

cc: @ryw @vikramkoka @vparekh94 if you guys have any thoughts on the matter.

vikramkoka:
> Awesome @kaxil! This is super useful. My only question is if it should exist in this repo or not. These dashboards are shippped to every install, including installations where we don't even use our terraform (openshift, home-grown, etc). In some of these cases, Velero won't be present and I'm guessing that this dashboard will essentially be empty.
> 
> I'm worried about users logging into Grafana and interpreting that as if something is wrong. We recently had this happen with some charts that were designed for Celery workers, but are empty when using KubernetesExecutor.
> 
> Just thinking out loud here.. maybe we could include this dashboard in our helm charts and mount it into the same directory, but wrap that mount block with something like `{{ if .Values.global.velero.enabled }}` or something. It could be disabled by default and our cloud terraform could set the `enabled` flag here to include.
> 
> I've wanted to actually move all the dashboards to our helm charts to make modifications easier, but we actually use these charts in our `examples` directory in `astronomer/astronomer`, for usage with our "demo" `docker-compose` version. This is probably out of date and not many people even know about it. Not sure if its worth us keeping around or not.
> 
> cc: @ryw @vikramkoka @vparekh94 if you guys have any thoughts on the matter.

I share the concern. Please keep this in a private cloud-only repo for now.

kaxil:
Closed in favor of https://github.com/astronomer/helm.astronomer.io/pull/198

PR #501, opened 2019-10-02
Add memory, % of limit graph for Prometheus

sjmiller609:


sjmiller609:
image

schnie:
Nice, looks good!

docker/vendor/grafana/include/prometheus.json
```diff
@@ -1023,13 +1042,104 @@
...
+          "legendFormat": "{{pod_name}}",
+          "refId": "A",
+          "step": 2
+        }
+      ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "Memory, % of limit",
```
schnie:
```suggestion
      "title": "Memory Usage",
```
Just to be consistent with the other similar graphs on other dashboards.

I like to make these titles as simple as possible and elaborate in the little tooltip when you hover over the _i_ icon on the charts.

@sjmiller609 would you mind adding a tooltip part to this, that goes in line with the other ones?

Downstream enterprise customers that may not have all the information and context, so a friendly explanation can help folks figure out all these charts.

schnie: Actually maybe add one to the PV chart above it, while you're in there if you don't mind. Must have forgotten that one. Been trying to standardize everything across these.

docker/vendor/grafana/include/prometheus.json
```diff
@@ -1023,13 +1042,104 @@
...
+      "pointradius": 5,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "container_memory_usage_bytes{component_name=\"prometheus\"} / container_spec_memory_limit_bytes{component_name=\"prometheus\"} * 100",
```
sjmiller609: we should use a different metric than container_memory_usage_bytes because it includes cached memory when it should not.

schnie: What do you mean by cached memory here?

sjmiller609: the kernel sometimes writes infrequently used memory pages to disk and removes them from memory.

sjmiller609: well actually, that's just my interpretation, I don't really know what it means. seems like cache does not make sense when the storage is slower. that's just what the metric is called "You might think that memory utilization is easily tracked with container_memory_usage_bytes, however, this metric also includes cached (think filesystem cache) items that can be evicted under memory pressure. The better metric is container_memory_working_set_bytes as this is what the OOM killer is watching for." https://blog.freshtracks.io/a-deep-dive-into-kubernetes-metrics-part-3-container-resource-metrics-361c5ee46e66

sjmiller609:
image

docker/vendor/grafana/include/airflow-overview.json
```diff
@@ -1716,7 +1716,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "container_memory_usage_bytes{deployment=~\"$deployment\", pod_name!~\".*-worker-.*\",component_name != \"POD\",image!=\"\", container_name!=\"istio-proxy\"} * 100 / container_spec_memory_limit_bytes{deployment=~\"$deployment\", pod_name!~\".*-worker-.*\",component_name != \"POD\",image!=\"\", container_name!=\"istio-proxy\"}",
+          "expr": "container_memory_working_set_bytes{deployment=~\"$deployment\", pod_name!~\".*-worker-.*\",component_name != \"POD\",image!=\"\", container_name!=\"istio-proxy\"} * 100 / container_spec_memory_limit_bytes{deployment=~\"$deployment\", pod_name!~\".*-worker-.*\",component_name != \"POD\",image!=\"\", container_name!=\"istio-proxy\"}",
```
sjmiller609: I modified the memory monitor on the airflow dashboard as well.

sjmiller609: image

schnie:
@sjmiller609 should we update the other charts that use this metric?
https://github.com/astronomer/astronomer/search?q=container_memory_usage_bytes&unscoped_q=container_memory_usage_bytes

docker/vendor/grafana/include/prometheus.json
```diff
@@ -1023,13 +1042,104 @@
         "alignLevel": null
       }
     },
+    {
+      "aliasColors": {},
+      "bars": false,
+      "dashLength": 10,
+      "dashes": false,
+      "datasource": "Prometheus",
+      "description": "Percent of memory limit used by Prometheus. Hitting 100% will cause termination.",
```
sjmiller609: added

PR #502, opened 2019-10-14
SLO Dashboard

sjmiller609:
Starting point for availability and latency SLO
image

Zoom on the client and server errors:
image

- Availability is moving average over 30 days the % HTTP 5xx (note 4xx NOT included)
- latency is p99, excluding requests to grafana and excluding requests of HTTP method PATCH


sjmiller609:
The latency panel is too slow to load. I need to optimize the query before we merge this

PR #504, opened 2019-10-15
Use latest of commander

sjmiller609:
depends on https://github.com/astronomer/commander/pull/19 and that pipeline completing

### DIFF
--- a/docker/platform/commander/Dockerfile
+++ b/docker/platform/commander/Dockerfile
@@ -13,7 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM astronomerinc/ap-base:0.10.1-alpha.19
+FROM astronomerinc/ap-commander:latest
+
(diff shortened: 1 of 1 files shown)


docker/platform/commander/Dockerfile
```diff
@@ -24,38 +25,3 @@ LABEL io.astronomer.docker.component="commander"
 ARG VERSION="v0.10.1-alpha.19"
```
sjmiller609: what about this?

schnie:
@sjmiller609 these tags are automatically replaced as part of the drone pipeline https://github.com/astronomer/astronomer/blob/master/.drone.yml#L159-L174

docker/platform/commander/Dockerfile
sjmiller609: > @sjmiller609 these tags are automatically replaced as part of the drone pipeline https://github.com/astronomer/astronomer/blob/master/.drone.yml#L159-L174

ok cool, i'll leave it then

PR #506, opened 2019-10-22
Uptime dashboard

sjmiller609:
Closes https://github.com/astronomer/issues/issues/464

See it live on prod here https://grafana.gcp0001.us-east4.astronomer.io/d/Mh_HjSoZz/uptime-dashboard?orgId=1&refresh=15m


sjmiller609:
I would like to add one more. I have an idea for monitoring the health of task runs - we can make an SLO for task completion based on the top 10% of success rate by deployment.

i.e. the top 10% of deployments, selected by task success rate, what is that rate?

sjmiller609:
I just pushed a change that ignores 503 only from the webserver, and has a few more panels.

schnie:
@sjmiller609 since this will be going out to Enterprise users as well, we may want to separate out any istio related stuff since that won't be guaranteed to exist (and most likely won't) in an enterprise env.

PR #509, opened 2019-11-01
Create a pip constraints file to fix Snowflake/Azure isssue

ashb:
This file (and the env var) makes pip install this specific version of
the module when ever it might be installed.

See https://pip.pypa.io/en/stable/user_guide/#constraints-files for more
info on this feature

The env var is like setting
`--constraint=/usr/local/share/astronomer-pip-constraints.txt`
automatically on every pip command


ashb:
In use:

```
root@311c1067014f:/usr/local/airflow# pip install azure-storage-blob
Collecting azure-storage-blob<12.0 (from -c /usr/local/share/astronomer-pip-constraints.txt (line 2))
```

PR #511, opened 2019-11-06
Add pip-constraints to Debian airflow image too

ashb:
We added this to our alpine image in #509, which was merged before the
debian image was added in #508


docker/airflow/debian-1.10.5/include/pip-constraints.txt
```diff
@@ -0,0 +1,2 @@
+# Constraint the version pip will install, if it ever needs to install a module
+azure-storage-blob<12.0
```
andriisoldatenko: from azure.storage.blob import BlockBlobService
ImportError: cannot import name ‘BlockBlobService’ ?

PR #517, opened 2019-11-14
Move `airflow initdb` to helm InitContainers

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/507
related PR https://github.com/astronomer/helm.astronomer.io/pull/238

### DIFF
--- a/docker/airflow/1.10.5/include/entrypoint
+++ b/docker/airflow/1.10.5/include/entrypoint
@@ -19,9 +19,6 @@ if [[ -n $AIRFLOW__CORE__SQL_ALCHEMY_CONN  ]] && [[ $CMD =~ ^(webserver|schedule
   while ! nc -w 1 -z "${HOST}" "${PORT}"; do
     sleep 0.001
   done
-
-  echo "Initializing airflow database..."
-  airflow initdb || true
 fi
 
 if [[ -n $AIRFLOW__CELERY__BROKER_URL ]] && [[ $CMD =~ ^(scheduler|worker|flower)$ ]]; then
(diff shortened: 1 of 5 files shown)


schnie:
@andriisoldatenko looks good. I'm going to let this one sit for a bit, until we get the other stuff merged so we don't break `master`.

docker/airflow/1.10.5/include/entrypoint
```diff
@@ -19,9 +19,6 @@ if [[ -n $AIRFLOW__CORE__SQL_ALCHEMY_CONN  ]] && [[ $CMD =~ ^(webserver|schedule
   while ! nc -w 1 -z "${HOST}" "${PORT}"; do
     sleep 0.001
```
andriisoldatenko: https://github.com/astronomer/astronomer/pull/517/files#diff-445ef4fcea54585dda41d282aff7723cR13-R19
do we really need this block here?

schnie: Without it, certain places, like the CLI (docker-compose), and in a new deployment, the postgres container, or pgbouncer may not be up and accepting traffic. This prevents the containers from exiting / crashing and allows for them to gracefully attach to the db/bouncer.

schnie:
@andriisoldatenko @ashb I'm worried about the CLI experience with this change. When using the local dev environment and this image, I'm thinking that the migrations wouldn't run since it relied on this. What can we do to serve both prod / dev environments?

Also, @andriisoldatenko would you mind moving this PR to this repo? https://github.com/astronomer/ap-airflow

The plan is to decompose this repo into several other repos, one of which is this airflow repo.

andriisoldatenko:
@ashb @schnie i was wondering to run it in `docker-compose.yaml`similar to `create_user`
https://github.com/astronomer/astro-cli/blob/3203c4e3cec87fc330929692f8d1bc2117041904/airflow/include/composeyml.go#L64
what do you think?

andriisoldatenko:
@schnie moved changes to https://github.com/astronomer/ap-airflow/pull/3

PR #518, opened 2019-11-14
Update registry to version 2.7.1

sjmiller609:
Closes https://github.com/astronomer/issues/issues/557


sjmiller609:
patch is already hotpatched to prod and stage. so we are good now

PR #522, opened 2019-12-04
Adds RHEL images

jbharath11:
There are 2 dockerfile scripts are for redis on ubi. 
First create image from redis-base-ubi and then refer that image in the redis-ubi to build the ap-redis-ubi


schnie:
@jbharath11 can we combine these scripts into one? Maybe just add the labels to the base image. Then it can mimic the same exact file structure.

PR #523, opened 2019-12-12
Bump fluentd to 1.7

andriisoldatenko:


schnie:
@andriisoldatenko were you able to reproduce the issue? I assume this fixes it? Doesn't seem like anything here wouldn't be backward compatible on older kube versions right?

banderson131:
Thanks for helping with this one!

PR #528, opened 2020-01-23
Add optional fluentd to s3 config

ianstanton:
* Adds the option to additionally send logs to an AWS S3 bucket
* Logs will still be sent to elasticsearch
* Users can enable by adding the following to their `config.yaml`:
```
fluentd:
  s3:
    enabled: true
    role_arn: ~
    role_session_name: ~
    s3_bucket: <bucket-name>
    s3_region: <aws-region>
```
* Related to astronomer/issues#559 and https://github.com/astronomer/astronomer/pull/521
* In order to use multiple outputs, we needed to use `@type copy`
  * https://docs.fluentd.org/output/copy


schnie:
@ianstanton looks like this has some conflicts now. We still want this right?

sjmiller609:
> @ianstanton looks like this has some conflicts now. We still want this right?

🤚 I want it!

ianstanton:
@schnie @sjmiller609 I'll be patching this up this week

sjmiller609:
@ianstanton and I chatted in PM and he said wait on more changes for this one.

ianstanton:
@sjmiller609 The logs are compressed and stored as gzip files:

image

sjmiller609:
Can the user view the logs from their airflow tasks logging view?

ianstanton:
> Can the user view the logs from their airflow tasks logging view?

Yes

charts/fluentd/templates/fluentd-configmap.yaml
```diff
@@ -175,24 +175,48 @@ data:
...
+          retry_type exponential_backoff
+          flush_thread_count 2
+          flush_interval 5s
+          retry_forever
+          retry_max_interval 30
+          chunk_limit_size "#{ENV['OUTPUT_BUFFER_CHUNK_LIMIT']}"
+          queue_limit_length "#{ENV['OUTPUT_BUFFER_QUEUE_LIMIT']}"
+          overflow_action block
+        </buffer>
+      </store>
```
sjmiller609, line 200: can you give me a quick explanation for this part that is applied regardless of the configuration for s3? I would like a comparison to the existing behavior.

configs/fluentd-s3.yaml
```diff
@@ -0,0 +1,10 @@
+
+# This is an example configuration that allows for logs to be forwarded to an AWS S3 bucket.
+
+fluentd:  
+  s3:
+    enabled: true
```
sjmiller609, line 6:
```suggestion
    enabled: false
```

sjmiller609: I know it would overwrite... but just to be safe and consistent with it.

sjmiller609:
also, please remove the [WIP] marker when you consider prod-ready.

configs/fluentd-s3.yaml
ianstanton: Thanks, missed this!

charts/fluentd/templates/fluentd-configmap.yaml
ianstanton: Here we're using the fluentd's `copy` output plugin to send to multiple destinations. More info here https://docs.fluentd.org/output/copy

configs/fluentd-s3.yaml
ianstanton: This was to be used as an example config for enabling this feature. Should we set this as false? Should we have the example at all?

ianstanton:
@sjmiller609 are you able to see what the merge conflicts are? I'm not able to see the details, just the file.

sjmiller609:
> @sjmiller609 are you able to see what the merge conflicts are? I'm not able to see the details, just the file.

Hi Ian, will you please try this procedure?

```
# check that you are on your branch feature/fluentd-s3
git branch

# ensure latest master branch available locally
git fetch origin master

# start the merge conflict resolution process
git pull origin master
# above: this will show an error message, and indicate you need to resolve merge conflicts

# this will show in red the files with conflicts
git status

# for each file in conflict edit the file and resolve the conflict
vim ./file
# ... and add the file to the git staging area
git add ./file

# double check nothing left:
git status

# proceed with the merge
git merge --continue

# push to your branch
git push origin feature/fluentd-s3
```

ianstanton:
@sjmiller609 I've confirmed that default fluentd -> es functionality is still there when `fluentd.s3.enabled` is false. Let me know if you'd like to see a demo of this.

configs/fluentd-s3.yaml
ashb: I would say just have an example in the values.yaml file, no need for a separate file (and we should probably give a fuller example -- if just switching on enabled is enough to make it work we probably don't need this)

sjmiller609: I see, I didn't realize the file location

sjmiller609:
Thanks Ian!

PR #531, opened 2020-01-23
Istio sidecar

sjmiller609:
In the interest of reducing memory overhead for Istio:

Issue: https://github.com/astronomer/issues/issues/702
Sidecar docs: https://istio.io/docs/reference/config/networking/sidecar/
Hint provided by: https://discuss.istio.io/t/typical-memory-utilization/2504
Another discussion post (from me): https://discuss.istio.io/t/configuration-to-avoid-growing-proxy-memory-for-inaccessible-services/4867

(dupe of https://github.com/astronomer/helm.astronomer.io/pull/305)


charts/astronomer/templates/istio/sidecars.yaml
```diff
@@ -0,0 +1,57 @@
+{{- if .Values.global.istioEnabled }}
```
sjmiller609: @schnie I emergency deployed this change manually using kubectl apply during the change window on 1/23/20

schnie:
@sjmiller609 looks good. The only thing I'd say we may want to do is move `.Values.global.istioEnabled` to `.Values.global.istio.enabled` and also add the `rootNamespace` to that same `istio` block. This would keep `istio` config in one place, and it'd conform to the config format we've be standardizing on elsewhere.

sjmiller609:
good call!

sjmiller609:
@schnie ok taken care of

PR #535, opened 2020-01-28
Add domain to SSL

andriisoldatenko:


schnie:
@andriisoldatenko should we .gitignore those pem files?

andriisoldatenko:
nice catch!

bin/generate_ssl_keys
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # https://github.com/FiloSottile/mkcert
 mkcert -install
-mkcert app.local.astronomer-development.com "*.local.astronomer-development.com"
-cat app.local.astronomer-development.com+1.pem > /tmp/fullchain.pem
+mkcert app.local.astronomer-development.com local.astronomer-development.com "*.local.astronomer-development.com"
+cat app.local.astronomer-development.com+2.pem > /tmp/fullchain.pem
```
schnie: @andriisoldatenko should we just keep this at +1? Or is there a way to choose a distinct name that `mkcert` generates?

andriisoldatenko: @schnie nice catch done in 0600451006074b87f36f64c2c6b69c221765be3d

andriisoldatenko:
TODO rewrite of URL 
 nginx.ingress.kubernetes.io/configuration-snippet: |
         rewrite /preview https://test.app.example.com$uri permanent;

PR #536, opened 2020-01-28
Change default tlsSecret to astronomer-tls

sjmiller609:


schnie:
Let's think about this one a bit. If its possible, we're trying to get a version of the platform running without a tls secret, so users can get up and running quicker for 0.12. There's a lot of logic that keys on the presence of a value here that we may want to actually start using.

PR #538, opened 2020-01-29
v0.11.1 stability fixes

sjmiller609:


charts/astronomer/values.yaml
```diff
@@ -107,15 +107,17 @@ houston:
...
+      chart:
+        version: 0.11.0
 
   # Automatically upgrade Airflow deployments to the latest
   # version specified by Houston configuration.
   # This runs as a Job after helm upgrades.
   upgradeDeployments:
     # Enable this helm hook on upgrade
     enabled: true
-
```
schnie: @sjmiller609 can we revert this whitespace change?

schnie:
Looks good, just need to bump the subchart version. So this is essentially the blueprint for the process that we need to automate.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+<!--
+Thank you for contributing to Astronomer!
+-->
+
+**Checklist**
+
+- [ ]  Chart.yaml version updated
+- [ ]  For each modified subchart, charts/subchart/Chart.yaml version updated
```
sjmiller609: @schnie I also added this, hope that's alright

charts/astronomer/values.yaml
```diff
@@ -107,7 +107,10 @@ houston:
 
   # Any Houston configuration. Reference here:
   # https://github.com/astronomer/houston-api/blob/master/config/default.yaml
-  config: {}
+  config:
```
schnie: @sjmiller609 we can revert this back. This section is reserved for end-user overrides. Anything we want to package and ship with the distribution can be added here: https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/houston/houston-configmap.yaml#L54-L62, which already has 0.11.0.

PR #539, opened 2020-01-29
Update PULL_REQUEST_TEMPLATE.md

sjmiller609:
**Checklist**
n/a
- [ ]  Chart.yaml version updated
- [ ]  For each modified subchart, charts/subchart/Chart.yaml version updated

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,4 +5,4 @@ Thank you for contributing to Astronomer!
 **Checklist**
 
 - [ ]  Chart.yaml version and appVersion updated (matching)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,5 +4,5 @@ Thank you for contributing to Astronomer!
 
 **Checklist**
 
-- [ ]  Chart.yaml version updated
-- [ ]  For each modified subchart, charts/subchart/Chart.yaml version updated
+- [ ]  Chart.yaml version and appVersion updated (matching)
+- [ ]  For each modified subchart, charts/subchart/Chart.yaml version and appVersion updated (appVersion should match the corresponding image tag - in the case of the Astronomer subchart it should match the houston image version)
```
sjmiller609: consider if you agree with this or not, just my first thought, but I don't mind if we just make it always match

schnie: We don't have any `appVersion`s in the subcharts, so we can probably just skip it there.

PR #541, opened 2020-01-30
Circle release process

sjmiller609:
requires: GCP_TOKEN, GITHUB_TOKEN

**Checklist**
n/a
- [ ]  Chart.yaml version and appVersion updated (matching)
- [ ]  For each modified subchart, charts/subchart/Chart.yaml version updated


schnie:
@sjmiller609 awesome! So after we approve a build for a `release-x` branch this will package and update helm.astronmer.io with the `astronomer` umbrella chart right?

So users will be able to `helm repo add astronomer helm.astronomer.io && helm install astronomer`?

sjmiller609:
> @sjmiller609 awesome! So after we approve a build for a `release-x` branch this will package and update helm.astronmer.io with the `astronomer` umbrella chart right?
> 
> So users will be able to `helm repo add astronomer helm.astronomer.io && helm install astronomer`?

yep, that's the idea! working locally, but still worth keeping those gsutil backups for the first few runs until we are comfortable. There are GCP storage features we can use to make the tarballs immutable if we want to.

sjmiller609:
If you tried to merge it right now into a release branch and approve it, I expect it would fail on the 'publish-github-release' step because the version in Chart.yaml matches an existing release tag (v0.11.1)

schnie:
Cool, sounds good. We can merge now and try it as we go, like the airflow chart.

sjmiller609:
thanks greg, multi-kube version testing incoming shortly

PR #542, opened 2020-01-30
Test k8s versions 1.12 to 1.17

sjmiller609:
Closes: https://github.com/astronomer/issues/issues/653

The point of this PR is to run 'helm install' for kube versions 1.12 - 1.17 in CI. Along the way, I found some issues, and they are resolved in this PR as well.

**Checklist**


Makefile
```diff
@@ -20,6 +20,9 @@ lint:
 	mkdir -p ${TEMP}
 	cp -R ../astronomer ${TEMP}/astronomer || cp -R ../project ${TEMP}/astronomer
 	helm lint ${TEMP}/astronomer
+	for chart in $$(ls ${TEMP}/astronomer/charts); do
+	helm lint ${TEMP}/astronomer/charts/$$chart
+	done
```
sjmiller609: I found that subcharts were not being completely linted

charts/astronomer/values.yaml
```diff
@@ -2,6 +2,20 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+global:
```
sjmiller609: by adding the used global variables down in each subchart, this allows the charts to be valid independently, which is important for linting.

charts/elasticsearch/templates/data/es-data-statefulset.yaml
```diff
@@ -2,7 +2,11 @@
 ## Elasticsearch Data StatefulSet
 #################################
 kind: StatefulSet
+{{- if gt (.Capabilities.KubeVersion.Minor | atoi) 15 }}
+apiVersion: apps/v1
+{{- else }}
 apiVersion: apps/v1beta2
+{{- end }}
```
sjmiller609: this is how to split api b/w kube versions. I tried to avoid this when possible, and I'm using 'greater than' so it doesn't require more work moving forward into future versions. We only need to check the minor version because the minimum version is specified in Chart.yaml to >= 1.12 (so we don't need to worry about major version 0)

Chart.yaml
```diff
@@ -1,7 +1,8 @@
 apiVersion: v1
+kubeVersion: ">= 1.12.0"
```
sjmiller609: nice to note this feature of helm

sjmiller609:
@schnie this is good to go. It includes fixes for installing our platform on 1.16+, which I think currently does not work

.circleci/config.yml
```diff
@@ -114,7 +393,7 @@ commands:
           at: /tmp/workspace
       - checkout
       - run:
-          name: Install Airflow chart
+          name: Install Astronomerirflow chart
```
schnie:
```suggestion
          name: Install Astronomer chart
```

.circleci/config.yml
```diff
@@ -131,4 +410,4 @@ commands:
             set -xe
             pip3 install --user astronomer_e2e_test
             # We can remove --debug after we are happy with it
-            astronomer-ci --debug publish-github-release --github-repository $CIRCLE_PROJECT_REPONAME --github-organization $CIRCLE_PROJECT_USERNAME --commitish $CIRCLE_SHA1
+            astronomer-ci --debug publish-github-release --github-repository $CIRCLE_PROJECT_REPONAME --github-organization $CIRCLE_PROJECT_USERNAME --commitish $CIRCLE_SHA1
```
schnie: @sjmiller609 looks like we need a newline here.

sjmiller609: this is a jinja2 template output, I can modify the script to add a newline at the end?

sjmiller609:
> Looking good. Bumping the chart versions as part PR of this has me thinking though. I feel like we should be able to continue to merge things in, without worrying about versions, and then have the pipeline handle bumping these version numbers. This feels a little too tightly coupled and possibly prone to human error.

Makes sense, I can come up with something for that.

sjmiller609:
@schnie ready to go?

schnie:
I think everything looks good. I'm just hung up on the manual bumping of the subchart versions here. Like what if we go and follow up later today with another patch to `alertmanager`? This would not have been released yet, and we wouldn't really need to bump the version again.

sjmiller609:
> I think everything looks good. I'm just hung up on the manual bumping of the subchart versions here. Like what if we go and follow up later today with another patch to `alertmanager`? This would not have been released yet, and we wouldn't really need to bump the version again.

Please consider this proposal https://github.com/astronomer/issues/issues/746

schnie:
@sjmiller609 is there a way we can avoid having the `global` section duplicated in the subchart values.yamls?

Maybe a single file in `configs` or something that we can reference when needed? `helm` allows multiple `-f` flags.

sjmiller609:
> @sjmiller609 is there a way we can avoid having the `global` section duplicated in the subchart values.yamls?
> 
> Maybe a single file in `configs` or something that we can reference when needed? `helm` allows multiple `-f` flags.

just realized that -f works with 'helm lint'! Sure, I can just duplicate the globals section one time and -f that file in all the subchart linting steps.

charts/prometheus/values.yaml
```diff
@@ -47,4 +47,4 @@ persistence:
 ports:
   http: 9090
 
-ingressNetworkPolicyExtraSelectors: []
```
sjmiller609: I know we don't like unrelated changes... but we want newlines, right?

schnie: Yep

sjmiller609:
@schnie ok it's ready

schnie:
@sjmiller609 can you patch the conflicts here?

PR #546, opened 2020-02-04
Add KEDA

sjmiller609:
This PR will bundle KEDA into the Astronomer platform

**Checklist**


schnie:
Does the KEDA chart allow us to override the image that's running? We'll need a vendored image that we control for security patches and consistency.

sjmiller609:
Looks like I have some work to do on the CI. I'll get Dan's image into the vendored stuff and figure out what's going on to make the testing fail.

requirements.yaml
```diff
@@ -50,6 +50,11 @@ dependencies:
   # In-cluster DB, not recommended
   # for production
   - name: postgresql
-    condition: postgresql.enabled
```
sjmiller609: I noticed that this line is causing postgresql to install even when the tag is set to false. removing it makes it work properly. This is evident by checking CI runs to see that postgresql was running in the jobs we did not expect it would be.

charts/keda/values.yaml
```diff
@@ -0,0 +1,7 @@
+enabled: true
+
+keda:
+  image:
+    # Enable KEDA POC in cloud
+    keda: "dimberman/keda:master"
+    metricsAdapter: "dimberman/keda-metrics-adapter:master"
```
sjmiller609: I'm still working on this, I will get it vendored before we merge

sjmiller609:
I prefer we merge this one first because it will make sure keda is coming up on older kube versions: https://github.com/astronomer/astronomer/pull/542

charts/keda/values.yaml
sjmiller609: I can't find what you built your image from... can you PR to ap-vendor please @dimberman

schnie: @dimberman @sjmiller609 these are being published to dockerhub now.

dimberman: @schnie I've changed it to point to our docker image

requirements.yaml
schnie: @dimberman do you see this as well?

schnie:
@dimberman could you look at these conflicts today so we can merge and deploy out to the dev?

To standardize with the rest of the platform we may want to use `.Values.global.kedaEnabled`, rather than the `.Values.keda.enabled` to key on. We just flipped postgres to do the same.

dimberman:
@schnie fixed

values.yaml
```diff
@@ -297,4 +300,4 @@ kubed:
       memory: "512Mi"
     limits:
       cpu: "2"
-      memory: "1024Mi"
+      memory: "1024Mi"
```
sjmiller609:
```suggestion
      memory: "1024Mi"

```

PR #547, opened 2020-02-05
Fix CI

sjmiller609:
We found some issues in the airflow chart CI, I'm adding those same fixes over here. The release worked for airflow chart, evident by "helm search -l astronomer/airflow" showing a new version 0.11.1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,8 +32,8 @@ jobs:
       - run:
           name: Release the Astronomer chart
           command: |
-            printf "%s" $${GCP_TOKEN} > /tmp/gcs_token.json
-            gcloud auth activate-service-account --key-file=/tmp/gcs_token.json
+            set -e
+            pyenv global 2.7.12
             HELM_CHART_PATH='/tmp/workspace/astronomer-*.tgz' bin/release
 
   platform:
@@ -129,6 +129,6 @@ commands:
           name: Create a release on GitHub
           command: |
             set -xe
-            pip3 install --user astronomer_e2e_test
-            # We can remove --debug after we are happy with it
-            astronomer-ci --debug publish-github-release --github-repository $CIRCLE_PROJECT_REPONAME --github-organization $CIRCLE_PROJECT_USERNAME --commitish $CIRCLE_SHA1
+            pyenv global 3.7.0
+            pip install astronomer_e2e_test
+            astronomer-ci publish-github-release --github-repository $CIRCLE_PROJECT_REPONAME --github-organization $CIRCLE_PROJECT_USERNAME --commitish $CIRCLE_SHA1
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -32,8 +32,8 @@ jobs:
       - run:
           name: Release the Astronomer chart
           command: |
-            printf "%s" $${GCP_TOKEN} > /tmp/gcs_token.json
```
sjmiller609: two issues here:
- printf command modifies the string to require a sed command to adjust (can't tell you why). instead use 'echo'
- two $$ is Makefile syntax, not normal script

.circleci/config.yml
```diff
@@ -129,6 +129,6 @@ commands:
           name: Create a release on GitHub
           command: |
             set -xe
-            pip3 install --user astronomer_e2e_test
-            # We can remove --debug after we are happy with it
-            astronomer-ci --debug publish-github-release --github-repository $CIRCLE_PROJECT_REPONAME --github-organization $CIRCLE_PROJECT_USERNAME --commitish $CIRCLE_SHA1
+            pyenv global 3.7.0
+            pip install astronomer_e2e_test
```
sjmiller609: pyenv will set pip to the correct python version. However, I found this did not work with --user flag

.circleci/config.yml
```diff
@@ -129,6 +129,6 @@ commands:
...
+            astronomer-ci publish-github-release --github-repository $CIRCLE_PROJECT_REPONAME --github-organization $CIRCLE_PROJECT_USERNAME --commitish $CIRCLE_SHA1
```
sjmiller609: we do not need "--debug" any more.

bin/release
```diff
@@ -40,13 +40,21 @@ echo "========================="
 echo ""
 
 echo "Releasing chart..."
+
+# This is needed to avoid using the CI's built-in
+# gcloud configurations.
+export BOTO_CONFIG=/dev/null
```
sjmiller609: The is a conflict in the built-in gcloud that's already in CI and the binary installed by our script. This variable serves to avoid making use of Circle's configs, which were messing it up.

bin/release
```diff
@@ -40,13 +40,21 @@ echo "========================="
...
+
+echo ${GCP_TOKEN} > /tmp/gcs_token.json
+gcloud auth activate-service-account --key-file=/tmp/gcs_token.json
+
 set -xe
 # Set up the index.html file
 wget https://helm.astronomer.io/index.yaml -O /tmp/index.yaml.current
-helm repo index ./astronomer --url https://helm.astronomer.io --merge /tmp/index.yaml.current
+helm repo index . --url https://helm.astronomer.io --merge /tmp/index.yaml.current
```
sjmiller609: helm repo index should target the directory in which the zip file(s) live, not the unzipped directory.

PR #548, opened 2020-02-05
Remove norbit from charts and rename astro-ui

schnie:
This PR removes the `orbit` chart, and renames `norbit` to `astro-ui`. This makes `astro-ui` the default UI when installing the platform.

It's currently pulling from the `latest` tag since we don't have a published version yet. We'll update that once we are pushing a prod tag.


CHANGELOG.md
```diff
@@ -35,7 +35,7 @@ All notable changes to this project will be documented in this file.
 ## [0.10.1] - 2019-09-30
 - Upgrade to Airflow 1.10.5
 - Kubernetes 1.14+ support
-- Improved metrics in Grafana and Orbit
+- Improved metrics in Grafana and the UI
```
andriisoldatenko: not sure that it's good idea to edit old releases in CHANGELGO
@ash wdyt?
better to create new section 0.12 or so and add 1 line.

schnie: Yea went back and forth on that too. Figured as we move forward, orbit would loose meaning for folks inside and outside the org, so this would make it clearer. Open to change if others feel otherwise.

charts/astronomer/templates/ingress.yaml
```diff
@@ -42,24 +42,15 @@ spec:
...
-            serviceName: {{ .Release.Name }}-orbit
-            servicePort: orbit-http
-{{- if .Values.norbit.enabled }}
-  - host: beta.{{ .Values.global.baseDomain }}
-    http:
-      paths:
-        - path: /
-          backend:
-            serviceName: {{ .Release.Name }}-norbit
-            servicePort: norbit-http
```
sjmiller609: I prefer if we keep the "beta." functionality around. I hope we can enable the front end team to push to this unconstrained by a release process.

sjmiller609: for example, the service could have pull policy always and point to astro-ui:latest. And somehow gets triggered to update when a new one is published

sjmiller609: perhaps we just put it back in with a different change, that may be outside the scope of this change.

PR #549, opened 2020-02-06
[WIP] Tests to check prereqs for installation in various scenarios

verdverm:
This PR adds a script and set of bats-core tests which can be used to verify

- tools & versions
- cloud setup & permissions
- external component readiness (like DB)
- k8s features available

See the run.sh for an overview and help message. TL;DR:

```
./bin/prereq-tests/run.sh -TAGKD
```

We ought to be able to use this setup to extract the "available features"
and feed this into the actual installation process.

---

[bats-core](https://github.com/bats-core/bats-core)


schnie:
@verdverm trying to run this locally but don't get much output. Are there any `-K` (k8s) tests yet?

verdverm:
@schnie No k8s yet, was about to do that, keep having my cluster access removed...

verdverm:
probably need some config file for providing parameters to some of the tests, like checking for an existing DB. Any thoughts on this? (docs and a config file for the user to fill in?)

verdverm:
@schnie this is probably ready for review. We may want to (dis)able certain tests for flags so that the most common scenarios are the easiest.

bin/prereq-tests/run.sh
```diff
@@ -0,0 +1,119 @@
...
+    K) T_K8S=1 ;;
+    P) T_AP=1 ;;
+    E) T_EXT=1 ;;
+
+    h)
+      USAGE
+      exit 0
+      ;;
+
+    *)
```
schnie: @verdverm could we have it print the help text if no flag is specified?

PR #550, opened 2020-02-06
Always use the Astronomer Airflow entrypoint

sjmiller609:
Closes: https://github.com/astronomer/issues/issues/620
Evidence that it's working: https://github.com/astronomer/issues/issues/620#issuecomment-583064469

I think for the sake of tonight's release, I will be adding these configurations in with the helm configuration (via the terraform-google-astronomer-cloud's locals.tf file)

I will get this PR in to the ap-airflow chart to ensure that this hardcoded entrypoint will always be valid: https://github.com/astronomer/ap-airflow/pull/37

**Checklist**


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -105,6 +105,28 @@ data:
...
+              else:
+                  extra_labels["kubernetes-pod-operator"] = "True"
+              pod.labels.update(extra_labels)
+              pod.tolerations += []
+              pod.affinity.update({})
+              # Ensure our entrypoint is respected
+              if not pod.args:
+                pod.args = []
+              pod.args = pod.cmds + pod.args
+              pod.cmds = ["tini", "--", "/entrypoint"]
```
sjmiller609: this part is different than the default.

PR #552, opened 2020-02-06
Disable the pod disruption budgets by default

schnie:
**Checklist**

- [ ]  For each modified subchart, charts/subchart/Chart.yaml version updated to VERSION

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -106,6 +106,14 @@ data:
         # Scheduler configuration.
         scheduler:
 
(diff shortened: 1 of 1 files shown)


schnie:
@sjmiller609 this is where I think its a little weird right now. I merged your previous PR, which bumped this subchart version, but its only sitting on master and hasn't been pushed anywhere. 

I don't think I should necessarily also bump the version since we haven't released it to anything. But that requires the prior knowledge that I _just_ merged your PR. If it was a couple of days from now, or someone else, that link may not exist.

schnie:
The next steps on our pipeline will handle this right?

PR #559, opened 2020-02-12
[WIP] Add verify ssl

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/647


charts/postgresql/values.yaml
```diff
@@ -1,6 +1,8 @@
 # Configuration for postgres subchart
 enabled: true
 
+verify_ssl: false
```
schnie: @andriisoldatenko nit but could we make it `verifySSL` for consistency with the rest of the umbrella?

charts/postgresql/templates/_helpers.yaml
```diff
@@ -0,0 +1,7 @@
+{{- define "astronomer_bootstrap_url" -}}
+{{- if .Values.verify_ssl }}
+postgres://{{ .Values.postgresql.postgresqlUsername }}:{{ .Values.postgresql.postgresqlPassword }}@{{ .Release.Name }}-postgresql.{{ .Release.Namespace }}.svc.cluster.local:{{.Values.postgresql.servicePort}}?sslmode=verify-ca"
```
schnie: @andriisoldatenko what other modes are available? Should we support a custom `mode` option, rather than the boolean?

charts/postgresql/templates/astronomer-bootstrap-secret.yaml
```diff
@@ -7,5 +7,5 @@ kind: Secret
 metadata:
   name: astronomer-bootstrap
 data:
-  connection: {{ printf "postgres://%s:%s@%s-postgresql.%s.svc.cluster.local:%s" .Values.postgresql.postgresqlUsername .Values.postgresql.postgresqlPassword .Release.Name .Release.Namespace ( .Values.postgresql.servicePort | toString ) | b64enc | quote }}
+  connection:  {{ include "pgbouncer_users" . | b64enc | quote }}
```
schnie: @andriisoldatenko is this a copy/paste error?

charts/postgresql/templates/_helpers.yaml
```diff
@@ -0,0 +1,7 @@
+{{- define "astronomer_bootstrap_url" -}}
```
schnie: @andriisoldatenko the rest of the umbrella uses a dot pattern, maybe rename to `astronomer.boostrap`?

schnie:
@andriisoldatenko @ashb this PR really only affects the built-in postgres option, in which case, I'm not sure it really matters.

I think the main thing we needed to prove, was if a user specifies an `sslmode` like `verify-ca` or `verify-full`, does everything work the way we expect.

I really don't see why it wouldn't, but we just needed to test.

sjmiller609:
I think we discussed some changes in this PR required after prisma 2 migration, so marking WIP. feel free to correct or hit me up.

sjmiller609:
@andriisoldatenko lmk if this PR is no longer applicable

andriisoldatenko:
yeah i think we can close it.

PR #560, opened 2020-02-13
Alert Emails missing Workspace Label

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/764


andriisoldatenko:
image

ashb:
Do we need to worry about the failed builds? (I mean yes, we should obviously worry a little bit)

PR #563, opened 2020-02-13
Add node exporter subchart

sjmiller609:
This adds nodes metrics to our metrics stack.

Related issues:
- https://github.com/astronomer/issues/issues/784
- https://github.com/astronomer/issues/issues/769

This was tested locally on kind using this kind configuration:

image

Here is evidence that all nodes became prom targets:
image

Here are the pods that were running:
image

**Checklist**

- [x]  Chart.yaml version and appVersion updated to match VERSION

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,11 +41,9 @@ build:
 	mkdir -p ${TEMP}
 	cp -R ../astronomer ${TEMP}/astronomer || cp -R ../project ${TEMP}/astronomer
 	# Install external charts
-	for chart in $$(ls ${TEMP}/astronomer/charts); do
-	if [[ -f ${TEMP}/astronomer/charts/$$chart/requirements.yaml ]]; then
-	helm dep update ${TEMP}/astronomer/charts/$$chart
-	fi
-	done
+	helm dep update ${TEMP}/astronomer/charts/postgresql
+	helm dep update ${TEMP}/astronomer/charts/keda
+	helm dep update ${TEMP}/astronomer/charts/node-exporter
 	helm package ${TEMP}/astronomer
 
 .PHONY: build-index
--- a/Makefile
+++ b/Makefile
@@ -33,15 +33,19 @@ lint:
 	rm ${TEMP}/prometheus_alerts.yaml
 
 .PHONY: build
+.ONESHELL:
 build:
 	set -xe
 	helm repo add kedacore https://kedacore.github.io/charts
 	rm -rf ${TEMP}/astronomer || true
 	mkdir -p ${TEMP}
 	cp -R ../astronomer ${TEMP}/astronomer || cp -R ../project ${TEMP}/astronomer
-	# Install the external chart
-	helm dependency update ${TEMP}/astronomer/charts/postgresql
-	helm dependency update ${TEMP}/astronomer/charts/keda
+	# Install external charts
+	for chart in $$(ls ${TEMP}/astronomer/charts); do
+	if [[ -f ${TEMP}/astronomer/charts/$$chart/requirements.yaml ]]; then
+	helm dep update ${TEMP}/astronomer/charts/$$chart
+	fi
+	done
 	helm package ${TEMP}/astronomer
 
 .PHONY: build-index
--- a/Makefile
+++ b/Makefile
@@ -41,9 +41,11 @@ build:
 	mkdir -p ${TEMP}
 	cp -R ../astronomer ${TEMP}/astronomer || cp -R ../project ${TEMP}/astronomer
 	# Install external charts
-	helm dep update ${TEMP}/astronomer/charts/postgresql
-	helm dep update ${TEMP}/astronomer/charts/keda
-	helm dep update ${TEMP}/astronomer/charts/node-exporter
+	for chart in $$(ls ${TEMP}/astronomer/charts); do
+	if test -f ${TEMP}/astronomer/charts/$$chart/requirements.yaml; then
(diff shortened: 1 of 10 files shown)


sjmiller609:
I confirmed that CI has the node exporter running https://circleci.com/gh/astronomer/astronomer/1408?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Makefile
```diff
@@ -33,15 +33,17 @@ lint:
...
 	set -xe
 	helm repo add kedacore https://kedacore.github.io/charts
 	rm -rf ${TEMP}/astronomer || true
 	mkdir -p ${TEMP}
 	cp -R ../astronomer ${TEMP}/astronomer || cp -R ../project ${TEMP}/astronomer
-	# Install the external chart
-	helm dependency update ${TEMP}/astronomer/charts/postgresql
-	helm dependency update ${TEMP}/astronomer/charts/keda
+	# Install external charts
+	helm dep update ${TEMP}/astronomer/charts/postgresql
```
schnie: Should we simplify this to `helm dep update`, or is there a reason that we're listing them one by one?

andriisoldatenko: i think it won't work.

charts/node-exporter/values.yaml
```diff
@@ -0,0 +1,15 @@
+prometheus-node-exporter:
```
schnie: @sjmiller609 I think we'll want to maintain our ability to pin our stuff to certain node pools. Can we somehow pass some node pool settings down through to this chart, based on our `global` settings?

Makefile
sjmiller609: yeah, helm dep update at the top level isn't working. I can do a loop though.

charts/node-exporter/values.yaml
sjmiller609: Ah right. Thanks for the reminder greg. I will need to ensure that tolerations and affinity can be passed. We will make use of the toleration configuration for cloud.

sjmiller609: After considering this, I believe that no change is required. We can pass down tolerations from the top-level chart like this:
```
node-exporter:
  prometheus-node-exporter:
    tolerations: []
```
We do not want to use the global configuration "platformNodePool" because this should be applied to all nodes - we want to just pass in tolerations for all taints, and allow the daemonset to run on all nodes.

charts/node-exporter/values.yaml
```diff
@@ -0,0 +1,15 @@
...
+    type: ClusterIP
+    port: 9100
+    targetPort: 9100
+    # By default, the prometheus annotation is on the service,
+    # but I do not think this is the intended configuration.
+    # https://github.com/helm/charts/issues/20736
+    annotations: {}
+  podAnnotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "9100"
```
sjmiller609: @jpweber will you review this comment in particular? I wonder if you have had this experience in the past

jpweber: Short answer is both will work. But really it all depends on how your scrape config is setup to scrape the information. The default configuration with the scrape annotation is assuming you are using the `endpoints` role. If an end point is attached to a service all the service `__meta_` labels are included.

charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -188,6 +188,20 @@ data:
             regex: 'fluentd_tail_file_.*'
             action: drop
 
+      {{- if .Values.global.nodeExporterEnabled }}
+      - job_name: node-exporter
+        scrape_interval: 30s
+        kubernetes_sd_configs:
+          - role: pod
```
jpweber: This is where `endpoints` would be used if you wanted to use the default helm chart settings

charts/node-exporter/values.yaml
schnie: I'm just thinking of the enterprise use-case where we may or may not have control of an entire cluster, maybe a specific namespace and node pool. If we couldn't deploy the daemonset to all nodes, would it still be useful to have on the nodes that our stuff is running?

jpweber: If there are enterprise use cases where we are limited to what nodes we can be installed on then yes we should have the ability to add node selectors or modify the tolerations.

sjmiller609: we do have that capability (see my most recent comment in this thread)

charts/node-exporter/values.yaml
sjmiller609: ok I see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#endpoints

sjmiller609:
@schnie please squash-and-merge it

charts/node-exporter/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+name: node-exporter
+version: 0.11.2
```
schnie: @sjmiller609 should we go 0.12 here?

sjmiller609: sure

values.yaml
```diff
@@ -35,6 +35,8 @@ global:
   # This is not recommended for production.
   postgresqlEnabled: true
 
+  nodeExporterEnabled: true
```
schnie:
```suggestion
  # Enables a node exporter DaemonSet for node-level metrics
  nodeExporterEnabled: true
```

PR #566, opened 2020-02-15
Add support for Azure DB PostgreSQL - astronomer chart

ianstanton:
fix issue astronomer/issues#102


schnie:
@ianstanton I'm wondering if we could mimic the same format / conventions that we're using for houston here: https://github.com/astronomer/astronomer/pull/566/files#diff-252d098af651bc3cec78dcadc46dc053R60-R61.

This way we can really just treat it as a manual prisma connection override, rather than relying on the one created by the bootstraper - so we could skip the bootstrap. Hoping it should also let us drop the `azure`-specific naming and keep it generic looking.

I think we'd just need to add 2 more checks here for the prisma secrets: https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/prisma/prisma-deployment.yaml#L35.

I may be skipping over something, but generally what do you think?

charts/astronomer/values.yaml
```diff
@@ -158,6 +158,7 @@ prisma:
   #  requests:
   #   cpu: 100m
   #   memory: 128Mi
+  azureDb: false
```
schnie: @ianstanton maybe we rename this `prisma.backendSecretName` to match Houston up above, and make a comment that if you are overriding the prisma secret, you have to use the right format / names for the keys.

At a minimum, this would just let us drop the `azure`-specific naming here.

ianstanton:
cc @vishwas-astro

charts/astronomer/values.yaml
ianstanton: Pushed changes to reflect this

ianstanton:
> @ianstanton I'm wondering if we could mimic the same format / conventions that we're using for houston here: https://github.com/astronomer/astronomer/pull/566/files#diff-252d098af651bc3cec78dcadc46dc053R60-R61.
> 
> This way we can really just treat it as a manual prisma connection override, rather than relying on the one created by the bootstraper - so we could skip the bootstrap. Hoping it should also let us drop the `azure`-specific naming and keep it generic looking.
> 
> I think we'd just need to add 2 more checks here for the prisma secrets: https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/prisma/prisma-deployment.yaml#L35.
> 
> I may be skipping over something, but generally what do you think?

Are you thinking we can skip the bootstrapper init container entirely? https://github.com/astronomer/astronomer/blob/113b61f43be428faf01a68788fcbafa4aff35ad6/charts/astronomer/templates/prisma/prisma-deployment.yaml#L36-L55

charts/astronomer/templates/prisma/prisma-configmap.yaml
```diff
@@ -18,7 +18,21 @@ data:
...
+        port: $PRISMA_DB_PORT
+        user: $PRISMA_DB_USER
+        password: $PRISMA_DB_PASSWORD
+        database: {{ printf "%s_houston" (.Release.Name | replace "-" "_")  }}
+        {{ else if and (.Values.prisma.backendConnection) (not .Values.prisma.backendSecretName) }}
+        host: {{ .Values.prisma.backendConnection.host }}
+        port: {{ .Values.prisma.backendConnection.port }}
+        user: {{ .Values.prisma.backendConnection.user }}
+        password: {{ .Values.prisma.backendConnection.pass }}
+        database: {{ printf "%s_houston" (.Release.Name | replace "-" "_")  }}
```
schnie: @ianstanton since we duplicate this in both blocks, we could assign a variable up top first, and use it down here, like we do over here for example: https://github.com/astronomer/airflow-chart/blob/master/templates/scheduler/scheduler-deployment.yaml#L5-L15.

schnie:
@ianstanton now that I'm looking at this again, I'm wondering if making this under `prisma` is more confusing, since it's really just doing the same thing as the houston database configuration options, its just a secret way to pass the connection string field by field.

I wonder if we should make a new `houston.backendSecretIsConnString` (or something), that would default to `true` (current behavior), but if overridden to `false`, then your change to the prisma configmap could pull the fields out one by one using the expected field names. Then the existing initContainer logic could stay exactly the same.

To launch for an azure customer, you would just create the houston secret yourself, then pass a config like:

```
houston:
  backendSecretName: "my-custom-secret"
  backendSecretIsConnString: false
```

It'd be cool if we didn't also have to signal that we created the secret in this way, but I'm not sure how else to know. WDYT?

ianstanton:
@schnie Yep, this should be fine. Should we give the option to provide user specified values?

ianstanton:
@schnie something like this?
```
            {{ if not .Values.houston.backendSecretIsConnString }}
            - name: PRISMA_DB_HOST
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.houston.backendSecretName }}
                  key: host
            - name: PRISMA_DB_PORT
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.houston.backendSecretName }}
                  key: port
            - name: PRISMA_DB_USER
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.houston.backendSecretName }}
                  key: user
            - name: PRISMA_DB_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.houston.backendSecretName }}
                  key: password
            {{ end }}
```

ianstanton:
It looks like if we supply `.Values.houston.backendSecretName`, the `houston-bootstrap-serviceaccount` is not created and the houston pod doesn't show up:

```
    message: 'pods "illmannered-pug-houston-547c987fcf-" is forbidden: error looking
      up service account astro/illmannered-pug-houston-bootstrapper: serviceaccount
      "illmannered-pug-houston-bootstrapper" not found'
```

https://github.com/astronomer/astronomer/blob/113b61f43be428faf01a68788fcbafa4aff35ad6/charts/astronomer/templates/houston/houston-bootstrap-serviceaccount.yaml#L4

schnie:
@ianstanton ah, I think we need to adjust the `if` statements on the Houston RBAC objects.

Check out the Houston `initContainer`. It's looking at these variables here - https://github.com/astronomer/astronomer/blob/master/charts/astronomer/values.yaml#L59-L61

We should dupe that logic over to the Houston RBAC stuff.

ianstanton:
@schnie It looks like one of the houston environment vars is looking at the `backendSecret` when we still want it to look at `astronomer-bootstrap`

https://github.com/astronomer/astronomer/blob/113b61f43be428faf01a68788fcbafa4aff35ad6/charts/astronomer/templates/_helpers.yaml#L172-L176

schnie:
@ianstanton good eye. This part of the configuration is consumed by the nodejs `knex` library, for running migrations, outside of `prisma`. 

https://github.com/astronomer/houston-api/blob/master/config/default.yaml#L23-L45

schnie:
@ianstanton I think this should stay as is, assuming it works. Houston would fail to boot up when it runs its migrations if it didn't like the config.

ianstanton:
@schnie this did not work because it was looking for `connection` key in our custom secret which is in this format:

`kubectl create secret generic my-secret --from-literal host="" --from-literal port="" --from-literal user="" --from-literal password=""`

charts/astronomer/templates/houston/houston-bootstrap-role.yaml
```diff
@@ -1,7 +1,7 @@
 ################################
 ## Houston Bootstrap Role
 #################################
-{{- if and .Values.global.rbacEnabled (and (not .Values.houston.backendSecretName) (not .Values.houston.backendConnection)) }}
+{{- if and .Values.global.rbacEnabled (and (not .Values.houston.airflowBackendSecretName) (not .Values.houston.airflowBackendConnection)) }}
```
sjmiller609: Is this change to repair an existing bug in our chart, or it related to this PR? Do we need only need this role for airflow metadata db secrets?

schnie: I think this is actually a fix. We never hit it because we don't set either, but yes, even if we don't merge the rest of this PR, we should PR this line right @andriisoldatenko ?

schnie: Well actually these 3 files.

ianstanton:
Closing until we do further performance testing with Azure DB PostgreSQL

PR #570, opened 2020-02-19
ElasticSearch reconfiguration

sjmiller609:
Closes: https://github.com/astronomer/issues/issues/784

- make es-master service headless by setting spec.clusterIP: None
- rename service because spec.clusterIP is immutable
- change ping schedule so that Linux default TCP keep alive of 2 hours does not cause data nodes to disconnect from the master once every two hours
- reconfigure discovery for elastic search 7
- enable readiness probe for elasticsearch masters
- enable startup probes for elasticsearch clients

**Checklist**

- [x]  Chart.yaml version and appVersion updated to match VERSION

### DIFF
--- a/charts/elasticsearch/templates/client/es-client-deployment.yaml
+++ b/charts/elasticsearch/templates/client/es-client-deployment.yaml
@@ -93,7 +93,7 @@ spec:
             fieldRef:
               fieldPath: metadata.name
         - name: DISCOVERY_SERVICE
-          value: {{ template "elasticsearch.fullname" . }}-discovery
+          value: {{ template "elasticsearch.fullname" . }}-headless-discovery
         {{- range $key, $value :=  .Values.common.env }}
         - name: {{ $key | upper | replace "-" "_" }}
           value: {{ $value | quote }}
--- a/charts/elasticsearch/templates/client/es-client-deployment.yaml
+++ b/charts/elasticsearch/templates/client/es-client-deployment.yaml
@@ -125,6 +125,17 @@ spec:
             path: /_cluster/health?local=true
             port: {{ .Values.common.ports.http }}
           initialDelaySeconds: 90
+        {{- if gt (.Capabilities.KubeVersion.Minor | atoi) 15 }}
+        # This feature is available in Kubernetes 1.16+
+        startupProbe:
+          httpGet:
+            # without local=true, we are waiting for this pod's
+            # es-master to report that the cluster is ready.
+            path: /_cluster/health
+            port: {{ .Values.common.ports.http }}
+          failureThreshold: 30
+          periodSeconds: 10
+        {{- end }}
         volumeMounts:
         - mountPath: /usr/share/elasticsearch/config/elasticsearch.yml
           name: config
--- a/charts/elasticsearch/templates/client/es-client-deployment.yaml
+++ b/charts/elasticsearch/templates/client/es-client-deployment.yaml
@@ -115,7 +115,9 @@ spec:
           protocol: TCP
         readinessProbe:
           httpGet:
-            path: /_cluster/health
+            # local: If true, the request retrieves information from the local node only.
+            # Defaults to false, which means information is retrieved from the master node.
+            path: /_cluster/health?local=true
             port: {{ .Values.common.ports.http }}
           initialDelaySeconds: 5
         livenessProbe:
--- a/charts/elasticsearch/templates/es-configmap.yaml
+++ b/charts/elasticsearch/templates/es-configmap.yaml
@@ -45,9 +45,12 @@ data:
       ping_schedule: "5s"
 
     discovery:
-      zen:
-        ping.unicast.hosts: ${DISCOVERY_SERVICE}
-        minimum_master_nodes: ${NUMBER_OF_MASTERS:2}
+      seed_hosts:
+        {{- $releaseName := $.Release.Name }}
+        {{- $replicas := .Values.master.replicas }}
+        {{- range $i, $e := until (int $replicas) }}
+        - {{ $releaseName }}-elasticsearch-master-{{ $i }}
+        {{- end }}
 
     # Disable scripting for security.
     script.allowed_types: none
--- a/charts/elasticsearch/templates/es-configmap.yaml
+++ b/charts/elasticsearch/templates/es-configmap.yaml
@@ -46,11 +46,7 @@ data:
 
     discovery:
       seed_hosts:
-        {{- $releaseName := $.Release.Name }}
-        {{- $replicas := .Values.master.replicas }}
-        {{- range $i, $e := until (int $replicas) }}
-        - {{ $releaseName }}-elasticsearch-master-{{ $i }}
-        {{- end }}
+        - {{ template "elasticsearch.fullname" . }}-headless-discovery
 
     # Disable scripting for security.
     script.allowed_types: none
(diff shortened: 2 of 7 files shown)


charts/elasticsearch/templates/es-configmap.yaml
```diff
@@ -38,6 +38,12 @@ data:
         enabled: ${HTTP_CORS_ENABLE}
         allow-origin: ${HTTP_CORS_ALLOW_ORIGIN}
 
+    # Setting the ping schedule explictly ensures that
+    # tcp_keep_alive configuration does not cause components
+    # to have their connections terminated.
+    transport:
+      ping_schedule: "5s"
```
sjmiller609: I have manually deployed this config to prod, and I'm assessing if it will fix the issue. I tested it for a few hours on staging first.

charts/elasticsearch/templates/es-configmap.yaml
```diff
@@ -38,10 +38,19 @@ data:
...
     discovery:
-      zen:
-        ping.unicast.hosts: ${DISCOVERY_SERVICE}
-        minimum_master_nodes: ${NUMBER_OF_MASTERS:2}
+      seed_hosts:
+        {{- $releaseName := $.Release.Name }}
+        {{- $replicas := .Values.master.replicas }}
+        {{- range $i, $e := until (int $replicas) }}
+        - {{ $releaseName }}-elasticsearch-master-{{ $i }}
+        {{- end }}
```
sjmiller609: I noticed deprecation notices in the logs for ping.unicast.hosts and minimum_master_nodes.

charts/elasticsearch/templates/master/es-master-statefulset.yaml
```diff
@@ -141,11 +141,13 @@ spec:
...
-        #     path: /_cluster/health?local=true
-        #     port: {{ .Values.common.ports.http }}
-        #   initialDelaySeconds: 5
+        readinessProbe:
+          httpGet:
+            # local: If true, the request retrieves information from the local node only.
+            # Defaults to false, which means information is retrieved from the master node.
+            path: /_cluster/health?local=true
+            port: {{ .Values.common.ports.http }}
+          initialDelaySeconds: 5
```
sjmiller609: The metrics exporter is dropping occasionally. In the logs, I have found it's getting 503 from the master check. I expect the master it's routed to is not ready, serving on the port. this will take the master out of the astronomer-elasticsearch service that metrics exporter queries when it's not serving HTTP. There were also timeout logs, however and this change does not address that aspect of the metrics exporter failing to get metrics.

sjmiller609: Previously, this was commented out. I suspect the initial reason for this might be not longer applicable with the addition of another change in this PR "publishNotReadyAddresses: true"

charts/elasticsearch/templates/es-configmap.yaml
sjmiller609: I think it's working

image

sjmiller609: current state can be checked here if you are an admin on prod cloud https://grafana.gcp0001.us-east4.astronomer.io/d/qfG4Q31mb/platform-overview-draft-2?orgId=1&refresh=10s&fullscreen&panelId=62&from=now-3h&to=now

charts/elasticsearch/templates/es-configmap.yaml
```diff
@@ -38,10 +38,15 @@ data:
...
+    # to have their connections terminated.
+    transport:
+      ping_schedule: "5s"
+
     discovery:
-      zen:
-        ping.unicast.hosts: ${DISCOVERY_SERVICE}
-        minimum_master_nodes: ${NUMBER_OF_MASTERS:2}
+      seed_hosts:
+        - {{ template "elasticsearch.fullname" . }}-headless-discovery
```
sjmiller609: this is how discovery is done for es 7+. I saw deprecation notices for the old 'zen' configs.

charts/elasticsearch/templates/master/es-master-service.yaml
```diff
@@ -4,15 +4,19 @@
...
+  name: {{ template "elasticsearch.fullname" . }}-headless-discovery
   labels:
     tier: logging
     component: {{ template "elasticsearch.name" . }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
     role: master
+  annotations:
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
```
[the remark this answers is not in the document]
jpweber: depending on the prometheus config it may never be able to scrape metrics from the "service". By making it headless there is not an IP for prometheus to hit based on the service. It would have to know to scrape the actual endpoints referenced via the service, rather than the service IP itself. Again all depends on prometheus is configured wether this will work or not.

charts/elasticsearch/templates/client/es-client-deployment.yaml
```diff
@@ -115,14 +115,27 @@ spec:
...
+        {{- if gt (.Capabilities.KubeVersion.Minor | atoi) 15 }}
+        # This feature is available in Kubernetes 1.16+
+        startupProbe:
+          httpGet:
+            # without local=true, we are waiting for this pod's
+            # es-master to report that the cluster is ready.
+            path: /_cluster/health
+            port: {{ .Values.common.ports.http }}
+          failureThreshold: 30
+          periodSeconds: 10
```
jpweber: I'm curious about this failure threshold on the startup probe. These values state that we expect ES to take five minutes to start up. Does it actually take that long? That seems long to me, but every configuration is different. In general this should be pretty close to the real startup time. (threshold * seconds = amount of time we expect it to fail)

sjmiller609: I'm not sure. The reason I chose 5 minutes was because that's the default timeout for helm install. I figured we wouldn't want the service to fail, but rather to wait all the way up to its deadline? I am OK to reconfigure this though.

charts/elasticsearch/templates/master/es-master-service.yaml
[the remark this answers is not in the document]
sjmiller609: We don't have a metrics endpoint being scraped on the master service. We are using an exporter that scrapes from the ES ingress nodes' (we are calling the ingress nodes 'clients') service:

https://github.com/astronomer/astronomer/blob/113b61f43be428faf01a68788fcbafa4aff35ad6/charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml#L58-L64

https://github.com/astronomer/astronomer/blob/113b61f43be428faf01a68788fcbafa4aff35ad6/charts/elasticsearch/templates/client/es-client-service.yaml#L7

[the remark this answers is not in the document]
sjmiller609: I don't know if this is the best way to do it, it does seem like a lot of hops for a metrics scrape. I think it would go like this on every scrape.

prom > exporter > client > master (then all the way back)

charts/elasticsearch/templates/es-configmap.yaml
sjmiller609: image

charts/elasticsearch/templates/client/es-client-deployment.yaml
jpweber: We can leave it for now. I was just more curious than anything.

sjmiller609:
@schnie we can either cherry pick this one into 0.12 or merge this one, for just the critical fix: https://github.com/astronomer/astronomer/pull/572

I will go ahead and merge in the above

PR #573, opened 2020-02-21
Update nginx-headers-configmap.yaml

samblackk:
Fix https://github.com/astronomer/issues/issues/790

Whitelists `www.segment.com` to resolve this issue in `astro-ui`: 
```
A cookie associated with a cross-site resource at http://segment.com/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.
```

### DIFF
--- a/charts/nginx/templates/nginx-headers-configmap.yaml
+++ b/charts/nginx/templates/nginx-headers-configmap.yaml
@@ -12,7 +12,7 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   # Policy was generated using https://report-uri.com/home/generate
-  Content-Security-Policy: "default-src 'self' *.{{ .Values.global.baseDomain }}; script-src 'unsafe-inline' 'unsafe-eval' *.{{ .Values.global.baseDomain }} cdn.jsdelivr.net cdn.astronomer.io cdn.metarouter.io www.google-analytics.com app.appzi.io js.stripe.com widget.intercom.io js.intercomcdn.com cdn.lr-ingest.io; img-src 'self' data: *; connect-src *.{{ .Values.global.baseDomain }} wss://*.{{ .Values.global.baseDomain }} e.metarouter.io api-iam.intercom.io wss://nexus-websocket-a.intercom.io; style-src 'unsafe-inline' *.{{ .Values.global.baseDomain }} cdn.jsdelivr.net fonts.googleapis.com; frame-src js.stripe.com; font-src *.{{ .Values.global.baseDomain }} fonts.gstatic.com js.intercomcdn.com data:; worker-src blob:"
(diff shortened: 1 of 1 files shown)


samblackk:
Good catch. Looks we need `cdn.segment` as well. Matching patterns found in this line: 
https://github.com/astronomer/astro-ui/blob/staging/index.ejs#L21

`src="https://cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js"`

samblackk:
@schnie Updated.

schnie:
@petedejoy @samblackk what about the actual event API? The actual script gets downloaded from `cdn.segment.com`, but I'm thining that segment may also have `api.segment.com` for the events.

schnie:
@samblackk for the `connect-src` do we want `cdn.segment.com` or `api.segment.com`?  I think this is the equivalent of `e.metarouter.io`, which we already have.

https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/

samblackk:
@petedejoy do you have any input for ☝️ i'm honestly not sure - i never integrated the metarouter items. we're not importing a script like we were before, just calling the `cdn.segment.com` item.

petedejoy:
I _think_ @schnie is right, but not 100% certain- any harm in trying the api url and seeing if it fixes the issue?

charts/nginx/templates/nginx-headers-configmap.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
...
+  Content-Security-Policy: "default-src 'self' *.{{ .Values.global.baseDomain }}; script-src 'unsafe-inline' 'unsafe-eval' *.{{ .Values.global.baseDomain }} cdn.jsdelivr.net cdn.astronomer.io cdn.metarouter.io cdn.segment.com www.google-analytics.com js.stripe.com widget.intercom.io js.intercomcdn.com cdn.lr-ingest.io; img-src 'self' data: *; connect-src *.{{ .Values.global.baseDomain }} wss://*.{{ .Values.global.baseDomain }} e.metarouter.io cdn.segment.com api-iam.intercom.io wss://nexus-websocket-a.intercom.io; style-src 'unsafe-inline' *.{{ .Values.global.baseDomain }} cdn.jsdelivr.net fonts.googleapis.com; frame-src js.stripe.com; font-src *.{{ .Values.global.baseDomain }} fonts.gstatic.com js.intercomcdn.com data:; worker-src blob:"
```
schnie:
```suggestion
  Content-Security-Policy: "default-src 'self' *.{{ .Values.global.baseDomain }}; script-src 'unsafe-inline' 'unsafe-eval' *.{{ .Values.global.baseDomain }} cdn.jsdelivr.net cdn.astronomer.io cdn.metarouter.io cdn.segment.com www.google-analytics.com js.stripe.com widget.intercom.io js.intercomcdn.com cdn.lr-ingest.io; img-src 'self' data: *; connect-src *.{{ .Values.global.baseDomain }} wss://*.{{ .Values.global.baseDomain }} e.metarouter.io api.segment.com api-iam.intercom.io wss://nexus-websocket-a.intercom.io; style-src 'unsafe-inline' *.{{ .Values.global.baseDomain }} cdn.jsdelivr.net fonts.googleapis.com; frame-src js.stripe.com; font-src *.{{ .Values.global.baseDomain }} fonts.gstatic.com js.intercomcdn.com data:; worker-src blob:"
```

schnie:
Done, merging. Looks like we removed appzi @samblackk?

PR #580, opened 2020-03-03
Configurable replicas for commander, houston

sjmiller609:
not urgent

configurable replicas for houston, commander

CI working is evidence that change works.


sjmiller609:
@schnie this one would be useful for 0.12 but not necessary. If we don't have it, I will just bump back up to 3 replicas after we upgrade to 0.12.

sjmiller609:
covered by the PDB PR

PR #584, opened 2020-03-06
Reduce cache time for index file

sjmiller609:
Thanks Ash and Tony for this fix suggestion!

This is to help with CDN propagation for new helm chart versions. I'm only doing it on the index file, because other files (artifacts, i.e. versioned helm tarballs) are intended immutable.


sjmiller609:
> Probably want `application/x-yaml` for the content type

can I just delete that header?

PR #585, opened 2020-03-06
Added email template

vparekh94:
Adds template for adding SMTP information


configs/starter.yaml
```diff
@@ -20,3 +20,15 @@ nginx:
   loadBalancerIP: ~
   # set to 'true' when deploying to a private EKS cluster
   privateLoadBalancer: false
+
+#################################
+## SMTP Setup
+#################################
+
```
schnie:
```suggestion
```

PR #586, opened 2020-03-06
Bump versions

sjmiller609:
I pushed this to the release branch already.

- bump all Chart.yaml to 0.12.0 (until we figure out subchart versioning automation)
- checked all image versions


schnie:
Do we need to bump these? This will cause every pod to reboot on upgrade. Leaving them back will only update the ones needed to be less disruptive.

schnie:
I was on a call when I reviewed this, disregard.

PR #589, opened 2020-03-10
Prometheus lifecycle api

jpweber:
This PR adds the ability to enable/disable the prometheus lifecycle api endpoint. The big reason for this to be added is so we can live reload configMap changes without having to restart the prometheus pod 

Related Prometheus docs
https://prometheus.io/docs/prometheus/latest/configuration/configuration/

**Checklist**

### DIFF
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus
-version: 0.12.0
+version: 0.12.1-alpha.1
 description: Helm chart to deploy the Astronomer Platform Prometheus module
 keywords:
   - astronomer
--- a/charts/prometheus/Chart.yaml
+++ b/charts/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus
-version: 0.12.1-alpha.1
+version: 0.12.0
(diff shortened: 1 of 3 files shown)


jpweber:
This PR resolves https://github.com/astronomer/issues/issues/910

charts/prometheus/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus
-version: 0.12.0
+version: 0.12.1-alpha.1
```
schnie: @jpweber @sjmiller609 what should we do here?

@jpweber for some background, this chart is bundled into the published umbrella chart. We aren't actually publishing this chart directly.

Bumping these versions as part of commits always feels off to me. :thinking:

jpweber: got it. I was following the directions in the PR template, but I had to ask Steven about it so I might have gotten it wrong still. Just wasn't clear what I should be bumping.

sjmiller609: it honestly does not matter very much until the "approve release" button is pressed.

schnie: So when would this get set to something normal? I feel like we should just pull this out for now, what do ya'll think?

sjmiller609:
```suggestion
version: 0.12.0
```

PR #590, opened 2020-03-10
In CI, run all tiers in the same VM

verdverm:
This changes the CircleCI test grid to

- have tests for each k8s versions (currently 5)
- remove platform components dimension
- increase VM size to 2xlarge
- runs entire platform in Kind at once


.circleci/config.yml
```diff
@@ -41,212 +41,62 @@ jobs:
...
-          astronomer-tags: "logging"
-  kubed-1-12-10:
-    machine:
-      image: ubuntu-1604:201903-01
-    environment:
-      KUBE_VERSION: v1.12.10
-    steps:
-      - helm-install:
-          astronomer-tags: "kubed"
+          astronomer-tags: "platform postgresql monitoring logging kubed keda"
```
sjmiller609: nice!

sjmiller609: a little history - previously we were doing it separately because we did not have an option to run in bigger VMs.

.circleci/config.yml.j2
```diff
@@ -32,44 +32,19 @@ jobs:
       - run:
           name: Release the Astronomer chart
           command: |
-            printf "%s" $${GCP_TOKEN} > /tmp/gcs_token.json
-            gcloud auth activate-service-account --key-file=/tmp/gcs_token.json
```
jpweber: looking at the changset for these two lines I don't understand. gcloud auth stuff on one side and setting py environment on the other. For my own knowledge could you explain this one?

bin/run-ci
```diff
@@ -74,14 +75,15 @@ sleep 5
...
   NOW=$(date +%s)
   DURATION="$(($NOW-$START))"
-  echo "${DURATION} seconds have elapsed, timeout at 400 seconds"
-  if [ $DURATION -gt 400 ];then
+  echo "${DURATION} seconds have elapsed, timeout at $MAX_SECONDS seconds"
+  if [ $DURATION -gt $MAX_SECONDS ];then
     HELM_CODE=1
     RC=1
-    echo "Error (timeout): giving up after 400 seconds"
+    echo "Error (timeout): giving up after $MAX_SECONDS seconds"
```
jpweber: Dig this change here, nice cleanup

.circleci/config.yml.j2
verdverm: This was actually changed a month ago in #547 but only half merged, and now second half here. It's unneeded now

verdverm:
We do want to hold off on merging this PR.

I am going to test the `xlarge` VM size and see if we can use half the resources / cost

PR #593, opened 2020-03-11
Set postgresqlEnabled: false by default astronomer/issues#958

shmanu017:
**Checklist**

- [X]  Chart.yaml version and appVersion updated to match all VERSION

- Related to https://github.com/astronomer/issues/issues/958

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -33,7 +33,7 @@ global:
 
   # Enable default postgresql database.
   # This is not recommended for production.
-  postgresqlEnabled: true
+  postgresqlEnabled: false
 
   # Enables a node exporter DaemonSet for node-level metrics
   nodeExporterEnabled: true


schnie:
The reason this was added was so that users could simply run `helm install astronomer` and be up and running. This reverts the UX that we added in 0.12.

We could keep it as is, and update the _upgrade_ docs to include adding a `--set global.postgresEnabled=false` flag or this new value to the `config.yaml`. Or we could merge this change, and update our "getting started" docs to add `--set global.postgresEnabled=true` to the `helm install` command.

@vishwas-astro @vparekh94 any thoughts?

vparekh94:
@schnie yeah I talked with Vikram about this issue a bit, but I think in cloud environments (which is where we are pushing most enterprise users), the ideal way is to turn it off

- This _could_ be helpful for on-prem environments, but we're trying to push enterprise users away from on-prem.
- No one has trouble getting db creds in the cloud, not requiring this doesn't speed things up
- People would rather go with a managed db from the start than have to think through a db migration or swapping database secrets afterwards
- The UX when a deployment fails for whatever reason can be frustrating; subsequent "helm installs" will fail because the `astronomer-bootstrap` secret already exists

I think the second suggestion you have is the ideal UX:
` or we could merge this change, and update our "getting started" docs to add --set global.postgresEnabled=true to the helm install command.`

paolaperaza:
@sjmiller609 @schnie This change wasn't included in 0.12.1 was it?

schnie:
Doesn't look like it was cherry-picked over for it - nope. We can include it for the next patch release if we want, or def will go in 0.13.

sjmiller609:
we need to merge the whole release-0.12 into master and handle the merge as intended

schnie:
@sjmiller609 what do you mean? If we want to release this with 0.13, we don't have to do anything here. If we want to release it with a 0.12.2, we cherry-pick it over to `release-0.12`. I don't think we have to merge `release-0.12` to `master` for this.

sjmiller609:
hey I think I misunderstood your previous comment - I thought you meant did not exist in master branch, but did exist in release branch. I think we have some commits in release branch and not master that need to be handled.

vishwas-astro:
@schnie @ianstanton @sjmiller609 I thought we needed this for EKS terraform to start using 0.12.1. Not so?

paolaperaza:
@vishwas-astro Should add this to the board for v0.13 w/ label, if it missed v0.12.1?

PR #596, opened 2020-03-13
Add additional domain to policy for Segment API

ryanahamilton:
Another!

Adds `api.segment.io` to `connect-src` within `Content-Security-Policy`.

Resolves astronomer/issues#990.


schnie:
Ain't this fun?! :man_facepalming: 

@verdverm any idea what happened in these tests? Is this related to the recent merge? Looks like we're getting some errors that I wouldn't expect to see for this kind of change.

PR #597, opened 2020-03-13
Fix CI

sjmiller609:
I think this one broke CI just need a reconfiguration https://github.com/astronomer/astronomer/pull/593


sjmiller609:
only merge if CI passes.

schnie:
@sjmiller609 looks like this one is failing

PR #598, opened 2020-03-16
Prometheus alert rules fixes after testing in stage

jpweber:
I found some areas where alerts and/or recording rules did not have the correct values for job names  and because of release name not being part of the job name. I also fixed up the alert annotations so they match how our slack messages are displayed.

### DIFF
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -144,6 +144,59 @@ data:
             regex: ([^:]+)(?::\d+)?;(\d+)
             replacement: $1:$2
             target_label: __address__
+
+          {{- if not .Values.global.singleNamespace }}
+            # This allows us to collect all the metrics we want in the cloud environment
+            # does not limit to just the astronomer release namespace
+          - source_labels: [__meta_kubernetes_endpoint_address_target_kind, __meta_kubernetes_endpoint_address_target_name]
+            separator: ;
+            regex: Node;(.*)
+            target_label: node
+            replacement: ${1}
+            action: replace
+          - source_labels: [__meta_kubernetes_endpoint_address_target_kind, __meta_kubernetes_endpoint_address_target_name]
+            separator: ;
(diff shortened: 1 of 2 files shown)


jpweber:
This configuration can be viewed in stage. I've also hand jammed in some Grafana dashboards to reflect the new data we're gathering. They are all in a kubernetes folder in Grafana.

charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -285,11 +357,6 @@ data:
             regex: (.+)
             target_label: __metrics_path__
             replacement: /api/v1/nodes/${1}/proxy/metrics
-        metric_relabel_configs:
```
schnie: @jpweber what happens here if there are multiple platforms running under a single cluster? Will we start scraping all airflows, regardless of what platform deployed them?

jpweber: I think this might be confusing because of how the diff is displayed here. This removal is not part of the airflow job, but part of the `kubernetes-nodes` and `kubernetes-nodes-cadvisor` job. But it does say if we are not in a single namespace collect all the kubelet information. Which provides container metrics and some measure of node metrics.

jpweber: An alternate method here is to had some kind of `isCloud` flag and only open it up if we are in the cloud rather than relying on the single namespace variable.

schnie: Ah true, I misread the diff! :grimacing:  This is prob good then. I could see a separate flag, but this might be good for now - up to you. If we go separate flag, you could give it a descriptive name, rather than `isCloud`. Generally we've followed the same pattern elsewehre.

jpweber: If you're ok with it I think this can provide useful data to support if we get to the point someday where support can get access to this data in enterprise clusters to help debug things.

PR #601, opened 2020-03-18
Dropping things that start with id was causing duplicate records

jpweber:
With removing the restriction of astronomer namespace only it allows us to get node specific container metrics. the id label is part of this information, by dropping that label these were causing non-unique metrics because that id label was needed to keep them unique.

These needs to go in before we push 0.12.1 to prod

solves https://github.com/astronomer/issues/issues/1005

### DIFF
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -426,8 +426,8 @@ data:
             target_label: systemd_service_name
             replacement: '${1}'
 
-            # Drop some unnecessary labels
-          - regex: "^(id|name)"
+            # Drop some unnecessary labels.
+          - regex: "^name"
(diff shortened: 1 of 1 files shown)


schnie:
@jpweber is this already on master? Typically merge things to `master` and cherry-pick out to the release branch if its going to get shipped.

jpweber:
Because of previous problems I didn't want to trust master honestly. But also since this was updated a change that is already in the 0.12 release branch I branched off of that.

schnie:
I'm just concerned that commits like this wouldn't make it to `release-0.13` and beyond if its not on master also. So if we merge this, we'll want to cherry-pick it over to `master` then right?

jpweber:
Fortunately this is a super trivial change, so we have need to change which branch it merges to just say so and I can remake the PR

PR #604, opened 2020-03-18
Blackbox exporter addition to monitoring stack

jpweber:
This is a kind of big PR

What this adds

blackbox exporter chart
associated prometheus scrape configs
network policies to allow scraping from blackbox exporter
updates to global configuration because of a new chart in the platform

Things that need review

chart versioning. Should this be 0.12.2? or 0.12.1? or even wait for 0.13?
Things that are still in process
I don't love how I'm handling the enabling and disabling of different targets and network policies so I'm open to suggestions.


jpweber:
cancel this. Let me clean this up and make a new PR. It is too much of a mess right now.

PR #608, opened 2020-03-19
Fix CI - turn on postgresql

sjmiller609:
**Checklist**

- [ ]  For each modified subchart, charts/subchart/Chart.yaml version updated to VERSION


sjmiller609:
moving this to https://github.com/astronomer/astronomer/pull/609

PR #617, opened 2020-03-25
Patches from 0.12 release

sjmiller609:
The changes that made it into release-0.12 but we forgot to add into master branch. I checked out a new branch from the master branch and simply 'git pull origin release-0.12' to ensure all updates were added. I resolved the small amount of merge conflicts, it wasn't much. I think we just missed 4 changes.

changes from 0.12:

- commander and houston replicas configurable
- commander liveness probe adjustment
- small scape configuration adjustment
- image versions

Other:

- Remove PR template: I don't want to ask people to manage the version, I believe it's not necessary to worry about
- some bumps in chart.yaml versions. I just bumped any conflict to 0.13.0 except the top level which I set to 0.13.0-alpha.1, which I figure will be the next published version


Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: astronomer
-version: 0.12.0
-appVersion: 0.12.0
+version: 0.13.0-alpha.1
+appVersion: 0.13.0-alpha.1
```
sjmiller609: I figure set top level to this for now

charts/astronomer/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: astronomer
-version: 0.12.0
+version: 0.13.0
```
sjmiller609: we discussed that these don't matter very much, so if there was a merge conflict on one I just bumped to 0.13.0 and we can leave it all the way until the next patch version if we want.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -59,8 +59,7 @@ data:
         # This version number controls the default Airflow chart version that will be installed
         # when creating a new deployment in the system. This is also used to ensure all
         # child airflow deployments are kept up to date and on the latest version.
-        version: 0.12.0
-
+        version: 0.12.1
```
sjmiller609: this includes fix for api

PR #619, opened 2020-03-27
Ingress for monitoring

jpweber:
This PR creates ingress sources and modified network policies for prometheus and Alertmanager 
So they can be exposed (behind Houston auth) over the internet. 

Today in order to access these things in prod you must port forward with `kubectl`. Having this ability will enable the ability to add helpful links/buttons to our alert messages to silence alerts, go go straight to prometheus to see the raw data that triggered the alert etc.

### DIFF
--- a/bin/run-ci
+++ b/bin/run-ci
@@ -35,8 +35,10 @@ sleep 10
 # kill $FORWARD_PID
 
 # Test monitoring/alerting ingress - This is a stopgap until it can move in to full testing suite
-curl -f -s -o /dev/null -H "Host: alertmanager.local.astronomer-development.com"  http://172.17.0.1/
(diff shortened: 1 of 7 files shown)


jpweber:
resolves issue https://github.com/astronomer/issues/issues/1044

jpweber:
@sjmiller609 Are you looking for just a comment that echoes out for a human to do it? Or add in commands to actually test those endpoints?

sjmiller609:
> @sjmiller609 Are you looking for just a comment that echoes out for a human to do it? Or add in commands to actually test those endpoints?

a curl command or similar to test that the endpoint does the right thing and exists instead of 404 or whatever asfasdfasdf.basedomain would return

jpweber:
@sjmiller609 
curl checks for the new ingresses have been added and ran successfully in CI

PR #620, opened 2020-03-28
Add pod disruption budgets

sjmiller609:
The point of this PR is to make replicas configuration effective for HA during voluntary disruptions for the workloads that are already HA (or were very easy to make HA). There are other parts of the platform that are not HA yet and will take a bit more work (registry, prisma).

If workloads are tolerant to voluntary disruptions, taking less than 2 minutes to move, they are candidates to run on spot instances, and this can save 80% of compute cost.

Add HA for:
- astro-ui
- nginx-default-backend

Add PDB for:
- houston
- commander
- astro-ui
- es-client
- es-master
- es-data
- nginx
- nginx default backend
- prometheus

Default to HA configuration for:
- houston
- commander
- nginx

Add to script to enable local testing of cordon and drain

Video of testing: https://www.youtube.com/watch?v=BZPhy-PI4bw
- part 1: show cordon, drain with HA on
- part 2: show cordon, drain with HA off

A good chunk of the work for this issue: https://github.com/astronomer/issues/issues/1052

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 config.yaml
 config.yml
+dev-config.yaml
 !.circleci/config.yml
 certs
 /secrets
--- a/README.md
+++ b/README.md
@@ -90,6 +90,34 @@ export KUBE_VERSION='v1.16.3'
 bin/reset-local-dev
 ```
 
+#### Locally test HA configurations:
+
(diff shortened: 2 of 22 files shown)


sjmiller609:
> LGTM, would we ever want to change those percentages? Just wondering if those should be variables too.

good call

PR #624, opened 2020-04-01
Postgres metrics exporter

sjmiller609:
Closes: https://github.com/astronomer/issues/issues/950
Requires first: https://github.com/astronomer/ap-vendor/pull/13

The point of this component is to serve Postgres-related metrics data.

- [x] reviewed code for [existing helm chart](https://github.com/helm/charts/tree/master/stable/prometheus-postgres-exporter)
There were a few things missing:
    - support for connection string style DB auth
    - pod disruption budget
    - effective probes
- [x] **copy / pasted existing helm chart (consolidated to first commit)**
- [x] made it work with DB connection string
- [x] HA, disruption budgets by default
- [x] vendor image
- [x] effective probes
- [x] add network policies
- [x] add scrape config, intention is prom to scrape ClusterIP of service (HA exporter service)
- [x] add to exporter query for detecting 'version' issue in 'Deployment' table
- [x] add corresponding alarm to prometheus
- [ ] (Optional) add postgres dashboard in grafana


sjmiller609:
@schnie opening this before it's done to get early feedback. i.e. let me know if this whole plan is OK or not

charts/prometheus-postgres-exporter/templates/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if .Values.rbac.pspEnabled }}
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
```
jpweber, line 3: Not an issue but an FYI since we do not have pod security policies enabled at the cluster level this fill not have an effect.

charts/prometheus-postgres-exporter/templates/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if .Values.rbac.pspEnabled }}
```
jpweber, line 1: optional choice here, but as part of the bigger psp changes I added a global var to our config called `pspEnabled` true|false (no rbac in the name) . You could tie in to that one rather than the local one here.

charts/prometheus-postgres-exporter/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ServiceAccount
+metadata:
+  name: {{ template "prometheus-postgres-exporter.serviceAccountName" . }}
+  labels:
+    app: {{ template "prometheus-postgres-exporter.name" . }}
+    chart: {{ template "prometheus-postgres-exporter.chart" . }}    
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+{{- end -}}
+
```
andriisoldatenko, line 12:
```suggestion
  
  
```

andriisoldatenko: missing new line.

charts/prometheus-postgres-exporter/templates/podsecuritypolicy.yaml
sjmiller609: i'll follow up on this point before merge, thank you

sjmiller609: Hey actually I don't think it's on master yet, so I'll omit that for now

PR #625, opened 2020-04-03
Migrate to prisma 2

andriisoldatenko:
~~related PR https://github.com/astronomer/db-bootstrapper/pull/16~~ MERGED
~~related PR https://github.com/astronomer/houston-api/pull/422~~ MERGED


sjmiller609:
andrii said WIP, so changing title for now

charts/astronomer/values.yaml
```diff
@@ -22,25 +22,21 @@ images:
     tag: 3.11.5-4
     pullPolicy: IfNotPresent
   houston:
-    repository: astronomerinc/ap-houston-api
-    tag: 0.19.9
+    repository: asoldatenko/ap-houston-api
```
danielhoherd, line 25: Are we shipping code that references non-astronomer repositories?

charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -58,6 +63,13 @@ spec:
                 secretKeyRef:
                   name: astronomer-bootstrap
                   key: connection
+            # we need to hardcode it to be able to upgrade from prisma1 to prisma2
+            - name: SCHEMA_NAME
+              value: "houston$default"
```
danielhoherd, line 68: What is this syntax? what is `$default`? Can `default` be renamed to be more specific? (EG: default_namespace, default_podname)

jpweber: That is a Postgres-ism. Its literally what you would use in a cli command for the database as well.

charts/astronomer/values.yaml
jpweber: I think this is what @andriisoldatenko is trying to get at here based on the slack comment, its a special case different image to test from the houston feature branch that contains this. Because of how things are wired up, In order for us to test from one of our docker registries the prisma2 changes have to be merged to main branch on houston. Which we don't want yet because we want to test. 

Feel free to correct me if I'm wrong here andrii

jpweber: Although, I would think we should be able to push to our registry without going through CI. Maybe that isn't true.

danielhoherd: We definitely could push this to our own registry. I don't know if it makes that much of a difference though, and this might actually be better since it won't be exposed to folks browsing our docker hub tags (unlikely, but possible.)

charts/astronomer/templates/houston/api/houston-deployment.yaml
andriisoldatenko: @danielhoherd it's old prisma1 artifact... so prisma2 supports custom schema_name, we need this change to have ability to migrate existing platform to new platform with prisma2 in safe way.

andriisoldatenko:
> I think it needs patch to houston init container that waits on prisma, since that's no longer applicable. I think CI is valid failure, you can try locally with run-ci script if you want.

@sjmiller609 done.

andriisoldatenko:
> andrii said WIP, so changing title for now

PTAL!

charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -48,6 +48,11 @@ spec:
           resources:
 {{ toYaml .Values.houston.resources | indent 12 }}
           env:
+            - name: DATABASE_URL
+              valueFrom:
+                secretKeyRef:
+                  name: astronomer-bootstrap
```
kushalmalani, line 54: Can we use the secret name as "{{ template "houston.backendSecret" . }}" instead of hardcoding the name?

charts/astronomer/templates/houston/api/houston-backend-secret.yaml
```diff
@@ -14,7 +14,8 @@ metadata:
 type: Opaque
 data:
   {{- if .Values.houston.backendSecretConnection }}
-  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db) | b64enc | quote }}
+  # in prisma1 has hardcoded schema, so for now to migrate to prisma2 we need to hardcode it too.
+  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db) | b64enc | quote }}
```
adam2k, line 18: Does this mean in the future we need a follow up ticket to change the connection here?  Is this only necessary for the migration process?

adam2k (approved):
This LGTM as long as @kushalmalani's comment is resolved and we have a ticket to eventually replace the temporary test houston-api images back to the astronomer houston api images once the testing is completed.

sjmiller609 (commented):
How do we handle prisma 1 to prisma 2 data migration?

kushalmalani (approved):

sjmiller609 (approved):
I approve, but FYI I expect merging this will break dev cloud since migration steps required

charts/astronomer/templates/houston/migrations/hosuton-migrations-job.yaml
```diff
@@ -0,0 +1,96 @@
...
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": pre-upgrade,post-install
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  template:
+    metadata:
+      labels:
+        tier: astronomer
+        component: houston
```
sjmiller609, line 24:
```suggestion
        component: houston-db-migrations
```

sjmiller609: otherwise it can match unintended selectors like the pod disruption budget

PR #626, opened 2020-04-05
Add configuration to re-generate CA on each upgrade

sjmiller609:
This patch, in the form of an optional configuration to 0.12, allows upgrades from before 0.10 to work without first delete --purge of the whole platform. The reason the --purge was required before is that since 0.10, there is a new secret installed by the helm hook. Since the helm hook runs by default only on installation, then delete --purge + install is performed to get the jwt token resource to create.

The solution proposed by this change request is to provide an optional configuration that configures this helm hook to run on each upgrade, in addition to installation. Since the helm template generates a self-signed certificate on each application, it will replace Houston's certificate authority on each helm upgrade.

One example for why this is valuable is that it eliminates the requirement for users to update their DNS records after upgrading Astronomer. This is often a challenge in the corporate environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ config.yml
 !.circleci/config.yml
 certs
 /secrets
+**kubeconfig**
 
 # IDEs
 .idea/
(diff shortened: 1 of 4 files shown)


sjmiller609:
@ianstanton @shmanu017 this is the JWT token issue discussed during our work on 0.12 upgrade testing

PR #630, opened 2020-04-06
Add upgrade script

sjmiller609:
**Changes**

- delete 0.11 pre-upgrade script, replace with automated upgrade script generic to latest version of Astronomer
- update readme for generic version upgrade guidance

**What I like about this script:**
- it works to upgrade from any version of Astronomer to latest (as of 0.12, tested back to 0.9.6)
- takes a lot less time (30 min vs 2-4 hours) to upgrade users

**What I don't like about this script:**
- There is not presently a test to automatically validate that future versions can be upgraded using it
- There are a few wild workarounds included in the script, and those should be removed after fixing the upgrade behavior of Astronomer or learning how to do it better
- It's bash and I think that usually ends up hard to read / having strange environment-specific problems. bash also is usually hard to test properly, but I think smoke tests are appropriate in this case (i.e. just apply it and did it work yes/no). 

**What is the future for this?**

The plan is to be the _Astronomer Operator_, similar to how istio has a command line for installation and updates that they name an operator. [Istio defines the operator](https://istio.io/blog/2019/introducing-istio-operator/) as "Kubernetes operators provide a pattern for encoding human operational knowledge in software and are a popular way to simplify the administration of software infrastructure components."

Astronomer is immediately benefiting by encoding our operational knowledge in this script today. I expect getting into the habit of using a script for upgrades, and in each case adding to the same body of work, will serve us well to persist our operational knowledge in one place. This enables new hires to perform upgrades, and ultimately for users to upgrade themselves. We will likely migrate this to a different language after it's a cleaner procedure / that is justified.

This can be added into the Astro CLI.
```
astro cluster upgrade
```

**What is the short-term plan for this?**

The Astronomer upgrade squad will watch users follow the directions in the README, and take notes in each case. After a large enough sample size, we can start to leave users unattended. We will need to add a test to CI that confirms the upgrade script works from the most recently published version of Astronomer to the version of Astronomer being validated by CI.

**How does this script work?**

- check user's environment looks valid
- query user to backup their database
- fetch the latest airflow chart and astronomer chart versions
- ask for confirmation before initiating upgrade, displaying the versions
- perform fernet key persistence procedure (0.11 required)
- upgrade Astronomer using 'helm upgrade' with some configs: all houston jobs disabled, reset the certificate authority (0.10 required)
- helm delete --purge all the airflow deployments, then re-install with the latest airflow chart version
- Proxy with 'kubectl exec' through the Prisma pod to update the Astronomer DB 'version' field in 'Deployments' table (this is the wild workaround I am talking about). Surprisingly, I did not find any easier way to get the upgrade to work in practice.

Keep in mind the whole point here is to persist our operational knowledge like the 0.10 requirement to replace the CA and the 0.11 requirement to save fernet keys. Strange workarounds can also live here but temporarily until the platform can be changed or better options are available.


sjmiller609:
> Looks great.
> 
> Would calling `mutation upgradeDeployment` help avoid the hacks to prisma? https://github.com/astronomer/houston-api/blob/master/src/resolvers/mutation/upgrade-deployment/index.js#L6-L39
> 
> I think we started talking about it earlier, but I don't know if I caught it.

the issue I had with this is that I want to call helm install or upgrade exactly once per helm release. when doing more than once, i.e. reinstall, then call upgradeDeployments, I have found race conditions to display.

PR #632, opened 2020-04-07
Add example IKS fluentd config

ianstanton:
- Related to https://github.com/astronomer/issues/issues/664
- Related to https://github.com/astronomer/issues/issues/905


sjmiller609:
CI blocked, but this is a documentation-only change, so I am good with it.

PR #633, opened 2020-04-08
fix flags for es-exporter

andriisoldatenko:
TODO:
- [x] fix elasticsearch-exporter


charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
```diff
@@ -56,12 +56,12 @@ spec:
           image: {{ template "exporter.image" . }}
           imagePullPolicy: {{ .Values.images.exporter.pullPolicy }}
           command: ["elasticsearch_exporter",
-                    "-es.uri=http://{{ .Release.Name }}-elasticsearch:{{ .Values.common.ports.http }}",
-                    "-es.all={{ .Values.exporter.es.all }}",
-                    "-es.indices={{ .Values.exporter.es.indices }}",
-                    "-es.timeout={{ .Values.exporter.es.timeout }}",
-                    "-web.listen-address=:{{ .Values.exporter.service.httpPort }}",
-                    "-web.telemetry-path={{ .Values.exporter.web.path }}"]
+                    "--es.uri=http://{{ .Release.Name }}-elasticsearch:{{ .Values.common.ports.http }}",
```
andriisoldatenko, line 59: docs https://github.com/justwatchcom/elasticsearch_exporter

andriisoldatenko: also i can add {{ if }} to stay with this change, i was wondering do we need to export specific index?

charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
```diff
@@ -56,12 +56,12 @@ spec:
           image: {{ template "exporter.image" . }}
           imagePullPolicy: {{ .Values.images.exporter.pullPolicy }}
           command: ["elasticsearch_exporter",
-                    "-es.uri=http://{{ .Release.Name }}-elasticsearch:{{ .Values.common.ports.http }}",
-                    "-es.all={{ .Values.exporter.es.all }}",
-                    "-es.indices={{ .Values.exporter.es.indices }}",
```
ashb, line 61: I guess we don't use these values, but we could do

```
{{if .Values.exporter.es.indices }}
                     "-es.indices=", 
{{ endif }}
```

(or whatever the text/template syntax is for ifs). Do we need to?

ashb:
Can you update this PR title -- it's doing rather more than just flags for es-exporter now.

PR #637, opened 2020-04-10
Add namespace var to upgrade_version_in_astro_db

ianstanton:
- This function will fail if platform namespace is not named "astronomer"
- Use namespace var to handle any namespace name
- Related to astronomer/issues#1123


sjmiller609:
important in the short term and not related to CI, so merging.

PR #639, opened 2020-04-10
Update values.yaml

andriisoldatenko:


ashb:
What's the interaction between this and ap-vendor. Why do we need too update version in two places

andriisoldatenko:
@sjmiller609 wdyt?

schnie:
ap-vendor is what pushes the new 0.30.0 tag of the image. This is what pulls that tag into the default deployment. We've talked about automating this PR though, just haven't gotten to that part of the pipelines yet.

schnie:
I think this PR should also have gone to master and cherry-picked to the release branch, but we can cherry-pick this one to master once we merge it.

PR #640, opened 2020-04-11
Add list of other supported images

kaxil:
This PR adds `-number` images and **buster** images too

Issue: https://github.com/astronomer/issues/issues/1119


kaxil:
CircleCI failure says:

```
This job has been blocked because no user seats are available on your plan. Please purchase more user seats to continue building.
```

schnie:
Thanks. This may get crazy. Hopefully we can find a better way to enforce this soon.

ashb:
Yeah, we need to come up with a better plan for this _really_ soon, espeically when it comes to enterprise customers that won't get platform "updates" like this in a timely manner but will want the ability to use new versions of Airflow as we release it.

PR #642, opened 2020-04-13
Added Ingress certificate expiring alert

jpweber:
Also updated nginx 5xx alert while I was in there to fix a 4/5 typo.

### DIFF
--- a/charts/prometheus/templates/prometheus-alerts-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-alerts-configmap.yaml
@@ -1299,7 +1299,7 @@ data:
               10 minutes."
 
         - alert: NginxIngressHigh5XXRate
-          expr: sum by (ingress) (rate(nginx_ingress_controller_requests{status=~"^4.."}[2m])) / sum by (ingress) (rate(nginx_ingress_controller_requests{ingress!=""}[2m])) * 100 > 1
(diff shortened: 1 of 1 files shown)


charts/prometheus/templates/prometheus-alerts-configmap.yaml
```diff
@@ -1299,7 +1299,7 @@ data:
               10 minutes."
 
         - alert: NginxIngressHigh5XXRate
-          expr: sum by (ingress) (rate(nginx_ingress_controller_requests{status=~"^4.."}[2m])) / sum by (ingress) (rate(nginx_ingress_controller_requests{ingress!=""}[2m])) * 100 > 1
+          expr: sum by (ingress) (rate(nginx_ingress_controller_requests{status=~"^5.."}[2m])) / sum by (ingress) (rate(nginx_ingress_controller_requests{ingress!=""}[2m])) * 100 > 1
```
sjmiller609, line 1302: I think this change is correct, but I would like to mention this will be noisy because the "waiting for airflow to start up" is 503.

jpweber: yeah all good, I wanted to at least make this little fix because it was flat wrong before. But not take it anything in this PR because this pr wasn't about that. I agree we should limit the scope of the alert to not look at customer namespaces as a separate PR.

PR #644, opened 2020-04-14
Add secret block back into Houston environment

schnie:
@ashb added this block a while ago, but I think it was dropped during a refactor with the `houston_environment` helper.

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -171,10 +171,19 @@ azure:
 {{- end }}
 
 {{- define "houston_environment" }}
+{{- /* Dynamically created envs */ -}}
 {{- range $i, $config := .Values.houston.env }}
 - name: {{ $config.name }}
   value: {{ $config.value | quote }}
-  {{- end }}
+{{- end }}
+{{- /* Dynamically created secret envs */ -}}
+{{- range $i, $config := .Values.houston.secret }}
+- name: {{ $config.envName }}
+  valueFrom:
+    secretKeyRef:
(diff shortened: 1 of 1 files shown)


schnie:
cc @vparekh94 @paolaperaza @petedejoy related to slack convo

ashb:
Can we get this in 0.13 or an 0.12 point rel?

vishwas-astro:
@ashb lets get this into 0.13, deployed on stage today, tested etc and if all goes well cherry pick it into the 0.12 branch as well for Zapier

charts/astronomer/templates/_helpers.yaml
```diff
@@ -171,10 +171,19 @@ azure:
...
-  {{- end }}
+{{- end }}
+{{- /* Dynamically created secret envs */ -}}
+{{- range $i, $config := .Values.houston.secret }}
+- name: {{ $config.envName }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ $config.secretName }}
+      key: {{ default "value" $config.secretKey }}
+{{- end }}
```
andriisoldatenko, line 186: only concern, this block will be in all houston related pods e.g.: upgrade-deployments/cleanup etc.
wdyt to extract to new `{{ -define ` block?

schnie: I think that might be desirable, since they also pull configuration from the same Houston `config` module. I look at these new jobs as being _exactly_ the same Houston proper, just doing handling specialized things in the backgroud.

schnie:
@sjmiller609 want to cherry-pick this one?

PR #650, opened 2020-04-17
Prevent duplicate labels of kubernetes_executor

jpweber:
pods have the labels `kubernetes-executor` and `kubernetes_exector` when those get turned in to metric labels the dash becomes and underscore which causes duplicated labels. Prometheus must have unique labels per metric.


jpweber:
This change is already live I prod. This PR gets it in to the codebase so it doesn't get removed.

PR #655, opened 2020-04-23
Add ap-airflow 1.10.10 Images to Supported list

kaxil:
Issue: https://github.com/astronomer/issues/issues/1143
Related to https://github.com/astronomer/ap-airflow/pull/66

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -81,6 +81,18 @@ data:
         - version: 1.10.7
           channel: stable
           tag: 1.10.7-8-buster-onbuild
+        - version: 1.10.10
+          channel: stable
+          tag: 1.10.10-buster-onbuild
+        - version: 1.10.10
+          channel: stable
+          tag: 1.10.10-alpine3.10-onbuild
+        - version: 1.10.10
+          channel: stable
+          tag: 1.10.10-1-buster-onbuild
+        - version: 1.10.10
+          channel: stable
+          tag: 1.10.10-1-alpine3.10-onbuild
         # - version: 1.10.6
         #   channel: edge
         #   tag: 1.10.6-alpine3.10-onbuild


kaxil:
Test is failing because CircleCI JOB is unable to run: 

```
This job has been blocked because no user seats are available on your plan. Please purchase more user seats to continue building.
```

sjmiller609:
@schnie can you add kaxil to user seats?

schnie:
@sjmiller609 are we worried about letting folks push 1.10.10 images? With this change merged, a customer can create a new deployment and it's initial schema would be created using 1.10.10 right? What would happen if they push a 1.10.7 image up after?

kaxil:
Oh, is that so? I thought by default it would use the oldest version i.e. 1.10.5

schnie:
We switched it to do that, but that introduced some new errors in Cloud rollout. I forget the specifics, @sjmiller609 might know more.

@kaxil would Airflow blow up if a lower version was pushed? Ultimately we need the version awareness in some fashion.

kaxil:
> We switched it to do that, but that introduced some new errors in Cloud rollout. I forget the specifics, @sjmiller609 might know more.
> 
> @kaxil would Airflow blow up if a lower version was pushed? Ultimately we need the version awareness in some fashion.

Can we use the solution mentioned by @ashb at https://astronomerteam.slack.com/archives/C02NZELT2/p1587658963221200 ?

Downgrading from 1.10.10 to 1.10.7 cant be problematic.

sjmiller609:
> @sjmiller609 are we worried about letting folks push 1.10.10 images? With this change merged, a customer can create a new deployment and it's initial schema would be created using 1.10.10 right? What would happen if they push a 1.10.7 image up after?
> 
> (also bumped seats in circle)

All the concerns regarding cloud are not applicable to the master branch where we are using user-provided default image. There is no known issue with using user-provided images for init containers when users use AC.

I think this change is valid for master branch.

sjmiller609:
I double checked. master uses houston version defaulting to oldest config, and airflow-chart using user-provided images.

PR #656, opened 2020-04-24
Migrate to helm3

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/688


bin/install-ci-tools
```diff
@@ -2,7 +2,7 @@
 set -xeuo pipefail
 
 KIND_VERSION="0.7.0"
-HELM_VERSION="2.16.1"
+HELM_VERSION="3.1.3"
```
sjmiller609, line 5: I think we also need
```
helm repo add stable https://kubernetes-charts.storage.googleapis.com
```
somewhere in this file

bin/e2e-tests/run.sh
```diff
@@ -6,4 +6,4 @@ REPO_DIR=$DIR/../..
 
 echo "Running E2E Tests..."
 
-helm test astronomer
+helm3 test astronomer
```
sjmiller609, line 9:
```suggestion
helm test astronomer
```

charts/astronomer/templates/commander/commander-role.yaml
```diff
@@ -13,51 +13,117 @@ metadata:
...
+  verbs: ["create", "patch"]
+- apiGroups: ["extensions"]
+  resources: ["ingresses"]
+  verbs: ["create", "delete", "get", "patch"]
+- apiGroups: ["extensions"]
+  resources: ["ingresses/status"]
+  verbs: ["update"]
+- apiGroups: ["networking.k8s.io"]
+  resources: ["ingresses"]
+  verbs: ["get", "create"]
```
andriisoldatenko, line 102: @schnie it was before:
https://github.com/astronomer/astronomer/blob/66c63c55d85d3d426549fba5ccadb19e3faed652/tiller-rbac/tiller-role.yaml#L51-L53

bin/install-ci-tools
andriisoldatenko: fixed.

bin/e2e-tests/run.sh
andriisoldatenko: fixed.

sjmiller609:
CI passing!!! 🌮 🎉 ok let me take a look :)

dimberman:
@andriisoldatenko I attempted to follow these directions to upgrade I found that it wasn't respecting the dependencies.

I set the tag `keda: true` and `global.kedaEnabled=true` but keda wouldn't come come up.

I saw that with helm3 `requirements.yaml` needs to be moved into the `Chart.yaml`, but I'm still unable to control which platform components are downloaded. I attempted both by trying to update an 0.13 EKS deployment, and by running install on a blank kind cluster.

vishwas-astro:
@schnie can you please review?

@andriisoldatenko please address the changes requested by @sjmiller609

PR #658, opened 2020-04-28
Corrected labeling in airflow alerts

jpweber:
Resolution for issue https://github.com/astronomer/issues/issues/1214
related to https://github.com/astronomer/issues/issues/1189

change labels in airflow alerts from $labels.release to $labels.deployment. I think the ones with release were there in error as we do not have `release` available as a usable label.


schnie:
LGTM

PR #666, opened 2020-05-04
Moves default configuration out of Airflow chart to Houston default

schnie:
This PR is the Astronomer side of this PR to remove the platform refs from Airflow chart here: https://github.com/astronomer/airflow-chart/pull/66.

This essentially takes the old values that were hard-coded into the Airflow chart templates and uses the new helm variables to add them back in via the platform level.

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -103,6 +103,15 @@ data:
 
       # These values get passed directly into the airflow helm deployments
       helm:
+        env:
+          - name: ASTRONOMER__AIRFLOW__WORKER_LOG_RETENTION
+            value: "3"
+          - name: ASTRONOMER__JWT_SIGNING_CERT
+            value: "/etc/airflow/tls/tls.crt"
+          - name: ASTRONOMER__JWT_AUDIENCE
+            value: "deployments.baseDomain"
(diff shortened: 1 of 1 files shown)


andriisoldatenko:
@ashb yeah i see some errors:
```
 astronomer           astronomer-houston-upgrade-deployments-ffg4k              0/1     Error              0          11s
```
i think we need practical tests here

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -103,6 +103,15 @@ data:
...
       # These values get passed directly into the airflow helm deployments
       helm:
+        env:
+          - name: ASTRONOMER__AIRFLOW__WORKER_LOG_RETENTION
+            value: "3"
+          - name: ASTRONOMER__JWT_SIGNING_CERT
+            value: "/etc/airflow/tls/tls.crt"
+          - name: ASTRONOMER__JWT_AUDIENCE
+            value: "deployments.baseDomain"
+            value: {{ printf "deployments.%s" .Values.global.baseDomain }}
```
sjmiller609, line 113: here is the issue

sjmiller609, line 113:
```suggestion
            value: {{ printf "deployments.%s" .Values.global.baseDomain }}
```

PR #673, opened 2020-05-06
Add support for AC 1.10.7-10 Docker images

kaxil:
Based on https://github.com/astronomer/ap-airflow/pull/75


schnie:
We should talk about this ever-growing list at standup

PR #676, opened 2020-05-11
Adding prometheus label drop for a duplicate label

jpweber:
We started seeing duplicates for the label `label_kubernetes_pod_operator` that was causing kube-state-metrics  ingestion issues. A label drop was added for this label to get through the issue

### DIFF
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -210,6 +210,8 @@ data:
           # Pods have this label twice
           - regex: 'label_kubernetes_executor'
             action: labeldrop
+          - regex: 'label_kubernetes_pod_operator'
+            action: labeldrop
           # Required for multi-namespace mode
           - source_labels: [namespace]
             regex: "^{{ .Release.Namespace }}-(.*-.*-[0-9]{4})$"


charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -210,6 +210,8 @@ data:
           # Pods have this label twice
           - regex: 'label_kubernetes_executor'
             action: labeldrop
+          - regex: 'label_kubernetes_pod_operator'
+            action: labeldrop
```
schnie, line 214: @sjmiller609 @jpweber so these two drops affect only the ones that airflow adds itself, but ignore the ones that we add via our config file?

jpweber: it will drop any label that matches exactly `label_kubernetes_pod_operator` wherever it comes from. Same problem as the `label_kubernetes_executor` problem we had a few weeks back. I don't know what the further downstream impact is honestly.

schnie:
We use the dashed labels, so I think this is just removing the underscored ones that new builds of Airflow are generating.

https://github.com/astronomer/google-environments/blob/master/prod/cloud/locals.tf#L272-L273

PR #677, opened 2020-05-15
Fix registry configuration

sjmiller609:
'serviceName' should only exist when it's a stateful set. The conditional was not the negation of the conditional for the 'kind', which is:
```
{{- if or .Values.registry.gcs.enabled .Values.registry.azure.enabled }}
```

### DIFF
--- a/charts/astronomer/templates/registry/registry-statefulset.yaml
+++ b/charts/astronomer/templates/registry/registry-statefulset.yaml
@@ -28,7 +28,7 @@ spec:
       component: registry
       tier: astronomer
       release: {{ .Release.Name }}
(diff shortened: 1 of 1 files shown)


sjmiller609:
I guess what was happening was extra, irrelevant keys were ingored in helm2, but not helm3? I'm not sure why this was encountered when shipping helm3 stuff.

sjmiller609:
I went ahead and cherry picked this change into release-0.14 already because it's not a released version at this point

schnie:
Ah, yes I think we cleaned up this in a few other places too. Guess we missed this one. I believe it was just ignored in helm2.

andriisoldatenko:
yeah helm3 looks more strict.

PR #681, opened 2020-05-19
Update UI and API components for future 0.14.1 release

ryanahamilton:


ryanahamilton:
Will we have to reset all of these component values back to the last 0.13 versions for platform 0.14.0 release since the application updates won't be shipping until platform 0.14.1?

sjmiller609:
@ryanahamilton we can tee up 0.14.0 on release branch and have master for 0.14.1 until 0.14.0 is shipped

PR #684, opened 2020-05-21
Add platform NodeSelector and affinities to KEDA

dimberman:
All KEDA system components will launch with the platform
NodeSelector, Affinities, and tolerations

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -51,9 +51,12 @@ global:
   # nodeSelector, affinity, and tolerations values for airflow deployment pods are assigned via houston config.
   # See more information on pod / node assignment here: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
   platformNodePool:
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -198,8 +201,18 @@ prometheus:
...
 
-
+keda:
+  keda:
+    nodeSelector:
+      <<: *platformNodeSelector
+    affinity:
+      <<: *platformAffinity
+    tolerations:
+      *platformTolerations
```
schnie, line 215: Nice workaround

dimberman: @sjmiller609 found the stackOverflow post :).

YAML is magic. Not the good kind.

schnie: I guess we can use this pattern going forward for any translation to the 3rd party charts. :+1:

PR #686, opened 2020-05-22
Add support for AWS S3 registry backend

ianstanton:
- Add ability to use AWS S3 bucket as registry storage backend
- Example config:
```
astronomer:
  registry:
    s3:
      enabled: true
      accesskey: my-access-key
      secretkey: my-secret-key
      region: us-east-1
      bucket: astro-registry
```
- Driver documentation:
  - https://docs.docker.com/registry/configuration/
  - https://github.com/docker/docker.github.io/blob/master/registry/storage-drivers/s3.md
- Related to astronomer/issues#93

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -170,6 +170,17 @@ azure:
   realm: {{ .Values.registry.azure.realm }}
 {{- end }}
 
+{{- define "registry.s3Config" }}
+s3:
+  accesskey: {{ .Values.registry.s3.accesskey }}
+  secretkey: {{ .Values.registry.s3.secretkey }}
+  region: {{ .Values.registry.s3.region }}
+  bucket: {{ .Values.registry.s3.bucket }}
+  encrypt: {{ .Values.registry.s3.encrypt }}
+  keyid: {{ .Values.registry.s3.keyid }}
+  rootdirectory: {{ .Values.registry.s3.rootdirectory }}
+{{- end }}
+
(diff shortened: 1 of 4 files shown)


vparekh94:
LGTM! Only question is around how we'd want the customer to implement this.

Do we want them to leave that key in plaintext in their config file? Should this be a secret? Or use a role or something along those lines?

schnie:
> LGTM! Only question is around how we'd want the customer to implement this.
> 
> Do we want them to leave that key in plaintext in their config file? Should this be a secret? Or use a role or something along those lines?

I guess all of these integrations have some sort of secret associated with them, so that might be something to look at. Although helm3 stores it in secrets by default it will still live in plaintext in that config file, which is a valid concern.

Follow up commit to optionally mount all these configs from a pre-configured secret?

ianstanton:
> > LGTM! Only question is around how we'd want the customer to implement this.
> > Do we want them to leave that key in plaintext in their config file? Should this be a secret? Or use a role or something along those lines?
> 
> I guess all of these integrations have some sort of secret associated with them, so that might be something to look at. Although helm3 stores it in secrets by default it will still live in plaintext in that config file, which is a valid concern.
> 
> Follow up commit to optionally mount all these configs from a pre-configured secret?

Seems fair to me! Here is a note from the driver documentation on using a role:

> Note: You can provide empty strings for your access and secret keys to run the driver on an ec2 instance and handles authentication with the instance's credentials. If you use IAM roles, omit these keys to fetch temporary credentials from IAM.

PR #688, opened 2020-05-26
Make helm test failures fail CI

verdverm:


schnie:
So are these failures in circle "expected" now?

sjmiller609:
I think we need more visibility into what is failing. For example, [here is an output of a behave test in circle CI](https://app.circleci.com/pipelines/github/astronomer/commander/97/workflows/4a25dd61-5c32-43a0-a130-dc2ffc45193a/jobs/186/parallel-runs/0/steps/0-104). However, this seems to be at-odds with our ask of you to do with "helm test". Is there a way to get the best of both here, for example starting background process or something like this?
```
kubectl logs -f the-test-pod &
```

verdverm:
@sjmiller609 The example CircleCI link is a passing test. Behave will output more information (I coded it this way) when a failure actually happens. Behave hides stdout unless there is an error, then it prints it. There are extra print statements to provide context when there is an error.

Also note, that the `helm test` well be going way in favor of a `docker run`, largely due to security concerns around running docker-in-docker-in-kubernetes.

sjmiller609:
Great work, I can see the output of the tests! I don't know why it's crashlooping to start, I wonder if we can figure that out, but I think it's OK for now. Another question is - what if these tests fail, can we expect the pipeline to fail, as intended?

sjmiller609:
@verdverm I'm sampling one of these, and it appears to have failed, but it's not clear what failed. I suspect something is wrong because it looks like no tests failed, yet the CI is not passing
```
5 features passed, 0 failed, 0 skipped
 
38 scenarios passed, 0 failed, 0 skipped
 
155 steps passed, 0 failed, 0 skipped, 0 undefined
 
Took 0m19.940s
 
 
Exited with code exit status 1 
```

verdverm:
I noticed that last night, going to print the captured status, but there's a spot above that showing the test container in a crashloop, so I think it's in there somewhere actually

verdverm:
hmm, it seems the title might need to be negated. I'm not sure we can get any Helm3 test to pass if the following is an issue in Helm:

```
containerStatuses:
 
  - containerID: containerd://5b7fad7a5b74ca05ec62fe96056766613b7833e10b1f478ada0966d7c57b3aac
 
    image: docker.io/astronomerinc/ap-e2e-test:0.0.7
 
    imageID: docker.io/astronomerinc/ap-e2e-test@sha256:1b94b2a1061cf0c20a01c58063a25f2d2c3b5d8e4420aec05a4304d3ad1aa38c
 
    lastState:
 
      terminated:
 
        containerID: containerd://5b7fad7a5b74ca05ec62fe96056766613b7833e10b1f478ada0966d7c57b3aac
 
        exitCode: 0
 
        finishedAt: "2020-06-01T19:40:19Z"
 
        reason: Completed
 
        startedAt: "2020-06-01T19:39:58Z"
 ```

The test container **is** exiting with a zero status...

Helm test docs are barely anything, maybe we just move to `docker run` now?

verdverm:
might have found another unmentioned option / requirement on the restartPolicy

verdverm:
hey, look at that, it passed

PR #690, opened 2020-05-27
Revert "Moves default configuration out of Airflow chart to Houston default"

sjmiller609:
Reverts astronomer/astronomer#666


sjmiller609:
ty

PR #691, opened 2020-05-27
Moves default configuration out of Airflow chart to Houston default

schnie:
This PR is the Astronomer side of this PR to remove the platform refs from Airflow chart here: astronomer/airflow-chart#66.

This essentially takes the old values that were hard-coded into the Airflow chart templates and uses the new helm variables to add them back in via the platform level.

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -103,6 +103,15 @@ data:
 
       # These values get passed directly into the airflow helm deployments
       helm:
+        env:
+          - name: ASTRONOMER__AIRFLOW__WORKER_LOG_RETENTION
+            value: "3"
+          - name: ASTRONOMER__JWT_SIGNING_CERT
+            value: "/etc/airflow/tls/tls.crt"
+          - name: ASTRONOMER__JWT_AUDIENCE
+            value: "deployments.baseDomain"
+            value: {{ printf "deployments.%s" .Values.global.baseDomain }}
+
         # Webserver configuration.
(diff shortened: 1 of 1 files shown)


andriisoldatenko:
@sjmiller609 do we want this patch in 0.16.0?
cc @vishwas-astro

schnie:
We haven't merged this into our airflow-chart so this wouldn't work right without that.

I don't know if its worth it. I made all the changes to support the OSS chart donation, but I don't know if we should bother trying to merge into our version.

We'd have to do a platform and Airflow chart upgrade to roll this out, with no actual difference in functionality.

sjmiller609:
this seems like a good idea to me, i'll issue a review when it's not in draft mode

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -121,6 +129,54 @@ data:
...
+            from airflow import configuration as conf
+            from flask_appbuilder.security.manager import AUTH_REMOTE_USER
+            basedir = os.path.abspath(os.path.dirname(__file__))
+
+            # The SQLAlchemy connection string.
+            SQLALCHEMY_DATABASE_URI = conf.get('core', 'SQL_ALCHEMY_CONN')
+
+            # Flask-WTF flag for CSRF
+            CSRF_ENABLED = True
+
```
ashb, line 151:
```suggestion
```

This was already default, and this setting does nothing -- it should have been `WTF_CSRF_ENABLED`

danielhoherd:
Closing stale PR.

PR #695, opened 2020-05-29
Add support for AC 1.10.7-11 Docker images

kaxil:


kaxil:
I think most likely here: https://github.com/astronomer/terraform-google-astronomer-cloud/blob/master/locals.tf but looks like we don't currently override any Houston settings there.

kaxil:
ping @schnie , @sjmiller609

sjmiller609:
@ashb @kaxil hey where kaxil indicated is correct how I first set it up, but that was a bit awkward. Now it's here, https://github.com/astronomer/google-environments take a look at the readme, there is a different spot for shared config, env-specific config, and for secrets.

PR #698, opened 2020-06-02
Commander 0.15.1

sjmiller609:
Fix upgrades


sjmiller609:
I added to release branch 0.15

andriisoldatenko:
@schnie please review and merge.

sjmiller609:
@schnie we should consider pick into release 0.14 also because of the information we have from recent production issue

PR #704, opened 2020-06-04
Kubernetes 1.16+ compatibility for cloud providers

verdverm:
This fixes the platform running on GKE (possible all k8s) >= v1.16.x. At this point some resource APIs graduated, and helm charts need to check for this. There are more changes required from helm3 migration, and also the postgres chart is unmaintained at this point. 

**This likely effects all versions of Astronomer and may apply to more Kubernetes vendors. If it is from a combination of Kubernetes ^v1.16 and Helm ^v3, then it may be more wide spread.**

This PR

- assimilates the postgres chart (primary contributor to the changeset size)
- fixes several places where k8s version checks -> resource api version needed changing / updating,by adding an `_apivers.tpl` to each charts' template dir 
- adds some more options to `bin/install-platform` for running helm in `dryrun` or `template` modes
- adds a `bin/prep-k8s` which is extracted from another file to be reusable, because...

I installed `release-0.15` branch on GKE v1.16.8. I have not tested it at all yet, see below for current status. This was done on a single VM cluster and we probably need to expand to multi-pool GKE to be more realistic. Single node should be good as a first pass and there is a stage belos to represent a multi-pool e2e CI like passing grade. 

_This will need rebasing apparently, also need to think about if / what versions to backport this into._

---

Testing matrix:

| KubeVersion | KinD | GKE-s | GKE-c |
| --------------- | ------ | -------- | --------- |
| 1.14              | 👍 | 👍 | -  |
| 1.15              | 👍 | - | n/a |
| 1.16              | 👍 | 👍 | 👍 |
| 1.17              | 👍 | n/a | - |
| 1.18              | 👍 | n/a | n/a |

GKE-s: static version selection
GKE-c: release channel selection (stable,regular,rapid) (going down that column)

Detailed versions:

| KubeMInor     | KinD | GKE-s | GKE-c |
| ---------------   | ------ | -------- | --------- |
| 1.14 | 1.14.10 | 1.14.10-gke.36  | 1.14.10-gke.36  |
| 1.15 | 1.15.11 | 1.15.11-gke.13  | n/a |
| 1.16 | 1.16.9   | 1.16.8-gke.15    | 1.16.8-gke.15 |
| 1.17 | 1.17.5   | n/a                     | 1.17.5-gke.0 |
| 1.18 | 1.18.2   | n/a                     | n/a |


ianstanton:
@verdverm thanks for the fix. I tested this out and it looks to resolve the issue with helm3, but I get the following error when using helm2:
```
Error: render error in "astronomer/charts/elasticsearch/templates/master/es-master-statefulset.yaml": template: astronomer/charts/elastiet.yaml" at <.Capabilities.KubeVersion.Version>: can't evaluate field Version in type *version.Info
```

verdverm:
@ianstanton many of the charts and templates lack this check. There should be a story about going through methodically and making sure all of these are present and correct.

ianstanton:
It looks like the `Capabilities` object in Helm2 doesn't include the value `Version`. Using the `GitVersion` value seemed to solve this in Helm2:
```
{{- if semverCompare "^1.16-0" .Capabilities.KubeVersion.GitVersion }}
```
However, `GitVersion` is not a value for this object in Helm3.

verdverm:
yes, as @jpweber discovered, the GitVersion has been removed in Helm3.

Are we at the point where:

1. We need to add an or check to support both helm 2/3
2. Have separate charts because there are enough differences

Depends on how much / long we expect Helm2 to be around.

jpweber:
if we could drop support for anything lower than 1.14 we wouldn't need the checks. because they would all be able to use the new API. But I feel like we have someone on 1.12 which is why we have to do the conditional business.

verdverm:
The APIs on resources graduate at their own paces. I think this is a persistent situation all charts will need to be checking (or any similar tool to helm).

The one we are seeing here is NetworkPolicy graduate to v1 between k8s 1.14 & 1.16. I'm modelling the testing after the versions in CI and what is available in GKE.

I anticipate that we will actually need to keep KinD around for k8s 1.12 and be mindful of the differences between real k8s

verdverm:
It also looks like the checks need to be more sophisticated so as to support Helm 2&3

ianstanton:
@verdverm something else I found interesting - I did _not_ experience this issue on Azure or IBM Kubernetes v1.16.x.

verdverm:
@ianstanton was that helm 2, 3, or both?

ianstanton:
@verdverm both using helm2. I did not use the dependent postgresql chart on IBM, but believe I did on Azure.

verdverm:
@ianstanton I think we are pretty confident that this is a Helm 2 v 3 issues.

We need to add checks to certain Kubernetes Kinds where the resource version changes between k8s versions astro supports. The compounding issues is that Helm changed one of the variable names for accessing information needed in this check.

So what we need to do is something like:

- use a template where `apiVersion` is specified in chart templates: https://github.com/astronomer/astronomer/pull/704/files#diff-62365d0e90bddbafad626f5a94fa7234R3  is using the definition here https://github.com/astronomer/astronomer/pull/704/files#diff-e014fac9a9b86fb35d8f6d3163d82c96R42
- This is the recommended way to compare k8s versions now: https://github.com/astronomer/astronomer/pull/704/files#diff-da9eb35bc6033bc7eaa6ce6185555d1c

What we _also_ need to add is an `(or (<helm3 check>) (<helm2 check>))` in the check because helm3 uses `Version` where helm2 used a `GitVersion` iirc.

So what happens when the error shows up is that someone is installing AP with Helm3, the postgresql chart (depreciated) is downloaded, and when it hits the check, the variable is undefined and selects the false condition, which is the old NetworkPolicy apiVersion that was used k8s pre-1.16

The intention of this PR is to pull in the depreciated chart and fix these checks to support both helm 2/3, as a short-term fix in lieu the long-term replacement of this chart.

Makefile
```diff
@@ -13,24 +13,56 @@ OUTPUT := repository
 TEMP := /tmp/${DOMAIN}
 
 .PHONY: lint
+lint: lint-prep lint-astro lint-charts 
+#lint-prom (omitted)
```
verdverm, line 17: note, prom is currently disabled, there's a bug in it

Makefile
```diff
@@ -4,7 +4,7 @@ URL ?= https://${DOMAIN}
 BUCKET ?= gs://${DOMAIN}
 
 # List of charts to build
-CHARTS := astronomer nginx grafana prometheus alertmanager elasticsearch kibana fluentd kube-state
+CHARTS := astronomer nginx grafana prometheus alertmanager elasticsearch kibana fluentd kube-state postgresql
```
verdverm, line 7: should this be added here?

verdverm:
we have flakey tests, is that known?

sjmiller609:
@verdverm which tests are flakey? can you find an example in a different PR?

verdverm:
I saw `commander` and `es-nginx...` crash looping inconsistently

https://app.circleci.com/pipelines/github/astronomer/astronomer/788/workflows/dafba23c-1dec-4103-8537-1c7cea92861c

sjmiller609:
> I saw `commander` and `es-nginx...` crash looping inconsistently
> 
> https://app.circleci.com/pipelines/github/astronomer/astronomer/788/workflows/dafba23c-1dec-4103-8537-1c7cea92861c

hmm, that is concerning. I have not seen in other branches

sjmiller609:
for example, consider master branch CI history: https://app.circleci.com/pipelines/github/astronomer/astronomer?branch=master

verdverm:
The build history on this branch has used different image vers over time. I needed to rebase to deal with merge conflicts as master evolved underneath.

I'm going to push a small commit to update the `.helmignore` file. If that works, I'm down to ignore the flakes I saw prior.

Makefile
schnie: does `CHARTS` still get used?

bin/run-ci
```diff
@@ -23,32 +27,18 @@ sleep 5
...
+### Create initial user
+$DIR/create-initial-user "tester@astronomer.io" "password"-
 
-sleep 10
+### Test Platform
+$DIR/test-ap
 
-./bin/create-initial-user "tester@astronomer.io" "password"
-# kill $FORWARD_PID
+exit $?
```
verdverm, line 36: Note, I added this here as the tests that follow were not passing, may be more a dns thing than correctness. @jpweber did you say something about the status of these tests?

verdverm:
There are also a number of chart values which have changes in this PR. With the rebasing to master multiple times, can someone who knows more about what these should actually be today go through and (painstakingly) double check them?

verdverm:
welp, guess we got some flakes after all...

sjmiller609:
> welp, guess we got some flakes after all...
> 
> you can retrigger from failed builds to get them on to pass too

if it's from the e2e tests, let's put those back in commented out so we can get this moving faster

verdverm:
It looks to be flakey auth, the tests fail at different scenarios, but always an auth error with the CLI

verdverm:
The CLI could be getting out of date too

bin/run-ci
```diff
@@ -23,32 +27,18 @@ sleep 5
 
 $DIR/waitfor-platform
 
-exit 0
-
-### Test Prep
-echo "172.17.0.1 local.astronomer-development.com" | sudo tee -a /etc/hosts
-sudo /tmp/bin/kubectl port-forward -n astronomer svc/astronomer-nginx 80 443 &
-FORWARD_PID=$!
+### Create initial user
```
sjmiller609, line 30:
```suggestion
# TODO: turn on e2e tests

exit 0

### Create initial user
```

.circleci/generate_circleci_config.py
```diff
@@ -10,10 +10,10 @@
...
 # recent patch version on Dockerhub
 KUBE_VERSIONS = ['1.14.10',
-                 '1.15.7',
-                 '1.16.4',
-                 '1.17.2',
-                 '1.18.2']
+        '1.15.11',
+        '1.16.9',
+        '1.17.5',
+        '1.18.2']
```
sjmiller609, line 16: ty

bin/install-platform
```diff
@@ -46,19 +76,22 @@ else
...
   --set tags.platform=${platform} \
   --set tags.logging=${logging} \
   --set tags.monitoring=${monitoring} \
   --set tags.kubed=${kubed} \
-  --set global.postgresqlEnabled=true \
+  --set global.postgresqlEnabled=${postgresql} \
   --set tags.keda=${keda} \
   $HELM_CHART_PATH
 
+exit
```
sjmiller609, line 93: Hey why this change?

templates/_apivers.tpl
```diff
@@ -0,0 +1,82 @@
+{{/* Generic Kubernetes apiVersion helpers */}}
```
sjmiller609, line 1: this is a great way to do it, love it

bin/run-ci
sjmiller609: i'm just going to disable e2e testing for now

bin/install-platform
sjmiller609, line 93:
```suggestion
```

sjmiller609: I'm just going to delete it for now

sjmiller609: lmk if I missed something though

sjmiller609:
Great work Tony. Thank you for tracking down this issue and fixing it

PR #714, opened 2020-06-11
Fix for DB migration of Airflow version 1.10.6

ashb:
I went to check my hot fix for Houston had made it to master (it hadn't) and noticed this commit needed doing to.

(cherry picked from commit e1750670bdbb618f16c43507ce290da847f6243a)

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: astronomer
-version: 0.15.0-rc.3
-appVersion: 0.15.0-rc.3
+version: 0.15.1
+appVersion: 0.15.1
 description: Helm chart to deploy the Astronomer Platform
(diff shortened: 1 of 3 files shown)


schnie:
These make me worried. Is there a reason we can't merge things to master first as a practice?

sjmiller609:
it's a dupe of https://github.com/astronomer/astronomer/pull/711/files

sjmiller609:
@schnie the reason is because I can push to release branches but not master. in a hotfix scenario where I need to release immediately, I push directly to release branch then ship to cloud.

ashb:
@sjmiller609 Not entirely -- on master the version in the charts is still wrong.

sjmiller609:
deprecated - we have a different version of houston up

PR #719, opened 2020-06-12
Change houston charts from npm to yarn

adam2k:
This feature is linked to the houston-api PR located here: https://github.com/astronomer/houston-api/pull/334

I confirmed this fix by testing the following:

1. Build a docker image locally
2. Push the docker image to my personal docker hub registry
3. Change the `houston-api` image in the `charts/astronomer/values.yaml` to the newly created docker image
4. Run `./bin/reset-local-dev`
5. Check the k8s pods and logs to make sure everything is working correctly
6. Launch the `astro-ui`, log in and click test to make sure `houston-api` is responding correctly

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -23,7 +23,7 @@ images:
     pullPolicy: IfNotPresent
   houston:
     repository: astronomerinc/ap-houston-api
-    tag: 0.15.5
+    tag: master
     pullPolicy: IfNotPresent
   astroUI:
     repository: astronomerinc/ap-astro-ui
--- a/charts/astronomer/templates/houston/houston-check-updates-cronjob.yaml
+++ b/charts/astronomer/templates/houston/houston-check-updates-cronjob.yaml
@@ -42,7 +42,7 @@ spec:
               image: {{ template "houston.image" . }}
               imagePullPolicy: {{ .Values.images.houston.pullPolicy }}
               # If you supply only args for a Container, the default Entrypoint defined in the Docker image is run with the args that you supplied.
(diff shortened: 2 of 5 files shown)


adam2k:
@andriisoldatenko @sjmiller609 Since this PR relies on https://github.com/astronomer/houston-api/pull/334 how can I get these tests to pass?  I don't want to change the docker image link until we have a new houston-api image that's ready to release that is using yarn.

adam2k:
@schnie or @andriisoldatenko these are all the locations that I could location `npm` in astronomer.  I think this PR will fail until the houston-api changes are merged and the new image builds with yarn.  Can either of you check to see if this looks good to you so that as soon as houston-api is merged we can quickly merge this PR afterward?

andriisoldatenko:
@schnie let's merge houston-api first than rerun this CI fails.

adam2k:
I'm watching the master build on houston-api.  If these fail before the houston-api build finishes then I will re-run these tests.

charts/astronomer/values.yaml
```diff
@@ -23,7 +23,7 @@ images:
     pullPolicy: IfNotPresent
   houston:
     repository: astronomerinc/ap-houston-api
-    tag: 0.15.9
+    tag: master
```
adam2k, line 26: @sjmiller609 @schnie I changed this to master because the houston-api version that is using yarn is not tagged.  I figured we can replace this with the correct 0.16.x version when that is ready.  Is that all right with you two? CC: @andriisoldatenko

sjmiller609: discussing houston image release process in slack

PR #728, opened 2020-06-19
Increase default proxy timeouts

sjmiller609:
This helps avoid 504 for customers with a more significant latency between nodes (specifically between nginx ingress pods and the backend services)


sjmiller609:
I want to ensure these are reasonable. just because it fixed for one customer doesn't mean I should add in like this for default. sec

PR #732, opened 2020-06-22
enable e2e tests in ci

verdverm:


schnie:
What should we do here? Looks like they are failing.

vishwas-astro:
@verdverm can you please check why the tests are failing so we can get to the underlying problem and make these tests more stable?

verdverm:
FYI, discussed off of GitHub, in summary:

- tests are failing because the platform has evolved underneath, so they are out-of-date
- the new test setup is about ready to go and it will be faster to migrate to the new than repair the old
- we will be making a new story / PR for the change over

sjmiller609:
@verdverm we discussed the issue with cross-repo cloning and that issue was resolved by the commits I added that were peer programed between me and you yesterday. Now it is pending on you again to make the pipelines pass, please let me know if you need any help.

verdverm:
@sjmiller609 Are we trusting the cluster self-signed certs in CI?

astro-cli does not like http or self-signed certs, so we need to get them trusted in CI, unless astro-cli picked up that feature request for config value for being less restrictive here

verdverm:
The latest run of this, where the test is failing, passes against staging

verdverm:
Update: using Kind is proving troublesome as more issues and complications have arisen.

- We boot a VM and run Kind on it during CI. All operations happen and run on the same VM.
- Now that we want to push DAGs, we need to use the CLI. It requires TLS.
- Since this is all local, the solution to TLS has been to trust the self-signed certificate and add an entry to `/etc/hosts`.
- Kind appears to be also picking up this entry inside of k8s DNS, so calls which use the external name (registry i.e.) break within cluster. This error was hidden until now, as we were only experiencing this in CI. The issue can be seen in Houston's logs.

So we are at a crux, we need one of the following possible solutions:

- Find a different method for providing TLS to the CLI while not breaking DNS in Kind
- Update the CLI to support non-tls based requests ( it is not clear that this is the only place we will hit TLS required issues )
- Migrate from Kind to a single node k8s VM. Google Cloud Build would be much easier to do this with and can be a parallel build.

verdverm:
nice! I think we are ready to review and merge @schnie @sjmiller609 @vishwas-astro

PR #733, opened 2020-06-22
Update default tags for new repo defaults

ryanahamilton:
These two repos now use `main` as their default branch (and build images with that name).

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -23,7 +23,7 @@ images:
     pullPolicy: IfNotPresent
   houston:
     repository: astronomerinc/ap-houston-api
-    tag: master
+    tag: main
     pullPolicy: IfNotPresent
   astroUI:
     repository: astronomerinc/ap-astro-ui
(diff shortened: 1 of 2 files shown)


charts/astronomer/values.yaml
```diff
@@ -23,7 +23,7 @@ images:
     pullPolicy: IfNotPresent
   houston:
     repository: astronomerinc/ap-houston-api
-    tag: master
+    tag: main
```
ryanahamilton, line 26: Not clear to me why we're not using a versioned tag here?

jpweber: Looks like it was change recently so Adam could get something deployed. Similar things to what we've been doing on the dev cluster. This will have to change back to a version before the next release.

sjmiller609:
Can we just move to a versioned tag?

sjmiller609:
I think we do not want this change now that versions are correctly set in master branch https://github.com/astronomer/astronomer/blob/cfc9ee26648b6f06c44a5fa7912a3e8b0404c5a5/charts/astronomer/values.yaml#L15-L47

PR #735, opened 2020-06-25
Use on-demand instances for sensitive workloads

sjmiller609:
I figure just add the label no matter what because it's just one less thing to worry about configuring.


sjmiller609:
I'm thinking I should actually make this configurable instead. I'll make it like astronomer.registry.labels and so forth

PR #746, opened 2020-06-30
privilege escalation needs to be allowed for now

jpweber:
the debian buster images make use of gosu in a way that requires us to allow privilege escalation. This will need to be changed back after the airflow AC images have been changed.

This change has already been made live in prod by editing the restrictive policy.


jpweber:
This needs to be put on hold because we can't turn on the policies in prod right now because of customer pods that are not permitted by the policies. We need to treat this similar to an api change and give people time to change their things before we break them.

PR #749, opened 2020-07-01
Make the restrictive policy same as permissive for now

jpweber:
Our restrictive policy breaks some of the customer workloads because they are doing things that are against policy. We need to provide notice and give them time to make changes before we move the policy back to its normal restrictiveness. 

This change is already live in production.


jpweber:
This needs picked in to release-0.16 after meged

templates/psp/restrictive-psp.yaml
```diff
@@ -5,28 +5,50 @@ apiVersion: policy/v1beta1
...
   volumes:
-  - 'configMap'
-  - 'downwardAPI'
-  - 'emptyDir'
-  - 'persistentVolumeClaim'
-  - 'secret'
-  - 'projected'
-  allowedCapabilities:
   - '*'
 {{- end -}}
```
schnie, line 54:
```suggestion
{{- end -}}

```

PR #755, opened 2020-07-10
Upgrade to 0.16-LTS docs and automation

sjmiller609:
We are going to consider 0.16 the first LTS version. The plan will be to use this script to upgrade from any version to 0.16, then we will provide additional automation to ensure upgrades work in between each new LTS version.

@ianstanton please review the script and the README
@vishwas-astro please review the README

closes https://github.com/astronomer/issues/issues/1405


sjmiller609:
@danielhoherd I would like to discuss with you if using Ansible for LTS-to-LTS version upgrades is a good idea or not compared to a bash script. I think you have some experience using Ansible to do helm stuff, and I would like to consider the pros and cons here.

bin/migration-scripts/README.md
```diff
@@ -1,35 +1,47 @@
...
 
-## Schedule a change window
+## Astronomer versioning: moving to the long-term-support (LTS) software versioning model
+
+We have customers running Astronomer versions 0.7 through 0.16, and we currently support all of them. The wide variety of configurations and versions present in our customer's environments and the accelerating new minor version release cycle has presented challenges with regard to training support enginnering, backporting fixes, documenting change notes, and with upgrades. In July 2020, Astronomer decided to start following the long-term-support (LTS) model for enterprise-ready Astronomer versions. This will look like some of our minor versions receive continuous patching while the others do not. This allows our developers to focus on building new features and limiting the number of versions where they need to backport patches. In the near future, we will be only publicly publishing the LTS versions. The release cadence is not yet decided but it will be somewhere in between 1 month and 6 months, depending on how many LTS versions we plan to support at the same time.
```
vishwas-astro, line 7: Typo on "engineering"

sjmiller609, line 7:
```suggestion
We have customers running Astronomer versions 0.7 through 0.16, and we currently support all of them. The wide variety of configurations and versions present in our customer's environments and the accelerating new minor version release cycle has presented challenges with regard to training support engineering, backporting fixes, documenting change notes, and with upgrades. In July 2020, Astronomer decided to start following the long-term-support (LTS) model for enterprise-ready Astronomer versions. This will look like some of our minor versions receive continuous patching while the others do not. This allows our developers to focus on building new features and limiting the number of versions where they need to backport patches. In the near future, we will be only publicly publishing the LTS versions. The release cadence is not yet decided but it will be somewhere in between 1 month and 6 months, depending on how many LTS versions we plan to support at the same time.
```

sjmiller609:
I found some issues for an 0.12.8 upgrade I want to address

ES env vars during upgrade-to-same-version step in script for version 0.12.8
wait for prism to be ready in the event of slow cluster
  most recent patch version of 0.16 instead of 0.16.0

sjmiller609:
@ianstanton @danielhoherd this is ready for review again

Ian, please review for process
Daniel, please review for bash-isms / best practice

The script is now idempotent which is pretty important for usability. Previously when things would go wrong during one of the steps, I would need to edit the script live to comment out the no longer applicable parts after manually resolving whatever customer-specific issue.

danielhoherd:
@sjmiller609 I took some time to fix a few hundred shellcheck complaints and rework some bits of the script. I think it's good enough to approve now, but since I also just committed 100 lines, we'll need to test this out again and get approval from somebody else.

sjmiller609:
> @sjmiller609 I took some time to fix a many dozen shellcheck complaints and rework some bits of the script. I think it's good enough to approve now, but since I also just committed 100 lines, we'll need to test this out again and get approval from somebody else.

Hey @danielhoherd thank you, and I'm happy for your collaboration on this! Let's peer up today and give it a whirl :)

sjmiller609:
I only merged before CI passes because this change is not related to the chart

PR #758, opened 2020-07-13
Implement internal-helm repo as first stage

danielhoherd:
* Implement two-step helm repo process where all release builds go to
  internal-helm first, then are promoted to prod (existin helm repo)
* Rework release script as release-helm-chart. Break flow into
  functions. Support multiple helm repos. Add bash getopts.
* Rework circleCI config with second helm chart sequence where releases
  first go to internal-helm, then must be approved for release to
  existing public helm repo.
* Clean up some code by fixing shellcheck complaints, removing old
  commented sections, delete reference to deprecated doc.


sjmiller609:
You are intended to have merge permission after 1 successful review, so if that is not the case just let me know.

sjmiller609:
Please use "squash then merge" and make the commit title informative from the user perspective when you merge, for example

Only publish enterprise-ready Helm chart versions to helm.astronomer.io (#758)

PR #759, opened 2020-07-14
Alert routing changes

jpweber:
If a `platformCritical` value is present, creates a route to match `critical` `platform` alerts and sends them to a critical alert receiver that only sends to pager duty. 

The route is configured such that an alert that matches the `platformCritical` receiver can continue matching other routes such as `platform`. This allows us keep the critical receiver scoped to just pager duty handling and not have to add any of the slack configs that we already have for platform alerts.


jpweber:
solves https://github.com/astronomer/issues/issues/1430

jpweber:
Yes it will send to slack _because_ any critical platform alerts will also still match the "normal" platform alert receiver. That's what the `continue` keyword in the critical route does, allows the alert to "continue" on matching any additional  routes.

PR #760, opened 2020-07-14
Added env var for REGISTRY_HTTP_SECRET

jpweber:
In order to run docker registry with multiple replicas all pods must use the same http secret.

This http secret is generated by helm as a random string of 32 characters. This will be updated every time/anytime the helm release is installed or upgraded. 

I choose the random string route rather creating a secret and storing it so that we don't have to manage the secret in anyway. 

What this parameter is for, from the docker docs. 
> A random piece of data used to sign state that may be stored with the client to protect against tampering. For production environments you should generate a random piece of data using a cryptographically secure random generator. If you omit the secret, the registry will automatically generate a secret when it starts. If you are building a cluster of registries behind a load balancer, you MUST ensure the secret is the same for all registries.

### DIFF
--- a/charts/astronomer/templates/registry/registry-statefulset.yaml
+++ b/charts/astronomer/templates/registry/registry-statefulset.yaml
@@ -47,6 +47,9 @@ spec:
         - name: registry
           image: {{ template  "registry.image" . }}
           imagePullPolicy: {{ .Values.images.registry.pullPolicy }}
+          env: 
+          - name: REGISTRY_HTTP_SECRET
+            value: {{ randAlphaNum 32 }}
           resources:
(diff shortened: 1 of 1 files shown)


jpweber:
This work is in support of https://github.com/astronomer/issues/issues/954

jpweber:
@sjmiller609 I don't think we're at risk of losing any state here. Because in its current implementation we don't specify the http secret so it gets auto generated by the pod at startup, so its already changing anytime a pod is killed or upgraded. Also because this "secret" doesn't become part of the saved state in anyway. Think of it like a session or JWT. Its only using during that session's transfer to ensure the data hasn't changed between the client and server.

PR #769, opened 2020-07-22
Add $RELEASE_NAME var to kubectl scale commands

ianstanton:
- The `kubectl scale` commands on line 280 and 282 assume the release name is `astronomer`
- Add $RELEASE_NAME var to these commands


ianstanton:
It looks like this was resolved by https://github.com/astronomer/astronomer/pull/770. Closing.

PR #771, opened 2020-07-24
Support private CA certs

danielhoherd:
# Code Changes

* Switch to helm3 chart (helm API v2)
* Implement private CA certificate handling through secrets and helm values.yaml
* Add logic to `setup-kind` which adds custom CA to that workflow so we can test it

# More Info

We should document the steps to create the secrets that hold the certificates. Currently there is the requirement for each certificate to be contained in a secret with data `cert.pem`. I tried to make this more flexible but was burning too much time debugging `helm` since it does not perform `lookup` calls when validating the templates. The steps to create the secrets can be seen in the setup-kind additions.

I also switched this chart to a helm3 api v2 chart, which we can undo if we want. AFAIK we are only supporting helm3 from here on out, so this change seemed appropriate at this time, but I can understand if we want to do that in a separate issue/PR.


charts/astronomer/templates/_helpers.yaml
```diff
@@ -282,3 +282,23 @@ s3:
...
+
+{{- define "custom_ca_volumes"}}
+{{ if .Values.global.privateCaCerts }}
+{{ range .Values.global.privateCaCerts }}
+- name: {{ . }}
+  secret:
+    secretName: {{ . }}
+{{- end }}
+{{- end }}
+{{- end }}
```
schnie, line 304:
```suggestion
{{- end }}

```

Chart.yaml
```diff
@@ -1,8 +1,97 @@
...
+  - name: nats
+    condition: global.natsEnabled
+    tags:
+      - nats
+
+  # Nats-streaming
+  - name: stan
+    condition: global.stanEnabled
+    tags:
+      - stan
```
schnie, line 97:
```suggestion
      - stan

```

sjmiller609:
backported to 0.16

PR #776, opened 2020-07-28
Improve Astronomer 0.16 upgrade script

sjmiller609:
There is a situation with some users who have upgraded a few times where a historically present config is carrying through upgrades such that it is causing a collision in a yaml merge of two lists. This is solved by deleting elasticsearch statefulset but not the persistent volumes behind.

### DIFF
--- a/bin/migration-scripts/upgrade-to-lts.sh
+++ b/bin/migration-scripts/upgrade-to-lts.sh
@@ -396,7 +396,7 @@ function main {
 
   git_clone_if_necessary
 
-  kubectl delete sts --cascade=false --namespace $NAMESPACE $RELEASE-es-master $RELEASE-es-data || true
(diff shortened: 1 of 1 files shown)


bin/migration-scripts/upgrade-to-lts.sh
```diff
@@ -396,6 +396,8 @@ function main {
 
   git_clone_if_necessary
 
+  kubectl delete sts --cascade=false --namespace $NAMESPACE $RELEASE-es-master $RELEASE-es-data || true
```
danielhoherd, line 399:
```suggestion
  kubectl delete sts --cascade=false --namespace "$NAMESPACE" "${RELEASE}-es-master" "${RELEASE}-es-data" || true
```

sjmiller609:
Hey, awesome Daniel. I will start doing that from here on out, and I will start using that shell check tool also

PR #782, opened 2020-07-30
Fix/nats houston connection

adam2k:
Add env vars for NATS Cluster ID and NATS URL to use for Houston (similar to the Houston Worker).

Related https://github.com/astronomer/houston-api/pull/391


adam2k:
@andriisoldatenko I needed to rebase.  Now this PR show only the changes that I'm proposing.

PR #786, opened 2020-08-03
bump grafana

andriisoldatenko:
https://github.com/astronomer/ap-vendor/pull/44/files


jpweber:
The tests failed with an image pull backoff. Is that because the image isn't published yet?

andriisoldatenko:
@jpweber it's a good question why it didn't publish new tag... to docker hub

@sjmiller609 any thoughts?

PR #790, opened 2020-08-04
Add a pull request template

sjmiller609:
The point of this PR is to help answer a question like "what patches went into 0.16.4"?


.github/pull_request_template.md
```diff
@@ -0,0 +1 @@
+Please make the PR title informative to the user experience. For example when adding a feature, say "Add metrics tab for Deployments" instead of "Bump Astro-UI to version X.Y.Z". Bugs are sometimes harder to make customer-facing because they are very detailed, but do what you can. Instead of "Fix metrics cardinality bug" (user does not know what 'cardinality' means and does not matter to them), say "Fix bug slowing down the metrics tab". The PR title will be used for the commit message when it's merged, and these commit messages are used to generate customer-facing release notes.
```
andriisoldatenko, line 1: @adam2k can you suggest template?

adam2k, line 1: @sjmiller609 we had some parallel thinking.  I don't know if you wanted to be more granular with this and divide some of this into sections.  I have an example in Houston-API https://github.com/astronomer/houston-api/blob/main/.github/pull_request_template.md I think it'd be nice if we have a similar format throughout our repos.

adam2k: 😂 we had the same idea!

adam2k:
@sjmiller609 we are getting more input on the Houston-API template right now, so we should have some improvements soon.

adam2k:
@sjmiller609 the template is updated in Houston-API (main) if you want to take a look now and do something similar: https://github.com/astronomer/houston-api/blob/main/.github/pull_request_template.md

sjmiller609:
merged before CI because it only changes PR template, and it's best to get this in quickly because we are in a release cycle where a lot of people are PRing

sjmiller609:
backported to 0.16

PR #794, opened 2020-08-05
Add Pod Disruption Budget for NATS and STAN

adam2k:
**Context**
Astro CLI is occasionally reporting `Error: connection closed` on `create` and `delete` deployment commands.  This most likely affects Astro UI as well, but errors there haven't been reported yet. 

**Propsed Solution**
Currently NATS and STAN can be down for a short amount of time and our system will continue to report that it's healthy.  We need NATS and STAN to have `maxUnavailable: 1` to function similarly to Commander.

### DIFF
--- a/charts/astronomer/templates/houston/worker/houston-worker-pod-disruption-budget.yaml
+++ b/charts/astronomer/templates/houston/worker/houston-worker-pod-disruption-budget.yaml
@@ -1,0 +1,18 @@
+###################################################
+## Astronomer Houston Worker Pod Disruption Budget
+###################################################
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ .Release.Name }}-houston-worker-pdb
+spec:
+  {{- if lt ( .Values.houston.worker.replicas | int ) 4 }}
+  maxUnavailable: 1
+  {{- else }}
+  maxUnavailable: {{ .Values.houston.worker.replicas.maxUnavailable }}
+  {{- end }}
+  selector:
+    matchLabels:
+      tier: astronomer
+      component: houston-worker
+      release: {{ .Release.Name }}
--- a/charts/astronomer/templates/houston/worker/houston-worker-pod-disruption-budget.yaml
+++ b/charts/astronomer/templates/houston/worker/houston-worker-pod-disruption-budget.yaml
@@ -6,10 +6,12 @@ kind: PodDisruptionBudget
 metadata:
   name: {{ .Release.Name }}-houston-worker-pdb
 spec:
-  {{- if lt ( .Values.houston.worker.replicas | int ) 4 }}
-  maxUnavailable: 1
+  {{- if lt ( .Values.houston.worker.replicas | int ) 2 }}
+  minAvailable: 0
+  {{- else if lt ( .Values.houston.worker.replicas | int ) 4 }}
+  minAvailable: 1
   {{- else }}
-  maxUnavailable: {{ .Values.houston.worker.replicas.maxUnavailable }}
+  minAvailable: "25%"
   {{- end }}
   selector:
     matchLabels:
--- a/charts/nats/templates/pod-disruption-budget.yaml
+++ b/charts/nats/templates/pod-disruption-budget.yaml
@@ -9,6 +9,6 @@ spec:
   maxUnavailable: 1
   selector:
     matchLabels:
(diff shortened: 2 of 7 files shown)


charts/nats/templates/pod-disruption-budget.yaml
```diff
@@ -0,0 +1,14 @@
...
+##############################
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ .Release.Name }}-nats-pdb
+spec:
+  maxUnavailable: 1
+  selector:
+    matchLabels:
+      tier: logging
```
jpweber, line 12: The tier here and on the Stan file should be something other than logging.

sjmiller609: good call https://github.com/astronomer/astronomer/blob/b0abcf6c5f0765d58309ff97bfa07dfe051a5e20/charts/nats/templates/statefulset.yaml#L32

jpweber: Doesn't look like any of the other nats charts have a `tier:` label. So just make this `nats` for both the nat template and stan template since we think of them both as part of the nats layer

adam2k: I realized I do need to add the Houston Worker too.

jpweber: Cool LGTM

adam2k: oops...just saw the next comment.   I'll change from `astronomer` to `nats`.

sjmiller609:
for example, this PDB:
```
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: nginx-pdb
spec:
  minAvailable: 1
  selector:
    matchLabels:
      app: nginx
```
is saying 'make sure at least 1 pod with label app: nginx is up at all times'

I think that our label selectors are not matching the pod spec. For example, stan's labels are:
```
Labels:         app=astronomer-stan
                chart=stan-0.4.2
                controller-revision-hash=astronomer-stan-84d667b54c
                security.istio.io/tlsMode=istio
                statefulset.kubernetes.io/pod-name=astronomer-stan-0
```

adam2k:
> I'm not sure the labels are matching

Are you talking about for STAN `component: {{ template "stan.name" . }}` or for NATS: `component: {{ template "nats.name" . }}` or both?

sjmiller609:
I think 'release' label should be added here: https://github.com/astronomer/astronomer/blob/b0abcf6c5f0765d58309ff97bfa07dfe051a5e20/charts/nats/templates/statefulset.yaml#L32 and we should match 'release' and 'app' labels. same for other components.

charts/astronomer/templates/houston/worker/houston-worker-pod-disruption-budget.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ .Release.Name }}-houston-worker-pdb
+spec:
+  {{- if lt ( .Values.houston.worker.replicas | int ) 4 }}
+  maxUnavailable: 1
+  {{- else }}
+  maxUnavailable: {{ .Values.houston.worker.replicas.maxUnavailable }}
+  {{- end }}
```
sjmiller609, line 13:
```suggestion
  {{- if lt ( .Values.houston.worker.replicas | int ) 2 }}
  minAvailable: 0
  {{- else if lt ( .Values.houston.worker.replicas | int ) 4 }}
  minAvailable: 1
  {{- else }}
  minAvailable: "25%"
  {{- end }}
```
I think that Values.houston.worker.replicas.maxUnavailable is not an existing configuration, so this doesn't seem quite right. But also, do we need this availability level for this component if the expected behavior is dequeuing from nats? i.e. if the pods are down that doesn't interrupt service because they come back up, then continue dequeuing.

adam2k: ah..that's true.  They will be able to replay from NATS.  

I was mainly trying to match this with what we were doing for Houston-API.  I think that the workers should be highly available because eventually deployments won't go out if there aren't any workers up to receive the messages and run the commander calls to deploy.

PR #795, opened 2020-08-05
fix unit tests

andriisoldatenko:


jpweber:
Already ahead of it and its in an running. :)

PR #797, opened 2020-08-06
Update astronomer patch version

adam2k:
Add the 0.18.1 patch version.

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -23,7 +23,7 @@ images:
     pullPolicy: IfNotPresent
   houston:
     repository: astronomerinc/ap-houston-api
-    tag: 0.18.0
+    tag: 0.18.1
     pullPolicy: IfNotPresent
   astroUI:
     repository: astronomerinc/ap-astro-ui


charts/astronomer/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: astronomer
-version: 0.18.0
+version: 0.18.1
```
sjmiller609, line 3: This is not necessary because 0.18.0 is not released https://github.com/astronomer/astronomer/releases
```suggestion
version: 0.18.0
```

Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: astronomer
-version: 0.18.0
-appVersion: 0.18.0
+version: 0.18.1
+appVersion: 0.18.1
```
sjmiller609, line 4:
```suggestion
version: 0.18.0
appVersion: 0.18.0
```

kushalmalani:
We dont need this PR then?

Chart.yaml
adam2k: How do we pick up the Houston API changes if it's not going off of the version number here?

sjmiller609:
this is the line that should change: https://github.com/astronomer/astronomer/blob/4021e01077d714b70d7bfeb0e800f94bd5c4892f/charts/astronomer/values.yaml#L26

sjmiller609:
the reasoning here is that we are currently seeking a stable code version in order to release 0.18.0 of astronomer helm chart. the first patch of houston is requiring a patch. so, we have a new code version 0.18.1 of houston that is now added into 0.18.0 astronomer chart. we will keep patching the software until it all works together, then we can release astronomer chart 0.18.0

PR #799, opened 2020-08-06
Network policy adjustments

sjmiller609:
## Description

This PR adds a test to notify the developer when there is a new network port that has become accessible to an arbitrary pod. It tidies up a few network policies that were missing.

I mentioned in slack that I would default false for the default deny policy, but since it is namespace scoped, I figure default true is appropriate.

Here is a sample failure, showing the results of the network scan: https://app.circleci.com/pipelines/github/astronomer/astronomer/1190/workflows/d91661bb-0f67-425f-bdeb-9f0b1a8f404d/jobs/10296/parallel-runs/0/steps/0-104

notes:

on both eks and kubernetes-in-docker, network policies do not work by default

## 🎟 Issue(s)

Closes: https://github.com/astronomer/issues/issues/1134

## 🧪  Testing

I have added a functional testing system to check the network is not leaving open ports. The functional testing will help identify if any new policy is breaking something, but a complete end to end test is really necessary to check everything.

## 📸 Screenshots

## 📋 Checklist


sjmiller609:
something is breaking on account of adding the default deny, I need to reproduce on my end and dig in a bit more. I didn't notice the issue on aws

danielhoherd (approved):

.gitignore
```diff
@@ -1,11 +1,14 @@
...
+*.key
 config.yaml
 config.yml
 dev-config.yaml
 !.circleci/config.yml
 **kubeconfig**
 certs
 /secrets
+test-*
+*__pycache__*
```
danielhoherd, line 11:
```suggestion
__pycache__
```

AFAIK you can just have `__pycache__` here. Was there another use case than matching a whole dir name?

sjmiller609:
for some reason the prometheus scape on fluentd is non-deterministically failing, but I don't know why yet. I haven't reproduced locally, and when i'm in CI with ssh it is not reproduced (yet).

sjmiller609:
I think I got it. The problem is that there is a race condition between prometheus and fluentd. If prometheus is up first, then the initial scrape fails on fluentd, which is OK, but the test fails because it runs before the next scrape is performed. By setting a sleep to twice the scrape configuration in prometheus, fluentd shows up in the targets list as expected.

PR #800, opened 2020-08-07
Fix Houston API pod disruption budget to not match Houston cron jobs

sjmiller609:
closes: https://github.com/astronomer/issues/issues/1229

houston PDB is accidentally patching the labels on the these jobs, this makes the houston rollout get 'stuck' for users with low replica count. For example if PDB says maxUnavailable: 1, then it will remain stuck as long as any of these jobs are in a non-ready state, including "completed"

I'm going to leave this in draft, I want to see if I can use the new unit testing framework to capture the logic 'no pods match these labels except the ones we want'

This is optionally in 0.18 and ought to be backported to 0.16, since it is a bug fix.

### DIFF
--- a/charts/astronomer/templates/houston/cronjobs/houston-expire-deployments-cronjob.yaml
+++ b/charts/astronomer/templates/houston/cronjobs/houston-expire-deployments-cronjob.yaml
@@ -20,7 +20,7 @@ spec:
         metadata:
           labels:
             tier: astronomer
-            component: houston
+            component: houston-expire-deployments
             release: {{ .Release.Name }}
           {{- if .Values.global.istio.enabled }}
           annotations:
--- a/charts/astronomer/templates/houston/cronjobs/houston-expire-deployments-cronjob.yaml
+++ b/charts/astronomer/templates/houston/cronjobs/houston-expire-deployments-cronjob.yaml
@@ -8,7 +8,7 @@ metadata:
   name: {{ .Release.Name }}-houston-expire-deployments
   labels:
     tier: astronomer
-    component: houston
+    component: houston-expire-deployments
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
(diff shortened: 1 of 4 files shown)


jpweber:
Good find on this one.

jpweber:
sort of thinking out loud here, the cleanup job doesn't talk to anything except the kube api really right? With the change of the labels it made me thinking about houston network policies and does this need to talk to anything that it won't have an allow policy for now. I don't _think_ it does though so we're probably good on that front.

sjmiller609:
> sort of thinking out loud here, the cleanup job doesn't talk to anything except the kube api really right? With the change of the labels it made me think about houston network policies and does this need to talk to anything that it won't have an allow policy for now. I don't _think_ it does though so we're probably good on that front.

they do talk to the DB through prisma, and I think your are correct this would cause an issue there. I will confirm that in the network policy pr that I also have open

charts/astronomer/tests/houston_pdb_test.yaml
```diff
@@ -0,0 +1,46 @@
...
+        value: houston
+      template: houston/cronjobs/houston-cleanup-deployments-cronjob.yaml
+    - notEqual:
+        path: spec.jobTemplate.spec.template.metadata.labels.component
+        value: houston
+      template: houston/cronjobs/houston-check-updates-cronjob.yaml
+    - notEqual:
+        path: spec.selector.matchLabels.component
+        value: houston
+      template: houston/worker/houston-worker-deployment.yaml
```
sjmiller609, line 46: it would be nice if I could accomplish "no other pods have the labels", but I don't think this kind of logic can be done, instead it's only best at asserting directly with 'full path' you might call it

charts/astronomer/templates/houston/cronjobs/houston-expire-deployments-cronjob.yaml
```diff
@@ -20,7 +20,7 @@ spec:
         metadata:
           labels:
             tier: astronomer
-            component: houston
+            component: houston-expire-deployments
```
sjmiller609, line 23: writing the unit test was helpful, it made me notice that we also need this label on the pod, just just the cronjob

sjmiller609:
@danielhoherd will you please review this one when you get the chance?

sjmiller609:
picked back to 0.16 https://github.com/astronomer/astronomer/commit/5058697f0ed8b74e850a7e48221cfc587624f7d8

PR #801, opened 2020-08-07
Minor fixes in LTS upgrade script

sjmiller609:
closes: https://github.com/astronomer/issues/issues/1590
closes: https://github.com/astronomer/issues/issues/1582


bin/migration-scripts/upgrade-to-lts.sh
```diff
@@ -296,6 +298,12 @@ function helm2_to_3 {
 function interactive_confirmation {
   echo
   echo
+  echo "Is tiller in the namespace $TILLER_NAMESPACE ?"
+  read -r -p "(y/n) " CONT
+  if ! [ "$CONT" = "y" ]; then
+    echo "Please modify the TILLER_NAMESPACE environment variable at the top of the script."
```
danielhoherd, line 304: Instead of asking people to edit the script, we should handle the custom namespace through either an env var, a command like arg, or both.

bin/migration-scripts/upgrade-to-lts.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+TILLER_NAMESPACE=kube-system
```
danielhoherd, line 3:
```suggestion
: "${TILLER_NAMESPACE:=kube-system}"
```

bin/migration-scripts/upgrade-to-lts.sh
sjmiller609, line 304:
```suggestion
    echo "Please modify the TILLER_NAMESPACE environment variable"
```

PR #805, opened 2020-08-11
Bump UI and API images for 0.16 patch release

ryanahamilton:


vishwas-astro:
@jpweber - Looks like the Astronomer chart version is already updated in the 0.16 release branch by Steven here: https://github.com/astronomer/astronomer/commit/4dca3b51ecf8950144350f7d4a4cd01a910c5b60 and we may not need to up it to 0.16.6.

PR #806, opened 2020-08-12
add missing keda namespaceLabels

andriisoldatenko:
based on slack conversation:
```
workers:
  autoscaling:
    enabled: false
    maxReplicas: 10
    minReplicas: 1
    targetCPUUtilization: 80
    targetMemoryUtilization: 80
  keda:
    cooldownPeriod: 30
    enabled: true
    namespaceLabels:
      name: astronomer # <-- add this line
    pollingInterval: 5
```
for ref https://github.com/astronomer/issues/issues/1625

### DIFF
--- a/charts/astronomer/templates/houston/api/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/api/houston-configmap.yaml
@@ -207,6 +207,8 @@ data:
           # This is here for upgrading to 0.11.1+ of Airflow.
           keda:
             enabled: false
+            namespaceLabels:
+              name: {{ .Release.Name }}
 
           # This is here for upgrading to 0.11.2+ of Airflow.
           safeToEvict: true


charts/astronomer/templates/houston/api/houston-configmap.yaml
```diff
@@ -207,6 +207,8 @@ data:
           # This is here for upgrading to 0.11.1+ of Airflow.
           keda:
             enabled: false
+            namespaceLabels:
+              name: {{ .Release.Name }}
```
schnie, line 211: Just to be sure, we want Release.Name (as you have) for this and not Release.Namespace right?

andriisoldatenko: good point... @dimberman wdyt?

dimberman:
@schnie @andriisoldatenko it should be whatever the label is on the astronomer namespace. If I recall that might be somewhere in the terraform as the helm chart doesn't create the namespace.

charts/astronomer/templates/houston/api/houston-configmap.yaml
danielhoherd: Houston does it exactly like this. https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/houston/api/houston-configmap.yaml#L176-L177

schnie:
Feels like we should document this behavior and configuration somewhere. If someone doesn't use the terraform, they'll have to take an additional step to add the namespace label right?

charts/astronomer/templates/houston/api/houston-configmap.yaml
schnie: Yep @danielhoherd that link represents labels that we want to apply to every airflow namespace. I believe the new usage here under KEDA is to tell the `pgbouncer` `NetworkPolicy` to allow traffic from the KEDA controller inside the platform namespace. It needs to run a query on the deployment dbs, so it connects to these to do so.

andriisoldatenko:
@schnie not sure that i fully understand what’s need to be done from my side…
not sure where and what document and how to check terraform part…

schnie:
I think we need to document this part on our installation docs - 


```
$ kubectl label namespace astronomer name=astronomer
namespace/default labeled
$ kubectl describe namespace astronomer
Name:         default
Labels:       name=astronomer
Annotations:  <none>
Status:       Active

No resource quota.

No LimitRange resource.
```

@andriisoldatenko is this what a user would need to do to use this? Looks like this label was somehow applied to dev.

schnie:
I updated the notion doc with this step https://www.notion.so/astronomerio/Using-KEDA-on-Astronomer-e15473b165d14443b17311def6979924

schnie:
Does anyone know how this label was added to dev? I can't find it in our terraform, so I'm wondering if it was manually applied.

If it doesn't exist, we should automate this step in our terraform.

jpweber:
It might have just been added manually via kubectl  "for testing".

vishwas-astro:
@jpweber before we forget about it, we need to make sure to add all the necessary things for KEDA support into all our terraforms. Do you mind creating an issue for that please?

schnie:
Created the issue over here https://github.com/astronomer/issues/issues/1654

vishwas-astro:
Never mind. @schnie already created it.

PR #809, opened 2020-08-17
Fix a few places where release name and namespace were conflated

sjmiller609:
wrt https://github.com/astronomer/issues/issues/1607

### DIFF
--- a/charts/stan/templates/statefulset.yaml
+++ b/charts/stan/templates/statefulset.yaml
@@ -66,14 +66,14 @@ spec:
             - "dockerize"
           args:
             - "-wait"
-            - "tcp://astronomer-nats:4222"
+            - "tcp://{{ .Release.Name }}-nats:4222"
             - "-timeout"
             - "1m"
           image: astronomerinc/ap-base:3.12-1
           imagePullPolicy: IfNotPresent
           env:
             - name: NATS__URL
-              value: nats://astronomer-nats:4222
+              value: "nats://{{ .Release.Name }}-nats:4222"
       containers:
         - name: stan
           image: {{ .Values.stan.image }}
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,8 @@ unittest-charts:
 
 .PHONY: lint-charts
 lint-charts:
+	# Check that nothing accidentally is using release name instead of namespace for metadata.namespace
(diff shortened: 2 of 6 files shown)


sjmiller609:
I'm considering for a few minutes how we might use the unit testing for this situation in the future, but I'm not sure if it works well for not explicting calling out specific resources, i.e. some logic like "nothing should have namespace: {{ .Release.Name }}"

sjmiller609:
I think the issue with the unit testing is that it does not play nicely to match many template files at once. It assumes that your assertion is only intended to apply to one template.

danielhoherd:
> What are we doing elsewhere for these metadata namespace fields? I wonder if we should remove them altogether. Wouldn't everything automatically be set to the right namespace?

Seems to me like this would be a good approach except where an alternate namespace is explicitly needed.

danielhoherd:
> I wonder if we should remove them altogether. Wouldn't everything automatically be set to the right namespace?

https://github.com/helm/helm/issues/5465 Says to not include them.

> In general, templates should not define a namespace. This is because Helm installs objects into the namespace provided with the --namespace flag.

schnie:
Cool, seems like we should just rip them out then?

sjmiller609:
> Cool, seems like we should just rip them out then?

Can we make that a separate github issue?

schnie:
I'm just talking about the ones that are changed in this PR. Can we just remove these, rather than changing them? The rest of the charts should already be OK and not have these specified.

sjmiller609:
> I'm just talking about the ones that are changed in this PR. Can we just remove these, rather than changing them? The rest of the charts should already be OK and not have these specified.

ok sgtm, thanks

charts/stan/templates/statefulset.yaml
```diff
@@ -66,14 +66,14 @@ spec:
             - "dockerize"
           args:
             - "-wait"
-            - "tcp://astronomer-nats:4222"
+            - "tcp://{{ .Release.Name }}-nats:4222"
```
sjmiller609, line 69: also noticed this spot was hardcoded

schnie:
LGTM

sjmiller609:
> LGTM

thank you, i'll merge if CI passes

sjmiller609:
picked into 0.16 https://github.com/astronomer/astronomer/commit/67cf7efcdc04c529663d12b025167d90dad64d7d

PR #811, opened 2020-08-18
Fix scrape configuration for blackbox exporter

sjmiller609:
Found another issue related to conflating namespace and release name.

Closes: https://github.com/astronomer/issues/issues/1680

image

### DIFF
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -564,7 +564,7 @@ data:
           - source_labels: [__param_target]
             target_label: instance
           - target_label: __address__
-            replacement: {{.Release.Name}}-prometheus-blackbox-exporter.{{.Release.Name}}:9115  # The blackbox exporter's real hostname:port.
+            replacement: {{.Release.Name}}-prometheus-blackbox-exporter.{{.Release.Namespace}}:9115  # The blackbox exporter's real hostname:port.
 
 
       {{- if .Values.tcpProbe.enabled}}
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -12,7 +12,7 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 data:
-  prometheus.yaml: |-
+  config: |-
     global:
       scrape_interval: 30s
       evaluation_interval: 30s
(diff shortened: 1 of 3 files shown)


sjmiller609:
gonna add a unit test

charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 data:
-  prometheus.yaml: |-
+  config: |-
```
sjmiller609, line 15: this change was made in order to enable the unit test

charts/prometheus/tests/prometheus-config-configmap_test.yaml
```diff
@@ -3,7 +3,28 @@ suite: Test prometheus-config-configmap
...
+      namespace: my-namespace
+    set:
+      global.blackboxExporterEnabled: true
+      global.veleroEnabled: true
+      global.prometheusPostgresExporterEnabled: true
+      global.nodeExporterEnabled: true
+      tcpProbe.enabled: true
+    asserts:
+      - notMatchRegex:
+          path: data.config
```
sjmiller609, line 26: this is the part that was not working with 'prometheus.yaml' key, including a dot in the key name

charts/prometheus/tests/prometheus-config-configmap_test.yaml
```diff
@@ -3,7 +3,28 @@ suite: Test prometheus-config-configmap
...
+          pattern: '.*\.my-release\..*'
+      - notMatchRegex:
+          path: data.config
+          pattern: '[^\^]+my-namespace-.*'
```
sjmiller609: This tests regex matches the prometheus config to error if something like this is detected:

'my-namespace-'
or
'.my-release.'
or
'.my-release:'

typically surrounded by dots or colon means it's supposed to be namespace, and also typically if a dash follows then it should be release name.

charts/prometheus/tests/prometheus-config-configmap_test.yaml
```diff
@@ -3,7 +3,31 @@ suite: Test prometheus-config-configmap
...
+    asserts:
+      - notMatchRegex:
+          path: data.config
+          pattern: '.*\.my-release\..*'
+      - notMatchRegex:
+          path: data.config
+          pattern: '.*\.my-release:.*'
+      - notMatchRegex:
+          path: data.config
+          pattern: '[^\^]+my-namespace-.*'
```
sjmiller609, line 33: This tests regex matches the prometheus config to error if something like this is detected:

'my-namespace-'
or
'.my-release.'
or
'.my-release:'

typically surrounded by dots or colon means it's supposed to be namespace, and also typically if a dash follows then it should be release name.

PR #821, opened 2020-08-24
Fix network policy to allow houston upgrader to connect to DB

sjmiller609:
## Description

The houston upgrader pod needs network policy to allow connection. This bug is caused by my other PR that adjusted the labels on the houston upgrader pod to not match the houston pod disruption budget.

## 🎟 Issue(s)

hot fix before release, no associated issue

## 🧪  Testing

I turned this part on by default in CI runs.

## 📸 Screenshots

### DIFF
--- a/bin/install-platform
+++ b/bin/install-platform
@@ -88,8 +88,8 @@ helm ${ACTION} \
   --set tags.kubed=${kubed} \
   --set global.postgresqlEnabled=${postgresql} \
   --set tags.keda=${keda} \
+  --set astronomer.houston.upgradeDeployments.enabled=true \
   $HELM_CHART_PATH
-
 
 if [ ! $? -eq 0 ]; then
   set +x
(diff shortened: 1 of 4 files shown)


sjmiller609:
I think I can add a helm upgrade right after helm install in CI to accomplish testing for this same kind of failure (helm upgrader won't start). I will do that.

bin/install-platform
```diff
@@ -88,9 +88,9 @@ helm ${ACTION} \
   --set tags.kubed=${kubed} \
   --set global.postgresqlEnabled=${postgresql} \
   --set tags.keda=${keda} \
+  --set astronomer.houston.upgradeDeployments.enabled=true \
```
sjmiller609, line 91: This line enables the upgrader to run in CI. The upgrader also has the post-install hook.

sjmiller609:
backported to both 0.16 and 0.19

PR #823, opened 2020-08-25
Reconnect correctly for NATS in Houston

adam2k:
## Description

Fixes issues with Houston not correctly reconnecting with NATS.

## PR Title

## 🎟 Issue(s)

Resolves astronomer/issues#1722

## 🧪  Testing

The NATS implementation is changing.  We have tested these connections as thoroughly as possible in dev to make sure that the connections are handled correctly. 

N/A

This easiest to way test this is to delete all of the `astrodev-nats-X` pods in dev and view the Houston Worker & Houson API logs to make sure that the re-connection events are smooth (you should not see several "Attempting to reconnect.." logs).

## 📸 Screenshots

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -23,7 +23,7 @@ images:
     pullPolicy: IfNotPresent
   houston:
     repository: astronomerinc/ap-houston-api
-    tag: 0.19.6
+    tag: 0.19.7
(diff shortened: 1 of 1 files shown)


andriisoldatenko (approved):

sjmiller609:
picked into 0.19

PR #824, opened 2020-08-25
Change to NATS servers array instead of single NATS url

andriisoldatenko:
## PR Title

Switch to NATS__SERVERS, related to https://github.com/astronomer/houston-api/pull/436
## 🎟 Issue(s)

Resolves astronomer/issues#1722

## 🧪  Testing

```
helm install --dry-run --debug -f ./configs/local-dev.yaml astronomer --namespace astronomer --set global.postgresqlEnabled=true --set global.stanEnabled=true . | grep -A2 NATS__SERVERS
install.go:158: [debug] Original chart version: ""
install.go:175: [debug] CHART PATH: /Users/andrii/workspace/src/github.com/astronomer/astronomer
install.go:180: [debug] WARNING: This chart or one of its subcharts contains CRDs. Rendering may fail or contain inaccuracies.
            - name: NATS__SERVERS
              value: '["astronomer-nats-0.astronomer-nats.astronomer.svc:4222","astronomer-nats-1.astronomer-nats.astronomer.svc:4222","astronomer-nats-2.astronomer-nats.astronomer.svc:4222"]'
            - name: NATS__CLUSTER_ID
--
            - name: NATS__SERVERS
              value: '["astronomer-nats-0.astronomer-nats.astronomer.svc:4222","astronomer-nats-1.astronomer-nats.astronomer.svc:4222","astronomer-nats-2.astronomer-nats.astronomer.svc:4222"]'
            - name: NATS__CLUSTER_ID
```


charts/astronomer/templates/_helpers.yaml
```diff
@@ -304,3 +304,12 @@ s3:
...
 {{- end }}
+
+{{- define "houston.nats.servers"}}
+{{- $ns := dict "servers" (list) -}}
+{{- $name := printf "%s-nats" .Release.Name -}}
+{{- range $i, $e := until (.Values.global.natsReplicas | int) -}}
+{{- $noop := printf "\"%s-%d.%s.%s.svc:4222\"" $name $i $name $.Release.Namespace | append $ns.servers | set $ns "servers" -}}
+{{- end }}
+{{- printf "'[%s]'" (join "," $ns.servers) -}}
+{{- end }}
```
schnie, line 315:
```suggestion
{{- end }}

```

values.yaml
```diff
@@ -51,6 +51,7 @@ global:
 
   # Used to enable nats-server
   natsEnabled: true
+  natsReplicas: 3
```
schnie, line 54: @andriisoldatenko what do you think about moving this config to the nats sub package? These global, top-level flags are mostly used to enable/disable the package itself from the umbrella installation.

andriisoldatenko: sgtm. but can you pls suggest how to reference to it? i mean how to share it between nats and houston charts?

charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -95,10 +95,6 @@ spec:
             failureThreshold: {{ .Values.houston.readinessProbe.failureThreshold }}
           env:
             {{- include "houston_environment" . | indent 12 }}
-            - name: NATS__URL
```
andriisoldatenko, line 98: it's already inside `houston_environment `

values.yaml
andriisoldatenko: @schnie ☝️

schnie: Ohh - hmm, missed that part. Should we maybe have a `nats` block, like the istio block up above?

charts/astronomer/templates/_helpers.yaml
```diff
@@ -304,3 +306,12 @@ s3:
...
 {{- end }}
 {{- end }}
+
+{{- define "houston.nats.servers"}}
+{{- $ns := dict "servers" (list) -}}
+{{- $name := printf "%s-nats" .Release.Name -}}
+{{- range $i, $e := until (.Values.global.natsReplicas | int) -}}
+{{- $noop := printf "\"%s-%d.%s.%s.svc:4222\"" $name $i $name $.Release.Namespace | append $ns.servers | set $ns "servers" -}}
+{{- end }}
+{{- printf "'[%s]'" (join "," $ns.servers) -}}
```
adam2k, line 316: Because this is formatted as: 
```
'["astronomer-nats-0.astronomer-nats.astronomer.svc:4222","astronomer-nats-1.astronomer-nats.astronomer.svc:4222","astronomer-nats-2.astronomer-nats.astronomer.svc:4222"]'
```

It will require the application code to parse this JSON object or is this coming in as the same format as the default yaml?  

```
servers: ["nats://localhost:4222"]
```

charts/astronomer/templates/_helpers.yaml
```diff
@@ -237,8 +237,10 @@ s3:
   value: {{ .Release.Name }}-registry
 - name: REGISTRY__PORT
   value: {{ .Values.ports.registryHTTP | quote }}
-- name: NATS__URL
-  value: nats://{{ .Release.Name }}-nats:4222
+- name: NATS__SERVERS
+  value: {{ template "houston.nats.servers" . }}
+- name: NATS__CLUSTER_ID
+  value: {{ .Release.Name }}-stan
```
adam2k, line 243: @andriisoldatenko this is where we are getting logs for `stan.clusterID` value from in Houston API.  I don't have a way with the NATS library to receive the clusterID for NATS too.  We're only getting the URLs that we are using for NATS from the `servers` reference OR from the NATS auto discovery (what you were demoing yesterday with the automatic update of the server urls).  So it seems like STAN is giving us the NATS pod urls and then we are able to make that NATS client connection from the library directly to NATS Server.

charts/astronomer/templates/_helpers.yaml
andriisoldatenko: we need json string here more details in https://github.com/astronomer/houston-api/pull/436/files#diff-b3e5513342454bc95092ef963d64b431R16

charts/astronomer/templates/_helpers.yaml
adam2k: This looks like it'll resolve correctly with the changes were recently merged within Houston that changes the schema to JSON.

schnie:
This looks fine. I think most of the env vars we specify in Houston are secrets and get mounted from a secret. Any other values that need to be configured are set up here in Houston's configmap - https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/houston/api/houston-configmap.yaml.

This obviously works fine, but just one small convention we have now.

andriisoldatenko:
@schnie my fault let me fix it!

charts/astronomer/templates/_helpers.yaml
adam2k: This isn't a requested change.  I'm wondering if we should change what we're logging in Houston API/Houston Worker and if it's important to show the cluster ID in the logs like they are currently implemented:

Example (note: `astrodev-stan`):
```
2020-08-24T18:27:05 INFO Reconnected to cluster astrodev-stan using client houston-deployment-deleted-worker-66dc5f1b-5ca4-4cb6-9513-fc4776e63e8a and url nats://astrodev-nats:4222
```

charts/astronomer/tests/houston_api_houston-deployment_test.yaml
```diff
@@ -6,3 +6,33 @@ tests:
   asserts:
     - isKind:
         of: Deployment
+- it: should work with 0 nats servers
```
adam2k, line 9: Should this be?

```Suggestion
it: should not work with 0 nats servers
```

andriisoldatenko:
@schnie @adam2k @sjmiller609 @danielhoherd PTAL!

charts/astronomer/templates/_helpers.yaml
adam2k: Removed the `clusterID` here https://github.com/astronomer/houston-api/pull/438

andriisoldatenko:
depend on https://github.com/astronomer/houston-api/pull/438#pullrequestreview-475644439

charts/astronomer/tests/houston_api_houston-deployment_test.yaml
adam2k: This is good as-is.  I talked to @danielhoherd about this test.

sjmiller609:
I am good with this change, but I would like for us to add the functional test to check the nats configuration does not include "url" option. I think it should be just a small python function to parse the nats config, similar to this test: https://github.com/astronomer/astronomer/blob/3e6df588c78814f38e442759595f9782131ce533/bin/functional-tests/test_astronomer.py#L24 It will be very nice to check the running system to ensure the config has ended up as intended

andriisoldatenko:
@sjmiller609 good point, but i'm not familiar with your e2e test framework. 
@vishwas-astro wdyt? should we add now and block release or create follow-up ticket and add tests?

vishwas-astro:
> @sjmiller609 good point, but i'm not familiar with your e2e test framework.
> @vishwas-astro wdyt? should we add now and block release or create follow-up ticket and add tests?

I think we should add a couple of tests. @sjmiller609, if @andriisoldatenko is not familiar with the test framework, could you help with or pair with him / @adam2k to write the tests?

sjmiller609:
no problem, I will submit a commit

adam2k:
> > @sjmiller609 good point, but i'm not familiar with your e2e test framework.
> > @vishwas-astro wdyt? should we add now and block release or create follow-up ticket and add tests?
> 
> I think we should add a couple of tests. @sjmiller609, if @andriisoldatenko is not familiar with the test framework, could you help with or pair with him / @adam2k to write the tests?

@vishwas-astro the tests are in place now.  They're under `charts/astronomer/tests/houston_api_houston-deployment_test.yaml`.  Are you looking for additional tests?

bin/functional-tests/test_astronomer.py
```diff
@@ -36,8 +36,15 @@ def test_prometheus_user(prometheus):
     assert user == "nobody", \
         f"Expected prometheus to be running as 'nobody', not '{user}'"
 
-def test_houston_test_fixture(houston_api):
-    data = houston_api.check_output("whoami")
+def test_houston_config(houston_api):
+    """ Make assertions about Houston's configuration
+    """
+    data = houston_api.check_output("echo \"config = require('config'); console.log(JSON.stringify(config))\" | node -")
```
sjmiller609, line 42: @adam2k @schnie ty for your help on this bit of shell-ism. Should be easy now to add just one more test for other houston config things in future.

andriisoldatenko:
@sjmiller609 you turn!

sjmiller609:
picked into 0.19

PR #825, opened 2020-08-26
Collect metrics from Houston

sjmiller609:
## Description

Collect metrics from Houston's metrics endpoint

## 🎟 Issue(s)

Closes: https://github.com/astronomer/issues/issues/1727

## 🧪  Testing

Prometheus misconfiguration is the biggest risk. Adding to the functional testing to check that the metrics are collected as intended.

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -268,11 +268,11 @@ prometheus-blackbox-exporter:
   # Configure resources
   resources:
     requests:
-      cpu: "5m"
-      memory: "7Mi"
+      cpu: "50m"
+      memory: "70Mi"
     limits:
-      cpu: "15m"
-      memory: "20Mi"
+      cpu: "100m"
+      memory: "200Mi"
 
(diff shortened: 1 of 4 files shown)


values.yaml
```diff
@@ -268,11 +268,11 @@ prometheus-blackbox-exporter:
...
     requests:
-      cpu: "5m"
-      memory: "7Mi"
+      cpu: "50m"
+      memory: "70Mi"
     limits:
-      cpu: "15m"
-      memory: "20Mi"
+      cpu: "100m"
+      memory: "200Mi"
```
sjmiller609, line 275: I found the functional test that checks if all prometheus targets are available was failing non deterministically because this component was crashing OOM.

sjmiller609: this change is not related to the rest of the pr

PR #829, opened 2020-08-27
Postgres and node exporters PSPs controlled by global enable disable flag.

jpweber:
## Description
The posgres exporter and the node exporter pod security policies were both found to not honor the global `pspEnabled: <bool>` variable and were being applied when they should not have been. The Postgres exporter template and values have been modified to honor this global flag. The node exporter subchart was being pulled in as a requirement and was not actually in our repository. In order to make the required changes the node exporter chart has been pull in to our repository so that the templated could be modified. 

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/1687

## 🧪  Testing

Risks associated with this PR include the change not working, which would result in the current bug staying as is. 
While low, because the node exporter chart is now included rather than being pulled from an upstream public chart, there is a risk of node exporter doing something different than we expected. 

## 📸 Screenshots

With global `pspEnabled: false`
Screen Shot 2020-08-28 at 2 30 27 PM

With global `pspEnabled: true`
Screen Shot 2020-08-28 at 3 02 43 PM

Node exporters are running without issue
Screen Shot 2020-08-28 at 3 04 47 PM

Prometheus can scrape node exporter
Screen Shot 2020-08-28 at 3 06 01 PM

## 📋 Checklist


danielhoherd:
The prometheus-node-exporter chart is an external chart we pulled in? I don't see any unittests, and if we're going to maintain this we should add them.

jpweber:
Yes I pulled it in because I needed to modify the template part itself. I'll get working on adding tests then. Been struggling to actually get helm unittest to work with helm 3

jpweber:
@danielhoherd tests added please re-review.

jpweber:
This also needs back ported. to 0.16.x

PR #833, opened 2020-09-02
Kube-DNS metrics dashboard

sjmiller609:
## Description

Add a grafana dashboard for kube dns, also fix a bug in the scrape configuration

## 🎟 Issue(s)

Closes: https://github.com/astronomer/issues/issues/1696

## 🧪  Testing

The main risk is that grafana won't start (mitigated by smoke tests) or that the scrape config is still broken. The issue is that kind does not use kube-dns, so the scrape config can only be functionally validated in GKE.

## 📸 Screenshots

image

## 📋 Checklist

### DIFF
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -70,13 +70,10 @@ data:
         relabel_configs:
           - source_labels: [__meta_kubernetes_pod_container_name]
             action: keep
-            regex: "^kubedns"
-          - source_labels: [__meta_kubernetes_endpoint_ready]
-            action: keep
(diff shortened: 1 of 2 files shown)


charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -70,13 +70,10 @@ data:
         relabel_configs:
           - source_labels: [__meta_kubernetes_pod_container_name]
             action: keep
-            regex: "^kubedns"
-          - source_labels: [__meta_kubernetes_endpoint_ready]
-            action: keep
-            regex: "^true"
```
sjmiller609, line 76: this is the bug fix

PR #835, opened 2020-09-02
Minor bug fixes and updates for Astro CLI, Houston and Airflow Chart Version

adam2k:
## Description

Update to the latest version of the LTS CLI `0.16.4`, update Houston API to `0.16.12` and update the Airflow Chart version to `0.15.6`

## PR Title

## 🎟 Issue(s)

See the slack discussion #patchrelease-0-16-6

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -254,7 +254,7 @@ registry:
 
 install:
   resources: {}
-  cliVersion: 0.16.1
+  cliVersion: 0.16.4
   #  limits:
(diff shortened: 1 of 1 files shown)


andriisoldatenko (approved):

adam2k:
@kushalmalani is going to update some values on this PR.  Hold merge until he can add his changes and update the title.

kushalmalani (approved):

PR #837, opened 2020-09-04
Fixing setting mountAllFromSecretName on createDeployment

kushalmalani:
## Description

Fix setting mount all from secret name on create deployment

## PR Title

## 🎟 Issue(s)

Resolves astronomer/issues#XXXX

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist


sjmiller609:
i'll merge when CI is done

PR #838, opened 2020-09-08
Set StatefulSet to apps/v1 for k8s before 1.16 where it was removed. 

sophomeric:
It is available to use before that though.

https://v1-15.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#statefulset-v1-apps

Otherwise you can't update it to v1 before upgrading k8s itself, which is not compatible with the old apiVersion.

## Description

Set StatefulSet to apps/v1 before kubernetes 1.16. apps/v1beta2 was removed in 1.16, so you need to set it to apps/v1 before upgrading to 1.16.

## PR Title

Set StatefulSet to apps/v1 for k8s before 1.16 where it was removed. 

## 🧪  Testing

It is applied as apps/v1beta2 and could then stop working when you upgrade to Kubernetes 1.16.

## 📸 Screenshots

```
$ helm get all astronomer | egrep -A1 'kind: StatefulSet'
kind: StatefulSet
apiVersion: apps/v1beta2
$ helm list
NAME      	NAMESPACE 	REVISION	UPDATED                                	STATUS  	CHART            	APP VERSION
astronomer	astronomer	28      	2020-09-08 14:19:02.93415208 +0200 CEST	deployed	astronomer-0.16.4	0.16.4
```

https://github.com/FairwindsOps/pluto
```
$ pluto detect-helm --helm-version 3 -owide
NAME                                                 NAMESPACE      KIND                           VERSION                                REPLACEMENT                       DEPRECATED   DEPRECATED IN   REMOVED   REMOVED IN
astronomer/astronomer-registry                       astronomer     StatefulSet                    apps/v1beta2                           apps/v1                           true         v1.9.0          true      v1.16.0
```

This is checking helm data, as I did above it manually, to see what was actually applied. Using AWS EKS you can't actually query etcd directly to see what is stored there.

```
$ kubectl proxy --port 8080 &
$ curl -s http://127.0.0.1:8080/apis/apps/v1beta2/namespaces/astronomer/statefulsets/astronomer-registry | grep apiVersion
  "apiVersion": "apps/v1beta2",
$ !!:s/v1beta2/v1
curl -s http://127.0.0.1:8080/apis/apps/v1/namespaces/astronomer/statefulsets/astronomer-registry | grep apiVersion
  "apiVersion": "apps/v1",
```

And kube will respond to all supported apiVersions to make it even more confusing.

## 📋 Checklist

- [X] The PR title is informative to the user experience

### DIFF
--- a/templates/_apivers.tpl
+++ b/templates/_apivers.tpl
@@ -45,7 +45,7 @@ scheduling.k8s.io/v1beta1
 {{- end -}}
 
 {{- define "apiVersion.StatefulSet" -}}
-{{- if semverCompare "^1.16-0" .Capabilities.KubeVersion.Version -}}
+{{- if semverCompare "^1.15-0" .Capabilities.KubeVersion.Version -}}
 apps/v1
 {{- else -}}
 apps/v1beta2


jpweber:
I support this change. Sound reasoning.

jpweber (approved):

jpweber:
Not merging right away, waiting for review from @sjmiller609 and @danielhoherd

sophomeric:
I should also point out that DaemonSet and Deployment above it in the same file don't even have a conditional on them and are affected by the same apiVersion change. So maybe you would prefer to remove the conditional altogether?

jpweber:
@sophomeric yeah that conditional is on most if not all resource types. With regard to removing the conditional all together. It isn't something that has come up before and probably should be revisited again.

danielhoherd:
The change looks good to me, especially now that CI ran and passed.

Regarding the additional resources, we should either add them to this change set or we should remove the conditional altogether.

danielhoherd:
I've opened PR astronomer/astronomer/pull/839 to remove this logic entirely since it is not needed. Thanks for taking the time to submit a PR on this issue @sophomeric! 👏🙇‍♂️

PR #839, opened 2020-09-08
Remove apiVersion.StatefulSet, always use apps/v1

danielhoherd:
## Description

Remove helm template for apiVersion.StatefulSet which sometimes selected `apps/v1beta2` when `apps/v1` was available. This template was also only one of three places that should have had such a mechanic if it were useful (daemonset, statefulset, deployment).

This will be a no-op to the user, it merely takes out an unneeded conditional.

## PR Title

Remove apiVersion.StatefulSet, always use apps/v1

## 🎟 Issue(s)

Resolves astronomer/astronomer/pull/838 discussion points, and supersedes that PR.

## 🧪  Testing

There should be no risks associated with this change. the `apps/v1` endpoint has existed since way before any currently supported version.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
       - id: check-xml
       - id: check-yaml
         args: ['--allow-multiple-documents']
+        exclude: charts
       - id: debug-statements
       - id: detect-private-key
       - id: end-of-file-fixer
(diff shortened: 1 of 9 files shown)


sjmiller609 (approved):
Please backport to 0.16

sjmiller609:
backported to 0.16

danielhoherd:
cherry-pick to 0.16 was done in https://github.com/astronomer/astronomer/commit/4e4371ccf78392f055aeaef33d1601cebdee54bd , and I deleted .pre-commit-config.yaml which was included with the cherry-pick since that file did not exist in 0.16.

vishwas-astro:
@danielhoherd can you create a GH issue for this please? We have typically tracked and kept release notes for only GH issues in patch releases. 

Once you create issue, I will add "backport 0.16, 0.16.7" labels and mark it for 0.20 release.

PR #840, opened 2020-09-08
Use ubuntu 20.04 machine image

danielhoherd:
## Description

CI changes only. Use ubuntu 20.04 as machine image.

## 🧪  Testing

These changes are all related to CI, so the customer impact is zero as long as tests are passing.

## Related links

- <https://discuss.circleci.com/t/early-preview-new-ubuntu-20-04-linux-machine-executor-image/37140>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
 
   release-to-internal:
     machine:
-      image: ubuntu-1604:202004-01
+      image: ubuntu-2004:202008-01
     steps:
       - attach_workspace:
           at: /tmp/workspace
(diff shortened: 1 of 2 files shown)


danielhoherd:
The Ubuntu 20.04 machine image is currently in preview. I think it would be OK for us to use it anyway because our purposes are so limited, and I think a lot of what will change in the image isn't related to our CI workflows. At the same time, it's not an urgent change, so I don't really have strong opinions related to waiting for 20.04 image to become stable or merge now while it's in preview.

I have run into some issues where I wanted to use a shell command but the old syntax of ubuntu 16 didn't support it. EG: in ubuntu 16, when running `sudo -E` you have to inherit the entire environment, but in 20.04 you can inherit only specific variables from the environment. Also I think there were some syntaxes with the find command, but I don't recall. So really just small things.

jpweber (approved):

PR #843, opened 2020-09-10
Disallow publishing of future releases to public helm repository

danielhoherd:
## Description

Disallow publishing of future releases to public helm repository, github tag, and github release.

## 🎟 Issue(s)

Resolves astronomer/issues#1771

## 🧪  Testing

AFAIK it's not possible to test this without committing to at least two release branch (eg: release-0.16 and release-0.19). It's possible the regular expressions do not match due to the regex dialect used by CircleCI, but I have made an effort to reference other people's regex examples and think that this is correct.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,7 @@ jobs:
             pyenv global 2.7.17
             bin/release-helm-chart /tmp/workspace/astronomer-*.tgz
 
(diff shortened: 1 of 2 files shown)


jpweber (approved):

danielhoherd:
This change was cherry-picked to branches release-{16..19}

PR #844, opened 2020-09-10
1771 add missing slash

danielhoherd:
This is a quick follow up to #843 which left out a trailing slash at the end of the regex. I also found a way to test this, by pushing it as a future release branch. This was done so [here](https://app.circleci.com/pipelines/github/astronomer/astronomer?branch=release-0.999). The failing tests are due to me deleting the branch before they could run. The important details is that there are "approve-internal-release" and "release-to-internal" steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ workflows:
           filters:
             branches:
               only:
-                - '/release-0\.\d+'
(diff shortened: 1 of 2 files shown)


jpweber (approved):

danielhoherd:
This change was cherry-picked to release-0.{16..19}

PR #845, opened 2020-09-10
Add patch version upgrade testing

sjmiller609:
## Description

To make sure "helm upgrade" works between patch versions

## 🎟 Issue(s)

Closes https://github.com/astronomer/issues/issues/1622

## 🧪  Testing

This change is to add tests

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/bin/functional-tests/README.md
+++ b/bin/functional-tests/README.md
@@ -2,9 +2,10 @@
 
 ## Prerequisites
 
-- Kubernetes cluster, with Astronomer installed
 - Helm 3 installed as 'helm3'
 - Astronomer-internal is added to Helm 3 repositories and up-to-date
+- Kubernetes cluster, with Astronomer already installed from this chart in test
+- HELM_CHART_PATH environment variable is set to a path to the chart in test
 
 ## How does it work?
 
--- a/bin/functional-tests/README.md
+++ b/bin/functional-tests/README.md
@@ -4,6 +4,7 @@
 
 - Kubernetes cluster, with Astronomer installed
 - Helm 3 installed as 'helm3'
+- Astronomer-internal is added to Helm 3 repositories and up-to-date
 
 ## How does it work?
 
--- a/bin/functional-tests/README.md
+++ b/bin/functional-tests/README.md
@@ -1,0 +1,70 @@
+# Astronomer functional testing
+
+## Prerequisites
+
+- Kubernetes cluster, with Astronomer installed
+- Helm 3 installed as 'helm3'
+
+## How does it work?
+
+This is a pytest suite that assumes the current shell is authenticated to a kubernetes cluster running Astronomer. It is not safe to run against a production system and does perform mutations on the system. This is designed to work running locally using Kind (kubernetes in docker).
+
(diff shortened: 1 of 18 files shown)


bin/functional-tests/test_versioning.py
```diff
@@ -0,0 +1,103 @@
...
+    command = \
+        "helm3 upgrade --reset-values " + \
+        f"-f {config_path} " + \
+        f"-n {namespace} " + \
+        f"--version={major}.{minor}.{patch} " + \
+        f"{release_name} " + \
+        helm_chart_path
+    print(command)
+    print(check_output(command, shell=True))
+    print("The upgrade worked!")
```
sjmiller609, line 103: here is the upgrade test. it tests both downgrade and upgrade

bin/functional-tests/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+This function name starts with "test_", a pytest convention to indicate this is a pytest function. Pytest function parameters are test fixtures. Since this test has a parameter 'prometheus', pytest will match this name to a pytest fixture of the same name, and provide it as an argument in this function. In this case, it will consume the following test fixture:
+
+```
+@pytest.fixture(scope='session')
+def prometheus(request):
+    yield testinfra.get_host(f'kubectl://astronomer-prometheus-0?container=prometheus&namespace=astronomer')
+```
+
+This test fixture is making use of both 'pytest' and 'testinfra'. The pytest.fixture decorator indicate to pytest that this is a fixture. testinfra is used to simplify the connection logic. We can see that the test function above makes use of the testinfra host object (the result of the get_host call that is provided by this fixture) with the 'check_output' function:
```
danielhoherd, line 43:
```suggestion
This test fixture is making use of both 'pytest' and 'testinfra'. The `@pytest.fixture` decorator indicate to pytest that this is a fixture. testinfra is used to simplify the connection logic. We can see that the test function above makes use of the testinfra host object (the result of the get_host call that is provided by this fixture) with the `check_output` function:
```

bin/functional-tests/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+This line:
+```
```
danielhoherd, line 46: Switching the leading fence to \`\`\`python would give syntax highlighting

bin/functional-tests/test_versioning.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
```
danielhoherd, line 1: IMHO this should be run through `black` to clean up the formatting. We could add black to pre-commit, which would make it easy for us all to forget about how to format our code in favor of having it automatically formatted.

```yaml
  - repo: https://github.com/psf/black
    rev: 20.8b1
    hooks:
      - id: black
```

danielhoherd (approved):
Looks great! I left a few small improvement suggestions in case you want to do them.

bin/functional-tests/test_versioning.py
sjmiller609: Some pathing situation I think I need to get black configured correctly to find at the path. I'm going to skip for now

PR #849, opened 2020-09-16
Migrate to Prisma 2 for Houston API, Astro UI and minor fixes for Astro CLI

adam2k:
## Description

Release Houston API and Astro CLI version 0.20.0

## PR Title

## 🎟 Issue(s)

This is for the 0.20.0 release

## 🧪  Testing

We need to make sure the manual database changes go out to production.  Follow the same steps that were used while testing on AWS, Dev and eventually Staging.

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -27,7 +27,7 @@ images:
     pullPolicy: IfNotPresent
   astroUI:
     repository: astronomerinc/ap-astro-ui
-    tag: 0.19.3
+    tag: 0.20.0
     pullPolicy: IfNotPresent
   dbBootstrapper:
     repository: astronomerinc/ap-db-bootstrapper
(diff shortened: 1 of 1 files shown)


adam2k:
Please hold on merge until Houston API 0.20.0 finishes building.

charts/astronomer/values.yaml
```diff
@@ -23,7 +23,7 @@ images:
     pullPolicy: IfNotPresent
   houston:
     repository: astronomerinc/ap-houston-api
-    tag: main
+    tag: 0.20.0
```
vishwas-astro, line 26: @ryanahamilton @samblackk @adam2k can we update astro-ui in this PR itself? Please change title of PR accordingly.

adam2k: It's updated now.

sjmiller609 (approved):
Hey looks good, feel free to merge when the images are released. CI is pretty slow today for some reason. It's ok to merge even if CI is not working in this situation because the existing code base is using mutable tags anyways.

PR #857, opened 2020-09-17
Add Support for AC 1.10.12-1

kaxil:
We haven't released AC 1.10.12-1 yet but just created a PR so that we can release it with the next platform release. So when we do release AC 1.10.12-1 the platform already supports it.

cc @paolaperaza


vishwas-astro:
Based on https://astronomerteam.slack.com/archives/G017CC53EJG/p1600376063099200 - We can go ahead and ship this as part of 0.20 release even though 1.10.12-1 is not yet out.

sjmiller609 (approved):

PR #864, opened 2020-09-21
Upgraded Prometheus to version 2.21.0

jpweber:
## Description

This PR upgrades the version of prometheus in the astronomer chart to be version 2.21.0. The init container that was doing a `chown` as removed in favor of specifying user and group via a security context. 

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/1772

## 🧪  Testing

Risks when upgrading is changes or is processed differently. This was accounted for by deployed a full platform in AWS test cluster and creating deployments than checking that all the metrics we depend on exist. This was done by running all the queries that are used in Houston-api and ensuring they produce results.

All queries responded correctly except for those that are airflow metrics that are currently not producing results because of an unrelated issue. 

helm-unittest has been added for the new security context fields.

### DIFF
--- a/bin/functional-tests/README.md
+++ b/bin/functional-tests/README.md
@@ -6,8 +6,10 @@
 - Astronomer-internal is added to Helm 3 repositories and up-to-date
 - Kubernetes cluster, with Astronomer already installed from this chart in test
 - HELM_CHART_PATH environment variable is set to a path to the chart in test
+- pytest https://docs.pytest.org/en/stable/getting-started.html
+- testinfra https://testinfra.readthedocs.io/en/latest/
 
-## How does it work?
+## How it works
 
 This is a pytest suite that assumes the current shell is authenticated to a kubernetes cluster running Astronomer. It is not safe to run against a production system and does perform mutations on the system. This is designed to work running locally using Kind (kubernetes in docker).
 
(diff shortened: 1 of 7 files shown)


bin/functional-tests/README.md
```diff
@@ -6,8 +6,10 @@
 - Astronomer-internal is added to Helm 3 repositories and up-to-date
 - Kubernetes cluster, with Astronomer already installed from this chart in test
 - HELM_CHART_PATH environment variable is set to a path to the chart in test
+- pytest https://docs.pytest.org/en/stable/getting-started.html
+- testinfra https://testinfra.readthedocs.io/en/latest/
 
-## How does it work?
+## How does it work
```
danielhoherd, line 12: If we're taking away the question marks should we change this to a statement?
```suggestion
## How it works
```

bin/functional-tests/README.md
```diff
@@ -58,7 +60,7 @@ Long story short, testinfra allows for connections to various places (in this ca
 
 pytest fixtures can also be used to provide connections to kubernetes, for example a kubernetes client.
 
-## What does it do?
+## What does it do
```
danielhoherd, line 63: If we're taking away the question marks should we change this to a statement?
```suggestion
## What it does
```

danielhoherd (changes requested):

PR #870, opened 2020-09-24
Use NLB as default AWS load balancer

ianstanton:
## Description

- Use Network Load Balancer (NLB) as default AWS LB instead of Classic Load Balancer (default)
- AWS recommends migrating to the newer generation of load balancers

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/1546

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/charts/nginx/templates/nginx-service.yaml
+++ b/charts/nginx/templates/nginx-service.yaml
@@ -12,7 +12,7 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
   annotations:
-
+    service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
     {{- if .Values.privateLoadBalancer }}
     # This annotation is detected by EKS to indicate that
     # it should provision a private ("internal") load balancer


vishwas-astro:
@ianstanton 

As per astronomer/issues#1546

`Move to use Application load balancer instead of classic load balancer as classic load balancer will be phased out in sometime and if that is done we will loose oour environments. Also there are other benefits of using the application load balancer like we can map multiple certs to same LB.`

In this PR here, looks like you are migrating to the Network Load Balancer and not the Application Load Balancer. Just wanting to make sure this is intentional, meaning the LB does not need to be application aware and not need to do request routing based on headers and such.

cc @vparekh94

ianstanton:
@vishwas-astro The intent here was to configure our helm chart to use a Network Load Balancer by default, not an Application Load Balancer.  When this PR was drafted, defaulting to NLB was the easiest acceptable alternative to Classic Load Balancer.

@pgagnon and @vparekh94 may have done some additional research since then. Please chime in if you think we should instead default to ALB.

vparekh94:
@pgagnon with regards to your implementation that let us use ACM certs, did you use an NLB? My gut is that most users will want to use whatever LB allows them to use ACM certs.

pgagnon:
> @pgagnon with regards to your implementation that let us use ACM certs, did you use an NLB? My gut is that most users will want to use whatever LB allows them to use ACM certs.

@vparekh94 Yes the ACM trick requires NLB. There's really no drawback to switching from CLB to NLB as a default at this point, in my opinion.

vishwas-astro:
Thanks @vparekh94 ..
@ianstanton do you want to move this out of draft mode?

@jpweber @danielhoherd @schnie can you review this please?

danielhoherd (approved):
Small change, LGTM.

pgagnon (approved):
LGTM

PR #871, opened 2020-09-24
Add network policy rules

sjmiller609:
## Description

add ingress rules to commander and registry from houston job pods

## 🎟 Issue(s)

no issue associated

## 🧪  Testing

we detected this issue during 0.20 release process. For other network policies, I usually add a function test. But these are jobs, so the pods aren't around all the time, so they aren't available for the functional test. I will add some helm chart unit tests.

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/charts/astronomer/templates/registry/registry-networkpolicy.yaml
+++ b/charts/astronomer/templates/registry/registry-networkpolicy.yaml
@@ -24,6 +24,16 @@ spec:
     - podSelector:
         matchLabels:
           tier: astronomer
+          component: houston-update-check
+          release: {{ .Release.Name }}
+    - podSelector:
+        matchLabels:
+          tier: astronomer
+          component: houston-expire-deployments
+          release: {{ .Release.Name }}
+    - podSelector:
+        matchLabels:
+          tier: astronomer
           component: houston-db-migrations
           release: {{ .Release.Name }}
     - podSelector:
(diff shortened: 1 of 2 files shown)


charts/astronomer/templates/registry/registry-networkpolicy.yaml
```diff
@@ -21,6 +21,16 @@ spec:
...
+    - podSelector:
+        matchLabels:
+          tier: astronomer
+          component: houston-update-check
+          release: {{ .Release.Name }}
+    - podSelector:
+        matchLabels:
+          tier: astronomer
+          component: houston-expire-deployments
+          release: {{ .Release.Name }}
```
andriisoldatenko, line 33: what about `houston-cleanup`?

andriisoldatenko: @sjmiller609 do you mind to add https://github.com/astronomer/astronomer/blob/b8e083fd0cd7fac110e781174d1b5f9b899a0504/charts/astronomer/templates/houston/cronjobs/houston-cleanup-deployments-cronjob.yaml#L26 too?

andriisoldatenko (changes requested):

charts/astronomer/templates/registry/registry-networkpolicy.yaml
sjmiller609: hey actually I think it's down a few lines in this file

sjmiller609: line 52

andriisoldatenko (approved):
LGTM

PR #873, opened 2020-09-25
livenessProbe and readinessProbe added to Stan server.

jpweber:
## Description

The nats Stan component did not previously have liveness or readiness probes. These are being added here. 

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/1672

## 🧪  Testing

If the probe was misconfigured it could either never pass the probe checks or would pass incorrectly, false positive. 

helm-unittest tests have been added to check for the existence of livenessProbe and readinessProbe
and that the path/port are correct


danielhoherd (approved):

danielhoherd:
Thanks for going the extra mile with several useful tests!

PR #878, opened 2020-10-01
Fix Token Validation Bug in Houston API

andriisoldatenko:
## Description

related to https://github.com/astronomer/houston-api/pull/477


adam2k (approved):
🎉

PR #881, opened 2020-10-05
Astronomer upgrade and rollback scripts for 0.16 to 0.23

sjmiller609:
Work in progress, building the LTS to LTS upgrade scripts. We need upgrades to be self-service by our customers, so we are investing time to make the process more seamless.

The automation should be used by our customers unattended by Astronomer support, and thereby must support both upgrade and rollback. The upgrade will use the following design:

Execution environment:

- Inside Pod: more consistent environment compared to customers' workstations
- Run with Ansible: easier to maintain than Bash, encourages self-documentation and informative logging
- Persistent volume: we will store the backups on a block volume (most general storage option in Kubernetes)

Upgrade procedure:

- pre-flight checks
- helm backup
- DB backup
- helm upgrade

Rollback procedure:

- User provides which backup to restore from
- DB restore
- Helm rollback
- All pods restart

CI additions:
- build, test, publish for upgrader docker image


sjmiller609:
@danielhoherd @ianstanton This PR is a work in progress, but you may be interested to review what's going on so I'm pinging you

sjmiller609:
I mentioned in infra team meeting that upgrade, rollback works but then the next upgrade fails. I found this is because new types like this:
```
CREATE TYPE "houston$default"."Role" AS ENUM ('SYSTEM_ADMIN', 'SYSTEM_EDITOR', 'SYSTEM_VIEWER', 'WORKSPACE_ADMIN', 'WORKSPACE_EDITOR', 'WORKSPACE_VIEWER', 'DEPLOYMENT_ADMIN', 'DEPLOYMENT_EDITOR', 'DEPLOYMENT_VIEWER', 'USER');
```
are not reverted by pg restore, even when using --clean on pg restore. To resolve this issue, we can wrap the creation in an SQL conditional.

bin/migration-scripts/lts-to-lts/0.16-to-0.20/playbooks/files/prisma1_to_prisma2.sql
```diff
@@ -12,8 +12,17 @@ ALTER TABLE "houston$default"."DockerImage" alter column labels type jsonb using
...
+
+DO $$ BEGIN
+IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'Role') THEN
+    CREATE TYPE "houston$default"."Role" AS ENUM ('SYSTEM_ADMIN', 'SYSTEM_EDITOR', 'SYSTEM_VIEWER', 'WORKSPACE_ADMIN', 'WORKSPACE_EDITOR', 'WORKSPACE_VIEWER', 'DEPLOYMENT_ADMIN', 'DEPLOYMENT_EDITOR', 'DEPLOYMENT_VIEWER', 'USER');
+END IF;
+
+IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'InviteSource') THEN
+  CREATE TYPE "houston$default"."InviteSource" AS ENUM ('SYSTEM', 'WORKSPACE');
+END IF;
+END $$;
```
sjmiller609, line 24: this is the part that fixes the second-time-apply issue

andriisoldatenko: good catch!

andriisoldatenko: @sjmiller609 ty!

sjmiller609:
One issue addressed in the most recent commit is making sure we always restore from backup the Helm backend secret. This is important in the situation that the max revision history has pruned the revision we want to restore to.

The other situation is during pg_restore, you can get an error 'could not execute query: ERROR:  database \"xxxxx_houston\" is being accessed'. This issue is caused by attempting to run pg_restore before connections from houston api, workers, or prisma pods are closed. Also, we need to avoid connecting to the houston DB with the restore command itself. I found that --db-name argument during a pg_restore should be "postgres" rather than the name of the DB that is being restored. This is an odd behavior, but it is how you avoid connecting to the houston DB.

sjmiller609:
Ok it is feature complete. Now I just need to check for various edge cases and provide more documentation.

sjmiller609:
It's working fairly well now, with a few items outstanding.

- [This issue](https://github.com/astronomer/issues/issues/2251) is happening, but I can ignore for now and I'm confident it will get it fixed eventually. This part is currently in the way of Airflow helm chart version upgrade
- I need to account for rollback of Airflow in addition to rollback of Astronomer, but I think it will be a very similar process (pending the above item)
- I need to create a build and release system: currently hardcoded to use container from my personal dockerhub
- I ought to create a testing system such that we can be continually confident that it works, even when new patches are released or if we want to change something in the upgraee script

sjmiller609:
still need to add docker build a publish for upgrader image

bin/lts-upgrade-0-23-test
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC1090
+set -euo pipefail
+
+BIN_DIR="${0%/*}"
```
danielhoherd, line 5: I would put this below REPO_DIR as `BIN_DIR="${REPO_DIR}/bin"`

sjmiller609:
Hi @krisdock can you please review this change? This is for 0.16 to 0.23 upgrade automation. I want to start working with CRE to knowledge sync on performing LTS enterprise upgrades.

danielhoherd (approved):

bin/migration-scripts/lts-to-lts/0.16-to-0.23/README.md
```diff
@@ -0,0 +1,96 @@
...
+All lines should say "yes".
+
+## Take an external DB backup
+
+Backup your database using your cloud provider's tool or make a request to your database administrator.
+
+## Check on the status of your pods
+
+All pods should be "Running" or "Completed". If you have any pods that are crashing but this is expected for some reason and you want to proceed anyways, then make note of which pods are crashing before the upgrade has been performed.
+
```
krisdock, line 51: @sjmiller609 should we add a section recommending a velero backup as well?

sjmiller609: yeah, I think that's a good call

PR #882, opened 2020-10-05
changing the inviteToken and was still able to signup on the platform. #1891 (bump chart to 0-20-1)

andriisoldatenko:
## Description
changing the inviteToken and was still able to signup on the platform. #1891
resolves https://github.com/astronomer/issues/issues/1891

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: astronomer
-version: 0.20.0
-appVersion: 0.20.0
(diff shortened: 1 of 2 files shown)


jpweber (changes requested):
Also need to update the `charts/astronomer/Chart.yaml ` file

andriisoldatenko:
> charts/astronomer/Chart.yaml

good catch @jpweber !
done.

jpweber (approved):

PR #894, opened 2020-10-13
Fix Houston Knex Migration Error

adam2k:
## Description

Fix the migration script for Houston API.  Prevent the `workspaceId` null error database constraint error from throwing errors during our migration.

## 🎟 Issue(s)

Resolves astronomer/issues#1926

## 🧪  Testing

The migrations should successfully run when there are `InviteToken.workspaceId` null values in the database (you'll need to seed your DB with invites and then make the workspaceIds null manually).

## 📋 Checklist

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -23,7 +23,7 @@ images:
     pullPolicy: IfNotPresent
   houston:
     repository: astronomerinc/ap-houston-api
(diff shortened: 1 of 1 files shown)


andriisoldatenko (approved):

andriisoldatenko:
@adam2k let's wait on CI before merging

PR #898, opened 2020-10-15
Only auto test latest k8s version.

danielhoherd:
## Description

Automatically test only the latest k8s version on each run. Testing all runs is now a manual step that must be run before releasing to internal.

Update kind versions to latest available in https://hub.docker.com/r/kindest/node/tags

Added a final newline to the rendering. Ran `black` on generate_circleci_config.py via pre-commit hooks.

This should be cherry-picked to all branches.

## PR Title

CI changes only. No-op for all installations.

## 🎟 Issue(s)

Related to astronomer/issues#1932

## 🧪  Testing

[Iterated on several configs](https://app.circleci.com/pipelines/github/astronomer/astronomer?branch=1932-reduce-ci-resource-usage) before arriving at the final config. "Test all platforms" is a manual step that can be approved immediately at the start of the run, essentially giving pre-approval to test all versions without needing to wait for early steps to succeed.

## 📸 Screenshots

Screen Shot 2020-10-14 at 6 56 47 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,46 +131,45 @@ workflows:
       - build-artifact:
           requires:
             - lint
+            - unittest-charts
+
+      - test-all-platforms:
+          type: approval
+          requires:
+            - build-artifact
 
       - platform-1-14-10:
           requires:
-            - build-artifact
-            - unittest-charts
+            - test-all-platforms
 
       - platform-1-15-11:
           requires:
-            - build-artifact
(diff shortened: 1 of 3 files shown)


danielhoherd:
I had also considered putting the step that requires human intervention as a step between approving the internal release and actually releasing it. That way would not let us test all versions except on release branches, but would make this change not require an additional human intervention step.

jpweber (approved):

PR #901, opened 2020-10-16
Consolidate AC version list

andriisoldatenko:
## Description

Consolidate AC version list #901
- create new cronjob to run `yarn check-airflow-updates`
## 🎟 Issue(s)

Resolves astronomer/issues#1728


charts/astronomer/values.yaml
```diff
@@ -180,6 +180,17 @@ houston:
...
+  # This runs as a CronJob
+  updateAirflowCheck:
+    # Enable check updates CronJob
+    enabled: true
+
+    # url for update service
+    url: https://updates.astronomer.io/astronomer-certified
+
+    # Default here is to run at midnight every night
+    schedule: "0 0 * * *"
```
danielhoherd, line 193: Given that this will run on every installation of Astronomer, we should use a mechanism to deterministically create a random minute offset to spread the requests out over a larger amount of time. A good mechanism to do this is to seed the random number generator with a piece of data that will be different on each installation while being consistent on a single installation. I've done it in ansible like this:
```yaml
demo_cron_minute: "{{ 59 | random(seed=inventory_hostname) }}"
demo_cron_hour: "{{ 23 | random(seed=inventory_hostname) }}"
```
I'm not sure offhand what the equivalent golang template syntax would be, or what value we could feed to it.

danielhoherd: Closest thing I can find that is similar is this: https://github.com/FairwindsOps/charts/blob/ed2f9c160f99d9833110e4ef4403aa2f370125f9/stable/insights-agent/templates/_specs.yaml#L2

Even just having the random minute would be better than nothing, regardless of whether it's deterministic.

danielhoherd: I just ran a bunch of tests with helm `randNumeric` in various ways, and seems that `mul` and `div` functions both produce 0 way more than other values. I tested on two physical machines, one with linux and one with macos 11, and got the same results. In some of my tests, 0 gets an order of magnitude more than the next highest number, and usually twice what the next number has.

Anyhow, long story short, I think this is our best bet for randomizing the minute: `{{ div (randNumeric 2) 2 }}` which doesn't hit the last 10 minutes of the hour, but produces the least 0-weighted results, and takes a lot of the load off of the 0 minute.

danielhoherd, line 193:
```suggestion
    schedule: "{{ div (randNumeric 2) 2 }} 0 * * *"
```

andriisoldatenko: @danielhoherd what do you think just change to "1 0 * * *" ?

andriisoldatenko:
@schnie @danielhoherd @adam2k PTAL.

charts/astronomer/values.yaml
danielhoherd: That is not going to spread the request load out over time. If we don't pick a random time, all installed hosts will request the resource at the same time, potentially overloading the server. However, after talking to Vishwas and looking at the number of clients we have it's not really a big deal right now.

andriisoldatenko:
@danielhoherd @adam2k @schnie @jpweber PTAL!
houston-api PR has been merged!

jpweber (approved):

danielhoherd (approved):

PR #905, opened 2020-10-21
Improve CPUThrottlingHigh prometheus alerts

jpweber:
## Description

Add airflow tier specific alert for CPU Throttling. This alert includes all the metadata required for houston to receive the alerts. 
Modified platform tier CPU Throttling alert that was incorrectly ignoring namespaces with a regex

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/1844

## 🧪  Testing

prom tool to validate alerts syntax. Needs to be deployed on a real cluster to truly confirm.

### DIFF
--- a/charts/prometheus/templates/prometheus-alerts-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-alerts-configmap.yaml
@@ -36,9 +36,10 @@ data:
             tier: airflow
             component: deployment
             workspace: {{ printf "%q" "{{ $labels.workspace }}" }}
-            deployment: {{ printf "%q" "{{ $labels.deployment }}" }}
+            # deployment label isn't populated by the query. However release and deployment are the same thing
+            deployment: {{ printf "%q" "{{ $labels.release }}" }}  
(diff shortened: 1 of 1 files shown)


danielhoherd (approved):

charts/prometheus/templates/prometheus-alerts-configmap.yaml
```diff
@@ -101,11 +103,24 @@ data:
             tier: airflow
             component: container
             workspace: {{ printf "%q" "{{ $labels.workspace }}" }}
+            deployment: {{ printf "%q" "{{ $labels.deployment }}" }}
```
vishwas-astro, line 106: @jpweber when should it be $labels.deployment and when should it be $labels.release?

jpweber: They should generally always be deployment, but some of the queries don't actually have deployment as part of their labels. So we use release there. They hold the same data so they are interchangeable as long as one of them isn't empty.

PR #907, opened 2020-10-22
Move helm-chart-released-already check into release script.

danielhoherd:
## Description

Move the check for an already published helm chart farther down the line so chart version mismatches don't block code testing.

## 🧪  Testing

Manually tested the release process on my laptop with an already published version.

### DIFF
--- a/bin/functional-tests/README.md
+++ b/bin/functional-tests/README.md
@@ -69,6 +69,5 @@ This test file makes assertions about the configuration of the platform by live-
 ### test_versioning.py
 
 This test file makes assertions about versioning:
-  - This version should not already by published
   - Patch versions support downgrade and upgrade
   - Versioning is in a valid configuration
--- a/bin/functional-tests/test_versioning.py
+++ b/bin/functional-tests/test_versioning.py
@@ -11,6 +11,7 @@ git_root_dir = os.path.join(
     os.path.dirname(os.path.realpath(__file__)),
     "..","..")
 
+
 def test_astro_sub_chart_version_match():
     """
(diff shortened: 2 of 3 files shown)


sjmiller609 (approved):

sjmiller609:
In my opinion, we should leave the test how it is for the sake of convenience. But if you feel it is more convenient this way, that's fine by me. The way I thought of it is that any code addition is invalid if the version is already published because published versions are immutable except for patch version zero, which we overwrite sometimes. If the current behavior is too annoying because it takes a long time to fail for some small thing, I would think it may be better to move the test earlier in the process so it fails CI faster rather than waiting until the final moment of release to error.

jpweber:
@danielhoherd is your take on this, let it run the tests and then if it trips this alert towards the end it prevents it from going somewhere, but you can at least see if your tests work before having to go figure out the version number stuff?

danielhoherd:
Right. The way it is now, as soon as we do a release, tests on every branch except release branches will fail because they all point to a release that has already occurred.

I think the question can be answered by asking ourselves "Where in the release process do we want to bump the version number?" If the answer is "before doing a release" then my change is relevant. If the answer is "right after doing a release" then the existing tests are OK, but still require everybody to rebase against master after a release and will require whoever does a release to actually make that change.

jpweber:
I'm on the side of leave it to the end. Chart versions in the master branch are pretty irrelevant until a release branch is made. This feels like it fits more as a gate to releasing a helm chart than a gate to pushing code to master/main/features/bugfix branches

jpweber (approved):

PR #910, opened 2020-10-26
Remove unmaintained tests

sjmiller609:
## Description

Remove e2e testing automation that is no longer used and has security vulnerabilities in an image scan.

## 🎟 Issue(s)

A customer is asking for removal of this image because it is not passing scan. I know it's no longer being used, so we should delete it.

## 🧪  Testing

This change only deletes test code.


danielhoherd (approved):

andriisoldatenko:
@sjmiller609 no chance to fix tests?

sjmiller609:
> @sjmiller609 no chance to fix tests?

These test were written in a programming language called 'hof' that we decided to not go with long term.

PR #916, opened 2020-10-27
Changed default airflow chart version to 0.18.0

jpweber:
## Description
Updated the default airflow version used with the astronomer chart to be 0.18.0


jpweber:
taking a different route.

PR #917, opened 2020-10-28
Optionally copy a private certificate to all nodes

sjmiller609:
## Description

Private certificate authorities are used by enterprises to authenticate services and for TLS encryption in their networks. In addition, maintaining control of a root certificate authority trusted by all clients in an enterprise network allows enterprises to decrypt and inspect TLS traffic.

In most cases, enterprises will install this certificate into all their nodes. However, in other cases they may use default images for their kubernetes nodes. In this case, the Docker (or containerd) daemon must be configured to accept this certificate, such that image pulls from the Astronomer registry will work.

## 🧪  Testing

I have tested manually on AKS to make sure image pull works after "astro deploy" with a private cert. Complete manual steps are documented here: https://github.com/astronomer/labs/tree/master/astronomer-with-private-ca (work in progress on this doc, but mostly done - just need to update how to set the config newly defined in this PR).

Also, I have added a unit test that asserts this feature is conditionally enabled as expected.

## 📋 Checklist

### DIFF
--- a/templates/trust-private-ca-on-all-nodes/daemonset.yaml
+++ b/templates/trust-private-ca-on-all-nodes/daemonset.yaml
@@ -1,0 +1,62 @@
+################################
+## DaemonSet to mount the private root CA
+##
+## This can be used by enterprise with private
+## CAs that do not already install their root
+## certificate on the kubernetes nodes.
+#################################
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ .Release.Name }}-private-ca
+  labels:
+    tier: platform
+    component: private-ca
+    release: {{ .Release.Name }}
+spec:
+  selector:
+    matchLabels:
+      tier: platform
+      component: private-ca
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        tier: platform
+        component: private-ca
+        release: {{ .Release.Name }}
+    spec:
+      serviceAccountName: {{ .Release.Name }}-private-ca
+      containers:
+      - name: sleeper
+        image: alpine:latest
+        command:
+          - "/bin/bash"
+          - "-c"
+        args:
+          - "cp /private-ca-certs/* /host-trust-store/ && sleep infinity"
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+        - name: hostcerts
+          mountPath: /host-trust-store
+        {{ range $secret_name := (.Values.global.privateCaCerts) }}
+        - name: {{ $secret_name }}
+          mountPath: /private-ca-certs/{{ $secret_name }}.pem
+          subPath: cert.pem
+        {{- end }}
+        resources:
+          requests:
+            cpu: 1m
+            memory: "1Mi"
+          limits:
+            cpu: 10m
+            memory: "1Mi"
+      volumes:
+      - name: hostcerts
+        hostPath:
+          path: {{ .Values.global.privateCaCertsAddToHost.hostDirectory }}
+      {{ range $secret_name := (.Values.global.privateCaCerts) }}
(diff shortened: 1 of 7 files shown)


templates/trust-private-ca-on-all-nodes/daemonset.yaml
```diff
@@ -0,0 +1,72 @@
...
+            cpu: 1m
+            memory: "25Mi"
+          limits:
+            cpu: 50m
+            memory: "50Mi"
+      terminationGracePeriodSeconds: 1
+      volumes:
+      - name: hostcerts
+        hostPath:
+          path: {{ .Values.global.privateCaCertsAddToHost.hostDirectory }}/registry.{{ .Values.global.baseDomain }}/
```
jpweber, line 66: This path must already exist on the host to mount it. If the job of this container is to add certs to the host, I have a feeling the `/registry.{{ .Values.global.baseDomain }}/` part won't exist the firs time this runs. 

you could get around this by adding `type: DirectoryOrCreate` on the line below path. That should create an empty dir if it doesn't exist.

jpweber (changes requested):

templates/trust-private-ca-on-all-nodes/daemonset.yaml
```diff
@@ -0,0 +1,72 @@
...
+          - "-c"
+        args:
+          - "ls -ltrah /private-ca-certs/* && cp -v /private-ca-certs/* /host-trust-store/ && sleep infinity"
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+        - name: hostcerts
+          mountPath: /host-trust-store
+        {{ range $secret_name := (.Values.global.privateCaCerts) }}
+        - name: {{ $secret_name }}
+          # mush be named as *.crt to be noticed by Docker
```
danielhoherd, line 50:
```suggestion
          # must be named as *.crt to be noticed by Docker
```

danielhoherd (approved):

templates/trust-private-ca-on-all-nodes/psp.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: {{ .Release.Name }}-private-ca
+spec:
+  privileged: false
+  allowPrivilegeEscalation: false
+  volumes:
+  - 'secret'
+  - 'emptyDir'
+  allowedHostPaths:
+    - pathPrefix: /etc/docker/certs.d
```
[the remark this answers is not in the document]
sjmiller609: yes thank you, let me fix that

templates/trust-private-ca-on-all-nodes/daemonset.yaml
sjmiller609: I'm not sure if this is correct because it is working, and I would expect it to fail in the case that pre-existing is required since it is mounting the domain name directory. [Docker bind mounts](https://docs.docker.com/storage/bind-mounts/) create directories on the host if they do not exist.

sjmiller609: I went with the continuously refresh option. Also, I found unexpectedly that restartPolicy: Never is not allowed with daemonsets

tests/private-ca-ds.yaml
```diff
@@ -0,0 +1,22 @@
+---
+suite: Test fluentd-daemonset
```
jpweber, line 2: suite name. :)

jpweber (approved):

sjmiller609:
backported to 0.16

sjmiller609:
Here is documentation on how to use private CA with Astronomer https://github.com/astronomer/labs/tree/master/astronomer-with-private-ca

PR #924, opened 2020-11-02
Fix conditionally disabling subcharts

sjmiller609:
## Description

The build in CI was being performed with Helm 2, and that broke the Helm 3 conditional subchart logic. Instead, all subcharts were unconditionally enabled.

## 🧪  Testing

I added a functional test to protect against regression

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,11 +30,10 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Install ci tools
-          command: ./bin/install-ci-tools 1
-      - run:
           name: Package the Astronomer chart
-          command: make build
+          command: |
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -29,12 +29,11 @@ jobs:
...
       - checkout
-      - run:
-          name: Install ci tools
-          command: ./bin/install-ci-tools 1
       - run:
           name: Package the Astronomer chart
-          command: make build
+          command: |
+            source ./bin/install-ci-tools 1
+            make build
```
sjmiller609, line 36: This change is what fixes the issue. previously, "helm" in path for the build was Helm 2

sjmiller609:
backported to 0.16

PR #932, opened 2020-11-10
Openshift support

sjmiller609:
## Description

Ian did the bulk of this PR. This PR serves to get helm install to work out of the box on Openshift, and to get the configuration documented.

## 🎟 Issue(s)

Closes: https://github.com/astronomer/issues/issues/2040
Closes: https://github.com/astronomer/issues/issues/2039
Closes: https://github.com/astronomer/issues/issues/2038
Closes: https://github.com/astronomer/issues/issues/2036
Closes: https://github.com/astronomer/issues/issues/2035
Closes: https://github.com/astronomer/issues/issues/2034
Closes: https://github.com/astronomer/issues/issues/2074

## 🧪  Testing

We are resorting to manual testing for now, because these changes are outside of the standard configuration. We should have automated, functional testing on openshift eventually.

## 📋 Checklist


schnie:
:boom: Looking great

charts/nginx/templates/nginx-service.yaml
```diff
@@ -41,11 +41,11 @@ spec:
     component: ingress-controller
     release: {{ .Release.Name }}
   ports:
-    - port: {{ .Values.ports.http }}
+    - port: 80
```
sjmiller609, line 44: it is only supported to use load balancer with external port 80, 443 due to CORS. However, it is not necessary for the external port to match the port used by the container, so we hard code here what is required and change the default nginx container ports to non privileged ports.

charts/nginx/templates/nginx-deployment.yaml
```diff
@@ -36,12 +36,6 @@ spec:
       containers:
         - name: nginx
           securityContext:
-            allowPrivilegeEscalation: true
-            capabilities:
-              add:
-                - NET_BIND_SERVICE
-              drop:
-                - ALL
```
sjmiller609, line 44: changing the default container ports of nginx to non privileged ports ought to remove this requirement, but this was not tested on a normal kube cluster with psp enabled. It will be discovered in dev, where psps are enabled, if there is an issue.

values.yaml
```diff
@@ -53,7 +53,7 @@ global:
   prometheusPostgresExporterEnabled: false
 
   # Enables blackbox exporter
-  blackboxExporterEnabled: true
+  blackboxExporterEnabled: false
```
sjmiller609, line 56: we should default to what enterprises mostly do, in this case most enterprises do not have use for bbe

sjmiller609:
The config example is multi-step and requires a narrative, so we are excluding the example config for openshift in this PR. Instead, the example will be in the PR to the docs repository.

charts/nginx/templates/nginx-deployment.yaml
sjmiller609: @jpweber this part is questionable I think, i'm not sure if we ought to drop or not. Just let me know what you think. If we do drop it, and it causes an issue we can always do a quick patch - there is very little chance of this slipping under the radar because if it's broken I think ingress won't work.

charts/nginx/values.yaml
```diff
@@ -55,8 +55,8 @@ hstsIncludeSubdomains: false
 serverNameHashBucketSize: 256
 
 ports:
-  http: 80
-  https: 443
+  http: 8080
+  https: 4433
```
danielhoherd, line 59: 8443 is a more common https alternate port, though not official.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -206,6 +206,11 @@ data:
         postgresql:
           enabled: false
 
+        # If security context constraints are enabled, enable SCCs for airflow-chart
```
danielhoherd, line 209: I wonder if these should be golang template comments, not yaml comments, so we don't end up with just a floating text in the yaml output. Not a huge deal though, and seems less readable.
```suggestion
        {{- /* If security context constraints are enabled, enable SCCs for airflow-chart */ }}
```

sjmiller609: I think we should keep it consistent with the other comments, but I move it into the helm conditional

jpweber (approved):

charts/nginx/templates/nginx-deployment.yaml
ashb: I think this is a great idea - the less privilege we need the better!

charts/astronomer/templates/commander/commander-role.yaml
```diff
@@ -129,4 +129,9 @@ rules:
 - apiGroups: ["policy"]
   resources: ["poddisruptionbudgets"]
   verbs: ["create", "delete", "get"]
+{{- if .Values.global.sccEnabled }}
+- apiGroups: ["security.openshift.io"]
+  resources: ["securitycontextconstraints"]
+  verbs: ["create", "delete"]
+{{- end }}
```
ashb, line 136: We probably need this for it to work, but I imagine some enterprises balking about this.

Is this just needed to support the "sudo" use in the Debian airflow images, or something else too?

sjmiller609: this one is necessary so that commander have permission to create security context constraints in airflow namespaces

sjmiller609: this resource https://github.com/astronomer/airflow-chart/blob/master/templates/airflow-scc-anyuid.yaml

ashb (approved):
Loogs good to me, but I may be slightly out of context on everything here.

schnie (approved):

PR #936, opened 2020-11-11
More compatibility for Astronomer logging

sjmiller609:
## Description

I found that Airflow, when logging not in json mode, logs timestamps in a different format than is specified in our config file. We would not notice this error because most users (including our clouds) use json logging format for Airflow logs.

I think this pattern is not in the version of airflow I am using, 1.10.7:
```
%Y-%m-%dT%H:%M:%S.%NZ
```
This pattern is working:
```
%Y-%m-%dT%H:%M:%S.%N+%z
```
Example log line, copied from fluentd pod filesystem
```
 2020-11-11T19:22:45.526814634+00:00 stdout F 10.130.4.1 - - [11/Nov/2020:19:22:45 +0000] "GET /meteoroidal-kiloparsec-2912/airflow/health HTTP/1.1" 200 187 "-" "kube-probe/1.17+"
```

## 🎟 Issue(s)

Not related to any particular issue, but discovered during openshift work

## 🧪  Testing

This change adds an additional pattern match for fluentd parsing of logs, leaving all existing patterns in place. The risk is that it is a misconfiguration and logging will not work at all. I think our normal testing covers the risk of this change.

## 📋 Checklist

### DIFF
--- a/charts/fluentd/templates/fluentd-configmap.yaml
+++ b/charts/fluentd/templates/fluentd-configmap.yaml
@@ -34,6 +34,13 @@ data:
           format /^(?<time>.+) (?<stream>stdout|stderr)( (?<logtag>.))? (?<log>.*)$/
           time_format %Y-%m-%dT%H:%M:%S.%NZ
         </pattern>
+         # Example log we are parsing
+         # 2020-11-11T19:22:45.526814634+00:00 stdout F 10.130.4.1 - - [11/Nov/2020:19:22:45 +0000] "GET /meteoroidal-kiloparsec-2912/airflow/health HTTP/1.1" 200 187 "-" "kube-probe/1.17+"
+         <pattern>
+           format /^(?<time>.+) (?<stream>stdout|stderr)( (?<logtag>.))? (?<log>.*)$/
+           # https://docs.ruby-lang.org/en/2.4.0/Time.html#method-c-strptime
+           time_format %Y-%m-%dT%H:%M:%S.%N+%z
+         </pattern>
       </parse>
       tag raw.kubernetes.*
       read_from_head true


sjmiller609:
I think this problem may be reproduced if running Astronomer in a cluster using containerd, but I'm not sure.

jpweber:
Is this the same root problem as here? https://github.com/astronomer/issues/issues/1893

jpweber (approved):

sjmiller609:
> Is this the same root problem as here? [astronomer/issues#1893](https://github.com/astronomer/issues/issues/1893)
> 
> I know we had a time logging issue so I wanted to see if you think its the same thing.

I took a peek, and I don't think so. But you can check by exec into fluentd pod and cat a random log file. if the timestamp looks like it only matches the new format, that will tell you. Also, fluentd will log a warning to indicate that the timestamp is not matched under the condition this issue is occuring.

danielhoherd:
@sjmiller609 have you tested this in multiple timezones? The addition of the `+` char is suspicious.
```
$ for TZ in Asia/Kolkata Europe/Kiev Europe/London US/Eastern US/Central US/Pacific ; do date "+%Y-%m-%dT%H:%M:%S.%N+%z $TZ"; done
2020-11-12T03:15:23.657380000++0530 Asia/Kolkata
2020-11-11T23:45:23.661661000++0200 Europe/Kiev
2020-11-11T21:45:23.666075000++0000 Europe/London
2020-11-11T16:45:23.670673000+-0500 US/Eastern
2020-11-11T15:45:23.675109000+-0600 US/Central
2020-11-11T13:45:23.678963000+-0800 US/Pacific
```
did you mean `%Y-%m-%dT%H:%M:%S.%N%z` ?
```
$ for TZ in Asia/Kolkata Europe/Kiev Europe/London US/Eastern US/Central US/Pacific ; do date "+%Y-%m-%dT%H:%M:%S.%N%z $TZ"; done
2020-11-12T03:14:56.952714000+0530 Asia/Kolkata
2020-11-11T23:44:56.957145000+0200 Europe/Kiev
2020-11-11T21:44:56.961685000+0000 Europe/London
2020-11-11T16:44:56.966839000-0500 US/Eastern
2020-11-11T15:44:56.971599000-0600 US/Central
2020-11-11T13:44:56.976091000-0800 US/Pacific
```

Also FWIW, capital Z means UTC, so yeah, it looks like we may have been not be accepting any timestamps on servers not configured to use UTC time.

danielhoherd:
Oh, it's also worth noting that the colon in the TZ is sometimes there, and sometimes not. 🙄

```
$ date "+%Y-%m-%dT%H:%M:%S+%z"
2020-11-11T13:50:30+-0800
$ date --iso-8601=seconds
2020-11-11T13:50:33-08:00
```

PR #939, opened 2020-11-13
Fix PSP on and Private CA on

sjmiller609:
Typo was not caught, specific to both PSP and private CA on.


sjmiller609:
I'm going to check if we can enable PSP in kind testing

danielhoherd (approved):

PR #943, opened 2020-11-17
Add edgeEnabled flag to top level

ryw:
Adds a top level flag to designate that the Astronomer deployment allows Astronomer Core Edge deployments. 

re: [Astronomer Core Action Plan](https://www.notion.so/astronomerio/Astronomer-Core-Action-Plan-fea56f20e06c44138ddb54d9fd230de6) and https://github.com/astronomer/issues/issues/1742

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -35,6 +35,9 @@ global:
   # Single namespace mode
   singleNamespace: false
 
+  # Enables the deployment of Astronomer Core Edge (astronomer/core:edge)
+  edgeEnabled: false
+
   # Enables necessary components for compatibility with Istio Service Mesh
   istio:
     enabled: false


schnie:
I think this config could live inside Houston's config, since that's where the logic for this lives. Houston would detect that we're allowing edge builds by this flag and allow those pushes to go through without blocking.

I think there are two categories of "edge" that we need to think about. First, builds that are pushed public by us that we intended to launch, but are acknowledging are edgy for whatever reason. Second, there are the builds that are not even in the system yet, which is more what @kaxil's DEV image issue is concerned with.

I think the behavior is slightly different for each, but they are related. @ryw do you want to address the first category first? These would be the builds that we push out public for folks to use, but they are marked edge.

ryw:
I'm thinking that both "edges" should be converged into one nightly build, per the process outlined in the [Astronomer Core Action Plan](https://www.notion.so/astronomerio/Astronomer-Core-Action-Plan-fea56f20e06c44138ddb54d9fd230de6#8c9fafcf8dcd4b8d97e0e245824f0d2d) — this merges @ernest-kr's circleci work and automates in at the beginning and end the work that @kaxil currently does manually.

ryw:
Just did quick call w/ @kaxil - and now i understand why we need both edge + dev — dev could be on older Airflow releases

* https://github.com/astronomer/ap-airflow/blob/master/.circleci/generate_circleci_config.py

ryw:
@schnie so should we add the two options to https://github.com/astronomer/houston-api/blob/main/config/default.yaml instead:

```
edgeEnabled: false
devEnabled: false
```

schnie:
Yea two flags or possibly an enum of sorts since there's kind of three distinct options.

ryw:
I'm thinking of the UI now -- how we would translate these values to a dropdown.

@schnie are you thinking of something like:

```
# extraDeploymentImages configuration.
# extraDeploymentImagesEnabled:
#   - edge
#   - dev
#   - custom
```

Or list out additional supported versions (which would be cumbersome):

```
# extraDeploymentImages configuration.
# extraDeploymentImagesEnabled:
#   - edge
#   - quay.io/astronomer/core:1.10.10-6.dev-alpine3.10-onbuild
```

schnie:
@ryw maybe something like this, underneath the already existing `deployments` block, where you can provide 1 of 3 options:

```
deployments:
  mode: NORMAL | EDGE | DEV
```

vishwas-astro:
STABLE | EDGE | DEV sounds good.

Assuming that EDGE includes STABLE and DEV includes STABLE and EDGE.

right @schnie ?

ryw:
@schnie + @vishwas-astro i like how you are both thinking about this. Closing this PR and opening another in houston.

schnie:
@vishwas-astro yep, they are just increasingly looser enforcement.

PR #945, opened 2020-11-18
Remove unneeded parentheses

jbampton:
## 📋 Checklist


danielhoherd:
I'm inclined to close this without merging because it's just a style preference change, not a functional change. Also, if we begin enforcing `pre-commit` runs across this project, it will automatically apply this change when it runs `black`. What do you think @jpweber?

jpweber:
I'll follow your lead on this one @danielhoherd as your more of a pythonista than I. I don't really know the ramifications of this change to be honest.

danielhoherd:
@jbampton thanks for the PR but we are closing it as it is a non-functional change. The style improvement is going to be handled by enforcing `black` on this repo.

PR #952, opened 2020-11-23
Add release-0.23 to publicly releasable branches

danielhoherd:
# Description

Only LTS release branches are able to be approved for public release. This adds release-0.23 to that list.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -184,7 +184,7 @@ workflows:
           filters:
             branches:
               only:
-                - '/release-0\.(12|13|14|15|16)/'
+                - '/release-0\.(12|13|14|15|16|23)/'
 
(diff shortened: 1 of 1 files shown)


vishwas-astro:
@danielhoherd We need to do a bunch of testing of 0.23.x against AWS/Azure etc before we can publish this. Thinking of end of 1st week of Dec as the time of releasing this to public. So, dont merge this PR until that time, I guess.

jpweber (approved):

danielhoherd:
Shoot, I just merged this without seeing @vishwas-astro's comment about not merging this yet. However this change does not mean that the chart will automatically be pushed to public release. That is still a manual step. We just need to not release to public yet.

PR #959, opened 2020-12-01
deprecate k8s 1.14 and 1.15

danielhoherd:
## Description

Deprecate Kube 1.14 and 1.15.

Related to astronomer/issues#2266

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,26 +70,6 @@ jobs:
             bin/release-helm-chart -p /tmp/workspace/astronomer-*.tgz
       - publish-github-release
 
-  platform-1-14-10:
-    machine:
-      image: ubuntu-2004:202008-01
-      resource_class: xlarge
-    environment:
-      KUBE_VERSION: v1.14.10
-    steps:
-      - helm-install:
-          astronomer-tags: "platform postgresql monitoring logging kubed keda"
-
-  platform-1-15-12:
-    machine:
-      image: ubuntu-2004:202008-01
-      resource_class: xlarge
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -184,15 +152,15 @@ workflows:
           filters:
             branches:
               only:
-                - '/release-0\.(12|13|14|15|16|23)/'
```
jpweber, line 187: Was this an accident that the 23 got removed, or is this intentional?

.circleci/config.yml
```diff
@@ -184,15 +152,15 @@ workflows:
...
+                - '/release-0\.(12|13|14|15|16)/'
 
       - release-to-public:
           requires:
             - approve-public-release
           filters:
             branches:
               only:
-                - '/release-0\.(12|13|14|15|16|23)/'
```
jpweber, line 195: same question here too, intentional change?

jpweber (changes requested):

.circleci/config.yml
danielhoherd: Oh shoot, yeah, I have no idea how that happened. Good catch!

danielhoherd: The problem was caused by me failing to update the .circleci/config.yml.j2 file when I originally added 23 to the supported list. I committed a change taken from ap-airflow that adds a pre-commit hook to automatically run the generation whenever those files change.

jpweber (approved):
LGTM

tommyk (approved):

PR #960, opened 2020-12-01
release_name-houston-backend secret is being deleted in error

andriisoldatenko:
## Description

fixes https://github.com/astronomer/issues/issues/2251

### DIFF
--- a/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
+++ b/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
@@ -17,6 +17,6 @@ data:
   # in prisma1 has hardcoded schema, so for now to migrate to prisma2 we need to hardcode it too.
   connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db) | b64enc | quote }}
   {{- else }}
-  connection: ""
+  connection: {{ randAlphaNum 5 | b64enc | quote }}
   {{- end }}
 {{- end }}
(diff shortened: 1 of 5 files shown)


sjmiller609:
Hey @andriisoldatenko  I was talking with the infra team on this one. They brought up a situation that I had not yet considered which is when a houston container restarts but the pod does not restart. This can happen in a few situation such as liveness probe failing. So, I think we can reproduce in test this error with the following process:

- helm upgrade or install
- without changing anything do helm upgrade --no-hooks (this ought to clear out the secret and NOT restart houston)
- instigate houston _container_ restart (but not a _pod_ restart - so db-bootstrapper does not run)

Then, we ought to find houston container starting and missing the backend secret.

sjmiller609 (approved):
I think this fix is good enough since we don't have an alternative right now, although I believe that helm upgrade will reset the connection string to a random string until it's replaced by any of the init containers. Thank you for the test of the houston container restart.

charts/astronomer/templates/houston/api/houston-backend-secret.yaml
```diff
@@ -17,6 +17,6 @@ data:
   # in prisma1 has hardcoded schema, so for now to migrate to prisma2 we need to hardcode it too.
   connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db) | b64enc | quote }}
   {{- else }}
-  connection: ""
+  connection: {{ randAlphaNum 5 | b64enc | quote }}
```
schnie, line 20: @andriisoldatenko @sjmiller609 so this is just adding a little 5 char string that will get overwritten soon after creation?

jpweber (approved):

PR #961, opened 2020-12-02
update prometheus ephemeral storage alert threshold

krisdock:
## Description

increasing the ephemeral storage alert to "5gb for 10 minutes"

this isn't a perfect threshold, but it's better than the legacy values we have currently. i'll be working to find a way to alert based on percentage rather than a hardcoded value. 

## 📋 Checklist


danielhoherd (approved):
I definitely like this accurate wording more. LGTM.

jpweber (approved):

PR #962, opened 2020-12-02
Test disappearing secret

sjmiller609:
This PR is into @andriisoldatenko 's branch where he is working on this issue: https://github.com/astronomer/issues/issues/2251

These commits include a test that reproduces an error under the following conditions:

- helm upgrade such that no houston pods are restarted (by using --no-hooks). This will clear out the connection secret with the current issue with helm upgrade clearing the 'connection' key in the release_name-houston-backend secret.
- delete houston container so that it restarts, it will start crashlooping because the connection secret is missing.

This test fails with error 'container not found' because the pod is crashlooping due to the secret missing. So, this test is reproducing the issue.

This test shows that using the init container for initializing the secret is not sufficient because the connection secret being cleared out at any time leads to houston failing if the container restarts for some reason.

Adding this test at @tommyk suggestion, thank you Tommy!

### DIFF
--- a/bin/functional-tests/test_config.py
+++ b/bin/functional-tests/test_config.py
@@ -13,6 +13,7 @@ import os
 import docker
 import time
 import yaml
+import testinfra
 from kubernetes import client, config
 from kubernetes.client.rest import ApiException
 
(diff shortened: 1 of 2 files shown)


bin/functional-tests/test_config.py
```diff
@@ -166,3 +168,47 @@ def test_prometheus_config_reloader_works(prometheus, kube_client):
...
+        helm_chart_path
+    print(command)
+    print(check_output(command, shell=True))
+    # Run the command twice to ensure the most
+    # recent change is a no-operation change
+    print(check_output(command, shell=True))
+    print("")
+    result = houston_api.check_output("env | grep DATABASE_URL")
+    # check that the connection is not reset
+    assert "postgres" in result, "Expected to find DB connection string before Houston restart"
```
sjmiller609, line 207: Currently, this line passes

bin/functional-tests/test_config.py
```diff
@@ -166,3 +168,47 @@ def test_prometheus_config_reloader_works(prometheus, kube_client):
...
+    # Kill houston in this pod so the container restarts
+    houston_api.check_output("kill 1")
+    # give time for container to restart
+    time.sleep(2)
+    result = houston_api.check_output("env | grep DATABASE_URL")
```
sjmiller609, line 212: ....but this line fails, because the container is crash looping after restart

sjmiller609: https://app.circleci.com/pipelines/github/astronomer/astronomer/1929/workflows/de9de9a1-6303-46dc-8757-80dfcbaf2d69/jobs/15108 container not found error - reproduced in any Astronomer installation by doing "helm upgrade" with --no-hooks such that houston does not restart (then the connection secret is cleared out), then do "kill 1" inside a houston container. Then the houston container in that pod will start crashlooping.

sjmiller609:
in pm andrii said go ahead an merge into his branch

PR #963, opened 2020-12-03
Allow for offline update of Airflow Releases

andriisoldatenko:
Resolves https://github.com/astronomer/issues/issues/2183

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -12,6 +12,38 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     heritage: {{ .Release.Service }}
 data:
+  airflow_releases.json: |
+    {
+      "version": "1.0",
+      "available_releases": [
+        {
+          "version": "1.10.7-15",
+          "level": "bug_fix",
+          "url": "https://github.com/astronomer/airflow/releases/tag/v1.10.7%2Bastro.15",
+          "release_date": "2020-09-22T15:52:00+00:00",
+          "tags": [
+            "1.10.7-15-alpine3.10-onbuild",
+            "1.10.7-15-buster-onbuild",
+            "1.10.7-15-alpine3.10",
+            "1.10.7-15-buster"
+          ],
+          "channel": "stable"
+        },
+        {
+          "version": "1.10.12-1",
+          "level": "new_feature",
+          "url": "https://github.com/astronomer/airflow/releases/tag/v1.10.12%2Bastro.1",
+          "release_date": "2020-09-29T22:15:00+00:00",
+          "tags": [
+            "1.10.12-1-alpine3.10-onbuild",
+            "1.10.12-1-buster-onbuild",
+            "1.10.12-1-alpine3.10",
+            "1.10.12-1-buster"
+          ],
+          "channel": "stable"
+        }
+      ]
+    }
   # These are system-specified config overrides.
   production.yaml: |
     webserver:
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -12,39 +12,11 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     heritage: {{ .Release.Service }}
 data:
-  airflow_releases.json: |
-    {
-      "version": "1.0",
-      "available_releases": [
-        {
-          "version": "1.10.7-15",
-          "level": "bug_fix",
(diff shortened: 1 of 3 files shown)


schnie (commented):
@andriisoldatenko yes I think so. Would it be possible to provide these overrides as YAML instead of JSON though? Just because it'll flow nice with the rest of the platform config.

I was also thinking that we would leave the override configmap empty by default, so that out of the box the file baked into the image would be used. If the user actually provides a value to us, then we would mount this file into the Houston pod. Maybe we need some helm logic to conditionally mount this in, _only_ when it's provided by the user.

andriisoldatenko:
> @andriisoldatenko would it be possible to provide these overrides as YAML instead of JSON though? Just because it'll flow nice with the rest of the platform config.
> 
> I was also thinking that we would leave the override configmap empty by default, so that out of the box the file baked into the image would be used. If the user actually provides a value to us, then we would mount this file into the Houston pod. Maybe we need some helm logic to conditionally mount this in, _only_ when it's provided by the user.

PTAL.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -12,7 +12,11 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     heritage: {{ .Release.Service }}
 data:
+  {{ if .Values.airflowReleasesConfig }}
+  airflow_releases.json:
```
schnie: @andriisoldatenko should we change this to yaml extension now?

andriisoldatenko: @schnie i fixed code, PTAL

andriisoldatenko:
@schnie 
practically LGTM

```
➜  astronomer git:(allow-for-offline-update-of-airflow-releases) ✗ k get configmaps astronomer-houston-config -o yaml          apiVersion: v1
data:
  airflow_releases.json: |
    {"available_releases":[{"channel":"stable","level":"new_feature","release_date":"2020-09-29T22:15:00+00:00","tags":["2.0.0-alpine3.10-onbuild"],"url":"https://github.com/astronomer/airflow/releases/tag/v1.10.12%2Bastro.1","version":"2.0.0"}]}
  local-production.yaml: |
    auth:
      local:
        enabled: true
 


/houston $ yarn check-airflow-updates --url https://111updates.astronomer.io/astronomer-certified2222
Explore the full API: http://pris.ly/d/client
info The preview flags `atomicNumberOperations` were removed, you can now safely remove them from your schema.prisma.
2020-12-07T16:09:14 INFO Starting update platform versions
2020-12-07T16:09:14 DEBUG Fetching airflow updates from https://111updates.astronomer.io/astronomer-certified2222
2020-12-07T16:09:14 ERROR Error: getaddrinfo ENOTFOUND 111updates.astronomer.io
2020-12-07T16:09:14 WARN Fallback to local file airflow_releases.json
2020-12-07T16:09:14 INFO Checking possible new release version 2.0.0
2020-12-07T16:09:14 INFO Checking possible new release tag 2.0.0-alpine3.10-onbuild
Done in 6.32s.


```

UI in local cluster:

image

charts/astronomer/templates/houston/houston-configmap.yaml
andriisoldatenko: I'm trying to respect original file from https://github.com/astronomer/houston-api/blob/main/airflow_releases.json,
so we define `airflowReleasesConfig` and than convert to json via `toJson`
```
data:
  {{ if .Values.houston.airflowReleasesConfig }}
  airflow_releases.json: |
    {{ .Values.houston.airflowReleasesConfig | toJson }}
  # These are system-specified config overrides.
  {{- end }}
```
and inside pod we can see correct file body and correct extension.
@schnie please correct me if i'm wrong.

andriisoldatenko:
```$ kubectl exec -it astronomer-houston-5f7f765f4b-qzvdb -n astronomer sh
/houston $
/houston $ cat airflow_releases.json
{"available_releases":[{"channel":"stable","level":"new_feature","release_date":"2020-09-29T22:15:00+00:00","tags":["2.0.0-alpine3.10-onbuild"],"url":"https://github.com/astronomer/airflow/releases/tag/v1.10.12%2Bastro.1","version":"2.0.0"}]}
/houston $
```

schnie (approved):

charts/astronomer/values.yaml
```diff
@@ -191,6 +191,39 @@ houston:
 
     # Default here is to run at midnight every night
     schedule: "0 0 * * *"
+  
+  # this option allow to overrides airflow releases config
```
vishwas-astro, line 195: this option allows to override airflow releases config

andriisoldatenko: done.

jpweber (approved):

PR #979, opened 2020-12-21
Modernize 16 ci

danielhoherd:
Modernize some CI aspects that need to be updated, but are not in sync with master. **This change should not be cherry-picked or merged to any other branch.**

- Use latest ap-build
- Do not install packages during CI runs (packages already existed in base image, plus this is a bad behavior)
- Replace deprecated and delete helm stable repo with merely deprecated one
- `set -x`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2.1
 jobs:
   lint:
     docker:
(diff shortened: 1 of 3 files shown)


jpweber (approved):
Good stuff.

PR #981, opened 2020-12-23
Allow registry to use workload identity

sjmiller609:
## Description

Create a service account for the registry so that the registry can be configured to access a cloud bucket using workload identity (mapping of cloud identity to k8s service account). Make the key file optional when using registry backed by a bucket.

## 🎟 Issue(s)

No issue, for a customer

## 🧪  Testing

Added helm unit tests

## 📋 Checklist


jpweber (approved):
LGTM

sjmiller609:
Looks like the downgrade then upgrade test revealed this change is not a valid patch. Let me check that out and see if it can be fixed.

PR #982, opened 2020-12-28
updated query to get correct number of running pods

tommyk:
## Description

fix a query that doesn't take `Completed` pods into account

## 🧪  Testing

tested against our local prometheus to verify counts were correct

### DIFF
--- a/charts/prometheus/templates/prometheus-alerts-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-alerts-configmap.yaml
@@ -1029,7 +1029,7 @@ data:
             runbook_url:
             summary: Too many pods on '{{`{{ $labels.node }}`}}'
           expr: |
-            sum by (node) ( kube_pod_info and on (pod) kube_pod_status_phase{phase="Running"} == 1 ) / on (node) kube_node_status_capacity_pods{job="{{.Release.Name}}-kube-state"} > 0.95
+            count by(node) ( (kube_pod_status_phase{job="astronomer-kube-state",phase="Running"} == 1) * on(instance,pod,namespace,cluster) group_left(node) topk by(instance,pod,namespace,cluster) (1, kube_pod_info{job="astronomer-kube-state"}) ) / max by(node) ( kube_node_status_capacity_pods{job="astronomer-kube-state"} != 1 ) > 0.95
           for: 15m
           labels:
             severity: warning
--- a/charts/prometheus/templates/prometheus-alerts-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-alerts-configmap.yaml
@@ -1029,7 +1029,7 @@ data:
             runbook_url:
             summary: Too many pods on '{{`{{ $labels.node }}`}}'
           expr: |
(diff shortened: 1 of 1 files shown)


charts/prometheus/templates/prometheus-alerts-configmap.yaml
```diff
@@ -1029,7 +1029,7 @@ data:
             runbook_url:
             summary: Too many pods on '{{`{{ $labels.node }}`}}'
           expr: |
-            max(max(kubelet_running_pod_count{job="kubernetes-nodes"}) by(instance) * on(instance) group_left(node) kubelet_node_name{job="kubernetes-nodes"}) by(node) / max(kube_node_status_capacity_pods{job="{{.Release.Name}}-kube-state"}) by(node) > 0.95
+            sum by (node) ( kube_pod_info and on (pod) kube_pod_status_phase{phase="Running"} == 1 ) / on (node) kube_node_status_capacity_pods{job="{{.Release.Name}}-kube-state"} > 0.95
```
danielhoherd, line 1032: Should this only include running? IE: what about pods that are initializing, crashlooping, etc.?

danielhoherd: Looks like [there are only 5 phases](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase):
- Pending
- Running
- Succeeded
- Failed
- Unknown

I guess "Completed" is "failed" or "succeeded". Should we include "Pending" in this list?

tommyk: after i did this, i found the following about fixing that query on the internets, which they basically have the same thing.  so not really sure about "not running" pods being in the count, thoughts?

https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/443/files

tommyk: one could ask even why 95% of pod capacity is "too many"

tommyk: this also correlates (from my testing) to be the same value as `k describe node <your node> | grep "Non-terminated Pods"`

danielhoherd (approved):

jpweber:
The original source of this alert is from the project you linked to @tommyk where they made a similar update. Part of me wants to use their change so we continue to match.

tommyk:
> The original source of this alert is from the project you linked to @tommyk where they made a similar update. Part of me wants to use their change so we continue to match.

that is fine, i found it afterwards or i probably would have just used it. i will test and update it in a second.

jpweber (approved):

tommyk:
> > The original source of this alert is from the project you linked to @tommyk where they made a similar update. Part of me wants to use their change so we continue to match.
> 
> that is fine, i found it afterwards or i probably would have just used it. i will test and update it in a second.

@jpweber updated the query to what comes out of 'kubernetes-mixin'

PR #985, opened 2021-01-04
updated alert to use Release.Name

tommyk:
# overview
added release name to drive the filter on nodes correctly

# testing
tested the output from `helm template --name-template astrodev` in dev and the query works


jpweber (approved):

danielhoherd (approved):
Coolio

PR #986, opened 2021-01-05
updated alert to use release name to drill into correct nodes. added …

tommyk:
this is the change for an alert change to make it more accurate.  it was taken from `kubernetes-mixin` repo.  i have tested this in dev and the query works correctly from what i can tell...


danielhoherd (approved):
```
 :::===== :::====  :::====  :::      ::: :::====
 :::      :::  === :::  === :::      ::: :::  ===
 ===      ===  === ===  === ===      === ===  ===
 ===      ===  === ===  === ===      === ===  ===
  =======  ======   ======  ======== ===  ======
```

vishwas-astro (approved):

PR #987, opened 2021-01-06
Merge master into release-0.23

danielhoherd:
Rather than cherry-picking changes that have been merged to master, let's merge the whole branch so that it stays in sync.


sjmiller609 (approved):
Looks good, but I would like my LTS upgrader thing in here also

danielhoherd:
> Looks good, but I would like my LTS upgrader thing in here also

I think anything that gets merged into master will be included. That is the normal behavior of PRs from feature->master, so I would expect it to work the same from master->feature

danielhoherd:
Oh, interesting... we can't not squash-and-merge here, because it's not enabled for the repo.

danielhoherd:
I manually merged these changes and pushed them, in order to avoid having to squash and merge.

ianstanton (approved):

PR #994, opened 2021-01-11
bump houston-api: Got invalid value for airflow chart version on prisma updateOneDeployment #2468

andriisoldatenko:
relates to https://github.com/astronomer/issues/issues/2468

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: astronomer
-version: 0.23.8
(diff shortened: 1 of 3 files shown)


danielhoherd (approved):

tommyk (approved):

tommyk:
with these "rc" images, does that mean we are going to have to change them and remove that later if we decide this isn't an "rc" build?

vishwas-astro (approved):

PR #995, opened 2021-01-12
update prometheus-alerts-configmap.yaml ephemeral storage alert threshold

krisdock:
further reducing unnecessary noise. the default ephemeral storage limit is 10gb. setting this alert threshold to 8.5gb.

## 📋 Checklist


danielhoherd (approved):
I thought we found a way to do these alerts by percentage. Can we do that here? If not then this LGTM.

krisdock:
@danielhoherd iirc i looked into it and found that it wasn't possible since the fs size reported in the container is actually the underlying node fs size. that makes percentage based wonky.

PR #998, opened 2021-01-12
Make platform KEDA 2.0 compliant

dimberman:
Uses KEDA 2.0 chart and docker images. Requires this PR to be merged and
released before it will work

https://github.com/astronomer/airflow-chart/pull/183

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -56,7 +56,7 @@ global:
   blackboxExporterEnabled: false
 
   # Used to enable Celery autoscaling
-  kedaEnabled: true
+  kedaEnabled: false
 
   # Used to enable nats-server
   nats:
--- a/values.yaml
+++ b/values.yaml
@@ -45,7 +45,7 @@ global:
 
   # Enable default postgresql database.
   # This is not recommended for production.
-  postgresqlEnabled: false
+  postgresqlEnabled: true
 
   # Enables a node exporter DaemonSet for node-level metrics
   nodeExporterEnabled: true
@@ -56,7 +56,7 @@ global:
   blackboxExporterEnabled: false
 
   # Used to enable Celery autoscaling
-  kedaEnabled: false
+  kedaEnabled: true
 
   # Used to enable nats-server
   nats:
--- a/values.yaml
+++ b/values.yaml
@@ -45,7 +45,7 @@ global:
 
   # Enable default postgresql database.
   # This is not recommended for production.
-  postgresqlEnabled: true
+  postgresqlEnabled: false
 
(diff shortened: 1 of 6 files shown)


values.yaml
```diff
@@ -45,7 +45,7 @@ global:
 
   # Enable default postgresql database.
   # This is not recommended for production.
-  postgresqlEnabled: false
+  postgresqlEnabled: true
```
danielhoherd, line 48: This should not change.
```suggestion
  postgresqlEnabled: false
```

values.yaml
```diff
@@ -56,7 +56,7 @@ global:
   blackboxExporterEnabled: false
 
   # Used to enable Celery autoscaling
-  kedaEnabled: false
+  kedaEnabled: true
```
danielhoherd, line 59: I'm not sure this should change... @sjmiller609 what do you think?

dimberman: @danielhoherd fixed

danielhoherd: This was not fixed. I'd like additional opinions on changing this default value. @schnie ?

danielhoherd: Oh wait, my bad. This was fixed by moving it to a different file!

danielhoherd (approved):

jpweber (approved):

danielhoherd:
Assigned to @dimberman to add more context.

dimberman:
Hi guys! So here's a quick writeup of the risk and strategy around this change:

Strategy:

So the big thing here is that both changes need to be released together. A v1 KEDA ScaledObject will not work on KEDA 2.0 as there are breaking changes, so the minimum Airflow chart version that has the v2 KEDA will need to be the minimum chart version that will be required with the new platform version. 

In terms of upgrading, I think that is dependent on how we do our upgrades (@danielhoherd @jpweber glad to set up a pow-wow to discuss how that works as I'm not personally sure). We'll need to test if simply helm upgrading is sufficient. It very well might be. My ideal case is that helm is able to upgrade dependencies as a part of the upgrade step (if they don't that's crappy on helm's part). Otherwise we might need to manually delete KEDA and reinstall it.

Risks: The main pain for a user is that if they have an airflow instance running KEDA 1.0 they're going to need to redeploy with the new chart version to upgrade. This shouldn't involve an upgrade of Airflow versions but might cause minor disruption/annoyance. We might be able to create a script that converts KEDA scaledObjects without redeploying if we want to avoid this.

Please let me know if this is enough info or if you'd like me to set up a meeting for us all to discuss this further.

danielhoherd:
@dimberman we're coming up on one year of creation of this PR. Is this still relevant?

PR #1004, opened 2021-01-14
trying to get the postgres exporter working

tommyk:
# overview

changes were needed to get the postgres operator up and running.  this still is not FULLY working, but i wanted to open up the PR to get some discussions going on how we can move forward.

the project that this uses is *mostly* dead, little patches here and there but seems like the guy that runs it doesn't have time (and not sure he wants help...).  just wanted to note that.

### testing

this was all run against our dev cluster, with me running local modifications testing to get this working.

### DIFF
--- a/charts/prometheus-postgres-exporter/templates/deployment.yaml
+++ b/charts/prometheus-postgres-exporter/templates/deployment.yaml
@@ -54,8 +54,6 @@ spec:
           {{- end }}
           {{- end }}
           env:
-          - name: PG_EXPORTER_AUTO_DISCOVER_DATABASES
-            value: "true"
           {{- if eq .Values.config.datasource.sslmode "disable" }}
           - name: SSL_MODE_OFF
             value: "1"
--- a/charts/prometheus-postgres-exporter/templates/deployment.yaml
+++ b/charts/prometheus-postgres-exporter/templates/deployment.yaml
@@ -86,16 +86,16 @@ spec:
             - name: {{ .Values.service.name }}
               containerPort: {{ .Values.service.targetPort }}
               protocol: TCP
-          # livenessProbe:
-          #   exec:
-          #     command:
-          #     - /healthcheck
-          #     - "{{ .Values.service.targetPort }}"
-          # readinessProbe:
-          #   exec:
-          #     command:
-          #     - /healthcheck
-          #     - "{{ .Values.service.targetPort }}"
+          livenessProbe:
+            tcpSocket:
+              port: {{ .Values.service.targetPort }}
+            initialDelaySeconds: 5
+            periodSeconds: 10
+          readinessProbe:
+            tcpSocket:
+              port: {{ .Values.service.targetPort }}
+            initialDelaySeconds: 5
+            periodSeconds: 10
           resources:
 {{ toYaml .Values.resources | indent 12 }}
           volumeMounts:
--- a/charts/prometheus-postgres-exporter/templates/deployment.yaml
+++ b/charts/prometheus-postgres-exporter/templates/deployment.yaml
@@ -85,19 +85,19 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:
-            - name: http
+            - name: {{ .Values.service.name }}
               containerPort: {{ .Values.service.targetPort }}
               protocol: TCP
-          livenessProbe:
-            exec:
-              command:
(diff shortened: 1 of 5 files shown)


charts/prometheus-postgres-exporter/templates/deployment.yaml
```diff
@@ -85,19 +83,19 @@ spec:
...
-            exec:
-              command:
-              - /healthcheck
-              - "{{ .Values.service.targetPort }}"
-          readinessProbe:
-            exec:
-              command:
-              - /healthcheck
-              - "{{ .Values.service.targetPort }}"
+          # livenessProbe:
```
tommyk, line 89: i took these out, i will look into if it has health checks and add them back (or if not remove the commented out section).

charts/prometheus-postgres-exporter/values.yaml
```diff
@@ -59,6 +59,11 @@ securityContext: {}
 config:
   datasource:
     connectionSecret:
+      # we have a problem here...
```
tommyk, line 62: this is something we need to discuss and come up with whatever solution we want to use.  i haven't figured out anything very good.

i got it working by modifying the `astronomer-bootstrap` secret value into a new temporary one to make it work for testing purposes, but that isn't going to work going forward...

jpweber: Need to have a think on this one

tommyk: at the easiest we could add a blank database called `airflow` i guess.... but i am unsure if not having that database was done on purpose or not

jpweber: Getting in to hack territory, but what about an init container on the exporter that reads in the `astronomer-houston-backend` secret, and mutates it to what the exporter needs it to look like, then creates a secret with that value?

tommyk:
https://github.com/astronomer/issues/issues/2473

charts/prometheus-postgres-exporter/values.yaml
sjmiller609: I don't like how this feature turned out when it was made in the first, place so sorry about that. I was having issues with consistently populating the configuration, considering for connection string variations or the other way of providing connection when using SSL. This is another part of the puzzle https://github.com/astronomer/ap-vendor/blob/main/postgres-exporter/include/entrypoint

charts/prometheus-postgres-exporter/values.yaml
```diff
@@ -78,16 +83,19 @@ config:
...
-    airflow_chart_version:
-      query: |
-        SET search_path TO houston$default, public; select count(*) from "Deployment" where version != (select mode() within group (order by version) as version_group FROM "Deployment")
-      master: false
-      cache_seconds: 30
-      metrics:
-        - unconverged:
-            usage: "GAUGE"
-            description: "How many rows in the 'version' column of 'Deployments' are not converged to the mode version"
+  # you have to be connected to the houston db for this to work, no place for to set DB
```
sjmiller609, line 86: I think that when I tested it the connection string included ..../astronomer_houston at the end, and so it would hook up to the houston db. Basically the issue was accounting for any valid format of the connection string

tommyk: yeah, its very hard to make this work and hopefully not  break anyones setups they have since this is an OSS helm package...

such a variance in how connection strings are constructed, what they have in them, etc makes it difficult to feel comfortable about any way to make this work for me.

tommyk:
ok, i am ready to move forward with this and get this released as a helm package, we can then start moving it through the environments...

# testing

i tested in dev w/ the following values added to the helm values for dev (this would get changed in google-environments once we get this published as a helm chart)
```yaml
prometheus-postgres-exporter:
  replicaCount: 1
  config:
    queries: ''
```

i also had to create a database on the dev db instance in gcp w/ the following command
```sql
CREATE DATABASE airflow;
```

jpweber (approved):

charts/prometheus-postgres-exporter/templates/deployment.yaml
jpweber: Easy fix for the probes should be something like this
```
    readinessProbe:
      tcpSocket:
        port: {{ .Values.service.targetPort }}
      initialDelaySeconds: 5
      periodSeconds: 10
    livenessProbe:
      tcpSocket:
        port: {{ .Values.service.targetPort }}
      initialDelaySeconds: 15
      periodSeconds: 20
```

PR #1012, opened 2021-01-27
Add connection limit to Houston DB

adam2k:
## Description

Add a connection limit parameter for the Houston Database connection

## 🎟 Issue(s)

Related to https://github.com/astronomer/db-bootstrapper/pull/30

Resolves astronomer/issues#2476

## 🧪  Testing

There are testing steps listed in the issue link.  We'll need to conduct further testing in `dev` and `staging`.

## 📋 Checklist

### DIFF
--- a/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
+++ b/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
@@ -15,7 +15,7 @@ type: Opaque
 data:
   {{- if .Values.houston.backendSecretConnection }}
   # in prisma1 has hardcoded schema, so for now to migrate to prisma2 we need to hardcode it too.
-  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default&%s" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db (.Values.houston.backendConnection.connectionLimit | toString)) | b64enc | quote }}
+  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db) | b64enc | quote }}
   {{- else }}
   connection: {{ randAlphaNum 5 | b64enc | quote }}
   {{- end }}
--- a/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
+++ b/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
@@ -15,7 +15,7 @@ type: Opaque
 data:
   {{- if .Values.houston.backendSecretConnection }}
   # in prisma1 has hardcoded schema, so for now to migrate to prisma2 we need to hardcode it too.
-  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default&%s" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db .Values.houston.backendConnection.connectionLimit) | b64enc | quote }}
+  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default&%s" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db (.Values.houston.backendConnection.connectionLimit | toString)) | b64enc | quote }}
   {{- else }}
   connection: {{ randAlphaNum 5 | b64enc | quote }}
   {{- end }}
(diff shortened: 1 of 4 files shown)


charts/astronomer/templates/houston/api/houston-backend-secret.yaml
```diff
@@ -15,7 +15,7 @@ type: Opaque
 data:
   {{- if .Values.houston.backendSecretConnection }}
   # in prisma1 has hardcoded schema, so for now to migrate to prisma2 we need to hardcode it too.
-  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db) | b64enc | quote }}
+  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default&%s" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db (.Values.houston.backendConnection.connectionLimit | toString)) | b64enc | quote }}
```
tommyk, line 18: unfortunately, this will not work.  we don't populate the "backendSecretConnection"

values.yaml
```diff
@@ -103,6 +103,10 @@ global:
   ssl:
     enabled: false
     mode: "prefer"
+  
+  connectionLimit:
+    enabled: true
```
andriisoldatenko, line 108: enabled: false

charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -71,6 +71,11 @@ spec:
             - name: SSLMODE
               value: {{ .Values.global.ssl.mode }}
             {{- end }}
+            # Connection Limit for Prisma Connection Pool: https://www.prisma.io/docs/concepts/components/prisma-client/deployment#database-connection-limit
+            {{ if .Values.global.connectionLimit.enabled }}
```
schnie, line 75: This probably doesn't need to be a global setting. Maybe push the config down to just the `astronomer` chart values.yaml? Nested under the `houston` top-level key.

charts/astronomer/values.yaml
```diff
@@ -82,6 +82,7 @@ houston:
     # host: ~
     # port: ~
     # db: ~
+    # connectionLimit: ~
```
schnie, line 85: `backendSecretName` and `backendConnection` are two ways to set the same value. `connectionLimit` isn't part of a connection object, so we may want to pull this out to a sibling to those fields.

values.yaml
```diff
@@ -103,6 +103,10 @@ global:
   ssl:
     enabled: false
     mode: "prefer"
+  
+  connectionLimit:
+    enabled: false
+    limit: 5
```
schnie, line 109: As I mentioned above, we can prob pull this out to be part of the `astronomer` subchart values.yaml

andriisoldatenko:
@adam2k locally it doesn't work for me may be bug somewhere in leveles https://github.com/astronomer/astronomer/blob/1a6e51f4a34be70a872272e2dd2303381bea6e33/charts/astronomer/templates/houston/api/houston-deployment.yaml#L75

```

/houston $ echo $DATABASE_URL
postgres://postgres:postgres@astronomer-postgresql.astronomer.svc.cluster.local:5432/astronomer_houston?schema=houston%24default
```

and secret value:
```
connection: postgres://postgres:postgres@astronomer-postgresql.astronomer.svc.cluster.local:5432/astronomer_houston?schema=houston%24default
```

andriisoldatenko:
my local settings:
```diff
diff --git charts/astronomer/values.yaml charts/astronomer/values.yaml
index d54bd4c4..85f8f6f5 100644
--- charts/astronomer/values.yaml
+++ charts/astronomer/values.yaml
@@ -30,9 +30,9 @@ images:
     tag: 0.23.3
     pullPolicy: IfNotPresent
   dbBootstrapper:
-    repository: quay.io/astronomer/ap-db-bootstrapper
-    tag: 0.23.0
-    pullPolicy: IfNotPresent
+    repository: asoldatenko/ap-db-bootstrapper
+    tag: add-connection-limit
+    pullPolicy: Always
   cliInstall:
     repository: quay.io/astronomer/ap-cli-install
     tag: 0.13.1
diff --git configs/local-dev.yaml configs/local-dev.yaml
index 3016773e..2c1c8e1f 100644
--- configs/local-dev.yaml
+++ configs/local-dev.yaml
@@ -61,6 +61,7 @@ astronomer:
         cpu: "0m"
         memory: "0Mi"
   houston:
+    prismaConnectionLimit: 5
     replicas: 1
     worker:
       replicas: 1
diff --git values.yaml values.yaml
index 7f1b8f7c..fddd0f8e 100644
--- values.yaml
+++ values.yaml
@@ -137,6 +137,7 @@ astronomer:
         memory: "1024Mi"

   houston:
+    prismaConnectionLimit: 5
     resources:
       requests:
         cpu: "500m"
(
```

schnie (approved):

charts/astronomer/values.yaml
```diff
@@ -56,6 +56,7 @@ astroUI:
   #   memory: 128Mi
 
 houston:
+  prismaConnectionLimit: 5
```
tommyk, line 59: this seems like we are going with the mind set "everyone, including our corporate customers, are going to have a connection limit set to 5".

is that what we want? or do we want to add the ability to control it if they so choose and leave behavior the same if they don't?

andriisoldatenko (approved):

charts/astronomer/values.yaml
adam2k: @tommyk we could set to `~` and you all can set it in cloud?  Does that work for you?

tommyk: my vote would be to not change the behavior and just allow users to override it if they choose.  

right now i think we are the only people really worried about this in our cloud, so changing it for everyone has a bigger potential ripple effect than leaving it the same imo. 

you guys can decide, i don't want to hold this up.

PR #1016, opened 2021-02-01
Remove irrelevant LTS tests from master branch

danielhoherd:
## Description

Remove LTS related CI jobs from master branch, since they won't work there now that master is on 0.24. These changes still allow LTS tests to be run on feature branches in case they pertain to 0.23.

## Screenshots

Master will look like

Screen Shot 2021-02-01 at 4 04 32 PM

Feature branches will look like

Screen Shot 2021-02-01 at 4 05 21 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,13 +144,13 @@ workflows:
           type: approval
           filters:
             branches:
-              not:
+              ignore:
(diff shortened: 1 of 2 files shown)


danielhoherd:
Closing this since we're not moving forward with 0.24 right now.

PR #1023, opened 2021-02-11
update nginx-es-networkpolicy.yaml to allow comms from webservers to elasticsearch-nginx pod

krisdock:
## Description

correct the selectors on the network policy so webservers can talk to the elasticsearch-nginx pod

## 🎟 Issue(s)

Resolves astronomer/issues#2598

## 📋 Checklist

### DIFF
--- a/charts/elasticsearch/templates/nginx/nginx-es-networkpolicy.yaml
+++ b/charts/elasticsearch/templates/nginx/nginx-es-networkpolicy.yaml
@@ -15,10 +15,9 @@ metadata:
 spec:
   podSelector:
     matchLabels:
-      tier: logging
-      component: {{ template "elasticsearch.name" . }}
+      tier: elasticsearch
+      component: es-ingress-controller
       release: {{ .Release.Name }}
-      role: nginx
   policyTypes:
   - Ingress
   ingress:


jpweber:
I haven't tested this yet to confirm, but if this fix is correct that means we have this issue same issue on our cloud too right? Can we confirm that we see the same behavior in prod cloud?

krisdock:
@jpweber i have no idea how this is working in prod cloud. 

```
> k get networkpolicy astronomer-elasticsearch-nginx-policy -o yaml|grep -A5 "^  podSelector"
  podSelector:
    matchLabels:
      component: elasticsearch
      release: astronomer
      role: nginx
      tier: logging
```

```
> k get po astronomer-elasticsearch-nginx-f98fcb9f4-djsbc -o yaml | grep -A5 labels
  labels:
    component: es-ingress-controller
    pod-template-hash: f98fcb9f4
    release: astronomer
    security.istio.io/tlsMode: istio
    tier: elasticsearch
```

from what i can tell, it should be broken since the networkpolicy selectors don't match the elasticsearch-nginx pod labels. unless there's something else allowing this and superseding this networkpolicy.

krisdock:
the only thing i can think of is that our gcp cluster has a default allow and ibm clusters have a default deny, but i'm not sure how to confirm that

danielhoherd (approved):

danielhoherd:
Merged this to master, [merged master to release-0.23](https://github.com/astronomer/astronomer/commit/d747429475244d9fc391a85b741a1fd6ef16ddb5), no merge conflicts.

PR #1024, opened 2021-02-15
Add service annotations to allow usage of ACM certificates on NGINX

pgagnon:
## Description

This PR adds the following:
- Allows the user to specify "ssl" as backend protocol. The user should provision a self-signed certificate to terminate SSL 
-  between the load balancer and the NGINX pod;
- Allows the user to specify "nlb" as load balancer type on the service resource;
- Allows the user to specify the ARN of a ACM certificate to terminate SSL at the load balancer.

## 📋 Checklist

Resolves astronomer/issues#2616

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -197,13 +197,13 @@ nginx:
   loadBalancerSourceRanges:
 
   # On AWS, set to "nlb" to provision a network load balancer for NGINX instead of a classic load balancer
-  loadBalancerType:
+  loadBalancerType: ~
 
   # Set to "ssl" if TLS termination is performed at the load balancer, but NGINX still expects to receive TLS traffic
-  loadBalancerBackendProtocol:
+  loadBalancerBackendProtocol: ~
 
   # If using ACM, enter the ARN of the certificate to use on the load balancer
-  loadBalancerAcmCertificate:
+  loadBalancerAcmCertificate: ~
 
 
 #################################
--- a/values.yaml
+++ b/values.yaml
@@ -196,6 +196,15 @@ nginx:
   # Used to restrict IPs which can reach the nginx ingress
   loadBalancerSourceRanges:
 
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -196,6 +196,15 @@ nginx:
   # Used to restrict IPs which can reach the nginx ingress
   loadBalancerSourceRanges:
 
+  # On AWS, set to "nlb" to provision a network load balancer for NGINX instead of a classic load balancer
+  loadBalancerType:
```
danielhoherd, line 200:
```suggestion
  loadBalancerType: ~
```

values.yaml
```diff
@@ -196,6 +196,15 @@ nginx:
...
+
+  # Set to "ssl" if TLS termination is performed at the load balancer, but NGINX still expects to receive TLS traffic
+  loadBalancerBackendProtocol:
```
danielhoherd, line 203:
```suggestion
  loadBalancerBackendProtocol: ~
```

values.yaml
```diff
@@ -196,6 +196,15 @@ nginx:
...
+
+  # If using ACM, enter the ARN of the certificate to use on the load balancer
+  loadBalancerAcmCertificate:
```
danielhoherd, line 206:
```suggestion
  loadBalancerAcmCertificate: ~
```

danielhoherd:
@pgagnon I'd love to see some tests with this. `helm-unittest` can be a bit finicky though, so if you're not familiar with it I'd take a look at the other cases in this repo where we use it. Feel free to ping me on slack if you get frustrated with it.

danielhoherd:
@pgagnon There is an unmerged doc that shows how to write tests for helm. Knocking them out should be pretty easy. https://github.com/astronomer/airflow-chart/blob/68eec2a633f9980c9d54f242897eee74c97cb37b/tests/README.md

PR #1033, opened 2021-03-04
Remove e2e test image from default values.yaml. Cleanup.

danielhoherd:
- Remove e2e test image from default values.yaml
- Fix whitespace issue in CI config.yaml
- Bump CircleCI image runner version

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -37,10 +37,6 @@ images:
     repository: quay.io/astronomer/ap-cli-install
     tag: 0.13.1
     pullPolicy: IfNotPresent
-  e2eTest:
-    repository: quay.io/astronomer/ap-e2e-test
-    tag: 0.0.7
-    pullPolicy: IfNotPresent
 
 astroUI:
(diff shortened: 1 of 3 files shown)


charts/astronomer/values.yaml
```diff
@@ -37,10 +37,6 @@ images:
     repository: quay.io/astronomer/ap-cli-install
     tag: 0.13.1
     pullPolicy: IfNotPresent
-  e2eTest:
```
tommyk, line 40: the helm chart itself had e2etest setup in it?!

tommyk (approved):

charts/astronomer/values.yaml
tommyk: what was using these values? does that need cleaned up too?

jpweber: I believe all the other parts were removed in https://github.com/astronomer/astronomer/pull/910. This is cleanup up pieces that one didn't get.

jpweber (approved):

PR #1039, opened 2021-03-18
Add ingressAnnotations. Improve other syntaxes.

danielhoherd:
Related to <https://github.com/astronomer/issues/issues/2640>, which talks about GCP Global Access, and offers a solution of allowing arbitrary ingress annotations.

# Description

* Add default null string for loadBalancerIP
* Add default null dict for loadBalancerSourceRanges
* Fix yaml anchor syntax for platformTolerations
* Whitespace
* Add test for these changes and other existing behaviors of the NGINX Service
* Auto-formatting applied by vscode `prettier` that removed some whitespace, re-aligned some braces.

### DIFF
--- a/charts/nginx/templates/nginx-service.yaml
+++ b/charts/nginx/templates/nginx-service.yaml
@@ -13,6 +13,11 @@ metadata:
     heritage: {{ .Release.Service }}
   annotations:
     service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
+{{- if .Values.ingressAnnotations -}}
+{{ range $key, $value := .Values.ingressAnnotations }}
+    {{ $key }}: {{ $value }}
+{{- end -}}
+{{- end -}}
     {{- if .Values.privateLoadBalancer }}
     # This annotation is detected by EKS to indicate that
     # it should provision a private ("internal") load balancer
--- a/charts/nginx/templates/nginx-service.yaml
+++ b/charts/nginx/templates/nginx-service.yaml
@@ -12,7 +12,6 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
   annotations:
-    service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
 {{- if .Values.ingressAnnotations -}}
 {{ range $key, $value := .Values.ingressAnnotations }}
     {{ $key }}: {{ $value }}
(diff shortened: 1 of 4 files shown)


danielhoherd:
May be related to https://github.com/astronomer/issues/issues/2700

charts/nginx/templates/nginx-service.yaml
```diff
@@ -13,6 +13,11 @@ metadata:
     heritage: {{ .Release.Service }}
   annotations:
     service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
```
danielhoherd, line 15: I feel like we should take this out and tell customers who need it to add it. This was very recently added.

danielhoherd: @ianstanton ^^^ ?

ianstanton: @danielhoherd I agree with this. Keeps things generic and gives customers the liberty to add this if they want. We'll need to add some instructions for this. @jwitz Who is in charge of document updates these days?

ianstanton, line 15:
```suggestion
```

ianstanton (changes requested):

charts/nginx/values.yaml
```diff
@@ -32,14 +32,17 @@ resources: {}
 #   cpu: 100m
 #   memory: 128Mi
 
+# dict of annotations to add to the ingress controller
```
rbankston, line 35: Would like to make sure that we have an example either here in the helm chart or in documentation somewhere for customers that aren't as familiar with Go for what a dict looks like in the language to lessen questions to CRE or in the forum

charts/nginx/templates/nginx-service.yaml
rbankston: what if service.beta.kubernetes.io/aws-load-balancer-type: "nlb" was removed from here and added a a default helm chart value as both documentation as code and to keep that default in place?

rbankston: Didn't like my other thought. Would like to see each of the deployment sections in the docs have common annotations for each cloud service noted there for customers to copy and remove mention of private from the sample yaml in favor of this method.

rbankston (approved):
These changes look great. Like removing the nlb section in favor of specifying it in the documentation

charts/nginx/templates/nginx-service.yaml
danielhoherd: @rbankston if you haven't yet, you may want to turn that into a ticket so the idea doesn't get lost when i resolve this conversation.

PR #1044, opened 2021-03-24
customRoutes should be a list not a dictonary, helm was throwing warn…

burandobata:
## Description

When defining customRoutes in values file helm is complaining with:
coalesce.go:200: warning: cannot overwrite table with non table for customRoutes (map[])

This is because in values.yaml customRoutes are initially defined as a dictionary but in fact they are a list.

## PR Title

Fix helm working about customRoutes in Alertmanager configuration

### DIFF
--- a/charts/alertmanager/values.yaml
+++ b/charts/alertmanager/values.yaml
@@ -51,7 +51,7 @@ receivers:
   # Configs for airflow alerts
   airflow: {}
 
-customRoutes: {}
+customRoutes: []
 # Example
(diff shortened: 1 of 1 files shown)


jpweber (approved):

danielhoherd (approved):

danielhoherd:
Thanks for the pr @burandobata!

danielhoherd:
I have stumped out a unittest for this case in a feature branch I am working on.

Issue #1035, opened 2021-03-10
externalTrafficPolicy in nginx-service.yaml template


burandobata:
Hey,

in nginx-service.yaml template of nginx chart we have possibility to set externalTrafficPolicy to "Cluster" only which is a default and doesn't make sense. This should allow to switch it to "Local". Can this switch be added?

I'm referring to this part:
  {{- if .Values.preserveSourceIP }}
  externalTrafficPolicy: "Cluster"
  {{- end }}

danielhoherd:
https://github.com/astronomer/astronomer/blob/0d69295ca7/charts/nginx/templates/nginx-service.yaml#L40-L42

danielhoherd:
It used to be `Local` https://github.com/astronomer/astronomer/commit/33431136b204601a97783a62ff139e08236473aa and I'm not sure why it changed. [The default does indeed appear to be `Cluster`](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip) and the only alternate value appears to be `Local`.

@schnie @jpweber any comments on this one? It seems like this should be done as an if/else block. Also, as noted by @burandobata, it seems that currently the condition isn't making a difference because it is just explicitly setting what is already the default. [We have it set to False in the chart](https://github.com/astronomer/astronomer/blob/0d69295ca7/charts/nginx/values.yaml#L51), which is a consideration for rolling out any changes to this section.

jpweber:
In a lot of cases people will want this to be local. As noted in the original comment this simply doesn't work. If you want to preserve the source IP you use `local` not `cluster`.  The logic of the if statement should be flipped.  `if .Values.preserveSourceIP` == true use `local`;  else use default (`cluster`)

danielhoherd:
Status update: this feature got caught in limbo due to lack of trustworthy testing. We have since switched to pytest where tests are *much* more reliable, but have not revisited this issue.

danielhoherd:
This was merged today and will go into 0.29.1.

PR #1052, opened 2021-04-06
Correct externalTrafficPolicy behavior. Add tests.

danielhoherd:
https://github.com/astronomer/astronomer/issues/1035 Pointed out that we are explicitly setting what is already the default behavior of nginx, and while doing so we are doing the opposite of what the configured behavior should be. This change set corrects the behavior and adds unittests for that aspect of the chart.


tommyk (approved):
might want to make sure this gets documented as i feel like its a breaking change you are making, right? 

(even though you are fixing a bug, it feels like the behavior will be the opposite as before?)

danielhoherd:
@jwitz @paolaperaza @krisdock what do you all think about this change?

danielhoherd:
Next up we should migrate the helm-unittest in this branch to be a pytest test. That way we can be more confident in this change.

danielhoherd:
If we are going to move forward with this the tests need to be rewritten in pytest.

pgvishnuram:
we can remove the helm test file for this change

danielhoherd:
@pgvishnuram I deleted the helm-unittest test. Do you think this is ready to merge? My worry is still that we are changing behavior on existing installations.

andriisoldatenko:
LGTM

andriisoldatenko (approved):

PR #1054, opened 2021-04-07
bump houston-api:0.23.22

andriisoldatenko:
resolves https://github.com/astronomer/issues/issues/2790


danielhoherd (approved):
I have verified that this fixes the problem.

PR #1056, opened 2021-04-08
LTS Upgrader: use private helm repo only when testing

danielhoherd:
# Description

Only use internal-helm repo when testing the LTS upgrader, use the public helm repo at any other time. This is done in order to prevent customers from picking up pre-release versions of the chart.

This is done through a new variable `USE_INTERNAL_HELM_REPO` which tells ansible (the component that performs the LTS upgrade) to use the internal helm repo instead of the public helm repo. This var is passed from [the test script into k8s](https://github.com/astronomer/astronomer/blob/e6ef341f/bin/lts-0-23-upgrade-tests/test_lts_upgrade.py#L60-L80) [into Docker into Ansible](https://github.com/astronomer/astronomer/blob/e6ef341f/bin/migration-scripts/lts-to-lts/0.16-to-0.23/playbooks/Dockerfile#L27).

Related to astronomer/issues#2596

# Testing

Normal testing env shows:
```
/lts-upgrade # circleci@default-37dd162a-e01a-4065-8720-9462520e9153:~$ kubectl exec -ti upgrade-astronomer-ztjsr -- sh
/lts-upgrade # echo $USE_INTERNAL_HELM_REPO
True
/lts-upgrade # helm repo list -o yaml
- name: astronomer
  url: https://internal-helm.astronomer.io
```

I modified the test script to always set `USE_INTERNAL_HELM_REPO: False`:
```
circleci@default-37dd162a-e01a-4065-8720-9462520e9153:~$ kubectl exec -ti upgrade-astronomer-s5v28 -- sh
/lts-upgrade # echo $USE_INTERNAL_HELM_REPO
False
/lts-upgrade # helm repo list -o yaml
- name: astronomer
  url: https://helm.astronomer.io
```

### DIFF
--- a/bin/lts-0-23-upgrade-tests/requirements.txt
+++ b/bin/lts-0-23-upgrade-tests/requirements.txt
@@ -1,5 +1,6 @@
-pytest==6.0.1
-testinfra==5.2.2
 docker==4.3.1
+GitPython==3.1.14
 kubernetes==11.0.0
+pytest==6.0.1
 PyYAML==5.3.1
+testinfra==5.2.2
--- a/bin/lts-0-23-upgrade-tests/test_lts_upgrade.py
+++ b/bin/lts-0-23-upgrade-tests/test_lts_upgrade.py
@@ -9,7 +9,7 @@ from packaging.version import parse as semver
 
 # The top-level path of this repository
 git_root_dir = Path(__file__).absolute().parent.parent.parent
-print(f"{git_root_dir=}")
+print(f"DEBUG: {git_root_dir=}")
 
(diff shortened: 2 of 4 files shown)


jpweber (approved):

vishwas-astro (approved):

vishwas-astro:
@danielhoherd - why is the unittest-charts step failing in CI?

danielhoherd:
@vishwas-astro it's not failing, i canceled it because that part of the tests weren't relevant to what i was iterating on. i'm re-running it now.

PR #1064, opened 2021-04-19
Add LTS 25, remove pre 16 versions from CI rules.

danielhoherd:
Tune up some LTS CI configs. I thought this would be bigger when I opened the original ticket.

Now that all releases will be public, we can reduce this logic back to a simple regex instead of matching against specific version numbers that were approved for public release.

Related to https://github.com/astronomer/issues/issues/2828

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,7 +220,7 @@ workflows:
           filters:
             branches:
               only:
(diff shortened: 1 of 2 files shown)


rahat-astro (approved):
I have reviewed the 2 files. Thanks.

PR #1071, opened 2021-04-27
added line to ignore rolling daemonsets for 10m before alerts fire fo…

tommyk:
## Description

rolling daemonsets are kicking off alerts due to the length of how many need updated.  it causes false alerts to go out.  this change ignores a daemonset for 10m after it starts rolling giving it time to move through nodes.
## 🎟 Issue(s)
https://github.com/astronomer/issues/issues/2902

## 🧪  Testing

i tested these queries side by side in our gen1 environment and this would reduce/eliminate the `calico-node` alerts we get from vertical scaling happening during expansion and contraction of the nodes.

### DIFF
--- a/charts/prometheus/templates/prometheus-alerts-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-alerts-configmap.yaml
@@ -736,6 +736,8 @@ data:
             kube_daemonset_status_number_ready{job="{{.Release.Name}}-kube-state"}
               /
             kube_daemonset_status_desired_number_scheduled{job="{{.Release.Name}}-kube-state"} < 1.00
+            and
+            increase(kube_daemonset_metadata_generation{job="{{.Release.Name}}-kube-state"}[10m]) == 0
           for: 15m
           labels:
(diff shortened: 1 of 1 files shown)


jpweber (approved):

tommyk:
@vishwas-astro @danielhoherd  if we merge this, will it deployed the next release?  i'd like to get it into dev and make sure its still working, and then most likely edit this in prod by hand until the next release that will include it.  just got another page about this...

vishwas-astro:
@tommyk - We have already cut the release-0.25 branch. We will have to merge this to main and then into the release branch as well for it to be deployed to cloud as part of 0.25.0 release.

Let's merge and test it in dev first.

tommyk:
> @tommyk - We have already cut the release-0.25 branch. We will have to merge this to main and then into the release branch as well for it to be deployed to cloud as part of 0.25.0 release.
> 
> Let's merge and test it in dev first.

sounds good, i will test in dev and then open up another PR for the merge into release branch after i validate.

PR #1073, opened 2021-04-29
adding docker.io in circleci configj2

rahat-astro:
## 📋 Checklist

Add the docker.io context after deleting the variables from the project local vars.
Related to astronomer/issues#2841

### DIFF
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -142,6 +142,7 @@ workflows:
             - lts-upgrade-0-23-test
           context:
             - quay.io
+            - docker.io
           filters:
             branches:
               only:


danielhoherd:
@rahat-astro I'm pretty sure we don't need this change. astronomer only interacts with quay.io, not docker. I see `DOCKER_REPO` in the circleci config, but that's combined with QUAY_* credentials. We are going to need the gcp-astronomer-prod context though, which includes the GCP_TOKEN.

danielhoherd:
I accidentally replicated this work in https://github.com/astronomer/astronomer/pulls, but added the GCP_TOKEN there, so I'm closing this PR in favor of that one.

PR #1074, opened 2021-04-29
Use CircleCI contexts

danielhoherd:
Use contexts for docker, gcp bucket, github. Relates to https://github.com/astronomer/issues/issues/2841


danielhoherd:
This was taken care of in a different PR.

PR #1094, opened 2021-05-12
Replace LTS 16-to-23 upgrade with 23-to-25

danielhoherd:
Now that we're onto a new LTS with 0.25, we need a new LTS upgrader! Instead of adding a new upgrader next to the old one, I have replaced the old 16-to-23 upgrader. This is because tests for 16-to-23 will only pass when the repo chart version is 0.23.

Alongside adding the upgrader, I'm doing some small improvements to helper scripts, and reorganizing some stuff.

Related to https://github.com/astronomer/issues/issues/2827


vishwas-astro:
@danielhoherd probably a naive question.

Does this mean that we wont be able to fix any more issues with our 0.16 to 0.23 upgrader scripts?

danielhoherd:
Not at all! To do that we would beach off of release-0.23, make changes, and merge back to release-0.23

vishwas-astro:
> Not at all! To do that we would branch off of release-0.23, make changes, and merge back to release-0.23

phew! thank you!

migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/rollback.yaml
```diff
@@ -52,11 +52,11 @@
...
-    msg: "DB username: {{ db_username }}"
-- debug:
-    msg: "DB hostname: {{ db_hostname }}"
-- debug:
-    msg: "DB port: {{ db_port }}"
+    msg:
+      - "DB hostname: {{ db_hostname }}"
+      - "DB port: {{ db_port }}"
+      - "DB username: {{ db_username }}"
+      - "DB password (sha512): {{ db_password | password_hash('sha512') }}"
```
vishwas-astro, line 59: do you want to log the password like this? What is the need to do that?

danielhoherd: good catch. i had put this in while troubleshooting with the intent to revert before pushing. i'll go ahead and remove that, especially given that this is not part of the issue i was troubleshooting.

migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml
```diff
@@ -233,17 +244,6 @@
...
-  register: prisma1_to_2
-  postgresql_query:
-    port: "{{ db_port }}"
-    autocommit: yes
-    login_host: "{{ db_hostname }}"
-    login_user: "{{ db_username }}"
-    login_password: "{{ db_password }}"
-    db: "{{ database_name }}_houston"
-    path_to_script: "files/prisma1_to_prisma2.sql"
-
```
vishwas-astro, line 246: @danielhoherd have you confirmed steps 2 and 3 below for the 0.23 to 0.25 upgrade? I dont think there were any Houston data migrations done between 0.23 and 0.25

danielhoherd: This file no longer exists.

migrations/scripts/upgrade-to-lts.sh
```diff
@@ -482,7 +407,7 @@ function main {
   echo "Done! Please contact Astronomer support if any issues are detected."
   echo
   echo "Please install the new CLI:"
-  echo "curl -sSL https://install.astronomer.io | sudo bash -s -- v0.16.1"
+  echo "curl -sSL https://install.astronomer.io | sudo bash -s -- v0.23.4"
   echo
   echo "You may choose to upgrade Airflow versions by changing your Dockerfile, for example:"
   echo "FROM quay.io/astronomer/ap-airflow:1.10.10-alpine3.10-onbuild"
```
vishwas-astro, line 413: you may want to change it to the most recent 2.0.x build of airflow and point to debian image instead of alpine as we are no longer building alpine images

migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml
vishwas-astro: but you have moved it to 0.23-to-0.25/playbooks/tasks/upgrade.yaml right? And retaining all the 6 steps of the upgrade process?

danielhoherd: I am still working on getting tests to pass. I haven't tuned the flow yet, but you're right, it looks like steps 2 and 3 can be deleted. Thanks for catching that!

vishwas-astro: @rbankston @andriisoldatenko - please review the steps below and confirm which ones are necessary for 0.23 to 0.25 upgrade

migrations/scripts/upgrade-to-lts.sh
danielhoherd: Done.

migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml
rbankston: If there isn't a database change how does https://github.com/astronomer/issues/issues/2920 get fixed?

danielhoherd: This test case is not caught by our upgrade testing because it requires an existing deployment to trigger the failure.

rahat-astro (approved):

rahat-astro (approved):

PR #1099, opened 2021-05-14
Improve LTS 23 to 25 upgrader

danielhoherd:
- Reformat `docker build` line (no functional changes)
- Show output of `helm history` several times during testing
- Use python list [-2] vs `tail` so we can more easily print helm history
- Delete a doc that was not needed (verified by CRE team)
- Point upgrade doc to official Astronomer docs page
- Expose `ASTRONOMER_UPGRADER_NAMESPACE` as env var so we can test that the upgrader is being run in `default`
- Add ansible.cfg with some helpful additions (no functional changes, just info display alterations)
- Add test that the upgrader is running in `default` namespace
- Add feature to use `helm install --devel` when using `USE_INTERNAL_HELM_REPO` so we pick up pre-release versions during testing
- Reformat `helm upgrade` lines so they're not so long

Closes astronomer/issues#2994


andriisoldatenko:
@danielhoherd can you please fix unittest-charts?

danielhoherd:
@andriisoldatenko they're not broken, i just canceled them because i didn't touch anything in the chart itself. i'm re-running the helm tests now.

andriisoldatenko:
Cool, after quick look LGTM.

PR #1102, opened 2021-05-20
🚧 POC of OpenShift Router -> Nginx Ingress -> Apps

ashb:
This draft PR proposes a mechnaism to allow the platform (and Airflow deployments) to run on an OpenShift cluster.

It operates under the assumption that we **will not** be able to make any global config changes to OpenShift components (such as the router) and so we have to work around some limitations.

This approach is under the assumption that the customer wishes all HTTP access to go via their existing configured OpenShift router. If they are happy to have an additional router/IngressController that is externally accessible to the cluster then there are better approaches we could take (i.e. Sharding Routes by labels.)

## Background

Under the current design of the platform, all of the apps (Kibana, Grafana, Registry) and all Airflow deployments are protected by the Nginx Ingress controller sending a _separate_ auth request to Houston before allowing the request to proceed to the backend service. This behaviour is a requirement to enforce authentication and authorization without heavy and serious rework of all the components.

## How this approach works

This aims to be the most minimally disruptive from an architectural point of view, and have all traffic still go through an single Nginx Ingress Controller.

The main difference here is that when `openShift.useRouter` value is set to true we make our Nginx Ingress controller only listen on a ClusterIP (instead of the more common LoadBalancer) meaning it is only accessible inside the cluster.

It then creates `route.openshift.io/v1.Route` resources to connect up the existing "public" ingress to send all traffic for relevant subdomains to the Nginx controller (which then sends it on to the backend service)

```
                                                                      ┌───────────────────┐
                  ┌───────────────┐        ┌──────────────────┐       │┌──────────────────┴┐
Incoming          │               │        │                  │       ││                   │
HTTP Request      │  OpenShift    │        │ Nginx Ingress    │       ││   Backend         │
      ───────────►│  Router       ├───────►│ Controller       ├──────►││   Service         │
                  │               │        │                  │       └┤                   │
                  └───────────────┘        └──────────────────┘        └───────────────────┘


```

## Drawbacks

There is a race condition when _first_ deploying the platform, where depending on which of the "internal" Ingress routes and the "public" onces get picked up by the Router.

This is because the Router only allows one Route to "own" a domain, so if the wrong route owns the domain in the router the internal ingress resource is the config which is used (and due to the NetworkPolicy we attach on those components, is inaccessible as they only allow traffic from our Nginx IC)

- If the external/public ones are picked up first, then this is the case we want. and everything works
- If the internal route is picked up first, we need to manually fix it up.

It can be checked by running `kubectl -n astronomer get route -l astronomerIngress=internal` and seeing if any of those are _not_ showing HostAlreadyClaimed.

The manual fix up steps are:

  - delete the conflicting ingress resource
  - run helm upgrade to get it back.

As soon as we delete the conflicting resource the Router will then give the host to the public route, so when it's re-created it will "loose" the race (as we want)

A better fix for this would be to set up the Router to use sharding so that ignores these internal Routes, but that requires a global config change to the cluster.

### DIFF
--- a/charts/alertmanager/templates/ingress.yaml
+++ b/charts/alertmanager/templates/ingress.yaml
@@ -2,6 +2,36 @@
 ## alertmanager Ingress
 #################################
 {{- if .Values.global.baseDomain }}
+{{- if .Values.global.openShift.useRouter }}
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: {{ template "alertmanager.fullname" . }}
+  labels:
+    tier: alertmanager-networking
+    component: alertmanager-ingress
+    chart: {{ template "alertmanager.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  host: {{ template "alertmanager.url" . }}
+  to:
+    kind: Service
+    name: "{{ .Release.Name }}-nginx"
+    weight: 100
+  wildcardPolicy: None
+  {{- if or .Values.global.acme .Values.global.tlsSecret }}
+  port:
+    targetPort: https
+  tls:
+    termination: passthrough
+    insecureEdgeTerminationPolicy: Redirect
+  {{- else }}
+  port:
+    targetPort: http
+  {{- end }}
+---
+{{- end }}
 kind: Ingress
 apiVersion: {{ template "apiVersion.Ingress" . }}
 metadata:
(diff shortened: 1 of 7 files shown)


ashb:
> defaulting to false but for Openshift this has to be set to "true" right?

Yes

vishwas-astro:
@ashb @pgvishnuram - can we close this PR?

pgvishnuram:
@ashb - this PR can be closed

PR #1103, opened 2021-05-20
POC on NginxAuthSidecar

pgvishnuram:
This draft PR proposes a way to allow traffic from openshift ingress to  connect platform services with sidecar pattern.

This approach is under the assumption that the customer wishes all HTTP access to go via their existing OpenShift router avoiding our nginx ingress installation.

## Background

Under the current design of the platform, all of the apps (Kibana, Grafana, Registry) and all Airflow deployments are protected by the Nginx Ingress controller sending a separate auth request to Houston before allowing the request to proceed to the backend service. This behavior is a requirement to enforce authentication and authorization without heavy and serious rework of all the components.

## How this approach works

This approach works by deploying a nginx sidecar which handles all the auth requests similar to a what ingress controller does but inside individual service. With this way openshift routes can be used with no additional effort of introducing new ingress mechanism.

## Drawbacks
- Introducing sidecar adds additional app deployment and config map configuration. 
- redirect config and other nginx location block configs will be managed seperately.
- when we use custom certs other than openshift default ones it creates additional responsibility to copy/sync  certs to each airflow deployments.
 

related to: astronomer/issues#3009


values.yaml
```diff
@@ -75,7 +75,12 @@ global:
   defaultDenyNetworkPolicy: true
 
   # Enable security context constraints required for OpenShift
-  sccEnabled: false
+  sccEnabled: true
+
+  # Deploy auth sidecar to use openshift native features
+  sidecar:
+    enabled: true
```
vishwas-astro, line 82: default is false I suppose

values.yaml
```diff
@@ -75,7 +75,12 @@ global:
   defaultDenyNetworkPolicy: true
 
   # Enable security context constraints required for OpenShift
-  sccEnabled: false
+  sccEnabled: true
```
vishwas-astro, line 78: default is false here as well i suppose

pgvishnuram: default is false for ocp it will be true

values.yaml
pgvishnuram: seems i have added my local values.yaml changes

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
```diff
@@ -26,12 +26,22 @@ spec:
           tier: monitoring
           component: prometheus
           release: {{ .Release.Name }}
+    {{- if .Values.global.sidecar.enabled  }}
```
danielhoherd, line 29: this value of

danielhoherd: I guess that since the old one was static we don't need to add logic to allow more general use.

values.yaml
pgvishnuram: updated values.yaml and pushed the changes with default values

values.yaml
pgvishnuram: updated values.yaml and pushed the changes with default values

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
pgvishnuram: okay daniel let me update those as well.

charts/astronomer/templates/ingress.yaml
```diff
@@ -13,13 +13,18 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     heritage: {{ .Release.Service }}
   annotations:
+    {{- if .Values.global.sidecar.enabled  }}
+    kubernetes.io/ingress.class: "default"
+    route.openshift.io/termination: "reencrypt"
+    {{- else }}
```
schnie, line 19: Could potentially decouple the openshift specifics here out to a separate flag.

Also could decouple and pass these annotations through as `extraAnnotations` on each component and document / provide sample configs.

schnie:
Looking good. Left a few in-line comments, but was wondering if we could maybe use a helm helper to consolidate and DRY up some of the NGINX configurations. Looks like they are pretty close, with a few small differences that maybe could be arguments on a helper or something?

Also wondering about the transfer of the TLS private key secret into the Airflow namespaces. This could potentially let a user launch a KPO and mount that secret, read that private key and then do something evil. Since this TLS cert is used cluster-wide, it would essentially allow someone to assume the identity of our services, which _could_ give them access to data they wouldn't have access to down the line, after they've crafted an attack using the cert. I don't know enough about the Openshift router to understand how or why this change was needed, but seems like a potential leak of data.

danielhoherd:
Thanks for the insight on the astronomer-tls cert @schnie, that is a huge red flag on this approach.

pgvishnuram:
This is the default behavior of any ingress controller. if we don't mention tls cert it will use the default configured one, in our case we are binded with our own *.basedomain cert . This is not applicable if we use some other ingress controller that is by default assigned with some other certificate.

danielhoherd:
> This is the default behavior of any ingress controller.

This is not how our existing ingress controller works though, is it? We are exposing a secret to users that was not previously exposed. We need to understand what is possible when a user gains access to that secret. If it breaks the isolation between namespaces of our user deployments, that would be a show-stopper.

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
andriisoldatenko, line 29:
```suggestion
    {{- if .Values.global.authSidecar.enabled  }}
```

pgvishnuram: Im have modified the entire chart doing on round of testing - will push with updated one.

pgvishnuram: sidecar name moved to authSidecar

pgvishnuram: sidecar name moved to authSidecar

charts/astronomer/templates/ingress.yaml
pgvishnuram: this change has been moved to extraAnnotations flag

andriisoldatenko:
@schnie fixed via https://github.com/astronomer/astronomer/pull/1103/commits/f6ee52f63b785f8a7896f443f26fd1ad86fa3467

> Looking good. Left a few in-line comments, but was wondering if we could maybe use a helm helper to consolidate and DRY up some of the NGINX configurations. Looks like they are pretty close, with a few small differences that maybe could be arguments on a helper or something?
> 
> Also wondering about the transfer of the TLS private key secret into the Airflow namespaces. This could potentially let a user launch a KPO and mount that secret, read that private key and then do something evil. Since this TLS cert is used cluster-wide, it would essentially allow someone to assume the identity of our services, which _could_ give them access to data they wouldn't have access to down the line, after they've crafted an attack using the cert. I don't know enough about the Openshift router to understand how or why this change was needed, but seems like a potential leak of data.

andriisoldatenko:
@danielhoherd @schnie @ashb @jedcunningham PTAL!

andriisoldatenko (approved):

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
```diff
@@ -26,12 +26,22 @@ spec:
...
           component: ingress-controller
           release: {{ .Release.Name }}
+    {{- end }}
     ports:
     - protocol: TCP
       port: {{ .Values.ports.http }}
+    {{- if .Values.global.authSidecar.enabled  }}
+    - protocol: TCP
+      port: 8080
+    {{- end }}
```
jedcunningham, line 46: I feel like we should have either/or, not both ports here?

charts/alertmanager/templates/alertmanager-service.yaml
```diff
@@ -22,3 +22,9 @@ spec:
       protocol: TCP
       port: {{ .Values.ports.http }}
       targetPort: {{ .Values.ports.http }}
+    {{- if .Values.global.authSidecar.enabled }}
+    - name: nginx-ui
+      protocol: TCP
+      port: 8080
+      targetPort: 8080
+    {{- end}}
```
jedcunningham, line 30: And here too.

charts/alertmanager/templates/alertmanager-sidecar.yaml
```diff
@@ -0,0 +1,41 @@
...
+  labels:
+    tier: monitoring
+    component: {{ template "alertmanager.name" . }}
+    chart: {{ template "alertmanager.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  default.conf: |
+    upstream astro-alertmanager {
+      server {{ .Release.Name }}-alertmanager:{{ .Values.ports.http }} ;
```
jedcunningham, line 18: And don't go back through the service here, use `localhost` instead.

Also, I feel like this filename should be `alertmanager-sidecar-configmap.yaml`, or be moved into `alertmanager-configmap.yaml`?

jedcunningham (commented):
Same comments apply to other components too.

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
pgvishnuram: This is to avoid internal service connection failure. if we connect to nginx-ui ( it needs an authenticated connection).

charts/alertmanager/templates/alertmanager-sidecar.yaml
pgvishnuram: @jedcunningham  both changes are added.

jedcunningham: I don't see the filename change?

jedcunningham: Looks good.

jedcunningham (approved):

charts/alertmanager/templates/alertmanager-sidecar-configmap.yaml
```diff
@@ -0,0 +1,41 @@
+################################
+## Alertmanager sidecar ConfigMap
```
jedcunningham, line 2:
```suggestion
## Alertmanager auth sidecar ConfigMap
```

And filename `alertmanager-auth-sidecar-configmap.yaml`.

jedcunningham (changes requested):
The comment applies to the other components as well.

values.yaml
```diff
@@ -77,6 +77,70 @@ global:
   # Enable security context constraints required for OpenShift
   sccEnabled: false
 
+  # Deploy auth sidecar to use openshift native features
+  authSidecar:
+    enabled: false
+    image: nginxinc/nginx-unprivileged
+    pullPolicy: IfNotPresent
+    default_nginx_settings: |
```
danielhoherd, line 85: I think these defaults should be in the template, not here. The config here should be.
```suggestion
    nginx_settings: ~
```
Then the template should have `{{- if .Values.global.authSidecar.nginx_settings }}`

values.yaml
```diff
@@ -77,6 +77,70 @@ global:
...
+      proxy_set_header            X-Auth-Request-Redirect $request_uri;
+      proxy_buffering             off;
+      proxy_buffer_size           4k;
+      proxy_buffers               4 4k;
+      proxy_request_buffering     on;
+      proxy_http_version          1.1;
+      proxy_ssl_server_name       on;
+      proxy_pass_request_headers  on;
+      client_max_body_size        1024m;
+    default_nginx_settings_location: |
```
danielhoherd, line 103: I think these defaults should be in the template, not here. The config here should be.
```suggestion
    nginx_settings_location: ~
```
Then the template should have `{{- if .Values.global.authSidecar.nginx_settings_location }}`

charts/alertmanager/templates/alertmanager-sidecar-configmap.yaml
andriisoldatenko: fixed via e670326f1432f6ed151c598e8e6c663dc648a322

jedcunningham: @andriisoldatenko, the same thing applies to the other components as well, e.g. the grafana configmap. I just didn't want to spam tons of comments.

andriisoldatenko (approved):

charts/prometheus/templates/prometheus-sidecar-configmap.yaml
```diff
@@ -0,0 +1,42 @@
...
+      listen [::]:8080  ;
+      location  = /auth {
+        proxy_set_header Host houston.{{ .Values.global.baseDomain }};
+        proxy_pass  https://houston.{{ .Values.global.baseDomain }}/v1/authorization;
+        {{ .Values.global.default_nginx_settings }}
+{{.Values.global.authSidecar.default_nginx_settings | indent 8  }}
+      }
+      location @401_auth_error {
+      	internal;
+      	add_header Set-Cookie $auth_cookie;
```
andriisoldatenko, line 32: @pgvishnuram can you please tweak formating? sounds like mixing spaces and tabs?

danielhoherd: I have a PR open that add pre-commit to CI, which has a hook to catch tabs. https://github.com/astronomer/astronomer/pull/1128

danielhoherd: I just merged that branch to master, then rebased this branch off master. Will run pre-commit on this branch, which will fix the problem.

PR #1108, opened 2021-05-25
Enable non-RFC address space for Alertmanager

fhoda:
## Description

Some enterprises use private networks outside of RFC 1918, for example Class E networks. These are necessary for enterprises that have exhausted or otherwise allocated all RFC 1918 address space. When this is done Alertmanager  crash loops because it is not able to use these address spaces. 

https://github.com/prometheus/alertmanager/issues/2438

This PR adds the ability to `enableNonRFC` address range usage by adding the related `env` and `args` to the helm chart. 

## Enable non-RFC address space for Alertmanager

🎟 Issue(s)

Resolves astronomer/issues#3028

## 🧪  Testing

- This has been tested in a sandbox on GKE as a helm chart, by both enabling and disabling the flag.
- Also has been tested with a customer deployment by editing the STS with the same values directly.
- Risks are that the values are not created properly thus non-RFC address spaces are still not supported by the helm chart.

## 📋 Checklist

### DIFF
--- a/charts/alertmanager/values.yaml
+++ b/charts/alertmanager/values.yaml
@@ -58,4 +58,4 @@ customRoutes: []
 #     tier: platform
 #     namespace: ^(test-ns.*)
 
-enableNonRFC: false
+enableNonRFC1918: false
--- a/charts/alertmanager/values.yaml
+++ b/charts/alertmanager/values.yaml
@@ -57,3 +57,5 @@ customRoutes: []
 #   match_re:
 #     tier: platform
 #     namespace: ^(test-ns.*)
+
+enableNonRFC: false
--- a/charts/alertmanager/templates/alertmanager-statefulset.yaml
+++ b/charts/alertmanager/templates/alertmanager-statefulset.yaml
@@ -42,7 +42,7 @@ spec:
           image: {{ template "alertmanager.image" . }}
           imagePullPolicy: {{ .Values.images.alertmanager.pullPolicy }}
           env:
-            {{- if .Values.enableNonRFC }}
+            {{- if .Values.enableNonRFC1918 }}
             - name: POD_IP
               valueFrom:
                 fieldRef:
@@ -61,7 +61,7 @@ spec:
           {{- if .Values.disableClustering }}
             - --cluster.listen-address=
           {{- end }}
-          {{- if .Values.enableNonRFC }}
+          {{- if .Values.enableNonRFC1918 }}
             - --cluster.advertise-address=$(POD_IP):9094
           {{- end }}
           ports:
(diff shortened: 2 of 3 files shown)


vparekh94:
@vishwas-astro doesn't look like I have permission to assign this to you, but heads up this should go into the next release. Looks like a pretty minor change

vishwas-astro:
> @vishwas-astro doesn't look like I have permission to assign this to you, but heads up this should go into the next release. Looks like a pretty minor change

@vparekh94 - the issue https://github.com/astronomer/issues/issues/3028 is already assigned to me. Will include it in 0.25.3.

charts/alertmanager/values.yaml
```diff
@@ -57,3 +57,5 @@ customRoutes: []
 #   match_re:
 #     tier: platform
 #     namespace: ^(test-ns.*)
+
+enableNonRFC: false
```
danielhoherd, line 61: Can we be more specific with this variable name?

```suggestion
enableNonRFC1918: false
```

danielhoherd:
@fhoda I would love to see some tests in this PR.

fhoda:
> @fhoda I would love to see some unit tests in this PR.

Happy to do so, just need to learn how to do that for helm charts :)

danielhoherd:
@fhoda Take a look at these files for reference, and if you need more help we can zoom and go over it:

- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/stan/tests/resources_test.yaml
- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/stan/tests/statefulset_test.yaml
- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/nats/tests/resources_test.yaml
- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/astronomer/tests/houston_pdb_test.yaml
- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/astronomer/tests/houston_api_houston-deployment_test.yaml
- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/nginx/tests/nginx-service_test.yaml
- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/prometheus/tests/prometheus-config-configmap_test.yaml
- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/prometheus/tests/prometheus-statefulset_test.yaml
- https://github.com/astronomer/astronomer/blob/20c82f9e2f/charts/prometheus/tests/prometheus-alerts-configmap_test.yaml

rbankston (approved):
Changes look good to me. Do like the name change for the variable

danielhoherd:
I met with @fhoda today and tried to make the tests work but the helm-unittest version we are using has bugs that are giving us false positives. I told him to commit a best effort test as bread crumbs for us to follow when we migrate to pytest. Related to https://github.com/astronomer/issues/issues/2701

charts/alertmanager/values.yaml
fhoda: changed it in my most recent commits.

fhoda:
I have pushed the unit tests as recommended by @danielhoherd and rebased with master.

danielhoherd (approved):

PR #1110, opened 2021-05-25
[bugfix] fixes elasticsearch on openshift

paulczar:
## Description

elasticsearch was failing on openshift clusters because it was trying to run as user 0 with privileged mode.  This adds a SCC with the correct permissions for it.

elasticsearch-client was failing because it needs to run as user 0 with privileged mode like the masters do. 

Logs of elasticsearch-client without ^ set:

```
kubectl logs astronomer-elasticsearch-client-656bfdf6c7-zhpb8
chroot: cannot change root directory to '/': Operation not permitted
```

## 📋 Checklist

### DIFF
--- a/tests/astronomer-scc-privileged_test.yaml
+++ b/tests/astronomer-scc-privileged_test.yaml
@@ -1,0 +1,20 @@
+---
+suite: Test astronomer-scc-privileged
+templates:
+  - scc/astronomer-scc-privileged.yaml
+tests:
+  - it: should exist
+    set:
+      global:
+        sccEnabled: True
+    asserts:
+      - isKind:
+          of: SecurityContextConstraints
+  - it: should not exist
+    set:
+      global:
+        sccEnabled: False
+    asserts:
+      - hasDocuments:
+          count: 0
+
--- a/charts/elasticsearch/templates/client/es-client-deployment.yaml
+++ b/charts/elasticsearch/templates/client/es-client-deployment.yaml
@@ -73,11 +73,12 @@ spec:
         command: ["sysctl", "-w", "vm.max_map_count=262144"]
         securityContext:
           privileged: true
-      serviceAccountName: {{ template "elasticsearch.fullname" . }}          
+      serviceAccountName: {{ template "elasticsearch.fullname" . }}
       containers:
       - name: es-client
         securityContext:
-          privileged: false
+          privileged: true
+          runAsUser: 0
           capabilities:
             add:
               - IPC_LOCK
--- a/templates/scc/astronomer-scc-privileged.yaml
+++ b/templates/scc/astronomer-scc-privileged.yaml
@@ -1,0 +1,39 @@
+{{- if .Values.global.sccEnabled }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  annotations:
+    release.openshift.io/create-only: "true"
+    "helm.sh/hook": "pre-install,pre-upgrade"
+  name: {{ .Release.Name }}-privileged
+priority: 10
+readOnlyRootFilesystem: false
+requiredDropCapabilities:
+- MKNOD
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: MustRunAs
+supplementalGroups:
+  type: RunAsAny
+users:
+- system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-elasticsearch
+volumes:
+- configMap
+- downwardAPI
+- emptyDir
+- persistentVolumeClaim
+- projected
+- secret
+defaultAddCapabilities: null
+fsGroup:
+  type: RunAsAny
+allowHostDirVolumePlugin: false
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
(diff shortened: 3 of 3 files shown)


andriisoldatenko:
Can you please add tests?

paulczar:
Ahh yeah I can look to, do you have details on the testing framework?  and what tests you'd expect?  a quick (non-exhaustive) grep shows that there's nothing testing for `scc` or `privileged` for the existing manifests, but I'm not familiar with the testing tools you're using so it could be just an issue with me.

danielhoherd:
@paulczar here's a write-up I did on how to develop tests https://github.com/astronomer/airflow-chart/blob/master/tests/README.md

paulczar:
Thanks,  I'll circle back to this in a couple of days when I have a few minutes.

danielhoherd:
> Thanks, I'll circle back to this in a couple of days when I have a few minutes.

@paulczar FYI there is a nuance to your situation that makes it deviate outside more simple test procedures. Our helm chart is an umbrella chart, and your test case requires setting global values. Usually, tests go into the subchart's test dir, EG: <https://github.com/astronomer/astronomer/tree/master/charts/elasticsearch/tests>. However, global values cannot be set in a sub-chart, and your change involves global values, so you have to put your tests into the [parent chart's test dir](https://github.com/astronomer/astronomer/tree/master/tests).

Here is an example of how to test global values in an umrella helm chart: <https://github.com/astronomer/astronomer/blob/master/tests/alertmanager-statefulset_test.yaml>

paulczar:
This now has a test for the SCC fix.

I had issues writing a test for the es client fix, I think because of the subdirectory, but since none of those files have existing tests I don't see a lot of value in fighting the test framework to add just that one thing.

pritt20:
Hey @paulczar ,

Thanks for the PR and details :) The reason why this was not included in Astronomer installation package since start just like other service accounts is that openshift admins do not allow us to provide privilege access to ElasticSearch stack without which stack will not work in current scenario and hence we advise client to keep `logging` flag as `false` in their config.

We are working towards an approach where we could make use of openshift native logging stack or may be modify the astronomer existing ES stack to run without privilege access, so that we can have it included in installation package for all clients.

do you think we can get some help from redhat to integrate openshift native logging solution with astronomer ? :)

paulczar:
hey @pritt20 that's a good point.  I think its useful to get this in so we have a /working/ stack for openshift admins that will allow this for POC etc, beyond that I'm pretty sure we can work on the integration with the native logging solution, which as long as you're logging to stdout etc it should just get picked up and then its up to the logging endpoint (could be es/kibana, could be cloudwatch in AWS, could be ?? in Azure ) to visualize the logs etc.

danielhoherd:
@paulczar regarding tests, since Andrii initially asked for tests, we have migrated this repo to a pytest based chart testing framework that should be a lot easier and a lot more flexible. I'm not asking you to rewrite the tests you already wrote, just letting you know that you aren't the only one who has had issues writing tests, and we have moved to something a lot better. Here's my writeup of the new system: https://github.com/astronomer/astronomer/blob/master/tests/README.md

rob-1126:
I've got a similar branch over at add-elastic-scc-policy that I'll be testing on a fresh openshift cluster tomorrow.

danielhoherd:
@rob-1126 If your branch could re-use anything in this branch, that would be cool so @paulczar could get some clout for participating in this feature and be included in the contributors history.

rob-1126:
I think we can, be glad to merge it into this branch if Paul doesn't mine. I did'nt get around to testing today but hopefully tomorrow.

rob-1126:
Previously working scripted resource creation of ARO cluster is failing with vague error messages. All quotes looked good at first glance.  Will have to revisit later. Maybe will work monday.

`+ az aro create --resource-group robaro07 --name robaro07 --vnet robaro07 --master-subnet robaro07-master --worker-subnet robaro07-worker --domain aro.robaro07.testing --apiserver-visibility Private --ingress-visibility Private --master-vm-size Standard_D8s_v3 --worker-count 6 --worker-vm-size Standard_D8as_v4
No --pull-secret provided: cluster will not include samples or operators from Red Hat or from certified partners.
(InternalServerError) Internal server error.
Code: InternalServerError
Message: Internal server error.`

tests/astronomer-scc-privileged_test.yaml
```diff
@@ -0,0 +1,20 @@
+---
+suite: Test astronomer-scc-privileged
```
andriisoldatenko, line 2: let's rewrite it with pytest.
@danielhoherd wdyt?

andriisoldatenko (changes requested):

tests/astronomer-scc-privileged_test.yaml
danielhoherd: I think this PR isn't needed anymore. We now have [`privileged: true` in master](https://github.com/astronomer/astronomer/blob/05f14f4570/charts/elasticsearch/templates/client/es-client-deployment.yaml#L78) and afaik we do not want `runAsUser: 0`, though maybe that is different with openshift. @pgvishnuram what do you think?

danielhoherd: I think we would want to solve this with `fsGroup`, not `runAsUser: 0`

PR #1113, opened 2021-05-27
Remove cruft. Make pre-commit pass on entire bin dir.

danielhoherd:
- Delete some unused scripts that are potentially dangerous
- Make create-initial-user.json valid json
- Add chmod +x where needed
- Echo out commands as instructions in `drain.sh` instead of executing them to make this script more safe
- Delete old bin/kind scripts for manual testing in GCP that AFAIK we have never used (we manually test in CircleCI)
- Delete `update-cloud-module` which is not used anymore and was only relevant to DroneCI
- Silence some shellcheck complaints by ignoring [SC1091](https://github.com/koalaman/shellcheck/wiki/SC1091)


isaulv (approved):
LGTM

PR #1115, opened 2021-05-28
add ability to use options for random data

andriisoldatenko:
resolves 3rd part of https://github.com/astronomer/issues/issues/3007

>Our chart uses Helm's random number generator in certain areas. This random number generator changes each time Argo evaluates those sections and marks the application as not in sync. It would be useful if there was an option to specify the variable instead of only having the Random Data generate those values. This is used in 6 places in our current repository but the houston-backend-secret and registry-statefulset appear to be the only places that a production grade Astronomer deployment would use.


rbankston (approved):
Looks good to me

danielhoherd:
Whoops! @andriisoldatenko I merged without seeing that tests were failing. Were you still working on this?

andriisoldatenko:
Yep I'll revert

PR #1118, opened 2021-06-03
Check chart for deprecated API versions

danielhoherd:
As we support new k8s versions we need a way to maintain confidence that our chart is functional with the API versions included in these new versions.

Originally I had looked into kubepug, but it did not function against files, instead requiring a running kubernetes system. This in itself was not bad because we test various versions in kind, but what was happening is the k8s api was mutating the manifest that we gave, which was creating false positives. EG: when we used `Ingress` `apiVersion: networking.k8s.io/v1`, what was created in the cluster was `apiVersion: extensions/v1beta1`, which was a violation.

Hat tip to @rbankston for showing me [`kubent`](https://github.com/doitintl/kube-no-trouble) which solves the above file problem.

Related to: <https://github.com/astronomer/issues/issues/3052>

- Install kubent alongside other ci tools
- Run kubent in CI on each version of tested k8s
- exclude templates from yaml pre-commit check
- Refactor some test code to be more concise, idiomatic
- Bump some test component versions
- Quote var in comparison
- Bump rbac everywhere to v1
- Use template for cronjob template everywhere
- Remove duplicate cronjob.apiVersion template
- Use apiregistration v1
- Remove templates handling old kubernetes version compatibilities


danielhoherd:
FYI checks are failing right now because kubepug is working and reporting failures that exist in our chart.

vishwas-astro:
@danielhoherd did you mean to close this?

danielhoherd:
@vishwas-astro no. I was trying to rename the branch to have the ticket number in it, but then I remembered I already had a PR open and was trying to find a way to re-target the PR to a different branch.

jpweber (approved):

tommyk (approved):

danielhoherd:
I've set up kubent to only look at the files generated by helm, not the cluster or helm installations. I've validated that it catches errors by checking out the astronomer v0.16.0 tag and running it in the same manner set up in CI:
```
$ helm template --kube-version=v1.14.0 -f values.yaml --set global.baseDomain=example.com . | kubent --cluster=false --helm2=false --helm3=false --filename -
7:14PM INF >>> Kube No Trouble `kubent` <<<
7:14PM INF version 0.4.0 (git sha 3d82a3f0714c97035c27374854703256b3d69125)
7:14PM INF Initializing collectors and retrieving data
7:14PM INF Retrieved 147 resources from collector name=File
7:14PM INF Loaded ruleset name=custom.rego.tmpl
7:14PM INF Loaded ruleset name=deprecated-1-16.rego
7:14PM INF Loaded ruleset name=deprecated-1-22.rego
__________________________________________________________________________________________
>>> Deprecated APIs removed in 1.16 <<<
------------------------------------------------------------------------------------------
KIND          NAMESPACE     NAME                                API_VERSION    REPLACE_WITH (SINCE)
StatefulSet   <undefined>   RELEASE-NAME-alertmanager           apps/v1beta2   apps/v1 (1.9+)
StatefulSet   <undefined>   RELEASE-NAME-elasticsearch-data     apps/v1beta2   apps/v1 (1.9+)
StatefulSet   <undefined>   RELEASE-NAME-elasticsearch-master   apps/v1beta2   apps/v1 (1.9+)
StatefulSet   <undefined>   RELEASE-NAME-prometheus             apps/v1beta2   apps/v1 (1.9+)
StatefulSet   <undefined>   RELEASE-NAME-registry               apps/v1beta2   apps/v1 (1.9+)
```

PR #1122, opened 2021-06-07
Support k8s 1.20, Ingress v1

danielhoherd:
1Ingress v1 introduces some syntax changes that are handled in this PR. One thing I am currently not sure about is if we need to change anything related to IngressClass. [The spec says the annotation is deprecated](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation), but I am not sure if the annotation we have been using is necessary for functionality or was more of a descriptive property for human consumption. The docs say that `ingressClassName` was added in 1.18, which we are already using, yet we do not have any instances of `ingressClassName` in our chart. I would appreciate any insights into how we are using these features.

- Add kubernetes 1.20
- Test first and last supported version of k8s, instead of two most recent
- Improve some testfile syntaxes
- Bump some test dependency versions
- Handle Ingress v1 differences from v1beta1, retroactive to 1.19
- Drop <=1.14 specific configs since we no longer support those versions


jpweber:
I don't think the ingress class annotation could actually provide any functionality because its set to `{release name}-nginx` which doesn't match an ingress controller. 

That annotation, and subsequently the `IngressClass` added to the `Ingress` spec is to inform an ingress controller if it should be concerned with it or not. For example if you want to run two ingress controller types you would use the ingressClass to define which controller should handle that particular ingress. 

For example if you had `nginx` and `istio` as ingress controllers in the same cluster; then you had a resource with something like `ingresClass == nginx` istio would ignore that Ingress Resource.

jpweber (approved):

tommyk:
did we test these changes deployed in a cluster to make sure it all still functions the same way? i know there a bunch of tests steven wrote, but not sure how far/deep they go about testing the ingress...  either way looks like it should work to me!

tommyk (approved):

danielhoherd:
I just ran a sanity check in dev, but that is on 1.18 so a lot of the contents here cannot be tested. It does give us confidence that I didn't break anything pre-1.19 related with these changes.

PR #1124, opened 2021-06-11
Add basic helm chart museum deployment and service for Airgapped / Network Restricted

fritz-astronomer:
## Description
For airgapped installs - one method to host our `airflow` helm chart is to include a helm chart museum locally in the client environment. This should be an opt-in only feature, that the client can then use `helm pull` / `helm push` to.

See more info here: https://www.notion.so/astronomerio/WIP-Airgapped-Install-744136e9a04041539283ff0ed910d9ba

More info is needed from chart maintainers around best practices for helm charts

## 🧪  Testing
- [ ] ` helm template astronomer --namespace=astronomer --set global.chartMuseumEnabled=true . | less` - everything shows up correctly

### DIFF
--- a/charts/chart-museum/templates/deployment.yaml
+++ b/charts/chart-museum/templates/deployment.yaml
@@ -1,0 +1,42 @@
+{{- if .Values.global.chartMuseumEnabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: helm-museum
+  namespace: astronomer
+spec:
+  selector:
+    matchLabels:
+      app: helm-museum
+  template:
+    metadata:
+      labels:
+        app: helm-museum
+    spec:
+      securityContext:
+        fsGroup: 1000
+      containers:
+      - name: helm-museum
+        image: ghcr.io/helm/chartmuseum:v0.13.1
+        env:
+          - name: STORAGE
+            value: "local"
+          - name: STORAGE_LOCAL_ROOTDIR
+            value: "/charts"
+        volumeMounts:
+          - mountPath: "/charts"
+            name: helm-chart-museum
+        resources:
+          requests:
+            memory: "32Mi"
+            cpu: "100m"
+          limits:
+            memory: "128Mi"
+            cpu: "500m"
+        ports:
+        - containerPort: 8080
+      volumes:
+        - name: helm-chart-museum
+          persistentVolumeClaim:
+            claimName: helm-chart-museum
+{{- end }}
--- a/charts/chart-museum/templates/deployment.yaml
+++ b/charts/chart-museum/templates/deployment.yaml
@@ -2,21 +2,36 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: helm-museum
+  name: {{ .Release.Name }}-chartmuseum
+  labels:
+    component: chartmuseum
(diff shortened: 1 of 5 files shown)


charts/chart-museum/templates/deployment.yaml
```diff
@@ -0,0 +1,42 @@
+{{- if .Values.global.chartMuseumEnabled }}
```
fritz-astronomer, line 1: I imagine that the namespace, securityContext, image + tag, limits+requests, and storage should all be configurable? Any other thoughts?

charts/chart-museum/templates/service.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: helm-museum
+spec:
+  type: ClusterIP
+  selector:
+    app: helm-museum
+  ports:
+  - port: 8080
+    targetPort: 8080
+{{- end }}
```
fritz-astronomer, line 13: could have an ingress, too, that'd probably make things easier. Don't know how much of a lift that'd be, to update other things to include it. It'd only be accessible on opt-in and via vpn or whatever the customer has configured
```
https://helm.BASEDOMAIN
```

fritz-astronomer:
I still need to bring up an airgapped enviroment and attempt to install this and do the whole process. That work is in progress, but this should be mostly correct, still, at least.

charts/chart-museum/templates/deployment.yaml
danielhoherd: Shouldn't the namespace be the same as the astronomer deployment? IE: if this is a component of astronomer, shouldn't it live with astronomer?

fritz-astronomer: Makes sense to me. I guess functionally it's similar to the `registry` which is also located there? I am not sure I fully grok when something is "astronomer" vs not - but that can be a convo for another day. Will move this

charts/chart-museum/templates/service.yaml
danielhoherd: We might as well reuse the existing ingress that hosts everything else like grafana, kibana, etc..

charts/chart-museum/templates/deployment.yaml
danielhoherd: Also yes on configurable limits, requests, and storage.

fritz-astronomer: looks like the houston stuff just didn't have a namespace - so I removed it

charts/chart-museum/templates/pvc.yaml
```diff
@@ -0,0 +1,17 @@
...
+    {{- range .Values.persistence.accessModes }}
+    - {{ . | quote }}
+    {{- end }}
+  resources:
+    requests:
+      storage: {{ .Values.persistence.size | quote }}
+    {{- if .Values.persistence.storageClassName }}
+      storageClassName: {{ .Values.persistence.storageClassName }}
+    {{- end }}
+{{- end }}
```
andriisoldatenko, line 17:
```suggestion
{{- end }}

```

fritz-astronomer:
After slack conversation - I think the consensus is that we don't want to add more to the helm chart and that would require support, this may be a client responsibility

PR #1129, opened 2021-06-17
Minor auth sidecar cleanup

jedcunningham:
A few renames and comment changes were missed in the original PR (#1103).

### DIFF
--- a/charts/grafana/templates/grafana-auth-sidecar-configmap.yaml
+++ b/charts/grafana/templates/grafana-auth-sidecar-configmap.yaml
@@ -1,4 +1,4 @@
-################################
+#################################
 ## Grafana auth sidecar ConfigMap
 #################################
 {{- if .Values.global.authSidecar.enabled  }}
--- a/charts/grafana/templates/grafana-auth-sidecar-configmap.yaml
+++ b/charts/grafana/templates/grafana-auth-sidecar-configmap.yaml
@@ -1,30 +1,29 @@
 ################################
-## Alertmanager sidecar ConfigMap
+## Grafana auth sidecar ConfigMap
 #################################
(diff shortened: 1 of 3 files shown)


charts/grafana/templates/grafana-auth-sidecar-configmap.yaml
```diff
@@ -1,5 +1,5 @@
 ################################
```
danielhoherd, line 1: Nitpick, but while we're cleaning up these three comments, can we make the hash lines consistent?
```suggestion
#################################
```

charts/kibana/templates/kibana-auth-sidecar-configmap.yaml
```diff
@@ -1,5 +1,5 @@
 ################################
```
danielhoherd, line 1: Nitpick, but while we're cleaning up these three comments, can we make the hash lines consistent?
```suggestion
#################################
```

charts/prometheus/templates/prometheus-auth-sidecar-configmap.yaml
```diff
@@ -1,5 +1,5 @@
 ################################
```
danielhoherd, line 1: Nitpick, but while we're cleaning up these three comments, can we make the hash lines consistent?
```suggestion
#################################
```

danielhoherd (approved):

charts/grafana/templates/grafana-auth-sidecar-configmap.yaml
jedcunningham: Works for me. They were consistently inconsistent everywhere so I left them 🤷‍♂️

PR #1132, opened 2021-06-21
Changes for new OSS based Helm chart

jedcunningham:
The Astronomer Helm chart is being changed to use the OSS chart as a subchart, and only add Astronomer specific extras on top of it.

These are the changes that are necessary to use the new chart.


danielhoherd:
This pr was closed because the branch was renamed to `3118-oss-airflow-chart` and there is no way to retarget a PR branch.

PR #1133, opened 2021-06-22
Allow publishing feature branches to our internal helm repository.

danielhoherd:
Related to astronomer/astronomer#3092

# Description

Allow publishing feature branches to our internal helm repository. This follows a similar pattern to what we have in airflow-chart, where the feature has been quite useful. I've been asked many times recently to publish to internal from feature branches, so this makes that a feature that will be present as a manual step for all feature branches.

# Screenshots
### Before
before
### After
after

### Before
before

### After
after

# Testing

I have not tested how this affects release branches, but the code appears to be fine because the chart is not modified at all, and the same logic that was there before is followed. This logic has been moved from being directly in the Makefile into a shell script, which can be better linted by `shellcheck`.

# Itemized changes
- Allow internal release steps to be run for non-release branches
- Tweak public release to rely on all platforms being run, but internal release only requires the two default platforms (first and last supported)
- Move chart build steps from Makefile into shell script
- Tweak chart build steps to alter the version strings of the Astronomer umbrella chart and the astronomer sub-chart when running on a non-release git branch
- Bump helm to [3.6.1 (security release)](https://github.com/helm/helm/releases/tag/v3.6.1)
- Bump gcloud just to get a new git sha to test against (got mixed up while testing a branch of a branch with the same sha)


jedcunningham (approved):

kaxil (approved):

danielhoherd:
I have verified that this works as expected on feature branches.
```
$ helm search repo --max-col-width=0 --devel -l internal/astro | head
2021-06-22T14:28:27-0700
NAME                          	CHART VERSION    	APP VERSION      	DESCRIPTION
astronomer-internal/astronomer	0.25.3-rc2       	0.25.3-rc2       	Helm chart to deploy the Astronomer Platform
...snip...
astronomer-internal/astronomer	0.25.2-build19280	0.25.2-build19280	2021-06-22T21:10:26+0000 https://git.io/Jnyax 3092-publish-to-internal https://circleci.com/gh/astronomer/astronomer/19280
```

PR #1134, opened 2021-06-22
Added the port-name of prometheus-data instead of prometheus in Prometheus Ingress

rbankston:
## Description

When using a newer k8s version the prometheus ingress port name isn't correct causing a 503 error because of the port name.

## 🎟 Issue(s)

Part of astronomer/issues#3126

## 🧪  Testing

Verified that my clusters prometheus page now works without manually changing the port name 

## 📸 Screenshots

Broken:

```
spec:
  rules:
  - host: prometheus.airflow.ralphbankston.com
    http:
      paths:
      - backend:
          service:
            name: uat-prometheus
            port:
              name: prometheus
        path: /
        pathType: Prefix
```
Fixed:

```
spec:
  rules:
  - host: prometheus.airflow.ralphbankston.com
    http:
      paths:
      - backend:
          service:
            name: uat-prometheus
            port:
              name: prometheus-data
        path: /
        pathType: Prefix
```
## 📋 Checklist

### DIFF
--- a/tests/test_prometheus_ingress.py
+++ b/tests/test_prometheus_ingress.py
@@ -9,7 +9,7 @@ from . import supported_k8s_versions
     supported_k8s_versions,
 )
 class TestIngress:
-    def test_basic_ingress(self, kube_version):
+    def test_prometheus_ingress(self, kube_version):
         # sourcery skip: extract-duplicate-method
         docs = render_chart(
             kube_version=kube_version,
--- a/tests/test_prometheus_ingress.py
+++ b/tests/test_prometheus_ingress.py
@@ -28,7 +28,7 @@ class TestIngress:
 
         if minor >= 19:
             assert doc["apiVersion"] == "networking.k8s.io/v1"
-            assert "RELEASE-NAME-prometheus-ingress" in [
+            assert "RELEASE-NAME-prometheus" in [
                 name[0]
                 for name in jmespath.search(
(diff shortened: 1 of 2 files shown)


danielhoherd:
@rbankston can we get a test for this? I can zoom with you if you want to go over how that might work.

rbankston:
@danielhoherd is there a test I didn't see from where this was added in https://github.com/astronomer/astronomer/pull/1122 ?

tests/test_prometheus_ingress.py
```diff
@@ -0,0 +1,57 @@
...
+import pytest
+from . import supported_k8s_versions
+
+
+@pytest.mark.parametrize(
+    "kube_version",
+    supported_k8s_versions,
+)
+class TestIngress:
+    def test_basic_ingress(self, kube_version):
```
danielhoherd, line 12:
```suggestion
    def test_prometheus_ingress(self, kube_version):
```

danielhoherd (changes requested):
Oooh, this change is less complicated than I thought. For some reason I thought this would be different in different versions of k8s, but it's a prom component version thing. Cool. Thanks for adding test! Just one small change and this can be merged.

rbankston:
Yeah definitely not a big change but for k8s 1.20 and 1.21 makes the prometheus.basedomain usable. That change has been committed to change the name of the function

PR #1141, opened 2021-06-29
syncing master back with release 0.25 branch

vishwas-astro:
So that Master is also referring to the right airflow and astronomer chart versions.


danielhoherd:
Will do this by hand

PR #1147, opened 2021-07-02
add ablity to use option when random data

andriisoldatenko:


vishwas-astro:
This PR looks like a no-op to me.

vishwas-astro:
What about other changes from this PR: https://github.com/astronomer/astronomer/pull/1115/files?

andriisoldatenko:
@vishwas-astro because i'm still working on it.

andriisoldatenko:
> What about other changes from this PR: https://github.com/astronomer/astronomer/pull/1115/files?

other changes already in master
```
https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/registry/registry-statefulset.yaml#L58-L63
```

andriisoldatenko:
closing since all commits already in master:

https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/_helpers.yaml#L64-L72
https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/registry/registry-statefulset.yaml#L58-L63

PR #1149, opened 2021-07-06
Execute git-rev-parse from the script dir, not PWD. Bump some CI things.

danielhoherd:
There are a bunch of places in this repo where we assign `GIT_ROOT` from the output of `git rev-parse --show-toplevel`, but this command is issued from the PWD of the shell that is issuing the command, which may not always be within the same repo as where the command resides, particularly if we are debugging and running these scripts manually. This change set specifies the repo to always be that of the script where `git rev-parse` is run from by issuing `-C <path>    Run as if git was started in <path> instead of the current working directory` from the directory of the script itself, determined by stripping `/*` from script name via `"${0%/*}"`.

Also bump some CI things.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 jobs:
   run_pre_commit:
     docker:
-      - image: quay.io/astronomer/ci-pre-commit:2021-06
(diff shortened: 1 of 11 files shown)


bote795 (approved):
lgtm

PR #1157, opened 2021-07-13
fix auth side car through nginx

aliotta:


danielhoherd (changes requested):
We currently don't have any tests for anything authSidecar related. I would love to see some tests. If you want to meet and go through it together we can do that. If you want to take a crack at it, check https://github.com/astronomer/astronomer/blob/master/tests/README.md and the rest of the tests dir for more examples.

PR #1160, opened 2021-07-19
Add ServiceType of NGINX to be configurable and externalIPs

rbankston:
## Description

Allows the NGINX ServiceType to be configurable within our chart. Default is left at LoadBalancer so only folks that need it would go looking for it.

Also adds ability to add an external ip to the NGINX service.

## 🎟 Issue(s)

Resolves astronomer/issues#3136
Resolves astronomer/issues#3204

## 🧪  Testing

Created tests for the nginx service for LoadBalancer, NodePort, and ClusterIP. Most people wouldn't notice this change unless it is needed for a customized on-prem configuration

Also added tests for adding an externalip

## 📸 Screenshots

LoadBalancer:
  sessionAffinity: None
  type: LoadBalancer

ClusterIP:
  sessionAffinity: None
  type: ClusterIP

NodePort:
  sessionAffinity: None
  type: NodePort

## 📋 Checklist

- [x] Passing tests

### DIFF
--- a/charts/nginx/templates/nginx-service.yaml
+++ b/charts/nginx/templates/nginx-service.yaml
@@ -32,7 +32,7 @@ metadata:
 spec:
   type: {{ .Values.serviceType }}
   {{- if .Values.externalIPs }}
-  externalIps: 
+  externalIps:
     - {{ .Values.externalIPs }}
   {{- end }}
   {{- if .Values.loadBalancerIP }}
--- a/charts/nginx/templates/nginx-service.yaml
+++ b/charts/nginx/templates/nginx-service.yaml
@@ -30,7 +30,7 @@ metadata:
     prometheus.io/scrape: "true"
     prometheus.io/port: {{ .Values.ports.metrics | quote }}
 spec:
-  type: LoadBalancer
+  type: {{ .Values.serviceType }}
   {{- if .Values.loadBalancerIP }}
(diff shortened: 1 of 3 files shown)


vishwas-astro:
@rbankston can you fix the pre-commit check failures here please?

vishwas-astro:
@fritz-astronomer and/or @pgagnon  please review this PR.

charts/nginx/templates/nginx-service.yaml
```diff
@@ -30,7 +30,11 @@ metadata:
     prometheus.io/scrape: "true"
     prometheus.io/port: {{ .Values.ports.metrics | quote }}
 spec:
-  type: LoadBalancer
+  type: {{ .Values.serviceType }}
```
andriisoldatenko, line 33: Looks good, but i was wondering how to check if user add by mistake wrong serviceType?

rbankston: Were you thinking of having a bunch of if statements and if the value was lb or something that we would put in LoadBalancer, and if they said nodeport we would put in NodePort, and ClusterIp?

I contemplated that but thought passing the value would look nicer and if a customer had an extension we wouldn't have to add explicit support for it.

andriisoldatenko (approved):

pgvishnuram (approved):

PR #1168, opened 2021-07-29
Drop k8s 1.16

danielhoherd:
## Description

k8s 1.16 has been dropped by all supported cloud providers, so we should drop support for it as well.

```
| Cloud provider | k8s version | Cloud provider EOL | Passed |
| -------------- | ----------- | ------------------ | ------ |
| Upstream       | 1.18        | 2021-04-30         | X      |
| AKS            | 1.18        | 2021-06            | X      |
| GKE            | 1.16        | 2021-07-03         | X      |
| EKS            | 1.16        | 2021-07-25         | X      |
| EKS            | 1.17        | 2021-09            |
| Upstream       | 1.19        | 2021-09-30         |
| EKS            | 1.18        | 2021-10            |
| GKE            | 1.17        | 2021-11            |
| Upstream       | 1.20        | 2021-12-30         |
```

Also notable is that 1.17 will be dropped soon as well.

## Itemized changes

- Delete unused kind-loop script
- Delete tests related to 1.16 (some references to 1.16 still exist in logic like `>=1.16<1.2x`)

Related to <https://github.com/astronomer/issues/issues/3280>


danielhoherd:
@jwitz @paolaperaza FYI we can adjust docs to reflect that we no longer support 1.16 now that it has been deprecated by all of the cloud providers we support and will be dropped from our support.

rdodev (approved):

PR #1174, opened 2021-08-09
Implement external elastic logging for platform

pgvishnuram:
* separate sub-chart for elastic-search proxy
* supports fluentd with enable/disable option
* Houston config support to use elastic-search proxy
* config changes for values.yml 
* support for elastic cloud
* support for aws elasticsearch/opensearch
* support for onprem/k8s installed elasticsearch 

## 🎟 Issue(s)

https://github.com/astronomer/issues/issues/3259

## 📸 Screenshots

image

## 📋 Checklist


pgvishnuram:
custom logging or external logging which name we can use for this feature

pgvishnuram:
Hi All - Added all changes including secrets from env - pls review my change and let me know if the changes looks good.

vishwas-astro:
@pgvishnuram can you remove the "Draft" state from the PR and update title to remove "WIP"?

vishwas-astro:
> custom logging or external logging which name we can use for this feature

This applies only to an external elastic stack right. Should we call it "external_es_logging"?
@danielhoherd @rbankston @pgagnon - do you have a preference?

charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
```diff
@@ -0,0 +1,73 @@
...
+    events { worker_connections 1024; }
+    error_log /dev/stdout info;
+    env ES_SECRET;
+    env ES_SECRET_NAME;
+
+    http {
+      access_log /dev/stdout;
+      client_max_body_size 1024M;
+
+      #upstream external-es-proxy {
```
bote795, line 27: can we remove this? (seems commented out)

charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
```diff
@@ -0,0 +1,109 @@
...
+          env:
+{{- if .Values.global.custom_logging.secret }}
+          - name: ES_SECRET
+            value: {{.Values.global.custom_logging.secret }}
+{{- else if .Values.global.custom_logging.secretName }}
+          - name: ES_SECRET_NAME
+            valueFrom:
+              secretKeyRef:
+                name: {{.Values.global.custom_logging.secretName }}
+                key: elastic
```
pgvishnuram, line 59: @danielhoherd - we need proper naming for this

values.yaml
```diff
@@ -137,6 +137,15 @@ global:
       limits: {}
       requests: {}
 
+  # External ES logging
+  custom_logging:
```
pgvishnuram, line 141: @danielhoherd : we are yet to decide on custom_logging or external_logging to keep same name across all places.

values.yaml
```diff
@@ -137,6 +137,15 @@ global:
...
+    enabled: false
+    scheme: https
+    host: ""
+    port: ""
+    secret: ""
+    #secretName: ~
```
pgvishnuram, line 147: @danielhoherd - vishwas asked to keep something like secretKey instead of secretName.

charts/external-es-proxy/values.yaml
```diff
@@ -0,0 +1,61 @@
+# Default values for external-es-proxy.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
```
pgvishnuram, line 7: i have hardcoded the images here.

charts/external-es-proxy/templates/external-es-proxy-env-configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: {{ template "external-es-proxy.fullname" . }}-setenv
+  labels:
+    tier: external-logging
+    component: {{ template "external-es-proxy.name" . }}
+    chart: {{ template "external-es-proxy.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+data:
+  setenv.lua: |-
```
pgvishnuram, line 15: i have added the script as a configmap

values.yaml
danielhoherd: IMHO `custom_logging` makes more sense since it may not be external. IE: we could use this for logging to other locally hosted services.

values.yaml
danielhoherd: What we call it depends on what it is. If you mean "key" as in `key: value` that's probably not the right term. If you mean "key" as in "cryptographic key" then yeah, that works.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -281,7 +281,22 @@ data:
     # single namespace mode. Fluentd requires a ClusterRole to
     # enrich logs with kubernetes metadata. In a single namespace environment,
     # we're likely not going to get a ClusterRole from admins.
-    {{- if not .Values.global.singleNamespace }}
+    {{- if  .Values.global.custom_logging.enabled }}
```
jedcunningham, line 284:
```suggestion
    {{- if .Values.global.custom_logging.enabled }}
```

charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
jedcunningham: Maybe we use `envFrom` to populate `ES_SECRET`, then we only have a single env var to consider? Might also remove the need for the lua?

charts/external-es-proxy/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
+
+service:
+  type: ClusterIP
+  port: 80
```
jedcunningham, line 40: This isn't used, right?
```suggestion
```

charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
pgvishnuram: we added lua because - nginx doesn't support direct env read in nginx config , and the value we pass inside the auth is base64, to handle these too we added lua.

charts/external-es-proxy/values.yaml
pgvishnuram: instead of 
service:
  type: ClusterIP
  port: 80
---
ports:
  http: 9200
  secure-http: 9201
  
what do you think

charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
```diff
@@ -0,0 +1,68 @@
...
+        proxy_http_version 1.1;
+        {{- if not .Values.global.istio.enabled }}
+        proxy_set_header Host $host;
+        {{- end }}
+        proxy_set_header Connection "Keep-Alive";
+        proxy_set_header Proxy-Connection "Keep-Alive";
+
+        location ~ ^/ {
+          access_by_lua_file  /usr/local/openresty/nginx/conf/setenv.lua;
+          proxy_pass {{.Values.global.custom_logging.scheme}}://{{.Values.global.custom_logging.host}}:{{.Values.global.custom_logging.port}};
```
rbankston, line 39: Could we include a mechanism for passing and configuring a self signed certificate authority as an option? A lot of customers that this applies to would likely have private ca certs that they use for their own infrastructure.

rbankston (changes requested):
Not sure if it is in scope for this solution and maybe I missed the ability to while reviewing it but it doesn't appear we have the ability to either disable ssl verification or include the private root ca that often gets passed into our charts.

charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
pgvishnuram: yes sure, i didn't verify with self-signed/private certs.let me add necessary changes .

pgvishnuram:
@danielhoherd : has raised an issue of sending data unencryptedly from our service to external es services.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -281,7 +281,22 @@ data:
...
+      enabled: true
+      client:
+        host: {{ printf "%s-external-es-proxy:9201" .Release.Name }}
+        log: error
+    {{- end }}
+
+    {{- if and (not .Values.global.singleNamespace ) (not .Values.global.custom_logging.enabled ) }}
       elasticsearch:
         enabled: true
         connection:
```
danielhoherd, line 302: Can we break this up into two `if` blocks that don't span two different levels of yaml? The way it is now is confusing. Having fewer conditionals is not as important here as readability. Let's try to make this as easy as possible to understand, even if there are redundant conditions.

danielhoherd (changes requested):
We need to have tests for these additions.

charts/astronomer/templates/houston/houston-configmap.yaml
danielhoherd: I still think this should be done.

values.yaml
danielhoherd: what about `token` or `credentials`? do either of those fit better?

danielhoherd (changes requested):
Adding tests to this PR will give us a lot of guidance on what the state of this feature is.

danielhoherd (changes requested):
LGTM but pre-commit is failing and there are merge conflicts.

Chart.yaml
```diff
@@ -84,3 +84,6 @@ dependencies:
     condition: global.stan.enabled
     tags:
       - stan
+  # external-es-proxy
+  - name: external-es-proxy
+    condition: global.custom_logging.enabled
```
thmzlt, line 89: Why is this underscore rather than snake case?

charts/external-es-proxy/templates/external-es-proxy-networkpolicy.yaml
```diff
@@ -0,0 +1,46 @@
...
+    release: {{ .Release.Name }}
+spec:
+  podSelector:
+    matchLabels:
+      tier: external-logging
+      component: {{ template "external-es-proxy.name" . }}
+      release: {{ .Release.Name }}
+  policyTypes:
+  - Ingress
+  ingress:
```
thmzlt, line 22: Do we know in advance the target ES instance address/host? This is so we could add egress rules as well.

charts/external-es-proxy/values.yaml
```diff
@@ -0,0 +1,68 @@
+# Default values for external-es-proxy.
```
thmzlt, line 1: Can we run this through a linter?

values.yaml
thmzlt: https://helm.sh/docs/chart_best_practices/values/#naming-conventions

thmzlt (changes requested):
I didn't see any blockers, but I wonder if we can do a better job with the nomenclature. There is "custom logging", "external logging", "external es proxy", etc. It would be a big win to consolidate that.

I think we should get in the habit of including a JSON schema file for new charts as we have just seen how beneficial it is.

values.yaml
pgvishnuram: so need to move to customlogging

pgvishnuram: @danielhoherd @vishwas-astro - pls confirm on the name

charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
danielhoherd: Seems like this conversation can be resolved?

pgvishnuram: Yes we now support trust certs as well

values.yaml
danielhoherd: Did we ever make a decision here?

values.yaml
danielhoherd: seems like `customLogging` is it

`git grep -l custom_logging | gsed -i 's/custom_logging/customLogging/g'`

pgvishnuram: Renamed to customLogging

values.yaml
pgvishnuram: yes - we are going with below formats
```bash
secret: "" - for base64 encode 
secretName: ~ - to provide k8s secret
awsSecretName: ~ - for aws credentials
awsIAMRole: ~ - for deployment/namespace level annotation 
awsServiceAccountAnnotation: ~ - for IRSA use case 
```

Chart.yaml
pgvishnuram: This one is addressed

thmzlt (approved):

values.yaml
danielhoherd: So the comments are correct? If so resolve this thread.

charts/external-es-proxy/values.yaml
pgvishnuram: astronomer managed non root images are added

pgvishnuram:
@danielhoherd can you do a rc release please

vishwas-astro:
> @danielhoherd can you do a rc release please

after reviewing/approving and merging this PR right?

PR #1175, opened 2021-08-10
Allow customization of cert-copy image

danielhoherd:
First crack at allowing customization of cert-copy image. I'm not sure everything here is in the right place, so feedback is welcome.

This change also alters the way that pytest reports failures, attempting to be more helpful. Here is an example exception:
```
ERROR: subprocess.CalledProcessError:
helm command: helm template --kube-version 1.20.0 RELEASE-NAME /Users/danielh/a/astronomer --set global.baseDomain=example.com --values /var/folders/9p/kbxv3fnn447706mqhnf7bncm0000z8/T/tmpmovq26rc --show-only templates/trust-private-ca-on-all-nodes/daemonset.yaml
Values file contents:
---------------------
global:
  privateCaCerts:
  - private-ca-cert-foo
  - private-ca-cert-bar
  privateCaCertsAddToHost:
    enabled: false
---------------------
error.output=b''
error.stderr=b'Error: could not find template templates/trust-private-ca-on-all-nodes/daemonset.yaml in chart\n'
ERROR: command is probably using templates with null output, which usually means there is a helm value that needs to be set to render the content of the chart.
command: helm template --kube-version 1.20.0 RELEASE-NAME /Users/danielh/a/astronomer --set global.baseDomain=example.com --values /var/folders/9p/kbxv3fnn447706mqhnf7bncm0000z8/T/tmpmovq26rc --show-only templates/trust-private-ca-on-all-nodes/daemonset.yaml
```

Related to <https://github.com/astronomer/issues/issues/3088>

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -19,6 +19,11 @@ global:
   privateCaCertsAddToHost:
     enabled: false
     hostDirectory: /etc/docker/certs.d
+    certCopier:
+      repository: alpine
+      tag: 3.14
+      pullPolicy: IfNotPresent
+
 
   # Use kube-lego
   acme: false
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-*       @danielhoherd @andriisoldatenko 
+*       @danielhoherd @andriisoldatenko
--- a/templates/trust-private-ca-on-all-nodes/daemonset.yaml
+++ b/templates/trust-private-ca-on-all-nodes/daemonset.yaml
@@ -35,7 +35,7 @@ spec:
       serviceAccountName: {{ .Release.Name }}-private-ca
       containers:
       - name: cert-copy
-        image: alpine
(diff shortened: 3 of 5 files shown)


andriisoldatenko (approved):

fritz-astronomer (approved):
setting things in `global` is still a bit of an odd concept but lgtm

values.yaml
```diff
@@ -19,6 +19,11 @@ global:
   privateCaCertsAddToHost:
     enabled: false
     hostDirectory: /etc/docker/certs.d
+    certCopier:
+      repository: alpine
+      tag: 3.14
```
vishwas-astro, line 24: @danielhoherd should this be 3.14.1?

danielhoherd: 3.14 is a floating tag, and matches whatever the most recent patch version of 3.14 is. we can do a more precise tag if we want.

vishwas-astro: Cool thanks @danielhoherd .. floating tag is fine.

PR #1182, opened 2021-08-14
global storage class change to apply on all platform deployments

pgvishnuram:
## Description

global storageclass values to reflect on all platform deployments

## 🎟 Issue(s)

https://github.com/astronomer/issues/issues/3313

## 📸 Screenshots

image

### DIFF
--- a/charts/alertmanager/templates/_helpers.tpl
+++ b/charts/alertmanager/templates/_helpers.tpl
@@ -54,29 +54,19 @@ Return  the proper Storage Class
 Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
 but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else logic.
 */}}
-{{- if .Values.global -}}
-    {{- if .Values.global.storageClass -}}
-        {{- if (eq "-" .Values.global.storageClass) -}}
-            {{- printf "storageClassName: \"\"" -}}
-        {{- else }}
-            {{- printf "storageClassName: %s" .Values.global.storageClass -}}
-        {{- end -}}
-    {{- else -}}
-        {{- if .Values.persistence.storageClassName -}}
-              {{- if (eq "-" .Values.persistence.storageClassName) -}}
-                  {{- printf "storageClassName: \"\"" -}}
-              {{- else }}
-                  {{- printf "storageClassName: %s" .Values.persistence.storageClassName -}}
-              {{- end -}}
-        {{- end -}}
+{{- if .Values.global.storageClass -}}
+    {{- if (eq "-" .Values.global.storageClass) -}}
+        {{- printf "storageClassName: \"\"" -}}
+    {{- else }}
+        {{- printf "storageClassName: %s" .Values.global.storageClass -}}
     {{- end -}}
 {{- else -}}
     {{- if .Values.persistence.storageClassName -}}
-        {{- if (eq "-" .Values.persistence.storageClassName) -}}
-            {{- printf "storageClassName: \"\"" -}}
-        {{- else }}
-            {{- printf "storageClassName: %s" .Values.persistence.storageClassName -}}
-        {{- end -}}
+          {{- if (eq "~" .Values.persistence.storageClassName) -}}
+              {{- printf "storageClassName: \"\"" -}}
+          {{- else }}
+              {{- printf "storageClassName: %s" .Values.persistence.storageClassName -}}
+          {{- end -}}
     {{- end -}}
 {{- end -}}
 {{- end -}}
--- a/charts/alertmanager/templates/_helpers.tpl
+++ b/charts/alertmanager/templates/_helpers.tpl
@@ -45,3 +45,38 @@ Image name.
 {{ define "alertmanager.url" -}}
 alertmanager.{{ .Values.global.baseDomain }}
 {{- end }}
+
+{{/*
+Return  the proper Storage Class
+*/}}
+{{- define "alertmanager.storageClass" -}}
+{{/*
+Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
+but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else logic.
+*/}}
+{{- if .Values.global -}}
+    {{- if .Values.global.storageClass -}}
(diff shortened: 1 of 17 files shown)


danielhoherd (changes requested):
I'd like to see tests with this. They can just cover these templates, and the `.Values.global.storageClass` and `.Values.persistence.storageClassName` present and absent cases.

charts/alertmanager/templates/_helpers.tpl
```diff
@@ -45,3 +45,38 @@ Image name.
...
+    {{- else -}}
+        {{- if .Values.persistence.storageClassName -}}
+              {{- if (eq "-" .Values.persistence.storageClassName) -}}
+                  {{- printf "storageClassName: \"\"" -}}
+              {{- else }}
+                  {{- printf "storageClassName: %s" .Values.persistence.storageClassName -}}
+              {{- end -}}
+        {{- end -}}
+    {{- end -}}
+{{- else -}}
```
danielhoherd, line 73: Is this section necessary? Can we just drop this `else` block and the `if .Values.global` check and use only lines 58-72?

pgvishnuram:
I will add some test case for storage class check.

pgvishnuram:
```python
def test_alertmanager_globalstorageclass():
        docs = render_chart(
            values={"global": {"storageClass": "vishnusre"}},
            show_only=["charts/alertmanager/templates/alertmanager-statefulset.yaml"],
        )

        assert len(docs) == 1
        doc = docs[0]

        assert doc["spec"]["volumeClaimTemplates"][0]["spec"]["storageClassName"] == "vishnusre"
```

charts/alertmanager/templates/_helpers.tpl
pgvishnuram:
```yaml
{{- if .Values.global.storageClass -}}
    {{- if (eq "-" .Values.global.storageClass) -}}
        {{- printf "storageClassName: \"\"" -}}
    {{- else }}
        {{- printf "storageClassName: %s" .Values.global.storageClass -}}
    {{- end -}}
{{- else -}}
    {{- if .Values.persistence.storageClassName -}}
          {{- if (eq "~" .Values.persistence.storageClassName) -}}
              {{- printf "storageClassName: \"\"" -}}
          {{- else }}
              {{- printf "storageClassName: %s" .Values.persistence.storageClassName -}}
          {{- end -}}
    {{- end -}}
{{- end -}}
```

rbankston (approved):
Looks great to me. The helper is tested and already used by the Postgres chart as is.

pgvishnuram:
@danielhoherd  Im not sure below is correct way 
```python

def test_alertmanager_wild():
        """Test globalstorageclass feature of alertmanager statefulset template"""
        supported_global_storage_options = ["-", "vishnusre"]
        for options in supported_global_storage_options :
          docs = render_chart(
              values={"global": {"storageClass": options}},
              show_only=["charts/alertmanager/templates/alertmanager-statefulset.yaml"],
          )
          return docs
          assert len(docs) == 1
          doc = docs[0]
          if options == "-" : 
            assert doc["spec"]["volumeClaimTemplates"][0]["spec"]["storageClassName"] == ""
          elif options == "vishnusre" :
            assert doc["spec"]["volumeClaimTemplates"][0]["spec"]["storageClassName"] == "vishnure"
```

tests/test_global_storageclass.py
```diff
@@ -0,0 +1,27 @@
...
+    "charts/elasticsearch/templates/data/es-data-statefulset.yaml",
+    "charts/prometheus/templates/prometheus-statefulset.yaml",
+    "charts/astronomer/templates/registry/registry-statefulset.yaml",
+]
+
+
+@pytest.mark.parametrize(
+    "supported_types, expected_output", [("-", ""), ("vishnure", "vishnure")]
+)
+def test_alertmanager_global_storageclass(supported_types, expected_output):
```
danielhoherd, line 16:
```suggestion
def test_global_storageclass(supported_types, expected_output):
```

tests/test_global_storageclass.py
```diff
@@ -0,0 +1,27 @@
...
+    """Test globalstorageclass feature of alertmanager statefulset template"""
+    docs = render_chart(
+        values={"global": {"storageClass": supported_types}},
+        show_only=chart_files,
+    )
+    assert len(docs) == 5
+    doc = docs[0]
+    assert (
+        doc["spec"]["volumeClaimTemplates"][0]["spec"]["storageClassName"]
```
danielhoherd, line 25: This test is only testing the first chart, because `doc = docs[0]`. Run this through a `all(...jmespath.search` style list comprehension or something so we check all the docs.

pgvishnuram: @danielhoherd : added some updates for the test cases that may not be perfect but runs through all files now.

tests/test_alertmanager.py
```diff
@@ -61,3 +62,29 @@ def test_alertmanager_rfc1918():
         )
         for value in item
     )
+
```
danielhoherd, line 65: All of this content in test_alertmanager.py can be removed since it is done in test_global_storageclass.py

tests/test_global_storageclass.py
danielhoherd: I slacked you some details about how to improve it using two different ways.

tests/test_alertmanager.py
pgvishnuram: @danielhoherd  - test cases removed from alert manager.

pgvishnuram: If it's all good we can merge this to master.

danielhoherd (approved):
```
 █████         █████████  ███████████ ██████   ██████
░░███         ███░░░░░███░█░░░███░░░█░░██████ ██████
 ░███        ███     ░░░ ░   ░███  ░  ░███░█████░███
 ░███       ░███             ░███     ░███░░███ ░███
 ░███       ░███    █████    ░███     ░███ ░░░  ░███
 ░███      █░░███  ░░███     ░███     ░███      ░███
 ███████████ ░░█████████     █████    █████     █████
░░░░░░░░░░░   ░░░░░░░░░     ░░░░░    ░░░░░     ░░░░░
```

PR #1183, opened 2021-08-16
Add docs folks to CODEOWNERS

danielhoherd:
Add docs folks to CODEOWNERS to keep them updated with changes to things they want to document.


jwitz (approved):
Yes this looks great @danielhoherd ! Happy to be an owner for these two files/ file types. This will make updating docs surrounding Helm much easier.

andriisoldatenko (approved):

PR #1198, opened 2021-08-26
Rollback 0.23 to 0.16 fix (Added Airflow chart block)

shmanu017:
Adding Deletion of Scheduler deployment box in current Rollback script from 0.23.14 to 0.16.15.
Ref: Zendesk [5619](https://astronomer.zendesk.com/agent/tickets/5619)


bin/migration-scripts/lts-to-lts/0.16-to-0.23/playbooks/tasks/rollback.yaml
```diff
@@ -239,11 +239,41 @@
 # which leads to 2 init containers wait-for-airflow-migrations(0.19.0) +
 # run-airflow-migrations when performing rollback to 0.15.7 airflow chart
 - name: Delete All Scheduler Deployemnts
-  kubernetes.core.k8s:
+  community.kubernetes.k8s:
     api_version: v1
     state: absent
-    definition: "{{ lookup('kubernetes.core.k8s', api_version='apps/v1', kind='Deployment', label_selector='component = scheduler') }}"
+    definition: "{{ lookup('community.kubernetes.k8s', api_version='apps/v1', kind='Deployment', label_selector='component = scheduler') }}"
```
danielhoherd, line 245: I was just running into a problem during testing another branch where this lookup was breaking when it came back empty. I was trying to come up with some logic to do the lookup, then only do this deletion when the lookup return had 0 length, but wasn't able to solve it before switching to something else. Here's the path I was taking:

```yaml
- name: "Get schedulers"
  community.kubernetes.k8s_info:
    api_version: apps/v1
    kind: Deployment
    label_selectors: 'component=scheduler'
  register: schedulers

- debug:
    var: schedulers

- debug:
    msg: "skipping schedulers because they are null"
  when: (schedulers["resources"]|length==0)
```

danielhoherd:
@shmanu017 did you want to complete this work? I think it's unnecessary since 0.16 will not be supported much longer.

PR #1203, opened 2021-08-31
Fix flaky test by merging master version of test_config.py

danielhoherd:
A flaky prometheus test was fixed in release-0.25, but was not backported to release-0.23. We are now experiencing that flaky test, so this backports the workaround by using the version of test_config.py that is currently found in master. Also bump some library versions and add the required pytest rerunfailures module.


danielhoherd:
I don't know why the license failure is happening. This same module is used in this repo in other branches.

andriisoldatenko (approved):

PR #1206, opened 2021-09-01
Bump a bunch of dev versions.

danielhoherd:
While testing some 0.16 upgrades I ran into issues and noticed that we have some really old versions of k8s in the release-0.16 branch, and are defaulting to an unsupported k8s version. This PR removes [unsupported versions, IE less than 16](https://www.astronomer.io/docs/enterprise/v0.16/resources/version-compatibility-reference), and bumps some other dev component versions. Also delete the unused kind-loop script rather than bump kind versions inside of it.


vishwas-astro:
I dont think we have the cycles to test and release an 0.16 patch. May be we keep this PR open and do this when we get a breather?

danielhoherd:
We could merge this but just not release a new version. I agree though that releasing a new version of 0.16 is not high priority right now.

PR #1207, opened 2021-09-02
 Support for AWS elasticsearch

pgvishnuram:
* integration with aws elasticsearch
* support for aws keys
* general configs

## 📋 Checklist

### DIFF
--- a/charts/external-es-proxy/values.yaml
+++ b/charts/external-es-proxy/values.yaml
@@ -9,6 +9,10 @@ images:
     repository: openresty/openresty
     pullPolicy: IfNotPresent
     tag: latest
+  awsproxy:
+    repository: abutaha/aws-es-proxy
+    pullPolicy: IfNotPresent
+    tag: v1.1
 
 imagePullSecrets: []
 nameOverride: ""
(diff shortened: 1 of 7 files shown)


pgvishnuram:
@rbankston - i have tested with AWS CREDS and AWS IAM as well - can you have a look once

charts/external-es-proxy/values.yaml
```diff
@@ -9,6 +9,10 @@ images:
     repository: openresty/openresty
     pullPolicy: IfNotPresent
     tag: latest
+  awsproxy:
+    repository: abutaha/aws-es-proxy
```
rbankston, line 13: Saw the license which is permissive. Should we package and use an astronomer/ap-aws-es-proxy  so in the field we don't hit docker hub issues and also are able to verify the provenance of the image? @danielhoherd

rbankston (approved):
The code looks good to me. I just had a comment about the docker image from aws-es-proxy to include it as an ap type image.

charts/external-es-proxy/values.yaml
danielhoherd: yes, absolutely. we should vendor this into ap-vendor.

pgvishnuram: will be adding it to ap-vendor

PR #1210, opened 2021-09-07
make fluentd to gather logs from all namespaces if preCreatedNamespaces is enabled

andriisoldatenko:
https://github.com/astronomer/issues/issues/3424

### DIFF
--- a/charts/fluentd/templates/fluentd-configmap.yaml
+++ b/charts/fluentd/templates/fluentd-configmap.yaml
@@ -102,6 +102,7 @@ data:
       @type grep
       <regexp>
         key $.kubernetes.namespace_name
+        # fluentd should gather logs from all namespaces if preCreatedNamespaces is enabled
         {{- if .Values.global.preCreatedNamespacesEnabled }}
         pattern "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
         {{ else }}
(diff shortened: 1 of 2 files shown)


rdodev (approved):

charts/fluentd/templates/fluentd-configmap.yaml
```diff
@@ -102,7 +102,11 @@ data:
       @type grep
       <regexp>
         key $.kubernetes.namespace_name
+        {{- if .Values.global.preCreatedNamespacesEnabled }}
```
bote795, line 105: i think you should add a comment on why we doing this

bote795 (approved):

values.yaml
```diff
@@ -63,6 +63,9 @@ global:
   # Used to enable Celery autoscaling
   kedaEnabled: false
 
+  # make fluentd to gather logs from all avaialble namespaces
+  preCreatedNamespacesEnabled: false
```
vishwas-astro, line 67: I am generally worried about using slightly different flag names for the same feature in different places. Eg: we are calling it manualNamespaceNames in Houston configs, and calling it preCreatedNamespacesEnabled here and possibly something else in Commander. Can we standardize on a name across all our configs?

vishwas-astro: Also, this is a global flag which is currently used in FluentD to gather logs from all namespaces, but it could be used in other places as well. So don't just restrict your comment for that.

andriisoldatenko: sounds good.

andriisoldatenko:
@vishwas-astro PTAL

vishwas-astro (approved):

PR #1211, opened 2021-09-07
add upgrade timeout to helm chart

aliotta:
Add upgrade timeout to helm chart

### DIFF
--- a/tests/test_astronomer_commander.py
+++ b/tests/test_astronomer_commander.py
@@ -23,15 +23,12 @@ def test_astronomer_commander_deployment(kube_version):
     assert "quay.io/astronomer/ap-commander:0.25.2" in jmespath.search(
         "spec.template.spec.containers[*].image", doc
     )
-    envVarNames = jmespath.search("spec.template.spec.containers[0].env[*].name", doc)
-    assert "COMMANDER_UPGRADE_TIMEOUT" in envVarNames
-    upgradeTimeoutIndex = envVarNames.index("COMMANDER_UPGRADE_TIMEOUT")
-    assert (
-        doc["spec"]["template"]["spec"]["containers"][0]["env"][upgradeTimeoutIndex][
-            "value"
-        ]
-        == "300"
-    )
+    assert len(doc["spec"]["template"]["spec"]["containers"]) == 1
+    env_vars = {
+        x["name"]: x["value"]
+        for x in doc["spec"]["template"]["spec"]["containers"][0]["env"]
+    }
+    assert env_vars["COMMANDER_UPGRADE_TIMEOUT"] == "300"
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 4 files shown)


tests/test_astronomer_commander.py
```diff
@@ -23,3 +23,28 @@ def test_astronomer_commander_deployment(kube_version):
...
+    assert len(docs) == 1
+    doc = docs[0]
+    assert doc["kind"] == "Deployment"
+    assert doc["apiVersion"] == "apps/v1"
+    assert doc["metadata"]["name"] == "RELEASE-NAME-commander"
+    assert "quay.io/astronomer/ap-commander:0.25.2" in jmespath.search(
+        "spec.template.spec.containers[*].image", doc
+    )
+    assert doc["spec"]["template"]["spec"]["containers"][0]["env"][6]["name"] == "COMMANDER_UPGRADE_TIMEOUT"
+    assert doc["spec"]["template"]["spec"]["containers"][0]["env"][6]["value"] == "300"
```
vishwas-astro, line 50: May be I misunderstand this @aliotta - but dont you want to assert it was 600 since that is what you are setting in values above?

aliotta: yeah test was a wip I still can't get the test to pass actually. Going to test locally to see that it works outside the test framework. If so then the bug is in the test framework itself.

aliotta: ah just needed the astronomer prefix should be good now.

vishwas-astro (approved):

tests/test_astronomer_commander.py
```diff
@@ -23,3 +23,35 @@ def test_astronomer_commander_deployment(kube_version):
     assert "quay.io/astronomer/ap-commander:0.25.2" in jmespath.search(
         "spec.template.spec.containers[*].image", doc
     )
+    assert (
+        doc["spec"]["template"]["spec"]["containers"][0]["env"][6]["name"]
```
danielhoherd, line 27: rather than using `[6]` here we should use `jmespath.search`

tests/test_astronomer_commander.py
```diff
@@ -23,3 +23,35 @@ def test_astronomer_commander_deployment(kube_version):
...
+        == "COMMANDER_UPGRADE_TIMEOUT"
+    )
+    assert doc["spec"]["template"]["spec"]["containers"][0]["env"][6]["value"] == "300"
```
danielhoherd, line 30: same `jmespath` thing here

tests/test_astronomer_commander.py
```diff
@@ -23,3 +23,35 @@ def test_astronomer_commander_deployment(kube_version):
...
+    assert len(docs) == 1
+    doc = docs[0]
+    assert doc["kind"] == "Deployment"
+    assert doc["apiVersion"] == "apps/v1"
+    assert doc["metadata"]["name"] == "RELEASE-NAME-commander"
+    assert "quay.io/astronomer/ap-commander:0.25.2" in jmespath.search(
+        "spec.template.spec.containers[*].image", doc
+    )
+    assert (
+        doc["spec"]["template"]["spec"]["containers"][0]["env"][6]["name"]
```
danielhoherd, line 54: same `jmespath` thing here

tests/test_astronomer_commander.py
```diff
@@ -23,3 +23,35 @@ def test_astronomer_commander_deployment(kube_version):
...
+        == "COMMANDER_UPGRADE_TIMEOUT"
+    )
+    assert doc["spec"]["template"]["spec"]["containers"][0]["env"][6]["value"] == "600"
```
danielhoherd, line 57: same `jmespath` thing here

tests/test_astronomer_commander.py
danielhoherd: Something like https://github.com/astronomer/astronomer/blob/533d047afc/tests/test_privateca.py#L57-L60

danielhoherd (changes requested):
the `jmespath` suggestions should be real quick, then this looks good! I appreciate that you added tests on the first pass!

PR #1212, opened 2021-09-07
Add upgrade timeout to commander helm release23

aliotta:
Add upgrade timeout to helm chart

### DIFF
--- a/tests/test_astronomer_commander.py
+++ b/tests/test_astronomer_commander.py
@@ -20,16 +20,10 @@ def test_astronomer_commander_deployment(kube_version):
     assert doc["kind"] == "Deployment"
     assert doc["apiVersion"] == "apps/v1"
     assert doc["metadata"]["name"] == "RELEASE-NAME-commander"
-
-    envVarNames = jmespath.search("spec.template.spec.containers[0].env[*].name", doc)
-    assert "COMMANDER_UPGRADE_TIMEOUT" in envVarNames
-    upgradeTimeoutIndex = envVarNames.index("COMMANDER_UPGRADE_TIMEOUT")
-    assert (
-        doc["spec"]["template"]["spec"]["containers"][0]["env"][upgradeTimeoutIndex][
-            "value"
-        ]
-        == "300"
-    )
+    
+    assert len(doc["spec"]["template"]["spec"]["containers"]) == 1
+    env_vars = {x['name']: x['value'] for x in doc["spec"]["template"]["spec"]["containers"][0]['env']}
+    assert env_vars['COMMANDER_UPGRADE_TIMEOUT'] == "300"
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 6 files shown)


tests/test_astronomer_commander.py
```diff
@@ -0,0 +1,61 @@
+from tests.helm_template_generator import render_chart
```
aliotta, line 1: These work locally at least. Not sure if we want to get them into the ci conf

tests/test_astronomer_commander.py
```diff
@@ -0,0 +1,61 @@
...
+
+    envVarNames = jmespath.search("spec.template.spec.containers[0].env[*].name", doc)
+    assert "COMMANDER_UPGRADE_TIMEOUT" in envVarNames
+    upgradeTimeoutIndex = envVarNames.index("COMMANDER_UPGRADE_TIMEOUT")
+    assert (
+        doc["spec"]["template"]["spec"]["containers"][0]["env"][upgradeTimeoutIndex][
+            "value"
+        ]
+        == "300"
+    )
```
danielhoherd, line 32: Do you think this is easier to read if we reshape the `env` list using a dict comprehension?
```suggestion
    assert len(doc["spec"]["template"]["spec"]["containers"]) == 1
    env_vars = {x['name']: x['value'] for x in doc["spec"]["template"]["spec"]["containers"][0]['env']}
    assert env_vars['COMMANDER_UPGRADE_TIMEOUT'] == "300"
```

tests/test_astronomer_commander.py
danielhoherd: I've actually [started work on backporting pytest to 0.23](https://github.com/astronomer/astronomer/commit/3cd724f8afcd80ccbe0aab4daca5bb7ad6646d4c). Let's worry about that in the other branch, and just accept this as groundwork for now, while not going as far as putting it in CI. If we do any more work on 0.23 we'll push for full pytest support.

tests/test_astronomer_commander.py
```diff
@@ -0,0 +1,61 @@
...
+        show_only=["charts/astronomer/templates/commander/commander-deployment.yaml"],
+    )
+
+    assert len(docs) == 1
+    doc = docs[0]
+    assert doc["kind"] == "Deployment"
+    assert doc["apiVersion"] == "apps/v1"
+    assert doc["metadata"]["name"] == "RELEASE-NAME-commander"
+    
+    envVarNames = jmespath.search("spec.template.spec.containers[0].env[*].name", doc)
```
danielhoherd, line 53: Same reshaping question applies here.

tests/test_astronomer_commander.py
danielhoherd: Also I added a length assertion here because we are making assumptions about what `[0]` is.

danielhoherd (approved):

PR #1218, opened 2021-09-13
Tweak lts 0 25 upgrader

danielhoherd:
During testing we ran into a few issues with the LTS upgrader. This solves some of those issues, while others remain outstanding with issues opened.

## Itemized changes
- Test 23-25 LTS upgrade using the latest k8s version supported by 23
- Make the upgrade test run automatically on every commit
- Bump 1.21 test version from 1.21.1 to 1.21.2
- bump all helm 3 versions to 3.6.3
- Add a dev helper script that sets up some env stuff in CircleCI ssh
- Change the method of discovering `USE_INTERNAL_HELM_REPO` in LTS upgrade from using a passed in env var to using an env lookup function
- Use a more ansible-native, intelligent method of waiting for airflow-update-check-first-run (replaces `sleep 30`)
- Regenerate requirements files
- Pass `upgrade_to_version` straight through without any tricks, let helm deal with the version


vishwas-astro:
lgtm .. may be @rbankston should review it too.

andriisoldatenko (approved):

migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml
```diff
@@ -7,9 +7,10 @@
 - name: Set up variables
   set_fact:
     backup_dir: "/astronomer-backups"
-    timestamp: "{{ lookup('pipe','date +%Y-%m-%d-%H-%M-%S') }}"
-    upgrade_to_version: "0.25"
-    upgrade_to_version_airflow: "0.19"
+    timestamp: "{{ lookup('pipe', 'date +%Y-%m-%d-%H-%M-%S') }}"
+    upgrade_to_version: "0.25.6"  # this must be a complete semver
```
rbankston, line 11: Why does this have to be a complete server? We have left the upgrade to version to 0.23 in the 0.16 to 0.23 version. This forces us to upgrade the upgraded to a new server version on every new release. If that is a thing we want then it works but wanted that to be acknowledged as work done on every patch release.

rbankston (approved):
Left a comment about specifying a specific server version but LGTM once that item is thought about

migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml
danielhoherd: @rbankston the two ways we've done it in the past were hacks that don't work in all cases.

more details:
* Here's where we define the major/minor version: https://github.com/astronomer/astronomer/blob/7d6fdb3a80/migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml#L11
* Here's the hack that finds the latest minor version: https://github.com/astronomer/astronomer/blob/7d6fdb3a80/migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml#L245
* Here's the commit that removed the old way: https://github.com/astronomer/astronomer/commit/d6deb7adefbfbb092b7d9f0af31aff43bb36389f#diff-efbe5156f62457226feb2e032b086805cd647894d38bbe907945340361df5d52

danielhoherd: More details and examples of the problem https://github.com/astronomer/issues/issues/3444

migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml
```diff
@@ -285,7 +288,6 @@
     --reset-values
     -f {{ astro_save_dir.path }}/helm-user-values.json
     --set astronomer.houston.upgradeDeployments.enabled=true
-    --set astronomer.airflowChartVersion={{ upgrade_to_version_airflow }}
```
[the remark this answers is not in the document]
danielhoherd: Merged deletion of upgrade_to_version_airflow into this branch.

migrations/scripts/lts-to-lts/0.23-to-0.25/playbooks/tasks/upgrade.yaml
danielhoherd: Oh, it looks like they now support version constraints. This was changed 6 months ago. https://github.com/helm/helm/commit/171321bb6cb51198bfb63cee5f82cbfea029409a

danielhoherd: Fixed by passing the variable straight through now that helm 3.6 supports ranges and no longer requires an exact version match.

vishwas-astro: Do we still need the exact patch version 0.25.6 then?

PR #1220, opened 2021-09-16
add triggerer to fluentd

aliotta:
Adds triggerer logs to fluentd such that the logs are sent to kibana


andriisoldatenko (approved):

bote795 (approved):

vishwas-astro:
@aliotta @bote795 Do you want to do this conditionally based on the feature flag?

bote795:
I don't think it will make a difference if we add a flag or not functionaly

danielhoherd (approved):

PR #1223, opened 2021-09-20
Add kubed grafana dashboard

danielhoherd:
## Description

Add a kubed grafana dashboard. Something we should've made a long time ago, but hey, better late than never!

## Testing

So far I've only manually imported this config into staging and prod, haven't seen this work via helm install yet.


vishwas-astro (approved):

andriisoldatenko (approved):
LGTM if works.

can you please add screenshot how does it look like?

danielhoherd:
Closed because these [dashboards are added directly to ap-vendor](https://github.com/astronomer/ap-vendor/tree/794e954f0f489d4d99ffd4bb1b3c8dbb39296578/grafana/include).

PR #1230, opened 2021-09-30
Add support for specifying nodePorts when seviceType is set to NodePort

rob-1126:
## Description

Add support for specifying the default ports used by services on the default ingress service via nginx(http|https|metrics)NodePort. Only valid when nginx.seviceType is set to NodePort.

## PR Title

Add support for specifying httpNodePort, httpsNodePort, and metricsNodePort

## 🎟 Issue(s)

Resolves astronomer/issues#3538

## 🧪  Testing

No identified risks. Basic unit test added for conformity. 

## 📋 Checklist

- [X ] The PR title is informative to the user experience

### DIFF
--- a/charts/nginx/values.yaml
+++ b/charts/nginx/values.yaml
@@ -35,6 +35,11 @@ resources: {}
 # Service type that Nginx can be. Default is LoadBalancer but "ClusterIP" and "NodePort" are both valid.
 serviceType: "LoadBalancer"
 
+# NodePort, requiers serviceType NodePort. Default of nil auto-assigns a port
+httpNodePort: ~
+httpsNodePort: ~
+metricsNodePort: ~
+
 # dict of annotations to add to the ingress controller
 ingressAnnotations: {}
 
--- a/tests/test_nginx_service.py
+++ b/tests/test_nginx_service.py
@@ -50,6 +50,27 @@ class TestNginx:
         doc = docs[0]
         assert doc["spec"]["type"] == "NodePort"
 
+    def test_nginx_type_nodeport_specifying_nodeports(self):
+        # sourcery skip: extract-duplicate-method
+        httpNodePort, httpsNodePort, metricsNodePort = [ 30401, 30402, 30403 ]
+        docs = render_chart(
+            values={"nginx": {
+             "serviceType": "NodePort",
+              "httpNodePort": httpNodePort,
+              "httpsNodePort": httpsNodePort,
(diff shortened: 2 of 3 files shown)


charts/nginx/values.yaml
```diff
@@ -35,6 +35,11 @@ resources: {}
 # Service type that Nginx can be. Default is LoadBalancer but "ClusterIP" and "NodePort" are both valid.
 serviceType: "LoadBalancer"
 
+# NodePort, requiers serviceType NodePort. Default of nil auto-assigns a port
```
danielhoherd, line 38:
```suggestion
# NodePort, requires serviceType NodePort. Default of nil auto-assigns a port
```

tests/test_nginx_service.py
```diff
@@ -50,6 +50,29 @@ def test_nginx_type_nodeport(self):
         doc = docs[0]
         assert doc["spec"]["type"] == "NodePort"
 
+    def test_nginx_type_nodeport_specifying_nodeports(self):
```
andriisoldatenko, line 53: can you please add test when httpsNodePort is empty?

danielhoherd, line 53: We should also have a check that these fields do not show up with `serviceType: "LoadBalancer"`. We may also need to adjust `nginx/templates/nginx-service.yaml` with logic that only does LB or nodePort, if those are mutually exclusive.

danielhoherd (changes requested):
Looks like a good start. Need to address some mutual exclusivity questions and add a test for when service is set to LB or something other than nodeport.

tests/test_nginx_service.py
```diff
@@ -50,6 +50,41 @@ def test_nginx_type_nodeport(self):
...
+        )
+
+        assert len(docs) == 1
+        doc = docs[0]
+        ports = doc["spec"]["ports"]
+        assert not [x for x in ports if "nodePort" in x]
+
+    def test_nginx_type_nodeport_doesnt_require_nodeports(self):
+        # sourcery skip: extract-duplicate-method
+        httpNodePort, httpsNodePort, metricsNodePort = [30401, 30402, 30403]
```
danielhoherd:
```suggestion
```

rob-1126:
@danielhoherd I think this should be ready for your review/approval. I had merged in your suggestion manually without using the github built-in accept suggestion functionality (it's pretty slick, I need to wise up).  @andriisoldatenko I added the tests you requested and have re-requested your review as well.

PR #1233, opened 2021-10-05
update image tags for release 0.26.0

aliotta:


vishwas-astro (approved):

danielhoherd:
We have a bunch of other components coming in that are built off of ap-base:3.14.2-1 so we should include those with this.

- <https://github.com/astronomer/houston-api/pull/788>
- <https://github.com/astronomer/commander/pull/65>
- <https://github.com/astronomer/ap-vendor/pull/161>
- <https://github.com/astronomer/db-bootstrapper/pull/42>
- <https://github.com/astronomer/astro-ui/pull/1205>

danielhoherd:
There is an LTS test that is failing. Is 0.26.0 an LTS? @vishwas-astro

PR #1242, opened 2021-10-12
protect internal houston alerts/registry events

andriisoldatenko:
https://github.com/astronomer/issues/issues/3579#issuecomment-939056273

### DIFF
--- a/charts/astronomer/templates/houston/ingress.yaml
+++ b/charts/astronomer/templates/houston/ingress.yaml
@@ -26,6 +26,7 @@ metadata:
         deny all;
         return 403;
       }
+    {{- end }}
 spec:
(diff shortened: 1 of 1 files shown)


schnie:
Looks good. I saw @pgvishnuram mention the auth sidecar consideration, which makes total sense. Are we doing this in addition to the pre-shared key on the registry? The registry is the main concern, so having an extra layer there will be nice, especially if a customer is using their own ingress. The alerts and elasticsearch endpoints aren't as much of a concern, given the scope of what an adversary could do. Additional layers there can come later on.

andriisoldatenko:
@danielhoherd / @pgvishnuram  PTAL

PR #1254, opened 2021-10-21
Make alertmanager HA

danielhoherd:
I tested this in QA by manually increasing replicas to 2 and checking prometheus.basedomain/status and it showed both alertmanagers as being available, so I think this is the only change we need to do in order to make alertmanager HA.

Related to: https://github.com/astronomer/issues/issues/949


danielhoherd:
I think a good way to move forward on this is to write a functional test for the success condition and make sure it passes.

PR #1266, opened 2021-11-04
bumping UI and houston for 0.27 RC1

rdodev:
## Description

Bumps image for UI and houston for RC1 of 0.27

## 📋 Checklist

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -24,11 +24,11 @@ images:
     # httpSecret: ~
   houston:
     repository: quay.io/astronomer/ap-houston-api
-    tag: 0.26.9
+    tag: release-0.27
     pullPolicy: IfNotPresent
   astroUI:
     repository: quay.io/astronomer/ap-astro-ui
-    tag: 0.26.7
+    tag: release-0.27
     pullPolicy: IfNotPresent
   dbBootstrapper:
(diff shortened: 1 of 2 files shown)


charts/astronomer/values.yaml
```diff
@@ -24,11 +24,11 @@ images:
     # httpSecret: ~
   houston:
     repository: quay.io/astronomer/ap-houston-api
-    tag: 0.26.9
+    tag: release-0.27
```
andriisoldatenko, line 27: 0.27.0?

andriisoldatenko (changes requested):

charts/astronomer/values.yaml
rdodev: As per CI logs `docker push astronomerinc/ap-astro-ui:release-0.27`

vishwas-astro:
@rdodev - you need to change the platform version here as well: https://github.com/astronomer/astronomer/blob/release-0.27/Chart.yaml#L3-L4

charts/astronomer/values.yaml
vishwas-astro: both 0.27.0 and release-0.27 exist in https://quay.io/repository/astronomer/ap-houston-api?tab=tags.

0.27.0 is explicitly this first version on the release branch and release-0.27 will be the tag published whenever the next houston build on release-0.27 branch happens.

PR #1270, opened 2021-11-05
Add JSON schema file

thmzlt:
Resolves astronomer/issues#3715. See https://github.com/astronomer/airflow-chart/pull/260#issuecomment-961965285 for context.

### DIFF
--- a/values.schema.json
+++ b/values.schema.json
@@ -1,0 +1,102 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "properties": {
+    "astronomer": {
+      "type": "object",
+      "properties": {
+        "houston": {
+          "type": "object",
+          "properties": {
+            "config": {
+              "type": "object",
+              "properties": {
+                "deployments": {
+                  "type": "object",
+                  "properties": {
+                    "helm": {
+                      "type": "object",
+                      "additionalProperties": false,
+                      "properties": {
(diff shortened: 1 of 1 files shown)


aliotta (approved):

values.schema.json
```diff
@@ -0,0 +1,102 @@
...
+                          "type": "boolean"
+                        },
+                        "workers": {
+                          "type": "object",
+                          "additionalProperties": false,
+                          "properties": {
+                            "affinity": {
+                              "type": "object"
+                            },
+                            "autoscaling": {
```
jedcunningham, line 48: `autoscaling` is the only thing that should be here. The others, if they exist, _should_ be an error as they should be under `airflow.workers` instead.

https://github.com/astronomer/airflow-chart/blob/d0c176ddaf42e1bd53e48c1c063057394aee464d/values.yaml#L340-L347

jedcunningham (changes requested):

values.schema.json
thmzlt: I searched for all instances of `.Values.workers` and added as properties. But I just realized that the others (that aren't autoscaling) appear in the podTemplate manifest which I assume takes its values from `airflow.workers`.

Thanks, Jed.

jedcunningham (approved):

PR #1273, opened 2021-11-09
Change cleanup cronjob minute to not coincide with */15 schedule

danielhoherd:
As a stopgap to https://github.com/astronomer/issues/issues/3741, this change alters the cronjob offset to not coincide with `*/15` which is commonly used as a dag schedule. This avoids running cleanup jobs on the same minute that dag pods are launching, reducing concurrent pod spikes by spreading workloads across time.


pgvishnuram:
@danielhoherd - are we going to include this in 0.27.1

vishwas-astro:
@pgvishnuram - Let's make this part of our Jan release (possibly 1.0)

danielhoherd:
This is an incredibly minor change. I actively make changes like this in prod on a regular basis to keep jobs from stacking up. We can include it at any time and it would be totally safe.

danielhoherd:
I've refactored this to use a hack-ish method for generating deterministic random-ish numbers from the release name, which is more towards the original goal of the ticket. I've also added tests.

bote795 (approved):

kaxil (approved):

PR #1274, opened 2021-11-09
add airflow to json schema

aliotta:
issue link: https://app.zenhub.com/workspaces/astro-platform-board-602c68a458670000110dd641/issues/astronomer/issues/3715

Description: This PR is meant to address the fact that currently we allow our users to specify defaults for the airflow chart at the astronomer/astronomer level. The defaults we let them apply currently are not verified against the airflow chart json schema which can lead to situations in which our users think their astronomer upgrade worked but in reality the platform deployment was a false positive. This can be as subtle as a specific feature they think is still working is no longer working to as verbose as failing airflow deployments when they create new ones or upgrade exisiting.

Solution: We have decided to pull in a tagged version of the json schema from the OSS side. We have also decided to add some more checks on the top level airflow params that currently the OSS chart does not specify.

Update notes: Long term the fact that the top level params are not specified in the chart is a defect known on the OSS side they are planning on fixing in a future release. When this change goes out we can drop those clauses from the astronomer/astronomer side. 

Additionally we will need to be diligent about updating the url ref everytime we modify the airflow chart version running forward.


danielhoherd:
@aliotta can you add some context around this change? What ticket is this addressing? How was this data generated? How will it be updated?

thmzlt:
> How was this data generated? How will it be updated?

Echoing these questions, and also pointing out that it should be run through a formatter (such as [prettier](https://prettier.io/)). A chunk of this diff is whitespace being removed.

values.schema.json
```diff
@@ -18,7 +18,143 @@
                       "additionalProperties": false,
                       "properties": {
                         "airflow": {
-                          "type": "object"
+                          "allOf": [
+                            {"$ref": "https://raw.githubusercontent.com/apache/airflow/main/chart/values.schema.json"},
```
aliotta, line 22: @jedcunningham I dislike that I'm pointing to main here as that changes frequently and I'd like to lock this to a specific releases json schema

aliotta: Can we ref a tag on a release branch instead?

aliotta: Would this link be the one to use? https://raw.githubusercontent.com/apache/airflow/2.2.1/chart/values.schema.json

Whats our risk of someone on OSS side changing the tag commit.

values.schema.json
```diff
@@ -18,7 +18,143 @@
...
+                            {
+                              "additionalProperties": false,
+                              "properties": {
+                                "affinity": {},
```
aliotta, line 26: these were pulled using the following command line script

`ag ".Values.\w+" -o --no-filename --no-numbers | sort | uniq `

aliotta:
> @aliotta can you add some context around this change? What ticket is this addressing? How was this data generated? How will it be updated?

Context added let me know what you think.

aliotta:
> > How was this data generated? How will it be updated?
> 
> Echoing these questions, and also pointing out that it should be run through a formatter (such as [prettier](https://prettier.io/)). A chunk of this diff is whitespace being removed.

whitespace changes have been removed

values.schema.json
jedcunningham: This is the one you want:
https://github.com/apache/airflow/blob/helm-chart/1.3.0/chart/values.schema.json

jedcunningham: The tag shouldn't change.

thmzlt:
@aliotta not sure if JSON Schema supports comments, but we might want to add the comment used to get that list of properties:

`ag "\.Values\.\w+" -o --nofilename --nocolor | sort | uniq`

thmzlt (approved):
:shipit:

aliotta:
> @aliotta not sure if JSON Schema supports comments, but we might want to add the comment used to get that list of properties:
> 
> `ag "\.Values\.\w+" -o --nofilename --nocolor | sort | uniq`

Added upgrade instructions into our readme.md

values.schema.json
jedcunningham: I think this is our only addition, and it should probably by allowed here as well:

https://github.com/astronomer/airflow-chart/blob/1909e227e1fd8539febc000e763ba63c20c361f6/values.yaml#L334-L335

aliotta: added as well as podMutation as discussed in PM

jedcunningham: Probably worth documenting these 2 explicitly, or how to find the delta.

README.md
```diff
@@ -128,7 +128,7 @@ gch tags/helm-chart/1.2.0
 ag "\.Values\.\w+" -o --no-filename --no-numbers | sort | uniq
 ```
 
-The values output by this command will need to be inserted manually into astronomer/values.schema.json at the astronomer.houston.config.deployments.helm.airflow.allOf parameter.
+The values output by this command will need to be inserted manually into astronomer/values.schema.json at the astronomer.houston.config.deployments.helm.airflow.allOf parameter. There are two additional params that need to be at this location outisde of what is returned from above. They are podMutation and useAstroSecurityManager. These can be found by running the same ag command against the astronomer/airflow-chart values.yaml file.
```
jedcunningham, line 131:
```suggestion
The values output by this command will need to be inserted manually into astronomer/values.schema.json at the `astronomer.houston.config.deployments.helm.airflow.allOf` parameter. There are two additional params that need to be at this location outside of what is returned from above. They are `podMutation` and `useAstroSecurityManager`. These can be found by running the same ag command against the astronomer/airflow-chart values.yaml file.
```

jedcunningham (approved):

PR #1276, opened 2021-11-12
add new field for settign sslmode

aliotta:
Currently the OSS chart only allows you to specify an sslmode on the resultBackendConnection and metadataConnection if you define the whole object. We need this param to allow our users to overide just this field and use what we specify for the other values.

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -68,6 +68,9 @@ data:
 
       # These values get passed directly into the airflow helm deployments
       helm:
+      {{- if and .Values.global.ssl.enabled .Values.global.ssl.mode }}
+        sslmode: {{ .Values.global.ssl.mode }}
+      {{- end }}
       {{- if .Values.global.sccEnabled }}
         # If security context constraints are enabled, enable SCCs for airflow-chart
         sccEnabled: true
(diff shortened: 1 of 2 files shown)


pgvishnuram (commented):
we need to add in houston-configmap to pick from 
https://github.com/astronomer/astronomer/blob/master/values.yaml#L182-L184 .

so one place to change everything - let me know your views

bote795 (approved):

aliotta:
> can we add in houston-configmap to pick from https://github.com/astronomer/astronomer/blob/master/values.yaml#L182-L184 .
> 
> so one place to change everything - let me know your views

makes sense to me. Added let me know what you think

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -68,6 +68,9 @@ data:
 
       # These values get passed directly into the airflow helm deployments
       helm:
+      {{- if and .Values.global.ssl.enabled .Values.global.ssl.mode }}
+        sslmode: {{ .Values.global.ssl.mode }}
```
andriisoldatenko, line 72: are you sure it's correct level? IMO it's wrong one.

andriisoldatenko (changes requested):

andriisoldatenko (approved):

PR #1279, opened 2021-11-16
bump to json version to v2

aliotta:
There are backwardly incompatible changes to the new desired update json. We are bumping the version so as not to break enterprise customers older code.

Related houston PR: https://github.com/astronomer/houston-api/pull/830


aliotta:
We decided to go another route on this one

PR #1296, opened 2021-12-07
authsidecar config missing in houston config has been re-added

pgvishnuram:
## Description

This PR fixes missing config's to enable authsidecar for airflow deployments 

## PR Title

Fix missing configs of authsidecar

## 🎟 Issue(s)

https://github.com/astronomer/issues/issues/3853


aliotta (approved):

danielhoherd (changes requested):
Looks good but needs tests.

pgvishnuram:
yes sure - i will need to add entire feature test - previously i have not added them

andriisoldatenko:
LGTM, could you pls fix CI platform-1-21-2 ?

tests/test_authsidecar.py
```diff
@@ -0,0 +1,169 @@
...
+                "charts/alertmanager/templates/alertmanager-statefulset.yaml",
+                "charts/alertmanager/templates/alertmanager-auth-sidecar-configmap.yaml",
+                "charts/alertmanager/templates/alertmanager-service.yaml",
+                "charts/alertmanager/templates/alertmanager-networkpolicy.yaml",
+                "charts/alertmanager/templates/ingress.yaml",
+            ],
+        )
+
+        assert len(docs) == 5
+        doc = docs[0]
```
danielhoherd, line 28: Why are we rendering 5 templates but only making assertions against 1?

tests/test_authsidecar.py
```diff
@@ -0,0 +1,169 @@
...
+        assert doc["kind"] == "StatefulSet"
+        assert doc["apiVersion"] == "apps/v1"
+        assert doc["metadata"]["name"] == "RELEASE-NAME-alertmanager"
+        assert "auth-proxy" == doc["spec"]["template"]["spec"]["containers"][1]["name"]
```
danielhoherd, line 32:
```suggestion
        assert doc["spec"]["template"]["spec"]["containers"][1]["name"] == "auth-proxy"
```

tests/test_authsidecar.py
```diff
@@ -0,0 +1,169 @@
...
+
+        assert "Service" == jmespath.search("kind", docs[2])
```
danielhoherd, line 34:
```suggestion
        assert jmespath.search("kind", docs[2]) == "Service"
```

tests/test_authsidecar.py
```diff
@@ -0,0 +1,169 @@
...
+        assert "RELEASE-NAME-alertmanager" == jmespath.search("metadata.name", docs[2])
```
danielhoherd, line 35:
```suggestion
        assert jmespath.search("metadata.name", docs[2]) == "RELEASE-NAME-alertmanager"
```

tests/test_authsidecar.py
```diff
@@ -0,0 +1,169 @@
...
+        assert "ClusterIP" == jmespath.search("spec.type", docs[2])
```
danielhoherd, line 36:
```suggestion
        assert jmespath.search("spec.type", docs[2]) == "ClusterIP"
```

tests/test_authsidecar.py
```diff
@@ -0,0 +1,169 @@
...
+                "charts/prometheus/templates/prometheus-statefulset.yaml",
+                "charts/prometheus/templates/prometheus-auth-sidecar-configmap.yaml",
+                "charts/prometheus/templates/prometheus-service.yaml",
+                "charts/prometheus/templates/prometheus-networkpolicy.yaml",
+                "charts/prometheus/templates/ingress.yaml",
+            ],
+        )
+
+        assert len(docs) == 5
+        doc = docs[0]
```
danielhoherd, line 58: Again, is there a reason we're rendering 5 templates and only asserting against 1? I feel like we should be making assertions about all of them if we're going to render them all.

danielhoherd (changes requested):
Looks generally good. I have a few suggestions on tests, but looking back at them I think they may just be [bike-shedding](https://en.wiktionary.org/wiki/bikeshedding). More importantly is a question about why we are rendering multiple templates but only making assertions against one. It would be great to make assertions against all relevant templates, and remove any templates that are not relevant to the test case.

tests/test_authsidecar.py
pgvishnuram: I am asserting statefulset and service

pgvishnuram: I still need to add more test cases to test against all the mentioned templates

vishwas-astro:
>  [bike-shedding](https://en.wiktionary.org/wiki/bikeshedding). 
so true in a lot of code reviews I have seen and possibly done myself :)

danielhoherd (approved):

PR #1295, opened 2021-12-07
Backport pytest to release-0.23

danielhoherd:
## Description

This change set backports our pytest chart tests to release 0.23. There are a few instances where we have had to make modifications to 0.23, and there are a few more in the pipeline, and having these tests in place will help us gain confidence in changes we make to this LTS release.

Having these tests in place is prerequisite for some customer upgrade work.

This change set is pretty safe. The only customer facing change is tweaking the rbac template. Everything else is just CI tweaks and test additions, and those additions mostly put us in a better testing stance going forward. Also, all of these pytest additions are taken from master branch, so they have already been reviewed.

## Itemized changes

- Backport all relevant pytest chart tests from master branch (some tests were dropped because they were for features that are newer than 0.23)
- Drops k8s 1.16 testing in CI because 1.16 is deprecated in all supported cloud environments (not possible to run 1.16 in any public cloud env)
- Bump ap-build image
- Bump ubuntu machine image and its python version used in testing
- Backport some CI changes from master branch: Makefile, build-helm-chart.sh, pre-commit
- Drop `apiVersion.rbac.v1beta2` template in favor of `apiVersion.rbac` which handles both v1 and v1beta1. There are some hard-coded apiVersion.rbac.v1 strings that should be fixed in newer versions. https://github.com/astronomer/issues/issues/3862

## Related to

- <https://github.com/astronomer/issues/issues/3855>
- <https://github.com/astronomer/issues/issues/3860>

PR #1299, opened 2021-12-07
Add flag to enable legacy Ingress API, support 1.19+ ingress

park-peter:
This PR includes changes to support 1.19+ ingress (networking.k8s.io/v1beta1) to 0.23, but more importantly, add a flag `useLegacyIngress` to allow customers who have updated their K8s cluster before upgrading Astronomer to 0.25. Users on later versions of K8s cannot upgrade to newer versions of 0.23 due to the ingress API differences. This PR includes similar code changes as PR #1122.

- Add `useLegacyIngress` flag, defaulting to false
- Handle Ingress v1 differences from v1beta1, retroactive to 1.19
- Drop some <=1.14 specific configs since we no longer support those versions

## Relevant links

- PR #1295 is a prerequisite
- Issue ticket to track this issue: https://github.com/astronomer/issues/issues/3860


.circleci/config.yml.j2
```diff
@@ -98,7 +98,7 @@ jobs:
           astronomer-tags: "platform postgresql monitoring logging kubed keda"
 {% endfor -%}
 
-{% for version in [kube_versions[1]] %}
+{% for version in [kube_versions[2]] %}
```
danielhoherd, line 101: This needs to be reverted before merging.

rbankston (approved):
Looks good to me

tests/test_houston_ingress.py
```diff
@@ -4,12 +4,44 @@
...
+        )
+    ]
+    assert "houston-http" in [
+        port[0]
+        for port in jmespath.search(
+            "spec.rules[*].http.paths[*].backend.servicePort", doc
+        )
+    ]
+
+
```
danielhoherd, line 38: @andriisoldatenko what do you think about this syntax? I'm not 100% sure that making functions like `ingress_assertions_v1` is a good idea. WDYT?

tests/test_ingress.py
```diff
@@ -4,12 +4,44 @@
...
+    ]
+    assert "astro-ui-http" in [
+        port[0]
+        for port in jmespath.search(
+            "spec.rules[*].http.paths[*].backend.service.port.name", doc
+        )
+    ]
+
+
+def ingress_assertions_v1beta1(doc):
```
thmzlt, line 23: I'm new to the test codebase so not sure if this is following a pattern, but this would read better to me if it was something more like `assert_ingress_v1beta1` (verb vs. noun).

thmzlt (approved):
Left a minor style comment. Looks good otherwise.

tests/test_ingress.py
pgvishnuram: should be good "test_ingress_assertions_v1"

tests/test_houston_ingress.py
pgvishnuram: in all places you have prefixed function with test_ , can we use the same here too

danielhoherd: If I prefix these functions with `test_` then pytest runs them as if they are individual test cases, and they are not.

tests/test_ingress.py
danielhoherd: These are helper functions, not individual test cases. We're not only asserting that it's v1beta1, but also that the data structure is the needed structure for the given version.

PR #1302, opened 2021-12-08
Turn registry.notifications.timeout from static value into a var

syamasakigoodrx:
## Description

add interpolation for notifications timeout in the astronomer-registry configmap, currently timeout is hardcoded

## 🎟 Issue(s)

timeout is hardcoded and cannot be overwritten from the values file

## 🧪  Testing

verified helm upgrade could successfully update configmap timeout using the values file


andriisoldatenko (approved):

andriisoldatenko:
@pgvishnuram if possible pls add test.

danielhoherd:
@syamasakigoodrx if you need help adding tests, here's a tutorial: https://github.com/astronomer/astronomer/blob/master/tests/README.md

danielhoherd (changes requested):
Needs tests.

@syamasakigoodrx if you need help adding tests, here's a tutorial: https://github.com/astronomer/astronomer/blob/master/tests/README.md

syamasakigoodrx:
test added to verify yaml is valid after making timeout variable configurable

tests/test_astro_yaml.py
```diff
@@ -0,0 +1,9 @@
+import pytest
```
danielhoherd, line 1: The filename `test_astro_yaml.py` is ambiguous.

tests/test_astro_yaml.py
```diff
@@ -0,0 +1,9 @@
...
+import os
+import yaml
+
+def test_astro_yaml():
```
danielhoherd, line 5: - The file name `test_astro_yaml.py` is ambiguous. You should probably add a test case to [test_astronomer_registry.py](https://github.com/astronomer/astronomer/blob/966fa36935/tests/test_astronomer_registry.py)
- The function name `test_astro_yaml` is ambiguous.
- This test should be parametrized to support all k8s versions.
- This test doesn't use the existing [`render_chart()`](https://github.com/astronomer/astronomer/blob/966fa36935/tests/helm_template_generator.py#L67) function that is used in all our helm tests.
- This test only checks for valid yaml, not a valid k8s manifest.
- This test doesn't check for presence or validity of the configured value.

Here is a tutorial that explains how these tests should be written: https://github.com/astronomer/astronomer/blob/master/tests/README.md#starting-a-new-pytest-file

See this example for guidance on refactoring this test to include the `render_chart()` function with `values` and `show_only` specified, and assertions made against the values that were given: https://github.com/astronomer/astronomer/blob/966fa36935/tests/test_registry_statefulset.py#L111-L120

One thing I find helpful when writing tests is to put a `breakpoint()` into the new test below `render_chart()` and run `venv/bin/python -m pytest tests --pdb -k registry` to get an interactive shell where I can craft assertions for the test against the `docs`.

I appreciate you taking the effort to add this feature and tests for it. We can't merge this though until test cases that cover the new feature.

danielhoherd (changes requested):
I added some feedback that needs to be addressed before we can merge. The tests are a good start but need to be refactored to be more standard for our test setup and to actually test the new feature, not just test for valid yaml.

tests/test_astronomer_registry_configmap.py
```diff
@@ -0,0 +1,25 @@
+from tests.helm_template_generator import render_chart
+from . import supported_k8s_versions
+import yaml
+import re
```
danielhoherd, line 4: @syamasakigoodrx Great work refactoring the tests! I think we just need one more thing:
```suggestion
import re
import pytest
```

danielhoherd (approved):
Looks good! Thanks for putting in the effort to implement this and add tests @syamasakigoodrx !

syamasakigoodrx:
@danielhoherd we currently run astronomer/astronomer helm deployments pinned to --version 0.25.4 (all in-house jobs etc have been verified against that version).  Would you recommend i open  a pr against that "tag" to make a "0.25.4-beta" version to avoid adding additional 0.27.0 update/commits?  Otherwise i can tell our engineering team they need to upgrade to the latest and do a full env verification...

danielhoherd:
@syamasakigoodrx personally i would recommend trying to keep up with the latest versions, so upgrading from 0.25 to 0.27 would be my recommendation. Also FYI 0.25.4 is not the latest version of 0.25. 

I'm not sure when we will be releasing a new version of 0.25. I can merge these changes into that branch right now though, so that the next version we release will get it. The next version we release of 0.27 will have these changes.

syamasakigoodrx:
cool.  I am still interested in a 0.25.x release if you do.  i tested updating our systems to 0.27.x looks like due to a constraint where we are pinning an older statsd (ie. with full metrics with wildcards vs the newer truncated ones) i was able to update with 25.x but not 26.x onward.  `deployments:
        helm:
          images:
            statsd:
              repo: quay.io/astronomer/ap-statsd-exporter
              tag: 0.11.0`

danielhoherd:
@syamasakigoodrx we will likely be releasing a new 0.25.x version soon with some security fixes and I'll make sure this change will go into that. I'm not sure what the timeline on that will be, but keep an eye out for it.

danielhoherd:
@syamasakigoodrx FYI this went out with 0.25.13

PR #1306, opened 2021-12-09
Bump ap-houston to 0.27.7 and ui to 0.27.5 to fix Astro-ui bug that still shows kabana on dropdown when customLogging enabled

bote795:
## Description

- fixes issue in astro-ui that stil shows kibana on drop down even when user has customLogging turned on

## 🎟 Issue(s)

Related astronomer/issues#3680

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -24,11 +24,11 @@ images:
     # httpSecret: ~
   houston:
     repository: quay.io/astronomer/ap-houston-api
-    tag: 0.27.5
+    tag: 0.27.7
     pullPolicy: IfNotPresent
   astroUI:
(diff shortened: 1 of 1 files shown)


danielhoherd (changes requested):
Please add information to the PR description. We see that versions are being bumped, but why? What do the new versions do? Links to the issues that required new versions of astro-ui and houston-api would be great.

pgvishnuram (approved):

bote795:
> Please add information to the PR description. We see that versions are being bumped, but why? What do the new versions do? Links to the issues that required new versions of astro-ui and houston-api would be great.

updated

danielhoherd (approved):

PR #1307, opened 2021-12-09
Vastly simplify the gh issue template [ci skip]

danielhoherd:
Because people rarely fill out any details in the PR template, often just leaving the boilerplate in place which just adds noise to the PR, I thought we could try vastly simplifying it. The changes here include text that indicates the PR should not be merged, leading people who want their code to be merged to actually replace that text. I'm not sure if this is too passive aggressive or if it will work. What do you all think?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,44 +1,11 @@
-<!--
-Please fill out the sections below, deleting anything that is irrelevant or empty.
--->
-
-
 ## Description
 
-<!--
-Describe the purpose of this pull request.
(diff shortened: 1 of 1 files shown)


vishwas-astro (approved):

kaxil (approved):

pgvishnuram (approved):

bote795 (approved):

krisdock (approved):
all about this

bote795:
we could also create a linter... but that does require more work

Issue #1308, opened 2021-12-09
Grafana configurations are not easily customizable


dintorf:
Currently I see no way to customize the `grafana.ini` in the helm chart. This makes it difficult to configure any settings for the grafana deployment (i.e. smtp settings for alerting, see [Ticket #7151](https://astronomer.zendesk.com/agent/tickets/7151)). 

A possible solution is to allow extra environment variables to be passed to the grafana configuration in the Astronomer helm chart values. For example:

```yaml
grafana:
  extraEnvVars: 
    - name: GF_SMTP_ENABLED
       value: true
    - name: GF_SMTP_HOST
       value: smtp.astronomer.io
    - name: GF_SMTP_USER
       value: grafana
    - name: GF_SMTP_PASSWORD
       valueFrom:
          secretKeyRef:
            name: grafana-smtp-secret
            key: smtp_password
```

This would allow you to populate those env vars into the deployment template and modify configurations for Grafana. I'd be willing to contribute this change if this is accepted.

danielhoherd:
This is probably a fairly simple change, with something like `{{ if .Values.grafana.extraEnvVars}}{{ .Values.grafana.extraEnVars | nindent 8 }}{{ end }}` or some such. Then of course [pytest chart tests](https://github.com/astronomer/astronomer/blob/master/tests/README.md) to make sure the added helm feature is generating valid k8s manifests.

dintorf:
That's what I was planning on adding, should I add this for any other charts while I'm at it?

danielhoherd:
@dintorf if you want to go that extra mile it would be useful :)

vishwas-astro:
@dintorf could you move this issue over to astronomer/issues repo please? That way it will be verified along with the rest of the issues in a release by QA.

dintorf:
sure thing @vishwas-astro 👍

PR #1312, opened 2021-12-13
1308-templating-grafana-extra-env-vars

dintorf:
## Description

Add extra environment variables to the `grafana-deployment` resource in order to set configuration settings in Grafana.

## Related Issues

- [#1308](https://github.com/astronomer/astronomer/issues/1308)

## Testing

Verified that extra environment variables are rendered when provided as a list to the `grafana` section of helm values, and no side effect occurs when not provided.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ server.key
 upgrade_test_config.yaml
 .unittest-requirements
 prof
+.python-version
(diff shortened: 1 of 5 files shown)


danielhoherd (approved):
Very nice, thanks for jumpstarting out Grafana tests @dintorf!

PR #1313, opened 2021-12-13
3870 migrate remaining helm unittests to pytest (part 1)

danielhoherd:
## Description

Finish migrating helm unittest tests to pytest chart tests (part 1). Originally I was trying to do the whole migration at once, but that PR would be HUGE and so I'm breaking it up. Future PRs for this will be smaller.

## Itemized changes

- Bump some CI images
- Add PodDisruptionBudget template and use it everywhere
- Migrate all remaining global helm-unittest tests (all `tests/*_test.yaml` files)
- no-op tweak to helm_template_generator.py (return immediately vs assign and return)
- Move to comprehensions from jmespath in a lot of places due to a variety of things. jmespath has nuanced quote behavior that is easy to get wrong. In many places jmespath wasn't even really helpful.
- Do straight object comparisons vs iterating sub-comparisons in a lot of places. This is a more complete test, though in some places it's a bit verbose. `json` was used in one test to get around having 100 extra lines of just test data structures.
- Add some additional tests that were not in helm-unittest.
- Add `astronomer.podLabels` comment where it was missing in `values.yaml`

## Related Issues

- https://github.com/astronomer/issues/issues/3870

## Testing

These are tests, and they all should pass. Helm unittest was so unreliable that it's unlikely we'll end up in a worse place by removing them.


charts/astronomer/values.yaml
```diff
@@ -227,7 +227,7 @@ houston:
     #   - 2.0.0-alpine3.10-onbuild
     #   channel: stable
 prisma:
-  podLabels: {}
+  podLabels: {} # This almost certainly does nothing. The right value is astronomer.podLabels
```
danielhoherd, line 230: @andriisoldatenko do you know anything about this?

charts/astronomer/values.yaml
```diff
@@ -259,7 +259,7 @@ registry:
   # k8s secret name for registry to use to authenticate to houston when sending notifications
   authHeaderSecretName: ~
   replicas: 1
-  podLabels: {}
+  podLabels: {} # This almost certainly does nothing. The right value is astronomer.podLabels
```
danielhoherd, line 262: @andriisoldatenko do you know anything about this? (2/2)

tests/test_ingress.py
```diff
@@ -20,38 +20,37 @@ def test_basic_ingress(self, kube_version):
...
-                )
-            ]
-            assert "astro-ui-http" in [
-                port[0]
-                for port in jmespath.search(
-                    "spec.rules[*].http.paths[*].backend.service.port.name", doc
-                )
-            ]
+        # This would be valid python, but we laod from json just to keep linters happy and the data more compact
+        expected_rules_v1beta1 = json.loads(
```
danielhoherd, line 32: @andriisoldatenko what do you think about this? There's a large data structure, and `black` wanted to break it up, and `flake8` was complaining about it being so long and stuff, so I am loading it from `json`. Seem OK?

charts/astronomer/values.yaml
andriisoldatenko: oh interesting your know what we are not using prisma pod anymore :) I assume we can get rid of it.

charts/astronomer/values.yaml
andriisoldatenko:
```
./charts/astronomer/templates/registry/registry-statefulset.yaml
37:{{- with .Values.podLabels }}
```
?

danielhoherd: @andriisoldatenko this is `astronomer.commander.podLabels`

tests/test_ingress.py
andriisoldatenko: yeah json looks okey for expected results, easy to read than big data structures in python.

charts/astronomer/values.yaml
danielhoherd: See https://github.com/astronomer/astronomer/blob/18770565233bac830319653df524b3e706ddbc67/tests/test_registry_statefulset.py#L111-L120 for `astronomer.podLabels` test in registry sts

charts/astronomer/values.yaml
danielhoherd: Oh, so we can get rid of this whole prisma section then! Even better!

tests/test_poddisruptionbudget.py
```diff
@@ -0,0 +1,53 @@
...
+        values={
+            "global": {
+                "prometheusPostgresExporterEnabled": True,
+                "useLegacyPodDisruptionBudget": True,
+            }
+        },
+    )
+    _, minor, _ = (int(x) for x in kube_version.split("."))
+
+    if minor < 25:
```
andriisoldatenko, line 50:
```suggestion
    K8S_MINOR_VERSION = 25
    if minor < K8S_MINOR_VERSION:
```

tests/test_poddisruptionbudget.py
```diff
@@ -0,0 +1,53 @@
...
+        kube_version=kube_version,
+        show_only=show_only,
+        values={
+            "global": {
+                "prometheusPostgresExporterEnabled": True,
+                "useLegacyPodDisruptionBudget": True,
+            }
+        },
+    )
+    _, minor, _ = (int(x) for x in kube_version.split("."))
```
andriisoldatenko, line 48: if you plan to reuse it you can try to use python package for semver or extract it to test utils.

tests/test_privateca.py
```diff
@@ -96,18 +96,20 @@ def test_privateca_daemonset_enabled_with_custom_image(self, kube_version):
...
+            == "snarks:boojums"
         )
 
 
 @pytest.mark.parametrize(
     "kube_version",
     supported_k8s_versions,
 )
-class TestPSP:
+class TestPrivateCaPsp:
```
andriisoldatenko, line 110:
```suggestion
class TestPrivateCApsp:
```

danielhoherd: Wouldn't `test private CA psp` be TestPrivateCaPsp? That's what https://textedit.tools/camelcase shows.

tests/test_poddisruptionbudget.py
danielhoherd: Yeah, I thought about that. I think we can put that into a future iteration. This exact line was copied from elsewhere. There is definitely room for improvement though.

tests/test_poddisruptionbudget.py
danielhoherd: Maybe we should rework the fixture to provide an object with more detailed info... Something to think about for a future iteration.

tests/test_privateca.py
andriisoldatenko: Typically you don't mix camelCase in abbreviations: good example

HTTPHeader instead of HttpHeader.


More examples https://stackoverflow.com/questions/15526107/acronyms-in-camelcase

It's not blocking review just FYI.

andriisoldatenko (approved):

PR #1314, opened 2021-12-14
Fix nats, stan and prometheus image names for environments using private registry

park-peter:
## Description

NATS, STAN, and Prometheus have wrong image names when `privateRegistry` is enabled. Normally, customers using private registry will mirror our public names. Having different names for only few of the images causes confusion and should be fixed.

## Related Issues

https://github.com/astronomer/issues/issues/3893


park-peter:
Closing and rebasing it to master instead.

PR #1315, opened 2021-12-14
Fix nats, stan, and prometheus image names for private registry

park-peter:
## Description

NATS, STAN, and Prometheus have wrong image names when `privateRegistry` is enabled. Normally, customers using private registry will mirror our public names. Having different names for only few of the images causes confusion and should be fixed. Should also be merged to 0.26, 0.25 and 0.23 releases.

## Related Issues

https://github.com/astronomer/issues/issues/3893

## Testing
Manual testing has been done to show that the configs do work in run time. It uses the same image names as the public images, should not need additional testing.

### DIFF
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 image:
-  repository: quay.io/prometheus/node-exporter
+  repository: quay.io/astronomer/node-exporter
   tag: v1.3.0
   pullPolicy: IfNotPresent
 
--- a/charts/prometheus-node-exporter/values.yaml
+++ b/charts/prometheus-node-exporter/values.yaml
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 image:
-  repository: quay.io/astronomer/node-exporter
+  repository: quay.io/prometheus/node-exporter
   tag: v1.3.0
   pullPolicy: IfNotPresent
 
--- a/charts/stan/templates/_helpers.tpl
+++ b/charts/stan/templates/_helpers.tpl
@@ -23,6 +23,14 @@ Return the list of peers in a NATS Streaming cluster.
 {{ print $replicas }}
 {{- end -}}
 
+{{ define "stan-exporter.image" -}}
+{{- if .Values.global.privateRegistry.enabled -}}
+{{ .Values.global.privateRegistry.repository }}/ap-nats-exporter:{{ .Values.images.exporter.tag }}
+{{- else -}}
+{{ .Values.images.exporter.repository }}:{{ .Values.images.exporter.tag }}
+{{- end }}
+{{- end }}
+
 {{ define "stan.image" -}}
 {{- if .Values.global.privateRegistry.enabled -}}
 {{ .Values.global.privateRegistry.repository }}/ap-nats-streaming:{{ .Values.images.stan.tag }}
--- a/charts/stan/templates/_helpers.tpl
+++ b/charts/stan/templates/_helpers.tpl
@@ -25,7 +25,7 @@ Return the list of peers in a NATS Streaming cluster.
 
 {{ define "stan.image" -}}
 {{- if .Values.global.privateRegistry.enabled -}}
-{{ .Values.global.privateRegistry.repository }}/ap-stan:{{ .Values.images.stan.tag }}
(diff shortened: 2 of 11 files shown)


charts/stan/templates/_helpers.tpl
```diff
@@ -25,7 +25,7 @@ Return the list of peers in a NATS Streaming cluster.
 
 {{ define "stan.image" -}}
 {{- if .Values.global.privateRegistry.enabled -}}
-{{ .Values.global.privateRegistry.repository }}/ap-stan:{{ .Values.images.stan.tag }}
+{{ .Values.global.privateRegistry.repository }}/ap-nats-streaming:{{ .Values.images.stan.tag }}
```
andriisoldatenko, line 28: it would be great to update tests as well.

rob-1126:
I'm working on a pair of tests now - one to check that the image name
begins with the override repository and the other that image names are
idempotent to the normal name when overridden with the quay rep we normally
use (aka named the same).

--Rob

charts/prometheus-node-exporter/values.yaml
```diff
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 image:
-  repository: quay.io/prometheus/node-exporter
+  repository: quay.io/astronomer/node-exporter
```
park-peter, line 5: Should revert. There is no repository called astronomer/node-exporter, we are pulling from prometheus repo.

charts/stan/templates/_helpers.tpl
park-peter: Added a test. Please review

tests/test_private_registry.py
```diff
@@ -0,0 +1,59 @@
...
+    public_repo_docs = render_chart()
+    private_repo_docs = render_chart(
+        values={
+            "global": {"privateRegistry": {"enabled": True, "repository": repository}}
+        },
+    )
+    # should be same number of images regardless of where they come from
+    assert len(public_repo_docs) == len(private_repo_docs)
+    search_string = "spec.template.spec.containers[*].image"
+    differtly_named_images = []
```
danielhoherd, line 18: Got a typo in several places in both tests
```suggestion
    differently_named_images = []
```

tests/test_private_registry.py
```diff
@@ -0,0 +1,59 @@
...
+        public_repo_images = jmespath.search(search_string, public_repo_doc)
+        private_repo_images = jmespath.search(search_string, private_repo_doc)
+        if public_repo_images is not None or private_repo_images is not None:
+            assert len(public_repo_images) == len(private_repo_images)
+            for public_repo_image, private_repo_image in zip(
+                public_repo_images, private_repo_images
+            ):
+                public_repo_image = public_repo_image.split("/")[-1]
+                private_repo_image = private_repo_image.split("/")[-1]
+                if public_repo_image != private_repo_image:
```
danielhoherd, line 29: This should probably be changed to `assert (comparison), "Error message"`

danielhoherd: Or... we could change line 36 to be `assert (len(differently_named_images) == 0), differently_named_images`

danielhoherd (changes requested):
Got a typo and a few suggestions. Looking pretty good though!

tests/test_private_registry.py
```diff
@@ -30,30 +34,34 @@ def test_private_repository_image_names_the_same_as_public_ones():
...
         if doc_images is not None:
             for image in doc_images:
                 if not image.startswith(repository):
                     print(
                         f"{image} did not begin with specified repository - {repository}"
                     )
-                    differtly_named_images.append(image)
-    assert len(differtly_named_images) == 0
+                    differently_named_images.append(image)
+    assert (len(differently_named_images) == 0, differently_named_images)
```
danielhoherd, line 67:
```suggestion
    assert (len(differently_named_images) == 0), differently_named_images
```

tests/test_private_registry.py
```diff
@@ -30,30 +34,34 @@ def test_private_repository_image_names_the_same_as_public_ones():
                     print(
                         f"image name differs when using a private repo named same as public - {public_repo_image} vs {private_repo_image}"
                     )
-                    differtly_named_images.append(
+                    differently_named_images.append(
                         (public_repo_image, private_repo_image)
                     )
-    assert len(differtly_named_images) == 0
+    assert (len(differently_named_images) == 0, differently_named_images)
```
danielhoherd, line 40:
```suggestion
    assert (len(differently_named_images) == 0), differently_named_images
```

danielhoherd (changes requested):
pre-commit is failing. also left two comments about invalid assert syntax.

danielhoherd (approved):

PR #1317, opened 2021-12-16
3870 migrate remaining helm unittests to pytest (part 2)

danielhoherd:
## Description

Migrate more helm-unittest tests to pytest, and add more tests while doing so.

There will be a part 3 to this. Might as well merge this work now.

## Related Issues

- Part 2 of https://github.com/astronomer/astronomer/pull/1313
- Related to https://github.com/astronomer/issues/issues/3870

## Testing

These are tests, and they are passing.


isaulv (approved):
LGTM. Only comment i have is that Yaml configuration stuff should be verified with a schema library instead of using isinstance checks.

danielhoherd:
> LGTM. Only comment i have is that Yaml configuration stuff should be verified with a schema library instead of using isinstance checks.

What specifically are you talking about? comment the line in https://github.com/astronomer/astronomer/pull/1317/files

bin/functional-tests/test_config.py
```diff
@@ -52,13 +52,13 @@ def test_houston_config(houston_api):
 
 
 def test_houston_can_reach_prometheus(houston_api):
```
tatiana, line 54: A PyTest feature that we may want to benefit from in the integration tests are the custom markers:
https://docs.pytest.org/en/6.2.x/example/markers.html

This would allow developers to run / not run specific tests (eg. Houston API or Nginx)

tatiana (commented):
Excellent work, @danielhoherd . I hadn't realized how high is the coverage of tests in Astronomer's Helm chart.

I noticed there are six classes decorated with `parametrize` - and each of them has around 2-3 methods. Currently, I understand that there are only three supported K8s versions. This Pytest will generate approximately 54 tests (6*3*3). Are we getting enough value of rendering charts for all the versions in all these classes? In the past, have we observed different behaviours in `render_chart` that would justify having individual tests perversion in all these classes/tests?

bin/functional-tests/test_config.py
```diff
@@ -152,12 +152,11 @@ def test_prometheus_config_reloader_works(prometheus, kube_client):
...
-    for i in range(12):
+    for _ in range(12):
         data = prometheus.check_output(
             "wget -qO- http://localhost:9090/api/v1/status/config"
         )
         j_parsed = json.loads(data)
-        # print(parsed['data']['yaml']['config']['global'])
         y_parsed = yaml.safe_load(j_parsed["data"]["yaml"])
         if y_parsed["global"]["scrape_interval"] != "30s":
             print(y_parsed["global"]["scrape_interval"])
```
jcomins-astro, line 162: in general, i think it would be better to use a logging library instead of print statements. but while we're using prints, should we print on line 165 as well? just to show that we're retrying?

bin/functional-tests/test_config.py
```diff
@@ -152,12 +152,11 @@ def test_prometheus_config_reloader_works(prometheus, kube_client):
         print(f"Exception when calling CoreV1Api->patch_namespaced_config_map: {e}\n")
 
     # This can take more than a minute.
-    for i in range(12):
+    for _ in range(12):
```
jcomins-astro, line 155: i'd rather see a polling library or helper function than loops with sleep in them. if we hang on line 156, the loop could potentially last much longer than 2 minutes

tests/helm_template_generator.py
```diff
@@ -65,15 +66,16 @@ def validate_k8s_object(instance, kube_version="1.21.0"):
...
 
 def render_chart(
-    name="RELEASE-NAME",
-    values=None,
-    show_only=None,
-    chart_dir=None,
-    kube_version="1.21.0",
-    baseDomain="example.com",
+    name: str = "RELEASE-NAME",
+    values: dict = {},
```
jcomins-astro, line 70: this and line 71 are bad. https://imgur.com/7UvO9VB
It's best to stick to None here, and there are other options like attr.s when you're initializing a class. Hit me up if you'd like more clarity

tests/helm_template_generator.py
```diff
@@ -93,6 +95,8 @@ def render_chart(
             "--values",
             tmp_file.name,
         ]
+        if namespace:
+            command.extend(["--namespace", namespace])
         if show_only:
```
jcomins-astro, line 100: set show_only default to `()` and delete this line

tests/helm_template_generator.py
jcomins-astro: this problem only happens with mutable objects, like dict and list. tuples are safe 
image

tests/test_kubed.py
```diff
@@ -0,0 +1,41 @@
...
+        """Test that helm renders a good chart for kubed with defaults."""
+        docs = render_chart(
+            kube_version=kube_version,
+            show_only=self.show_only,
+        )
+
+        assert len(docs) == 11
+
+    def test_kubed_disabled(self, kube_version):
+        """Test that helm renders does not render any k8s manifests when kubed is disabled."""
```
jcomins-astro, line 35: not sure if we're following any style guides. python black recommends 88 lines max.

tests/test_prometheus_alerts_configmap.py
```diff
@@ -1,50 +1,102 @@
...
-    assert isinstance(rule.get("expr"), str)
-    if rule["alert"] != "Watchdog":
-        assert isinstance(rule.get("for"), str)
-    assert isinstance(rule.get("labels"), dict)
-    assert isinstance(rule.get("annotations"), dict)
+        https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/
+        """
+        assert isinstance(rule.get("alert"), str)
+        assert isinstance(rule.get("expr"), str)
+        if rule["alert"] != "Watchdog":
```
jcomins-astro, line 32:
```
else:
    assert "alert" not in rule
```
?

tests/test_prometheus_alerts_configmap.py
```diff
@@ -1,50 +1,102 @@
...
-    assert doc["kind"] == "ConfigMap"
-    assert doc["apiVersion"] == "v1"
-    assert doc["metadata"]["name"] == "RELEASE-NAME-prometheus-alerts"
+        # Validate the contents of an embedded yaml doc
+        groups = yaml.safe_load(doc["data"]["alerts"])["groups"]
+        for group in groups:
+            assert isinstance(group.get("name"), str)
+            assert isinstance(group.get("rules"), list)
+            for rule in group["rules"]:
+                if rule.get("alert"):
```
jcomins-astro, line 58: I have a personal preference for `if "alert" in rule`. but it's just a nitpick

jcomins-astro: it might be nice to have a comment or docstring showing what a rule actually looks like

bin/functional-tests/test_config.py
jcomins-astro: i think it's better to leave them as separate test cases. they call different helpers and use different args. we can already select for test specific test cases to run with `-k` notation.

tests/test_prometheus_alerts_configmap.py
jcomins-astro: i think we could do something even cooler.
```
assert (rule["alert"] == "Watchdog") == ("for" not in rule)
```

bin/functional-tests/test_config.py
danielhoherd: what would the benefit of a logging library be in a test case?

bin/functional-tests/test_config.py
danielhoherd: Good catch, we can use pytest.flaky on this.

bin/functional-tests/test_config.py
danielhoherd: we were starting from very low confidence when we started with pytest chart testing, so running all versions made sense. it might make sense to default to not testing all versions of k8s, except in specific circumstances. i'll have to think more about that as these tests evolve. thanks for the feedback on that!

also i agree that having `-k` has served us pretty well. eg: `-k '1.21 and configmap_with_addition_alerts'` would only run one test even with the parametrized versions. i'm trying to think of a use case where having additional markers would help...

tests/helm_template_generator.py
danielhoherd: great catch! i wonder if there's a linter for this common pitfall...

tests/helm_template_generator.py
danielhoherd: Wouldn't setting show_only to `()` trigger the same empty data in assignment bug that you brought up earlier?

tests/test_kubed.py
danielhoherd: We're enforcing `black` via [pre-commit](https://github.com/astronomer/astronomer/blob/f67f926be1ac/.pre-commit-config.yaml) in CI 🤷‍♂️

```
$ black -l 88 tests
All done! ✨ 🍰 ✨
38 files left unchanged.
```

tests/test_prometheus_alerts_configmap.py
danielhoherd: Already there: https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/

tests/test_prometheus_alerts_configmap.py
danielhoherd: I like the first suggestion, but I think the second one would confuse people. Going with the first one.

bin/functional-tests/test_config.py
danielhoherd: actually i don't think pytest flaky is the right approach here...

danielhoherd:
I'm going to go ahead and merge this, but will keep the above suggestions in mind for part 3.

bin/functional-tests/test_config.py
jcomins-astro: tbh it probably depends on the environment. at my prev company, i had logging https://pypi.org/project/pytest-html/. html report was really nice for our UI tests since we were able to add screenshots next to our logs, but we probably don't need all this for this env

bin/functional-tests/test_config.py
jcomins-astro: https://github.com/justiniso/polling

tests/helm_template_generator.py
jcomins-astro: tuples are immutable, so it wouldn't. When we use `[]`, python stores a reference to the list's location as the default variable. When we use `()`, python will literally store an empty tuple.

bin/functional-tests/test_config.py
pgvishnuram: yes seems to be useful - on next phase lets optimise our test cases

PR #1334, opened 2022-01-07
Bump grafana to 8.3.3

danielhoherd:
## Description

This is a major version bump from grafana 7 to grafana 8.

## Related Issues

https://github.com/astronomer/issues/issues/3292

## Testing

No testing done for this specific version yet. We should definitely check this quite a lot before shipping it to customers.

Also it would be possible to write some functional tests for grafana, so maybe we should do that before merging this.


danielhoherd:
quay.io/astronomer/ap-grafana:8.3.3-1 has already been merged

PR #1335, opened 2022-01-10
setting proxy buffer size to 16k to avoid payload issues

pgvishnuram:
## Description

Due to payload size exceeds the default value end up in 502 error, this change fixes that issue by increasing 16k

## Related Issues

https://github.com/astronomer/issues/issues/3974

## Testing

NA

### DIFF
--- a/charts/astronomer/templates/houston/ingress.yaml
+++ b/charts/astronomer/templates/houston/ingress.yaml
@@ -26,6 +26,7 @@ metadata:
         deny all;
         return 403;
       }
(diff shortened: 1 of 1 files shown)


pgvishnuram:
@danielhoherd - do you need a test case for testing this value

danielhoherd:
> @danielhoherd - do you need a test case for testing this value

No. Because this doesn't have variable input I think this is fine as it is.

danielhoherd (approved):

PR #1337, opened 2022-01-13
rename ap-nginx-unprivileged to ap-auth-sidecar, enhance helm templates to reduce duplication

rob-1126:
## Description

Replace the hard-coded reference to nginxinc/nginx-unprivileged with a template that references an image in our quay repo or an image at global.privateRegistry.repository (when global.privateRegistry.enabled).

The existing behavior is to remain unaffected  (except for the new default location being at quay and the image name changing to ap-nginx-unprivileged). 

As such,  .Values.global.authSidecar.repository will continue to work as it currently does - over-riding this image value (regardless of whether custom repository is enabled)  and will continue to include the full repository name as it currently does.

## Related Issues

"vendor all images we use to simplify repository mirroring" - https://github.com/astronomer/issues/issues/3901

### no output normally, good, not enabled by default
helm template . |grep unpriv
### present when enabled looks good with default repo
helm template . --set global.authSidecar.enabled=True |grep unpriv
### following shows name and tag look good with custom repo
helm template . --set global.authSidecar.enabled=True --set global.authSidecar.repository=my-repo/astro-mirror/nginx-unprivileged|grep unpriv
### unit tests pass
venv/bin/python -m pytest -sv tests/test_authsidecar.py
### show-docker-images shows image
`make show-docker-images|grep unpriv lists the image`
was previously not present but should have been


charts/astronomer/templates/_helpers.yaml
```diff
@@ -41,6 +41,22 @@
...
+{{- else -}}
+quay.io/astronomer
+{{- end }}
+{{- end }}
+
+{{ define "authSidecar.image" -}}
+{{- if .Values.global.authSidecar.repository -}}
+{{ .Values.global.authSidecar.repository }}
+{{- else -}}
+{{ template "astronomer.platform_registry" . }}/ap-nginx-unprivileged
```
danielhoherd, line 56: This does not account for variety of the image name and does not include an image tag.

tests/test_authsidecar.py
```diff
@@ -4,6 +4,12 @@
 import jmespath
 import yaml
 
+current_tag_level = "1.21.4"
```
danielhoherd, line 7: Why is this variable `current_tag_level`? That doesn't seem to make sense here.

danielhoherd (changes requested):
Definitely needs more tests and a description ticket. I'm not sure what's even going on here. I see things changing, but I don't understand why they are changing and what the goal is.

charts/astronomer/templates/_helpers.yaml
rob-1126: Starting to dust off this ticket including giving it a proper description and existing the related upstream issue. Regarding this specific template - the optional .Values.global.authSidecar.repository (when set) must include the full path to the image and so users can over-ride the image name there. I wouldn't anticipate we'd want the portion after the default location to change - part of the purpose of this is to say 'look mirror everything here and mirror it with the existing names if you don't want to have to configure anything.'

rob-1126: @danielhoherd Added nine new tests, please review at your convenience.

tests/test_authsidecar.py
rob-1126: Updated to just explicitly declare the tag via fiat.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -41,6 +41,37 @@
 {{- end }}
 {{- end }}
 
+{{/*
+Returns a repository pre-fix to be used to construct image names throughout
```
danielhoherd, line 45:
```suggestion
Returns a repository prefix to be used to construct image names throughout
```

charts/astronomer/templates/_helpers.yaml
danielhoherd: @rob-1126 all of the other places we do this kind of thing are here: https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/_helpers.yaml#L12-L58

is there some reason we can't re-use that simpler snip of code that is working elsewhere?

tests/test_authsidecar.py
```diff
@@ -14,7 +19,9 @@ def test_authSidecar_alertmanager(self, kube_version):
         """Test Alertmanager Service with authSidecar."""
         docs = render_chart(
             kube_version=kube_version,
-            values={"global": {"authSidecar": {"enabled": True}}},
+            values={
+                "global": {"authSidecar": {"enabled": True, "tag": "placeholder-tag"}}
```
danielhoherd, line 23: This test has no assertions against this tag.

tests/test_authsidecar.py
```diff
@@ -189,3 +198,337 @@ def test_authSidecar_houston_configmap_with_annotation(self, kube_version):
...
+            "enabled": True,
+            "repository": "some.registry.internal/ap-nginx-unprivileged",
+            "tag": "placeholder-tag",
+            "port": 8084,
+            "pullPolicy": "IfNotPresent",
+            "annotations": {},
+        }
+        assert expected_output == prod["deployments"]["authSideCar"]
+
+    def test_authSidecar_houston_configmap_prioritizes_authSidecar_repository_over_privateRegistry_repository(
```
danielhoherd, line 320: This test name is defined twice.

danielhoherd (changes requested):
Got a few suggestions but it mostly looks good.

tests/test_authsidecar.py
rob-1126: removed the extraneous tag

tests/test_authsidecar.py
rob-1126: removed the duplicate test.

danielhoherd (approved):

pgvishnuram:
@danielhoherd - are we ship this change in 0.28.4

charts/astronomer/templates/_helpers.yaml
```diff
@@ -41,6 +41,37 @@
...
+{{ template "astronomer.platform_registry" . }}/ap-auth-sidecar
+{{- end }}
+{{- end }}
+
+{{/*
+Image of the authSidecar image (same as authSidecar.repository but with tag)
+*/}}
+{{ define "authSidecar.image" -}}
+{{ template "authSidecar.repository" . }}:{{ .Values.global.authSidecar.tag }}
+{{- end }}
```
danielhoherd, line 73: Why does this section need to be different from all of the other privateRegistry image templates?

danielhoherd:
@rob-1126 the title to this issue does not cover all of the changes that are included.

charts/astronomer/templates/_helpers.yaml
rob-1126: The templates we are currently using break image name over-riding when when privateRegistry is enabled.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -41,6 +41,37 @@
...
+the platform. Defaults to quay.io/astronomer but uses the value defined in
+global.privateRegistry.repository when enabled.
+*/}}
+{{ define "astronomer.platform_registry" -}}
+{{- if .Values.global.privateRegistry.enabled -}}
+{{ .Values.global.privateRegistry.repository }}
+{{- else -}}
+quay.io/astronomer
+{{- end }}
+{{- end }}
```
danielhoherd, line 55: Is this needed? The only place `astronomer.platform_registry` is used is `authSidecar.repository`.  I'm still having a hard time understanding why all of this cannot be the same as, say, `registry.image`. EG, structured in a way where the following big works:

```
{{ define "authSidecar.image" -}}
{{- if .Values.global.authSidecar.enabled -}}
{{ .Values.global.privateRegistry.repository }}/ap-auth-sidecar:{{ .Values.images.registry.tag }}
{{- else -}}
{{ .Values.images.registry.repository }}:{{ .Values.images.registry.tag }}
 {{- end }}
 {{- end }}
```

charts/astronomer/templates/_helpers.yaml
danielhoherd: Can you be more specific? Maybe it's something we should fix.

tests/test_authsidecar.py
```diff
@@ -187,3 +192,294 @@ def test_authSidecar_houston_configmap_with_annotation(self, kube_version):
...
+        )
+        assert (
+            container["image"]
+            == "some.registry.internal/ap-auth-sidecar:placeholder-tag"
+        )
+
+    def test_authSidecar_repository_doesnt_use_disabled_private_registry(
+        self, kube_version
+    ):
+        """Test authSidecar.image template doesnt use a value from privateRegistry when privateRegistry disabled"""
```
danielhoherd, line 458:
```suggestion
        """Test authSidecar.image template doesn't use a value from privateRegistry when privateRegistry disabled"""
```

pgvishnuram:
@danielhoherd This PR can be closed

PR #1339, opened 2022-01-18
set resources limits on node exporter and kube state metrics

pgvishnuram:
## Description

set resources limits to node exporter and kube state metrics

## Related Issues

https://github.com/astronomer/issues/issues/4056

## Testing

NA


danielhoherd (changes requested):
@pgvishnuram We should write tests for this. There's probably something we can copy paste real quick to cover this feature.

danielhoherd (approved):

PR #1340, opened 2022-01-18
disable postgres init container

pgvishnuram:
## Description

we no longer need to run init container with root privileges for in cluster postgres , we have already migrated our postgres image to run as non root. we can disable this in astronomer/astronomer postgres chart helm values

## Related Issues

https://github.com/astronomer/issues/issues/4087

## Testing

NA

### DIFF
--- a/charts/postgresql/values.yaml
+++ b/charts/postgresql/values.yaml
@@ -46,7 +46,7 @@ image:
 ## volumePermissions: Change the owner of the persist volume mountpoint to RunAsUser:fsGroup
 ##
 volumePermissions:
-  enabled: true
+  enabled: false
   image:
     ## ref: https://hub.docker.com/r/bitnami/minideb/tags/
     registry: docker.io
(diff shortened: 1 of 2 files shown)


danielhoherd (changes requested):
@pgvishnuram I'd like to see some tests with this.

Also, since we don't use this init container any more, do we even need it? Can we get rid of lines 48-68?

pgvishnuram:
I just thought of only disabling the init container configuration

danielhoherd:
> I just thought of only disabling the init container configuration

OK, that's fine, we can just leave it in place but disabled by default. Once there are tests we can merge this.

pgvishnuram:
@danielhoherd - is this test case is fine or you need to more validation

pgvishnuram:
@danielhoherd - is this test case is fine or you need to more validation

danielhoherd:
@pgvishnuram those tests are fine :) every little bit helps.

danielhoherd (approved):

PR #1345, opened 2022-01-24
Fix incorrect namespace for astronomer kube-state metrics resources

danielhoherd:
## Description

The values.yaml file has had an incorrect `kubeState` namespace for the kube-state chart for a few years. This was caught in testing, and was partially fixed. This change set changes the astronomer/astronomer values.yaml so what has been in there since 2019 is actually used.

Also refactor some pytest tests related to kube-state and node-exporter, removing jmespath which has some silently failing cases in favor of directly asserting against containers by their name using the `c_by_name` mechanic that is used in several other test files.

This was done as a response to our dev environment being broken due to kube-state not having enough resources because the custom vars here are not enough. I did set the defaults to 256Mi though, which is enough for our dev env, reported by @bote795.

## Related Issues

A lot of this work was originally done in astronomer/astronomer#1339 which was a response to the original discovery https://github.com/astronomer/issues/issues/4056

## Testing

```
testing high
testing low
pytest fast
helm-unittest slow
```


vishwas-astro (approved):

pgvishnuram:
Linking original issue - https://github.com/astronomer/issues/issues/4056

PR #1347, opened 2022-01-26
Allow nodeSelector, affinity, and tolerations configuration for Stan and Nats

dintorf:
## Description
Allow the ability to configure nodeSelector, affinity, and tolerations for Nats and Stan in order to restrict the Kubernetes nodes the Pods are scheduled on.

## Related Issues
https://github.com/astronomer/issues/issues/4137

## Testing
- Added `test_nats_sts.py` to validate generation of the statefulset yaml and nodeSelector, affinity, and tolerations are configured correctly
- Added test in `test_stan_sts.py` to validate nodeSelector, affinity, and tolerations are configured correctly

### DIFF
--- a/tests/test_nats_sts.py
+++ b/tests/test_nats_sts.py
@@ -30,12 +30,12 @@ class TestNatsStatefulSet:
             "quay.io/astronomer/ap-nats-server:"
         )
         assert c_by_name["nats"]["livenessProbe"] == {
-            "httpGet": {"path": "/streaming/serverz", "port": "monitor"},
+            "httpGet": {"path": "/", "port": 8222},
             "initialDelaySeconds": 10,
             "timeoutSeconds": 5,
         }
         assert c_by_name["nats"]["readinessProbe"] == {
-            "httpGet": {"path": "/streaming/serverz", "port": "monitor"},
+            "httpGet": {"path": "/", "port": 8222},
             "initialDelaySeconds": 10,
             "timeoutSeconds": 5,
         }
--- a/tests/test_nats_sts.py
+++ b/tests/test_nats_sts.py
@@ -69,11 +69,62 @@ class TestNatsStatefulSet:
 
     def test_nats_statefulset_with_affinity_and_tolerations(self, kube_version):
         """Test that nats statefulset renders proper nodeSelector, affinity, and tolerations"""
+        values = {
+            "nats": {
+                "nodeSelector": {"role": "astro"},
+                "affinity": {
+                    "nodeAffinity": {
+                        "requiredDuringSchedulingIgnoredDuringExecution": {
+                            "nodeSelectorTerms": [
+                                {
+                                    "matchExpressions": [
+                                        {
+                                            "key": "astronomer.io/multi-tenant",
+                                            "operator": "In",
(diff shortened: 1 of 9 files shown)


dintorf:
Odd, all tests passed locally

dintorf:
oops, changed the helm command to get around the template issue with RELEASE-NAME

danielhoherd:
@dintorf what's up with the PR title? seems completely unrelated.

dintorf:
not sure what happened there, I'll fix it

tests/test_nats_sts.py
```diff
@@ -0,0 +1,119 @@
...
+        c_by_name = {c["name"]: c for c in containers}
+        assert c_by_name["nats"]["resources"]["requests"]["cpu"] == "123m"
+        assert c_by_name["metrics"]["resources"]["requests"]["cpu"] == "234m"
+
+    def test_nats_statefulset_with_affinity_and_tolerations(self, kube_version):
+        """Test that nats statefulset renders proper nodeSelector, affinity, and tolerations"""
+        docs = render_chart(
+            kube_version=kube_version,
+            show_only=["charts/nats/templates/statefulset.yaml"],
+            values={
```
AmFlint, line 75: Maybe you could also add a test with the .Values.global default values for nodeSelector, affinity and tolerations ?

As you've added the behaviour to default to these values if chart-level values are not provided.

tests/test_nats_sts.py
```diff
@@ -0,0 +1,119 @@
...
+        assert spec["affinity"] != {}
+        assert (
+            len(
+                spec["affinity"]["nodeAffinity"][
+                    "requiredDuringSchedulingIgnoredDuringExecution"
+                ]["nodeSelectorTerms"]
+            )
+            == 1
+        )
+        assert len(spec["tolerations"]) > 0
```
AmFlint, line 119: Maybe here you could directly assert that spec["tolerations"] is the same as the one you've passed to the template.

```python
values = {} # the values defined above
docs = render_chart(values=values) #...

assert spec["tolerations"] = values["nats"]["tolerations"]
```

It would make sure that even if in the futur we add a default tolerations or injected from another variable, we're still testing the right behavior (injecting variables from the chart level values).

tests/test_stan_sts.py
```diff
@@ -62,3 +66,54 @@ def test_stan_statefulset_with_metrics_and_resources(self, kube_version):
...
+        assert spec["affinity"] != {}
+        assert (
+            len(
+                spec["affinity"]["nodeAffinity"][
+                    "requiredDuringSchedulingIgnoredDuringExecution"
+                ]["nodeSelectorTerms"]
+            )
+            == 1
+        )
+        assert len(spec["tolerations"]) > 0
```
AmFlint, line 119: Same comment as above, maybe we want to be more specific on the assertion here to test the right behavior.

CarpathianUA:
Interested in this PR to merged

danielhoherd (changes requested):
Very nice additions! I agree with @AmFlint that the extra tests he pointed out will top it off.

danielhoherd (approved):

PR #1351, opened 2022-01-31
Add missing security fixes release 0 25

pgvishnuram:
## Description

* kibana 7.10.2-2 -> 7.16.3
* elasticsearch 7.10.2-2 -> 7.16.3
* curator 5.8.4-7 -> 5.8.4-8
* db-bootstrapper 0.26.1 -> 0.26.2 
* nginx ingress 0.49.0-1 -> 0.50.0
* stan ap-base 3.15.0 -> 3.15.0-1

## Related Issues

https://github.com/astronomer/issues/issues/4123

## Testing

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -32,7 +32,7 @@ images:
     pullPolicy: IfNotPresent
   dbBootstrapper:
     repository: quay.io/astronomer/ap-db-bootstrapper
-    tag: 0.26.1
+    tag: 0.26.2
     pullPolicy: IfNotPresent
   cliInstall:
     repository: quay.io/astronomer/ap-cli-install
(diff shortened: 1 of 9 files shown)


pgvishnuram:
@danielhoherd - upstream has added log4j fixes by default , should we add the env vars ??

pgvishnuram:
@vishwas-astro

danielhoherd:
@pgvishnuram it doesn't make sense to require the env vars if we are using the fixed images. We can remove the env vars wherever the fixed images are being used.

Also we should not merge this until we also have fixes in place for 0.26, because doing so would break the upgrade path from 0.25 to 0.26 which is currently required (not tested any other way) to get to 0.27 or 0.28.

danielhoherd (approved):

PR #1362, opened 2022-02-07
Use lowercase helm data for helm 3.8 compatibility.

danielhoherd:
## Description

Use lowercase helm data for helm 3.8 compatibility.

> helm 3.8.0 has a regex that does not allow capitals in release names. I think this only affects our test environments.
> 
> ```
> $ helm-3.7.2 template RELEASE-NAME . >/dev/null ; echo $? ;
> 0
> $ helm-3.8.0 template RELEASE-NAME . >/dev/null ; echo $? ;
> Error: release name "RELEASE-NAME": invalid release name, must match regex ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ and the length must not be longer than 53
> 1
> ```
> 
> This [validation regex is 17 months old](https://github.com/helm/helm/blame/dff9ade9d6b10ad871a5/pkg/chartutil/validate_name.go#L73), so it looks like they are just now enforcing what has existed for a while.

This change will have to be ported to all supported branches. Unfortunately, test coverage changes drastically between branches, so if this change isn't handled by `git grep -l RELEASE | xargs gsed -i 's/RELEASE/release/g'` and other simple logic (which it should be) then we'll have to spend more time fixing those.

## Related Issues

https://github.com/astronomer/issues/issues/4124

## Testing

No additional testing should be necessary. If CI passes we should be good, because AFAIK this change only applies to CI and related test code.


bin/functional-tests/test_config.py
```diff
@@ -203,8 +203,8 @@ def test_houston_backend_secret_present_after_helm_upgrade_and_container_restart
         print("No NAMESPACE env var, using NAMESPACE=astronomer")
         namespace = "astronomer"
 
-    if not (release_name := getenv("RELEASE_NAME")):
-        print("No RELEASE_NAME env var, usgin RELEASE_NAME=astronomer")
+    if not (release_name := getenv("release_NAME")):
```
AmFlint, line 206: Same here

bin/functional-tests/test_versioning.py
```diff
@@ -52,8 +52,8 @@ def test_downgrade_then_upgrade():
         print("NAMESPACE env var is not present, using 'astronomer' namespace")
         namespace = "astronomer"
 
-    if not (release_name := getenv("RELEASE_NAME")):
-        print("No RELEASE_NAME env var, usgin RELEASE_NAME=astronomer")
+    if not (release_name := getenv("release_NAME")):
```
AmFlint, line 55: Same here

bin/functional-tests/conftest.py
```diff
@@ -10,9 +10,9 @@
     print("NAMESPACE env var is not present, using 'astronomer' namespace")
     namespace = "astronomer"
 
-if not (release_name := getenv("RELEASE_NAME")):
+if not (release_name := getenv("release_NAME")):
```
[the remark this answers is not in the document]
danielhoherd: Good catch! This was just a quick late-night start. I should've marked it as not ready for review yet since I was really just looking to see if CI passed. I'll circle back and give it a closer look when I have time.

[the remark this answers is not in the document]
danielhoherd: Or... I guess I'm mis-remembering since I specifically asked for review above, haha

pgvishnuram (approved):

bin/functional-tests/conftest.py
```diff
@@ -10,9 +10,9 @@
     print("NAMESPACE env var is not present, using 'astronomer' namespace")
     namespace = "astronomer"
 
-if not (release_name := getenv("RELEASE_NAME")):
```
AmFlint, line 13: the Name of the environment variable should remain `RELEASE_NAME`

bin/functional-tests/test_versioning.py
```diff
@@ -52,8 +52,8 @@ def test_downgrade_then_upgrade():
         print("NAMESPACE env var is not present, using 'astronomer' namespace")
         namespace = "astronomer"
 
-    if not (release_name := getenv("RELEASE_NAME")):
```
AmFlint, line 55: name of env var should remain `RELEASE_NAME`

bin/install-platform
```diff
@@ -6,7 +6,7 @@ set -x
 # The path to the working directory - the root of the repo
 GIT_ROOT="$(git -C "${0%/*}" rev-parse --show-toplevel)"
 HELM_CHART_PATH=${HELM_CHART_PATH:-$GIT_ROOT}
-RELEASE_NAME="${RELEASE_NAME:-astronomer}"
```
AmFlint, line 9: The name of the variable should remain `RELEASE_NAME`, we just want to change the value of this variable to all lowercase

bin/install-platform
```diff
@@ -73,7 +73,7 @@ set -x
 # shellcheck disable=SC2086
 if ! helm ${ACTION} \
   --namespace "$NAMESPACE" \
-  "$RELEASE_NAME" \
+  "$release_name" \
```
AmFlint, line 76: Var name -> Uppercase

bin/install-platform
```diff
@@ -91,5 +91,5 @@ then
   get_debugging_info
   exit 1
 else
-  helm history -n "$NAMESPACE" "$RELEASE_NAME"
+  helm history -n "$NAMESPACE" "$release_name"
```
AmFlint, line 94: var name -> uppercase

bin/test-ap
```diff
@@ -13,7 +13,7 @@ virtualenv --python=python3 /tmp/venv
 source /tmp/venv/bin/activate
 pip install -r "$GIT_ROOT/requirements/functional-tests.txt"
 export NAMESPACE=astronomer
-export RELEASE_NAME=astronomer
```
AmFlint, line 16: Same here, the ENV Var to export should be uppercase (`RELEASE_NAME`)

AmFlint (changes requested):

PR #1366, opened 2022-02-08
Release 0.28

bote795:
## Description

update main with versions from 0.28 release

## Related Issues

https://github.com/astronomer/issues/issues?q=is%3Aissue+label%3A0.28.0+

## Testing

0.28 is being tested by qa

### DIFF
--- a/configs/local-dev.yaml
+++ b/configs/local-dev.yaml
@@ -68,8 +68,11 @@ astronomer:
       requests:
         cpu: "0m"
         memory: "0Mi"
-
     config:
+      registry:
+        protectedCustomRegistryEnabled: true
+      deployments:
+        enableUpdateDeploymentImageEndpoint: true
       emailConfirmation: false
       auth:
         local:
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: astronomer
-version: 0.28.0-rc5
-appVersion: 0.28.0-rc5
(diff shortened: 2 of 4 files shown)


configs/local-dev.yaml
```diff
@@ -68,8 +68,11 @@ astronomer:
       requests:
         cpu: "0m"
         memory: "0Mi"
-
     config:
+      registry:
+        protectedCustomRegistryEnabled: true
+      deployments:
+        enableUpdateDeploymentImageEndpoint: true
       emailConfirmation: false
```
vishwas-astro, line 76: Do we need the above 4 lines? Is this a dev only set up which should get reflected in our dev cluster?

danielhoherd:
This merge is going the wrong way. We should not be merging the latest release into main, we should be merging main into the latest release branch. It's not important that all those rc changes go into the main branch history, and the other things that were bumped should have been bumped in main to begin with.

vishwas-astro:
@danielhoherd is correct. We should update main with the Houston and Astro UI upgrades first and then pick them to release branch

vishwas-astro:
We should still update main. No opinion on whether you do it this way or raise a different PR against main.

configs/local-dev.yaml
bote795: not sure why this was added.. but ill just close the issue and just update the versions..

bote795:
created following?: https://github.com/astronomer/astronomer/pull/1367

PR #1372, opened 2022-02-10
Add tests for stan

danielhoherd:
## Description

release-0.27 did not have tests for stan, and we ran into some problems with some release-0.28 in airgapped environments. This adds tests to make sure release-0.27 image customization works. The boilerplate tests were taken from release-0.28, but `test_stan_statefulset_with_custom_images` is not in release-0.28 and should be ported and updated for those changes.

Part of the problem here included imagePullPolicy being absent from a pod manager manifest. To check for this everywhere I wrote a test that checks all pod managers for imagePullPolicy. This test found 3 other places where it was missing, along with at least one other place where it would be missing if a feature that appears to be broken (but apparently quite unused) was enabled. I have fixed those three instances so those changes appear in this change set, and have [opened an issue for the broken feature](https://github.com/astronomer/issues/issues/4228).

## Itemized changes

- Add test for all pod managers to have imagePullPolicy
- Fix or add imagePullPolicy in nats, nginx, stan (total 4 containers)
- Add tests for nats statefulset

## Related Issues

https://github.com/astronomer/issues/issues/4229

## Testing

Testing is the name of the game here.

### DIFF
--- a/charts/nats/templates/statefulset.yaml
+++ b/charts/nats/templates/statefulset.yaml
@@ -132,7 +132,7 @@ spec:
       containers:
       - name: nats
         image: {{ include "nats.image" . }}
-        imagePullPolicy: {{ .Values.nats.pullPolicy }}
+        imagePullPolicy: {{ .Values.images.nats.pullPolicy }}
         ports:
         - containerPort: 4222
           name: client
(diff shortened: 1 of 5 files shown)


vishwas-astro:
@danielhoherd - any reason you are raising this against release-0.27 branch and not main/0.28? I thought we are running into airgapped install problems on 0.28 release.

vishwas-astro:
> @vishwas-astro yes. the problem came about because we did not have testing in place in past releases where it worked. i have made the past release work, and after this i will port these working tests to the new release, see what has broken, and fix the problem.

makes sense.

vishwas-astro:
not sure I am the best person to review this. but here it goes :)

vishwas-astro (approved):

PR #1384, opened 2022-02-22
make sysctl config configurable for elasticsearch

pgvishnuram:
## Description

* this change enables customer to enable|disable sysctl config option (defaults to true)
* applicable for master,data and client services

## Related Issues

Issue Link: https://github.com/astronomer/issues/issues/4226

## Testing

QA team can set elasticsearch.sysctlInitContainer.enabled=false to see if sysctl containers are not applied anymore.

### DIFF
--- a/charts/elasticsearch/values.yaml
+++ b/charts/elasticsearch/values.yaml
@@ -209,6 +209,7 @@ nginx:
   resources: {}
 
 # sysctl init container
+# Warning: sysctl changes affect the kernel, and thus, all containers running on the same node.
 sysctlInitContainer:
   enabled: true
   sysctlVmMaxMapCount: 262144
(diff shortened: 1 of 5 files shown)


charts/elasticsearch/values.yaml
```diff
@@ -207,3 +207,8 @@ nginx:
   fullnameOverride: ~
   ingressClass: ~
   resources: {}
+
+# sysctl init container
```
danielhoherd, line 211: We should put a warning here that sysctl affects the entire node, so all containers in all pods on the node will be affected by the sysctl changes.

danielhoherd, line 211:
```suggestion
# sysctl init container
# Warning: sysctl changes affect the kernel, and thus all containers running on the same node
```

danielhoherd (approved):

PR #1385, opened 2022-02-23
remove json schema from astronomer/astronomer to support airgapped customers

aliotta:
Description: Airgapped customers want the ability to helm upgrade w/o internet access. this PR removes the internet dependence by copying and pasting over the airflow chart json schema file and reffing it locally instead of via internet.

Issue: https://app.zenhub.com/workspaces/team-platform-apps-61d5f7ef8de56e0014840a75/issues/astronomer/issues/4289


jedcunningham:
I haven't tested it, so maybe it's been fixed now, but I don't think this will work: https://github.com/helm/helm/issues/10481

tl;dr: it looks in the pwd where you run helm, not in the chart's root.

aliotta:
> I haven't tested it, so maybe it's been fixed now, but I don't think this will work: [helm/helm#10481](https://github.com/helm/helm/issues/10481)
> 
> tl;dr: it looks in the pwd where you run helm, not in the chart's root.

Wow that's very frustrating

danielhoherd:
@aliotta can you link the issue in the description?

aliotta:
Spoke with @cmarteepants on this one. They are in favor of just dropping the json schema altogether.

"right. I think focusing on showcasing commander errors is the way to go, because that upgrade isn't the only situation where you see the helm commands failing, and only feedback being Airflows not coming up when creating new deployments, or code not updating when deploying an updated image to an existing deployment"

danielhoherd (approved):
YOLO? 🤷‍♂️

So long schema validation, we hardly knew ye.

PR #1395, opened 2022-03-03
Add Prometheus config file global.external_labels, remote_write

danielhoherd:
## Description

Add the ability to specify helm values to be passed into the prometheus config file's remote_write and global.external_labels sections.

## Itemized changes

- Add .Values.external_labels and .Values.remote_write sections to prometheus config file template
- Add tests for Values.external_labels, and .Values.remote_write
- Add commented out sections in charts/prometheus/values.yaml which indicate these available options, and links to prometheus documentation on how to use them
- Refactor adjacent config test test_prometheus_config_configmap_with_different_name_and_ns to not use regex, and instead do string comparisons against loaded prometheus config. This was mainly done for consistency.

## Related Issues

- https://github.com/astronomer/issues/issues/4186

## Testing

I have written unit tests that capture my understanding of the original request made by @fhoda.

Prometheus is tested in CI too, so if prometheus works there we know that the defaults didn't break anything, but I have not implemented a functional test for this feature. Doing so is not a trivial process with our current test process.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 jobs:
   run_pre_commit:
     docker:
-      - image: quay.io/astronomer/ci-pre-commit:2022-01
+      - image: quay.io/astronomer/ci-pre-commit:2022-03
     steps:
       - checkout
       - run:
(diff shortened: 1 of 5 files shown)


pgvishnuram (approved):

fhoda:
@danielhoherd for these types of changes that are hard to test, what do you think about some sort of smoke test included in these PRs? Something like include the values yaml submitted to helm and the output/rendered yaml. Idea is that if I've applied these configs in the wild I know exactly what it should look like for it to work, so I can validate based on that.

fhoda:
Also thank you for getting this in so quickly!

PR #1396, opened 2022-03-04
re-addValuesSchemaForOptionalValidation

aliotta:
## Description

Re-add previous json schema file that can be used optionally when needed. (it was previously removed b/c it blocked air-gapped customers)

## Issue

https://github.com/astronomer/issues/issues/4336


danielhoherd (changes requested):
Please add a link in the description to a ticket describing why this change is being made.

README.md
```diff
@@ -133,3 +133,7 @@ The values output by this command will need to be inserted manually into astrono
 ## License
 
 The code in this repo is licensed Apache 2.0 with Commons Clause, however it installs Astronomer components that have a commercial license, and requires a commercial subscription from Astronomer, Inc.
+
+## Optional schema validation
+
+The ./bin/values.schema.json file can be used to validate the helm values you are using work with the default airflow chart shipped with this repo. To use them move the file to from bin into the root of this directory and proceed with the helm lint, install, and upgrade commands as normal.
```
danielhoherd, line 139: `bin/values.schema.json` is the wrong place for this file, but I'm not sure what the right place is. Maybe `values.schema.json.example` ? We've abused the `bin` dir in the past and I have been working to reduce that dir to only executables.

cmarteepants (approved):
LGTM

PR #1399, opened 2022-03-07
Update commander RBAC: remove RBAC resources if clusterRoles disabled

AmFlint:
## Description

When a user deploys the Astronomer Platform, he wants to configure Commander in either of the following ways:
- Use default RBAC resources configured by this chart -> a ClusterRole and a ClusterRoleBinding that allows commander to manage Airflow deployments in all namespaces of the cluster.
- Use a custom RBAC configuration to enhance the security/access level of commander to only manage certain resources in specific namespaces

Before this PR, this chart either created `ClusterRole and ClusterRoleBinding` by default, or a `Role and Rolebinding` for commander's permissions. However as mentioned above, if the user creates his own custom permissions, he doesn't want to get a default `role and rolebinding` resources created, as he won't be using these.

Moreover, commander actually interacts with the namespace in which it's deployed, so creating `Role and Rolebinding` for commander to provide permissions in the same namespace doesn't make sense.

In this PR, I changed the behaviour of the parameter `global.clusterRoles`:
- If set to `true`, creates a ClusterRole and ClusterRoleBinding as mentioned in above scenario.
- If set to `false`, does not create any RBAC resources (instead of creating a Role + RoleBinding in commander's namespace).

## Related Issues

https://github.com/astronomer/issues/issues/4320

## Testing

To test this update:
- Deploy this helm chart with `--set global.clusterRoles=true --set global.rbacEnabled=true` to test that default ClusterRole and ClusterRoleBinding are created as expected:
```bash
helm template my-release . --set global.clusterRoles=true --set global.rbacEnabled=true --show-only charts/astronomer/templates/commander/commander-role.yaml --show-only charts/astronomer/templates/commander/commander-rolebinding.yaml

# this should display 2 resources: a ClusterRole and a ClusterRoleBinding
```
- Deploy this helm chart with either clusterRoles set to false or rbacEnabled set to false: should not create any RBAC resources for commander:
```bash
helm template my-release . --set global.clusterRoles=true --set global.rbacEnabled=false --show-only charts/astronomer/templates/commander/commander-role.yaml --show-only charts/astronomer/templates/commander/commander-rolebinding.yaml

# or
helm template my-release . --set global.clusterRoles=false--set global.rbacEnabled=true --show-only charts/astronomer/templates/commander/commander-role.yaml --show-only charts/astronomer/templates/commander/commander-rolebinding.yaml

# Both of these commands should result in no RBAC resources being created
```


AmFlint:
I'm closing this PR, since the same fix is included along many improvements in the following PR: https://github.com/astronomer/astronomer/pull/1407

Issue #1370, opened 2022-02-09
K8s v1.22 support


whyvez:
From looking at your [website](https://docs.astronomer.io/enterprise/version-compatibility-reference/), it looks like you don't yet support k8s `v1.22`. 

We are on GKE and use the `REGULAR` release channel in which `v1.22` is already available.

Can you provide more information about if or when you will support k8s `v1.22`?

danielhoherd:
There is a PR open to add this feature. We are doing extra work to validate it before merging, and updating nginx-ingress from 0.x to 1.1.x. This PR will be merged in the next few weeks. https://github.com/astronomer/astronomer/pull/1401

danielhoherd:
@whyvez 1.22, 1.23 support will be going out with 0.29, and will most likely be back-ported to 0.28 at some point.

PR #1401, opened 2022-03-09
3889 k8s 1.22, 1.23 support

danielhoherd:
## Description

Add support for k8s 1.22, 1.23. Originally this was slated for 1.22, but 1.23 is out now so why not support it too, especially because even 1.22 will be EOL by 2023 for [upstream](https://github.com/kubernetes/website/blob/main/content/en/releases/patch-releases.md#122) and [AKS](https://docs.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#aks-kubernetes-release-calendar).

## Related Issues

- <https://github.com/astronomer/issues/issues/3889>
- Closes <https://github.com/astronomer/astronomer/issues/1370>

## Testing

If CI is passing that is one thing, but we should also make sure to test this with an airflow deployment before merging, so we can make sure that our airflow-chart works in 1.22+ when deployed through commander.


charts/keda/requirements.yaml
```diff
@@ -1,5 +1,5 @@
 dependencies:
 - name: keda
-  version: 1.3.0
+  version: 2.6.2
```
pgvishnuram, line 3: are we trying to only put chart changes not images. whats the actual benefit out of it

danielhoherd: The old version had api incompatibilities.

vishwas-astro: Should we completely get rid of keda in our software?

danielhoherd: hah, i don't know. do we use it? do customers use it? i'm honestly not sure where we are at with this.

danielhoherd:
I'm pretty sure this is good to be merged. @pgvishnuram do you know of anything else we need to do before merging this?

charts/keda/requirements.yaml
danielhoherd: If we want to get rid of keda, we should do it in a different branch. I've opened an issue to track that discussion: <https://github.com/astronomer/issues/issues/4441>

pgvishnuram:
Are we going to support 1.18 - if not we can merge this change

danielhoherd:
@pgvishnuram I'd like to merge the removal of keda before merging this so that the keda upgrade included in this change set never makes it into the repo history. as for k8s 1.18, I'm not sure. More discussion on that point in <https://github.com/astronomer/issues/issues/3889#issuecomment-1081228761>

PR #1404, opened 2022-03-10
Remove nginx metrics port from being exposed by the ingress load balancer.

shmanu017:
## Description

- Removed the following from the current service definition

```
    - port: {{ .Values.ports.metrics }}
      targetPort: {{ .Values.ports.metrics }}
      protocol: TCP
      name: metrics
```

- Created the following new separate Service and Network Policy definitions for Nginx metrics 
 

## Related Issues

https://github.com/astronomer/issues/issues/3040

## Testing

- Updated existing nginx-service test case to included metrics service
- Added networkpolicy initial test to verify the types


tests/test_nginx_service.py
```diff
@@ -105,9 +105,9 @@ def test_nginx_type_nodeport_specifying_nodeports(self):
         doc = docs[0]
         ports = doc["spec"]["ports"]
         ports_by_name = {x["name"]: x["nodePort"] for x in ports}
+        print(f"port_names dict: {ports_by_name}")
```
danielhoherd, line 108: This should be removed

tests/test_nginx_service.py
```diff
@@ -120,3 +120,12 @@ def test_nginx_enabled_externalips(self):
         doc = docs[0]
         assert len(doc["spec"]["externalIps"]) > 0
         assert "1.2.3.4" in doc["spec"]["externalIps"]
+
+    def test_nginx_metrics_service_type(self):
+        NodeMetricsPort = 10254
```
danielhoherd, line 125: This is defined but never used.

tests/test_nginx_service.py
```diff
@@ -1,3 +1,4 @@
+from lib2to3.pgen2.token import EQUAL
```
danielhoherd, line 1: This does not appear to be used... what even is this? Never seen it before, looks odd.

danielhoherd (changes requested):
This looks like a good start. I get confused about the NetworkPolicy stuff I'd have to check those while the cluster was running to be sure. It would be great if we had functional tests for them, but we can work on that later.

The unit tests here need some work, but so far it's looking like a good start.

tests/test_nginx_service.py
shmanu017: Added an asset to compare the port

tests/test_nginx_networkpolicy.py
```diff
@@ -0,0 +1,16 @@
+from tests.helm_template_generator import render_chart
+
+
+class TestNginx:
```
danielhoherd, line 4:
```suggestion
class TestNginxNetworkPolicy:
```

tests/test_nginx_service.py
```diff
@@ -135,3 +133,13 @@ def test_nginx_enabled_externalips(self):
...
+
+    def test_nginx_metrics_service_type(self):
+        NodeMetricsPort = 10254
+        docs = render_chart(
+            show_only=["charts/nginx/templates/nginx-metrics-service.yaml"],
+        )
+        assert len(docs) == 1
+        doc = docs[0]
+        assert doc["spec"]["type"] == "ClusterIP"
+        assert doc["spec"]["ports"][0]["port"] == NodeMetricsPort
```
danielhoherd, line 145:
```suggestion
        doc = render_chart(
            show_only=["charts/nginx/templates/nginx-metrics-service.yaml"],
        )[0]
        assert doc["spec"]["type"] == "ClusterIP"
        assert doc["spec"]["ports"][0]["port"] == 10254
```

danielhoherd (approved):

PR #1407, opened 2022-03-14
Improve namespace pools configuration

AmFlint:
## Description

We want to re-work the namespace pools feature configuration.

At the moment, in order to use "namespace pools", users have to configure many different flags in many different scopes of this chart. We want to make it easier to configure this feature as it's very important for a lot of users that want to improve the security in their cluster.

More details in Issue 4358 linked bellow.

So, in this PR, when you configure the `global.namespacePools`, we automatically:
- configure houston's `production.yaml` configuration to use the appropriate feature flags (`hardDeleteDeployment`, `manualNamespacesEnabled` and the list of namespaces chosen by the user: `preCreatedNamespaces`)
- Configure a mandatory environment variable in Commander (container inside the Deployment object)
- Create appropriate RBAC resources, Role and RoleBinding for each provided namespaces in the namespace pool list. Also, if neither namespacePools and clusterRoles is enabled, don't create any RBAC resources for commander (fixes issue 4320).
- Create `Namespaces` resources as part of this helm chart if enabled.
-  Update fluentd Configuration to also handle `global.namespacePools.enabled` value.

In the futur, these changes allow us to improve the security of multiple places, for example fluent-d's configuration where we'll be able to target only provided namespaces in the pool (part of another issue).

## Related Issues

- https://github.com/astronomer/issues/issues/4358
- https://github.com/astronomer/issues/issues/4320: this PR will also fix this issue
- https://github.com/astronomer/issues/issues/4383

## Testing

There are multiple areas to test with these changes.

### Commander RBAC permissions

**First, test that if namespacePools is enabled and a list of namespaces is provided, we create a Role and RoleBinding for each of these namespaces**:
```bash
helm template . --kube-version=1.20.0 --show-only charts/astronomer/templates/commander/commander-role.yaml --show-only charts/astronomer/templates/commander/commander-rolebinding.yaml --set global.namespacePools.enabled=true --set global.namespacePools.namespaces.names="{my-ns-1,my-ns-2}"
```
This should display 2 Roles and 2 RoleBindings, 1 for each namespaces provided in `namespacePools.namespaces.names`. And RoleBinding.roleRef should point to the same role, and RoleBinding.subject should point to the serviceAccount created for commander in the release namespace.

**Then, also test that if namespacePools is disabled and clusterRoles is enabled, we create a ClusterRole and a ClusterRoleBinding for commander**:
```bash
helm template . --kube-version=1.20.0 --show-only charts/astronomer/templates/commander/commander-role.yaml --show-only charts/astronomer/templates/commander/commander-rolebinding.yaml --set global.namespacePools.enabled=false --set global.clusterRoles=true
```
This should display a ClusterRole and a ClusterRoleBinding.

**Finally, if both namespacePools and clusterRoles are disabled, do not create any RBAC resource**:
```bash
helm template . --kube-version=1.20.0 --show-only charts/astronomer/templates/commander/commander-role.yaml --show-only charts/astronomer/templates/commander/commander-rolebinding.yaml --set global.namespacePools.enabled=false --set global.clusterRoles=false
```

### Namespaces in the Namespace Pool

**First, check that if we enabled namespacePools and namespaces creation in the pool with a list of names, it creates the resources**:
```bash
helm template . --kube-version=1.20.0 --show-only charts/astronomer/templates/namespaces.yaml --set global.namespacePools.enabled=true --set global.namespacePools.namespaces.create=true --set global.namespacePools.namespaces.names="{my-ns-1,my-ns-2}"
```
This should display 2 namespaces, my-ns-1 and my-ns-2 as we provided these names as a value.

**Second, check that no namespaces are created if we disable create or disable namespacePools**:
```bash
helm template . --kube-version=1.20.0 --show-only charts/astronomer/templates/namespaces.yaml --set global.namespacePools.enabled=true --set global.namespacePools.namespaces.create=false --set global.namespacePools.namespaces.names="{my-ns-1,my-ns-2}"
```
and with namespacePools disabled:
```bash
helm template . --kube-version=1.20.0 --show-only charts/astronomer/templates/namespaces.yaml --set global.namespacePools.enabled=false --set global.namespacePools.namespaces.create=true --set global.namespacePools.namespaces.names="{my-ns-1,my-ns-2}"
```
Both of these commands should not display any resources.

### Houston ConfigMap 

**First, check that feature flags are added automatically if namespacePools is enabled**:
```bash
helm template . --kube-version=1.20.0 --show-only charts/astronomer/templates/houston/houston-configmap.yaml --set global.namespacePools.enabled=true --set global.namespacePools.namespaces.create=true --set global.namespacePools.namespaces.names="{my-ns-1,my-ns-2}"
```
The configmap will be displayed, you need to make sure that the following feature flags are set properly in the `production.yaml` property:
```yaml
# ...
deployments:
  hardDeleteDeployment: true
  manualNamespacesEnabled: true
  preCreatedNamespaces:
    - my-ns-1
    - my-ns-2
# ...
```

**Second, check that the feature flags are not added when namespacePools is disabled**:
```bash
helm template . --kube-version=1.20.0 --show-only charts/astronomer/templates/houston/houston-configmap.yaml --set global.namespacePools.enabled=false --set global.namespacePools.namespaces.create=true --set global.namespacePools.namespaces.names="{my-ns-1,my-ns-2}"
```
This will display the houston configmap, you need to make sure that the `hardDeleteDeployment`, `manualNamespacesEnabled` and `preCreatedNamespaces` are not set in the production.yaml property.

In a more functional oriented testing, we want to make sure that we are to achieve the following steps: https://docs.astronomer.io/software/namespace-pools/ by using this new values in the helm chart in a simpler way.


requirements/chart-tests.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.9
+# This file is autogenerated by pip-compile with python 3.8
```
AmFlint, line 2: Oh my bad, will fix tomorrow, the venv created was using python 3.8, I'll pip-compile it again with python 3.9

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -93,3 +93,7 @@ spec:
             - name: COMMANDER_AIRGAPPED
               value: "true"
             {{- end }}
+            {{ if .Values.global.features.namespacePools.enabled }}
```
danielhoherd, line 96: Interesting... `.global.features` could be a great mechanic to abstract things behind, and account for which features we have.

AmFlint: Yeah, Constance has good ideas about this !

charts/astronomer/templates/commander/commander-role.yaml
```diff
@@ -1,19 +1,33 @@
...
 ## Astronomer Commander Role/ClusterRole
 ########################################
-{{- $singleNamespace := .Values.global.singleNamespace }}
-{{- $clusterRoles:= .Values.global.clusterRoles }}
-{{- if and .Values.global.rbacEnabled }}
-kind: {{ if or ($singleNamespace) (not $clusterRoles) }}Role{{ else }}ClusterRole{{ end }}
+# Here, we either:
+# 1. Do not create any resource if rbacEnabled is disabled
+# 2. Create a Cluster Role if clusterRoles and rbacEnabled are enabled and namespacePools is disabled
+# 3. Create roles for each namespaces in the namespacePool if enabled.
```
danielhoherd, line 7: I think it might be good to structure our tests around these three different scenarios, since they're so complicated, and hard to match which test case is mapping up to which entry in this list.

AmFlint: I didn't really know how to approach testing on these fields:
- Test the RBAC field with these 3 distinct scenarios 
- Or Test the whole feature, like `namespacePools` and all resources impacted by this feature.

In the end I decided to go with the second option.
- First and Second scenario are tested in a commander related tests.
- Third scenario is tested in the namespace pools feature test file

If you want me to change it I can do it, but I believed grouping tests with other ones related functional-wise was better, as long as all test scenarios are covered

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -71,6 +71,12 @@ data:
       kibanaUIEnabled: false
       {{ end }}
 
+      {{ if .Values.global.features.namespacePools.enabled }}
+      hardDeleteDeployment: true
+      manualNamespaceNames: true
+      preCreatedNamespaces:{{ printf "\n" }}{{ toYaml .Values.global.features.namespacePools.namespaces.names | indent 6 }}
+      {{ end }}
+
```
ajayy004, line 79: If it's not enable don't we want to disable to make sure its not enabled within the service.

ajayy004: @cmarteepants are we keeping manualReleaseNames and webhook configs at the service (houston) layer?

AmFlint: We don't want to create breaking changes with this improvement.

So, if namespacePools is disabled, we just don't touch these houston parameters.

Also, the values inside .Values.astronomer.houston.config that you mentionned are still injected in a separate configuration file `local-production.yaml` which is more important (will take priority over the production.yaml created here), so users can still override this "default" configuration if they want to.

charts/astronomer/templates/commander/commander-role.yaml
danielhoherd: My original comment was mostly from the difficulty I had reading the tests and knowing which configuration was being tested. So, even if the tests passed, I had a hard time discerning if the tests were valid and valuable. If anything, we should add more comments about what is being tested, or break up `test_astronomer_commander_rbac_cluster_role` into test cases for specific configurations and add docstrings for those.

AmFlint: Oh I understand, I'll split the tests to make them easier to understand

ajayy004 (approved):

tests/test_astronomer_namespace_pools.py
```diff
@@ -0,0 +1,241 @@
...
+            "charts/astronomer/templates/commander/commander-rolebinding.yaml",
+        ],
+    )
+
+    assert len(docs) == 4
+
+    # assertions on Role objects
+    for i in range(0, 2):
+        role = docs[i]
+
```
danielhoherd, line 42: This is more idiomatic:
```suggestion
    for role in docs:
```

tests/test_astronomer_namespace_pools.py
```diff
@@ -0,0 +1,241 @@
...
+                    }
+                }
+            }
+        },
+        show_only=["charts/astronomer/templates/namespaces.yaml"],
+    )
+
+    assert len(docs) == 2
+    for i in range(0, 2):
+        namespace = docs[i]
```
danielhoherd, line 92:
```suggestion
    for doc in docs:
```

danielhoherd (approved):
Very nice. I love the test work, it's really easy to follow!

tests/test_astronomer_namespace_pools.py
AmFlint: The thing is I created another list `namespaces` to compare the name of the namespaces to be created (line 45)

tests/test_astronomer_namespace_pools.py
AmFlint: Same here, I need to compare this namespace resource with namespace name in the `namespaces` list variable.

If you want I can change it like you mentioned and use 
```python
assert namespace["metadata"]["name"] in namespaces
```

tests/test_astronomer_namespace_pools.py
danielhoherd: Ah, yeah, I missed that.

PR #1411, opened 2022-03-16
run elasticsearch with non root

pgvishnuram:
## Description

Currently elasticsearch chart requires elevated privileges to run elasticsearch services , this changes removes that behaviour and allows customer either to install with less privileges or with allowed escalated privileges if needed.

* remove chown init container which requires root privileges
* add security context and remove privileged requirement for es services

## Related Issues

GH Issue: https://github.com/astronomer/issues/issues/4227

## Testing

QA should able to install platform with logging without any issues and also able upgrade them from existing installations without breaking.


pgvishnuram:
@danielhoherd @vishwas-astro - I have opened a draft PR.

Need to restructure the current test to validate the new change

tests/test_elasticsearch.py
```diff
@@ -79,22 +79,46 @@ def test_elasticsearch_with_sysctl_disabled(self, kube_version):
...
         assert len(docs) == 3
-        doc = docs[0]
 
         # elasticsearch master
+        doc = docs[0]
         assert doc["kind"] == "StatefulSet"
-        assert "sysctl" not in jmespath.search(
-            "spec.template.spec.initContainers[*].name", docs[0]
-        )
+        assert jmespath.search("spec.template.spec.initContainers", docs[0]) is None
```
danielhoherd, line 86: Shorter and maybe more legible:
```suggestion
        assert not docs[0]['spec']['template']['spec']['initContainers']
```

Or you could wrap all of these up in:

```
assert not any([doc['spec']['template']['spec']['initContainers'] for doc in docs])
```

tests/test_elasticsearch.py
```diff
@@ -79,22 +79,46 @@ def test_elasticsearch_with_sysctl_disabled(self, kube_version):
...
-            "spec.template.spec.initContainers[*].name", docs[1]
-        )
+        assert jmespath.search("spec.template.spec.initContainers", docs[1]) is None
 
         # elasticsearch client
         doc = docs[2]
         assert doc["kind"] == "Deployment"
         assert jmespath.search("spec.template.spec.initContainers", docs[2]) is None
+
+    def test_elasticsearch_securitycontext_defaults(self, kube_version):
```
danielhoherd, line 98: Because this test case is all defaults, it's basically the same as the first test case in this file, so you could take these assertions and put them into that "default" test case, and just comment on what is being tested, or do it without comments since the meaning is implied by the code.

tests/test_elasticsearch.py
```diff
@@ -79,22 +79,22 @@ def test_elasticsearch_with_sysctl_disabled(self, kube_version):
...
         assert len(docs) == 3
-        doc = docs[0]
-
-        # elasticsearch master
-        assert doc["kind"] == "StatefulSet"
-        assert "sysctl" not in jmespath.search(
-            "spec.template.spec.initContainers[*].name", docs[0]
-        )
+        for doc_ids in range(len(docs)):
+            assert not docs[doc_ids]["spec"]["template"]["spec"]["initContainers"]
```
danielhoherd, line 83:
```suggestion
        for doc in docs:
            assert not doc["spec"]["template"]["spec"]["initContainers"]
```

tests/test_elasticsearch.py
```diff
@@ -79,22 +79,22 @@ def test_elasticsearch_with_sysctl_disabled(self, kube_version):
...
+            ],
         )
-
-        # elasticsearch client
-        doc = docs[2]
-        assert doc["kind"] == "Deployment"
-        assert jmespath.search("spec.template.spec.initContainers", docs[2]) is None
+        assert len(docs) == 3
+        for doc_ids in range(len(docs)):
+            assert docs[doc_ids]["spec"]["template"]["spec"]["securityContext"] == {
```
danielhoherd, line 98:
```suggestion
        for doc in docs:
            assert doc["spec"]["template"]["spec"]["securityContext"] == {
```

tests/test_elasticsearch.py
```diff
@@ -67,7 +67,7 @@ def test_elasticsearch_with_sysctl_defaults(self, kube_version):
         )
 
     def test_elasticsearch_with_sysctl_disabled(self, kube_version):
-        """Test  ElasticSearch with sysctl config/values.yaml."""
+        """Test ElasticSearch master,data and client with sysctl config/values.yaml."""
```
danielhoherd, line 70:
```suggestion
        """Test ElasticSearch master, data and client with sysctl config/values.yaml."""
```

danielhoherd (changes requested):
Looks mostly good. Got a few suggestions, then we can merge.

pgvishnuram:
@vishwas-astro - this is merged to master now

PR #1413, opened 2022-03-16
A global flag to disable astronomer components network policies

shmanu017:
## Description

Disable all the network policies from global configuration.

## Related Issues

https://github.com/astronomer/issues/issues/4386

## Testing


pgvishnuram:
@vishwas-astro pls add a tag when are we shipping this option

vishwas-astro:
@cmarteepants - I am thinking this can be shipped as part of 0.29.0 in May. Can you confirm if Ford needs this sooner than that? If so, we can try and squeeze this into next week's 0.28.4 release. My preference is to not pile this on to 0.28.4 though.

tests/test_global_network_policy_flag.py
```diff
@@ -0,0 +1,55 @@
+from cgitb import enable
+from tests.helm_template_generator import render_chart
+import pytest
+import jmespath
+from . import git_root_dir
+import git
+
+# External-es-proxy and prometheus-postgres-exporter are set false by default,
+# needs additional work on creating test cases for future
+show_only = [
```
danielhoherd, line 6: A better approach than enumerating these files would be to render the entire chart, and then filter for `kind: NetworkPolicy`. We do this in at least one other place: <https://github.com/astronomer/astronomer/blob/e69786dbd92d1c32ab63b2c2634ca32d9624f50a/tests/test_default_chart.py#L60-L123>

Doing it this way means we do not have to update this list of NetworkPolicies change, any NetworkPolicies that are misplaced will be caught, and we only have to render the chart once, which resulted in a big speed improvement.

So you could render the full chart without passing `show_only` and then use `trimmed_docs = [x for x in docs if docs['kind'] == 'NetworkPolicy']` to get only the NetworkPolicy docs to check.

danielhoherd (approved):
For 0.29.0 code freeze this is good. My only open thread are nits about test optimizations.

tests/test_global_network_policy_flag.py
shmanu017:
```
show_only = [
     str(path.relative_to(git_root_dir))
     for path in git_root_dir.rglob("./**/*.yaml")
     if "networkpolicy" in path.name
 ]
```
 I tried that via filtering out `netowrkpolicy` in the path but cant figure out the way to filter out  `External-es-proxy` and `prometheus-postgres-exporter` in show only as it can't accept AND operator.
 
 Will try the other solution later this week.

PR #1414, opened 2022-03-16
Fix helm 3.8 compatibility in release 0.25 tests

danielhoherd:
## Description

- Fix helm 3.8 compatibility in release-0.25 tests
- Fix already broken lts 23-to-25 upgrader (big changes to this process)
- Bump requirements
- Use newer ci images
- Use newer kind versions

The core changes here, switching a few things to support the new helm 3.8 regex, are small and already tested in other branches.

A lot of the changes in this PR are to fix the broken lts-23-to-25 upgrade process. This process is handled through ansible, and there were some ansible errors related to the upgrade process even though the upgrader had not changed. In order to solve these upgrader problems, I upgraded ansible to the latest version, switched from alpine to debian, implemented pip-compile for version tracking on the upgrader packages, and installed postgres-client 12 from postgres apt repos. The third vendor-provided postgresql client was necessary because debian bullseye includes postgresql-client 13, which had a version of pg_dump that is incompatible with postgresql 13, which is what we use in the test environment. This is a red flag for customers deployments, and is something we should look into more, or at least that CRE should be aware of.

I really wish these two changes were not mixed up together, but I'm not sure if it's worth the effort to untangle them and submit them as two separate PRs. I can do so if we think that is necessary, or we can merge these two changes separately. The main reason they got tangled up is because the upgrader uses helm, and so it needed to be changed to support helm 3.8, which required fixing the lts upgrader build that had broken due to code rot.

## Related Issues

- <https://github.com/astronomer/issues/issues/4582>
- <https://github.com/astronomer/issues/issues/4124>

## Testing

CI tests passing should be enough. It wouldn't hurt to run the upgrader through some more tests before we release a new version, but merging this does not mean we need to release a new upgrader version, so we can take more time to verify that if we want to.

Also I tested on helm 3.8 locally and tests passed.

## Merging

This should **only be merged into release-0.25**. None of these changes are relevant to other branches, either because they do not apply, or because they have already been made.


AmFlint (approved):
good one

isaulv (approved):

PR #1417, opened 2022-03-21
nginx ingress controller enhancements

pgvishnuram:
## Description

* add electionID flag to nginx controller
* update values files for overrides

## Related Issues

Issue Link: https://github.com/astronomer/issues/issues/4392

## Testing

Need to update

### DIFF
--- a/charts/nginx/values.yaml
+++ b/charts/nginx/values.yaml
@@ -76,4 +76,4 @@ ports:
   metrics: 10254
 
 # Election ID to use for status update
-electionID: ingress-controller-leader
+electionId: ingress-controller-leader
--- a/charts/nginx/values.yaml
+++ b/charts/nginx/values.yaml
@@ -74,3 +74,6 @@ ports:
   defaultBackendHTTP: 8080
   healthz: 10254
   metrics: 10254
+
+# Election ID to use for status update
+electionID: ingress-controller-leader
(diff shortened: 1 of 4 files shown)


pgvishnuram:
self note: add test cases to validate

danielhoherd (changes requested):
This could use some tests, even if just to validate that the electionID value works as expected. These test cases could be included inside of other tests cases for defaults and custom configs.

charts/nginx/values.yaml
```diff
@@ -74,3 +74,6 @@ ports:
   defaultBackendHTTP: 8080
   healthz: 10254
   metrics: 10254
+
+# Election ID to use for status update
+electionID: ingress-controller-leader
```
danielhoherd, line 79: AFAIK camelCase has `election_id` as `electionId` <https://capitalizemytitle.com/camel-case/>
```suggestion
electionId: ingress-controller-leader
```

danielhoherd:
@pgvishnuram tests failed due to prometheus not being available, which is a known flaky test. Can you bump [test_prometheus_targets](https://github.com/astronomer/astronomer/blob/774cf23ae7/bin/functional-tests/test_config.py#L66) to `@pytest.mark.flaky(reruns=20, reruns_delay=10)` in this branch? Seems like it would be worth waiting a little bit longer rather than re-running the entire workflow.

pgvishnuram:
@danielhoherd - Yea added test case to validate election-id

PR #1423, opened 2022-03-28
Namespace Pools: Optimize fluentd's configuration

AmFlint:
## Description

When using the Namespace Pools feature, we want to optimize fluentd to only retrieve Airflow logs from the namespaces provided in the pool, since Airflow deployments will ONLY be deployed in these namespaces.

I have also added tests to look for this configuration under the different scenarios.

## Related Issues

https://github.com/astronomer/issues/issues/4322

## Testing

To test this PR, we want to render the helm chart and check that the fluentd configuration is set properly according to our global values.

We want to make sure that fluentd is still working as expected, we are going to create a helm release with namespace pool enabled, and provide a list of namespaces to use and to create, and then we will need to make sure that fluentd does collect the logs from our Airflow deployments created in these namespaces:
- First step: create the release, you need to set the following values (here, we're enabling the namespacePools feature, and want helm to create 2 namespaces to be used in the pool: `my-ns-1` and `my-ns-2`)
```yaml
global:
  features:
    namespacePools:
      enabled: true
      namespaces:
        create: true
        names:
          - my-ns-1
          - my-ns-2
```
- Wait for the release to be deployed
- Check that the fluentd configmap is configured properly, by reading the rendered configuration inside the cluster:
  ```bash
  kubectl get cm -n <release-namespace> fluentd-configmap -o yaml
  ```
  You should be looking for the following content inside the output of the previous command:
  ```
  <filter kubernetes.**>
      @type grep
      <regexp>
        key $.kubernetes.namespace_name
        pattern ^(my-ns-1|my-ns-2)$
      </regexp>
      <regexp>
        key $.kubernetes.labels.component
        pattern ^(scheduler|webserver|worker|triggerer)$
      </regexp>
    </filter>
  ```
- Create Airflow deployments in your Astronomer platform inside these 2 namespaces
- Check that Logs from the airflow components show up in ElasticSearch (or whatever, we need to check that fluentd collects logs from Airflow components in the namespaces of the pool).

### DIFF
--- a/tests/test_fluentd.py
+++ b/tests/test_fluentd.py
@@ -60,6 +60,7 @@ def test_fluentd_clusterrolebinding(kube_version):
 
     assert len(docs) == 0
 
+
 @pytest.mark.parametrize(
     "kube_version",
     supported_k8s_versions,
@@ -75,7 +76,7 @@ def test_fluentd_configmap_manual_namespaces_enabled(kube_version):
                     "namespacePools": {
                         "enabled": False,
                     }
-                }
+                },
             }
         },
         show_only=["charts/fluentd/templates/fluentd-configmap.yaml"],
@@ -83,6 +84,7 @@ def test_fluentd_configmap_manual_namespaces_enabled(kube_version):
 
     expected_rule = "key $.kubernetes.namespace_name\n    # fluentd should gather logs from all namespaces if manualNamespaceNamesEnabled is enabled"
     assert expected_rule in doc["data"]["output.conf"]
+
 
 @pytest.mark.parametrize(
     "kube_version",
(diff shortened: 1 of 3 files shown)


andriisoldatenko (approved):

tests/test_fluentd.py
```diff
@@ -59,3 +59,55 @@ def test_fluentd_clusterrolebinding(kube_version):
...
+    expected_rule = "key $.kubernetes.namespace_name\n    # fluentd should gather logs from all namespaces if manualNamespaceNamesEnabled is enabled"
+    assert expected_rule in doc["data"]["output.conf"]
+
+
+@pytest.mark.parametrize(
+    "kube_version",
+    supported_k8s_versions,
+)
+def test_fluentd_configmap_manual_namespaces_and_namespacepools_disabled(kube_version):
+    """Test that when namespace Pools and manualNamespaceNamesEnabled aredisabled, helm renders a default fluentd configmap looking at an environment variable"""
```
danielhoherd, line 94:
```suggestion
    """Test that when namespace Pools and manualNamespaceNamesEnabled are disabled, helm renders a default fluentd configmap looking at an environment variable"""
```

danielhoherd (approved):
LGTM. Just got a comment typo suggestion that isn't crucial.

PR #1425, opened 2022-03-28
add runtime check cronjob

andriisoldatenko:
https://github.com/astronomer/issues/issues/4057

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -193,6 +193,17 @@ houston:
     # Default here is to run at midnight every night
     schedule: "0 0 * * *"
 
+  # Check for Astronomer Runtime Updates
+  # This runs as a CronJob
+  updateRuntimeCheck:
+    # Enable check updates CronJob
+    enabled: true
+
+    # url for update service
+    url: https://updates.astronomer.io/astronomer-runtime
+
+    # Default here is to run at midnight every night
+    schedule: "0 0 * * *"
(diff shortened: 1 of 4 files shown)


danielhoherd (changes requested):
Looks good so far. Needs tests.

pgvishnuram:
Lets add some default resource limits too.

charts/astronomer/values.yaml
```diff
@@ -193,6 +193,17 @@ houston:
...
+  # Check for Astronomer Runtime Updates
+  # This runs as a CronJob
+  updateRuntimeCheck:
+    # Enable check updates CronJob
+    enabled: true
+
+    # url for update service
+    url: https://updates.astronomer.io/astronomer-runtime
+
+    # Default here is to run at midnight every night
```
danielhoherd, line 205: We should pick a different minute to run, in order to avoid coinciding with every other job on the same node that runs on the hour. This will spread system load out more, reducing resource contention. Even 3 after would be better. Or 27 after. Or 43 after.

danielhoherd (changes requested):
Tests look good. I left a comment about changing the cronjob time to an odd minute. Also I agree with @pgvishnuram that we should add resource limits and requests.

charts/astronomer/values.yaml
andriisoldatenko: fixed.

danielhoherd (approved):

PR #1427, opened 2022-03-28
Remove keda

danielhoherd:
## Description

Remove keda from astronomer/astronomer helm chart.

## Related Issues

https://github.com/astronomer/issues/issues/4441

## Testing

The only testing I've done are the regular CI tests, which should be enough IMHO.


pgvishnuram:
@danielhoherd  Should we add steps or point how to integrate keda with astronomer if we are removing them completely

danielhoherd:
> @danielhoherd Should we add steps or point how to integrate keda with astronomer if we are removing them completely

I'm counting on the docs team to handle anything related to that.

cmarteepants (approved):

rbankston (approved):
Those changes look good to me.

jwitz:
Is this being introduced alongside a Software release, or is this already out?

jwitz:
As far as I know, the only non-release note docs we would have to update are any code configurations that mention keda as a small part of a larger config ([example](https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/google/cloud/example_dags/example_gcs.py))

jwitz:
Whoops wrong link! Meant here: https://docs.astronomer.io/software/namespace-pools#step-1-configure-namespaces

PR #1430, opened 2022-03-30
Add redirect section in registry configmap

neel-astro:
## Description
Changes:
- Added redirect section under storage in registry config map (Default value for `redirectDisabled` is set to False, same as what docker registry treats it in case it is not passed)

Currently, the astronomer registry will use redirection whenever the backend of the registry supports it. This PR is to allow users to disable registry redirection and pass all the data through the registry, this use case is useful in a couple of scenarios, one of them is when the backend storage for instance does not allow any other connection to push objects except the registry IP.

Reference: https://docs.docker.com/registry/configuration/#redirect

## Related Issues
https://github.com/astronomer/issues/issues/4183

## Testing
In order to test whether this change fixes the mentioned issue, one has to bring up an Astronomer cluster on AWS with a registry backend configured as an S3 bucket, with a similar bucket policy:

```JSON
{
    "Version": "2012-10-17",
    "Id": "Policy1648101441413",
    "Statement": [
        {
            "Sid": "Stmt1648101431372",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:*",
            "Resource": [
                "arn:aws:s3:::neel-dev-test-registry",
                "arn:aws:s3:::neel-dev-test-registry/*"
            ],
            "Condition": {
                "StringNotLike": {
                    "aws:userid": "AKxxxxxxxxx",
                    "aws:username": "neel"
                },
                "NotIpAddress": {
                    "aws:VpcSourceIp": [
                        "10.10.0.0/16",
                        "10.100.0.0/16"
                    ]
                },
                "StringNotEquals": {
                    "aws:username": "neel"
                }
            }
        }
    ]
}
```

When the platform is brought up without disabling the redirection, then the registry configmap will look something like:
Screenshot 2022-03-30 at 1 19 30 PM
(AWS user used in this screenshot has already been deleted)

And then when trying to push an image from local via `astro deploy` one will face the following error:
Screenshot 2022-03-30 at 1 19 11 PM

Same scenario when registry redirection is disabled by setting `redirect.disable` under registry in values.yaml as True and the platform is recreated, i.e.:
```YAML
  registry:
    replicas: 1
    redirect:
      disable: true
    s3:
      enabled: true
      bucket: neel-dev-test-registry
      region: us-east-1
```
Then one should be able to successfully push the image to the registry:
Screenshot 2022-03-30 at 12 46 47 PM

### DIFF
--- a/charts/astronomer/templates/registry/registry-configmap.yaml
+++ b/charts/astronomer/templates/registry/registry-configmap.yaml
@@ -35,6 +35,10 @@ data:
       filesystem:
         rootdirectory: /var/lib/registry
     {{- end }}
+    {{- if .Values.registry.redirectDisabled }}
+      redirect:
+        disable: {{ .Values.registry.redirectDisabled }}
+    {{- end }}
       delete:
         enabled: true
     http:
--- a/charts/astronomer/templates/registry/registry-configmap.yaml
+++ b/charts/astronomer/templates/registry/registry-configmap.yaml
@@ -35,9 +35,9 @@ data:
       filesystem:
         rootdirectory: /var/lib/registry
     {{- end }}
-    {{- if .Values.registry.redirectDisabled }}
+    {{- if (.Values.registry.redirect).disable }}
       redirect:
-        disable: {{ .Values.registry.redirectDisabled }}
+        disable: {{ .Values.registry.redirect.disable }}
     {{- end }}
       delete:
         enabled: true
--- a/charts/astronomer/templates/registry/registry-configmap.yaml
+++ b/charts/astronomer/templates/registry/registry-configmap.yaml
@@ -35,7 +35,7 @@ data:
       filesystem:
         rootdirectory: /var/lib/registry
     {{- end }}
-    {{- if (.Values.registry.redirect).disable }}
+    {{- if .Values.registry.redirect.disable }}
       redirect:
         disable: {{ .Values.registry.redirect.disable }}
     {{- end }}
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -310,6 +310,8 @@ registry:
   notifications:
     timeout: 30s
 
+  redirectDisabled: false
+
 install:
   resources: {}
   cliVersion: 0.27.0
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -310,7 +310,8 @@ registry:
   notifications:
     timeout: 30s
 
-  redirectDisabled: false
+  redirect:
+    disable: false
 
 install:
   resources: {}
--- a/tests/test_astronomer_registry_configmap.py
+++ b/tests/test_astronomer_registry_configmap.py
@@ -9,7 +9,7 @@ import re
     "kube_version",
     supported_k8s_versions,
 )
-def test_astronomer_registry_configmap(kube_version):
+def test_astronomer_registry_configmap_defaults(kube_version):
     """Test that helm renders a good configmap template for astronomer registry."""
     docs = render_chart(
         kube_version=kube_version,
@@ -24,14 +24,15 @@ def test_astronomer_registry_configmap(kube_version):
     assert doc["kind"] == "ConfigMap"
     assert doc["apiVersion"] == "v1"
     assert bool(re.match("[+]?\\d+s", timeout))
+    assert "redirect" not in parsed_config_yml["storage"]
 
 
 @pytest.mark.parametrize(
     "kube_version",
     supported_k8s_versions,
 )
-def test_astronomer_registry_redirect(kube_version):
-    """Test that helm renders redirect section in configmap of astronomer registry."""
+def test_astronomer_registry_redirect_disabled(kube_version):
+    """Test that helm renders astronomer registry configmap template with redirect disabled."""
     docs = render_chart(
         kube_version=kube_version,
         values={"astronomer": {"registry": {"redirect": {"disable": True}}}},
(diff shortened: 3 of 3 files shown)


pgvishnuram:
@neel-astro redirectDisabled we need to add in values.yaml in charts/astronomer/values.yml

neel-astro:
@pgvishnuram added `redirectDisabled` in values.yaml

charts/astronomer/templates/registry/registry-configmap.yaml
```diff
@@ -34,6 +34,10 @@ data:
     {{- else }}
       filesystem:
         rootdirectory: /var/lib/registry
+    {{- end }}
+    {{- if .Values.registry.redirectDisabled }}
```
danielhoherd, line 38: Should this be `.Values.registry.redirect.enabled` ? That seems more consistent with the data structure and naming that we use elsewhere.

neel-astro: That was my initial instinct, but will that confuse the user as values will have `redirect.enabled` but the registry config map will have it as `redirect.disable`?

danielhoherd: @neel-astro i'm missing some context. where is this `redirect.disable` in the registry config map? If possible we should also change that to `enabled`. `disabled = False` is a double negative, which is harder to grok.

https://schneide.blog/2014/08/03/dont-ever-not-avoid-negative-logic/

neel-astro: By `redirect.disable` I meant `disable` under `redirect`section in yaml ([code reference](https://github.com/astronomer/astronomer/pull/1430/files#diff-087010573e7ff5d6f9d72b6fc63bf83256d1e52d9e86611fa177bdee5da84cafR40)).
We won't be able to change that to `enable` since that is defined on the docker registry side (Reference: https://docs.docker.com/registry/configuration/#redirect)

danielhoherd: SMH. ok, well it would be better for us to keep what they are using. i think we should use `redirect.disable` though instead of `redirectDisabled`. what do you think?

neel-astro: sure `redirect.disable` feels better than `redirectDisabled`, I will make the required changes.

charts/astronomer/templates/registry/registry-configmap.yaml
```diff
@@ -34,6 +34,10 @@ data:
     {{- else }}
       filesystem:
         rootdirectory: /var/lib/registry
+    {{- end }}
+    {{- if (.Values.registry.redirect).disable }}
```
AmFlint, line 38: Why do you use `()` here ?

Also, I can see in many places in the Chart we have `feature.enabled`, I think it would be more consistent to change .disable to .enabled and set default to true

charts/astronomer/values.yaml
```diff
@@ -310,6 +310,9 @@ registry:
   notifications:
     timeout: 30s
 
+  redirect:
+    disable: false
```
AmFlint, line 314: see previous comment

charts/astronomer/templates/registry/registry-configmap.yaml
andriisoldatenko, line 38: why do we need brackets here? does it work without?

tests/test_astronomer_registry_configmap.py
```diff
@@ -24,3 +24,24 @@ def test_astronomer_registry_configmap(kube_version):
     assert doc["kind"] == "ConfigMap"
     assert doc["apiVersion"] == "v1"
     assert bool(re.match("[+]?\\d+s", timeout))
+
+
+@pytest.mark.parametrize(
+    "kube_version",
+    supported_k8s_versions,
+)
+def test_astronomer_registry_redirect(kube_version):
```
andriisoldatenko, line 33: does it make sense to add alternative test?

danielhoherd: @andriisoldatenko what are you proposing? There is only one other test case in this file, so it seems to me that having this test case that tests `registry.redirect.disable=True` makes sense.

charts/astronomer/templates/registry/registry-configmap.yaml
neel-astro: I think we can remove the brackets since there is a default value set (will try out the scenario), but the original idea was not to avoid a panic when there is no `redirect` section in values.yaml file itself.

neel-astro: removed the brackets

tests/test_astronomer_registry_configmap.py
```diff
@@ -24,3 +24,24 @@ def test_astronomer_registry_configmap(kube_version):
...
+        show_only=["charts/astronomer/templates/registry/registry-configmap.yaml"],
+    )
+
+    assert len(docs) == 1
+    doc = docs[0]
+    config_yml = doc["data"]["config.yml"]
+    parsed_config_yml = yaml.safe_load(config_yml)
+    redirectDisabled = parsed_config_yml["storage"]["redirect"]["disable"]
+    assert doc["kind"] == "ConfigMap"
+    assert redirectDisabled
```
danielhoherd, line 47: This seems more organized, and AFAIK it will work the same.

```suggestion
    config_yml = doc["data"]["config.yml"]
    assert doc["kind"] == "ConfigMap"

    parsed_config_yml = yaml.safe_load(config_yml)
    assert = parsed_config_yml["storage"]["redirect"]["disable"]
```

tests/test_astronomer_registry_configmap.py
danielhoherd: @neel-astro I think maybe what @andriisoldatenko is suggesting is we add a test with registry redirect enabled, but also disabled. Usually in a case like this we'll add the assertion for the default value in a more conglomerate test with a bunch of other defaults. So maybe add add the `partsed_config_yml` bit and assert False in the `test_astronomer_registry_configmap` test, which should probably be renamed to `test_astronomer_registry_configmap_defaults` or something.

neel-astro:
Have updated the PR with the recommendation, can we have a review again. Thanks for patience
@danielhoherd @andriisoldatenko

danielhoherd (approved):
LGTM!

PR #1435, opened 2022-03-31
Adding Docker Image Pull Test

mishah334:
## Description

Adding the docker image pull test.  This test ensures that the all declared docker images are pullable.

## Related Issues

https://github.com/astronomer/issues/issues/3968

## Testing

This change is about to add the test case.


tests/test_docker_images.py
```diff
@@ -0,0 +1,39 @@
+import docker
+from tests.helm_template_generator import render_chart
+import jmespath
+
+
+def test_docker_images(docker_client):
```
andriisoldatenko, line 6: wdyt to refactor it to pytest parametrized? so you can see each test for each image rather than doing for loop.

tests/test_docker_images.py
```diff
@@ -0,0 +1,39 @@
...
+    search_string = "spec.template.spec.containers[*].image"
+    docker_images = []
+    for public_repo_doc in public_repo_docs:
+        docker_image_sublist = jmespath.search(search_string, public_repo_doc)
+
+        if docker_image_sublist is not None:
+            docker_images = docker_image_sublist + docker_images
+
+    # Pulling docker image test
+    for docker_image in docker_images:
```
andriisoldatenko, line 29: pls let me know if you need any help.

andriisoldatenko (approved):

tests/test_docker_images.py
```diff
@@ -0,0 +1,39 @@
...
+    return docker_images
+
+
+# Pulling docker image test
+@pytest.mark.parametrize("docker_image", list_docker_images())
+def test_docker_image(docker_client, docker_image):
+    docker_image = docker_image.replace('"', "").strip()
+    try:
+        # Pulling docker image
+        docker_client.images.pull(docker_image)
```
danielhoherd, line 37: @mishah334 it looks like this pulls the image. it would be awesome if we could find a way to validate the image without having to pull it. pulling all of these images will add a ton of network consumption, so it would be great to avoid it if possible.

tests/test_docker_images.py
```diff
@@ -0,0 +1,41 @@
...
+    return docker_images
+
+
+# Pulling docker image test
+@pytest.mark.parametrize("docker_image", list_docker_images())
+def test_docker_image(docker_client, docker_image):
+    docker_image = docker_image.replace('"', "").strip()
+    try:
+        # Pulling docker image
+        docker_client.images.get(docker_image)
```
mishah334, line 37: @danielhoherd The call `image->get` will not pull but verify whether image and tag is exists or not.

tests/test_docker_images.py
```diff
@@ -0,0 +1,37 @@
...
+            docker_images = docker_image_sublist + docker_images
+
+    return docker_images
+
+
+@pytest.mark.parametrize("docker_image", list_docker_images())
+def test_docker_image(docker_client, docker_image):
+    docker_image = docker_image.replace('"', "").strip()
+    try:
+        docker_client.images.get_registry_data(docker_image)
```
mishah334, line 35: Again replaced the `image->get` to `image->get_registry_data`. The `image->get` works for only local images. The  build was working fine for sometime when switched to `image->get` but started failing after may be because of caching on Circle CI side.

tests/test_docker_images.py
danielhoherd: Nice!

tests/test_docker_images.py
danielhoherd: This was done well

```
tests/test_docker_images.py::test_docker_image[quay.io/astronomer/ap-nats-server:2.7.2-1]
```

tests/test_docker_images.py
```diff
@@ -0,0 +1,37 @@
...
+    for public_repo_doc in public_repo_docs:
+        docker_image_sublist = jmespath.search(search_string, public_repo_doc)
+
+        if docker_image_sublist is not None:
+            docker_images = docker_image_sublist + docker_images
+
+    return docker_images
+
+
+@pytest.mark.parametrize("docker_image", list_docker_images())
```
danielhoherd, line 31: I think we're going to have to mark this test as flaky. I've gotten a few 500 errors when testing:

`E           AssertionError: 'Error reading image: quay.io/astronomer/ap-houston-api:0.28.11 | Error: 500 Server Error for http+docker://localhost/v1.41/distribution/quay.io/astronomer/ap-houston-api:0.28.11/json: Internal Server Error ("Get "https://quay.io/v2/": dial tcp [2600:1f18:483:cf00:1a24:60a6:800:1de7]:443: connect: cannot assign requested address")`

```suggestion
@pytest.mark.flaky(reruns=5, reruns_delay=1)
@pytest.mark.parametrize("docker_image", list_docker_images())
```

tests/test_docker_images.py
danielhoherd: Trying to test this, I see that it requires the docker daemon to be running. It would be great if we could find a way to skip these tests if docker daemon is not present.

I'm thinking something like:

```python
try:
    docker.from_env().ping()
    docker_daemon_present = True
except (docker.errors.DockerException):
    docker_daemon_present = False
```

and then on the test:

```
@pytest.mark.skipif(not docker_daemon_present, reason="Docker daemon not available")
```

tests/test_docker_images.py
danielhoherd: Looks like the above change will require adding `pytest-rerunfailures` to `requirements/chart-tests.in` and then running `make update-requirements`

danielhoherd (changes requested):
This is looking great! I have some feedback that I think will wrap this up.

danielhoherd:
Hah, we just experienced a failure in the wild where this feature would have caught it at the first stage of the CI job. <https://app.circleci.com/pipelines/github/astronomer/astronomer/4228/workflows/1a7f7d4e-d40b-4a74-8f3f-67ebbc910861/jobs/28160>

tests/test_docker_images.py
mishah334: Done

tests/test_docker_images.py
mishah334: Done

danielhoherd (approved):
Looks great!

I ran into a bug with pytest-sugar while verifying this and opened up a PR to fix it so hopefully they merge that soon. https://github.com/Teemu/pytest-sugar/pull/234

PR #1437, opened 2022-03-31
bump commander to 0.28.3

AmFlint:
## Description

Update commander to version 0.28.3


pgvishnuram:
@vishwas-astro @danielhoherd 0.28.3 will be shipping 1.4.0 of astronomer/airflow chart

danielhoherd (approved):

danielhoherd:
@pgvishnuram yeah. AFAIK 1.4.0 has support for k8s 1.22, but does not require it. however afaik it drops support for 1.18 because of the ingress changes.

PR #1448, opened 2022-04-04
CI: check commander's airflow chart version matches astronomer/astronomer's

AmFlint:
## Description

We want to check the Airflow Chart version used in Commander when making changes to astronomer/astronomer (see the issue for more details).

In this PR, I added the job in CI that inspects commander's image to retrieve the airflow chart version used, and compare it with the default version set in astronomer/astronomer.

This will allow us to get notified when there's a mismatch of Airflow Chart Version between Commander and astronomer/astronomer before shipping a release.

## Related Issues

https://github.com/astronomer/issues/issues/4369

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,6 +164,16 @@ jobs:
             kubent --cluster=false --helm2=false --helm3=false --filename -
           when: always
 
+  check-commander-airflow-version:
+    docker:
+      - image: cimg/base:2022.03
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Check that commander image uses same Airflow chart version
+          command: make validate-commander-airflow-version
+
 workflows:
   version: 2.1
   build-and-release-helm-chart:
@@ -171,6 +181,7 @@ workflows:
       - lint
       - run_pre_commit
       - unittest-charts
+      - check-commander-airflow-version
       - build-artifact:
           requires:
             - lint
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -131,6 +131,16 @@ jobs:
           when: always
 {%- endfor %}
 
+  check-commander-airflow-version:
+    docker:
+      - image: cimg/base:2022.03
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Check that commander image uses same Airflow chart version
+          command: make validate-commander-airflow-version
+
 workflows:
   version: 2.1
   build-and-release-helm-chart:
@@ -138,6 +148,7 @@ workflows:
       - lint
       - run_pre_commit
       - unittest-charts
+      - check-commander-airflow-version
       - build-artifact:
           requires:
             - lint
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,10 @@ unittest-requirements: .unittest-requirements ## Setup venv required for unit te
 unittest-charts: .unittest-requirements ## Unittest the Astronomer helm chart
 	venv/bin/python -m pytest -n auto tests
 
+.PHONY: validate-commander-airflow-version
+validate-commander-airflow-version:
+	./bin/validate_commander_airflow_version
+
 .PHONY: lint-charts
 lint-charts: lint-prep ## Lint Astronomer sub-charts
 	# Check that nothing accidentally is using release name instead of namespace for metadata.namespace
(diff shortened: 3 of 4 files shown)


.circleci/config.yml
```diff
@@ -164,13 +164,25 @@ jobs:
             kubent --cluster=false --helm2=false --helm3=false --filename -
           when: always
 
+  check-commander-airflow-version:
+    docker:
+      # https://hub.docker.com/_/docker?tab=tags
+      - image: cimg/base:2021.04
```
danielhoherd, line 170: Any reason why we're using a year old tag? Could we use `2022.03` ?

.circleci/config.yml.j2
```diff
@@ -131,13 +131,25 @@ jobs:
           when: always
 {%- endfor %}
 
+  check-commander-airflow-version:
+    docker:
+      # https://hub.docker.com/_/docker?tab=tags
+      - image: cimg/base:2021.04
```
danielhoherd, line 137: Any reason why we're using a year old tag? Could we use `2022.03` ?

https://circleci.com/developer/images/image/cimg/base#image-tags

.circleci/config.yml.j2
```diff
@@ -131,13 +131,25 @@ jobs:
...
+          command: ./bin/validate_commander_airflow_version
+
 workflows:
   version: 2.1
   build-and-release-helm-chart:
     jobs:
       - lint
       - run_pre_commit
       - unittest-charts
+      - check-commander-airflow-version
```
danielhoherd, line 152: Given that this is going to stop the CI workflow if it fails, could this be implemented as a pytest test case? Or could this shell script be included as part of the `make unittest-charts` target?

pgvishnuram: we can adding it to makefile so that QA can also make a quick validation.
Pytest way seems good so all test cases are handled in single place

.circleci/config.yml.j2
AmFlint: Took it from CircleCI docs, I'll update to latest one, thanks for that catch

.circleci/config.yml.j2
AmFlint: I thought about making it as a test case, but I thought about the following:
- Somebody makes a change to another completely independent part of the chart, let's say FluentD's subchart
- Test run, and they fail because Airflow chart version mismatches in Commander and Astronomer
- Might create confusion for the user, and the reviewers. I believe the reaction should be "well, we'll create another PR to bump the airflow chart version either in Commander or in Astronomer". But the "tests" CI job will still fail until we do, leading to confusion.
- By creating a new job for this specific purpose, tests can still pass if versions mismatch, no confusion for devs + reviewers, we can decide to ignore this fail and create a second PR to bump up version (or do it in Commander).
- Another point that motivated this decision to split it up into a new job, is that we also have contributors from the community, who don't have access to commander's repository, so they can't even make the second PR to fix airflow chart version. So I wanted this check to be independant from the chart tests that the user might update/contribute to.

I'll create a Make command to run this script, to allow QA folks to test it easily. What do you think, do you want me to move it to a Pytest test case anyway ?

andriisoldatenko: +1 to add make subcommand and trigger it during `make unittest-charts` via dependency.

andriisoldatenko: regarding pytest not sure it will slow down chart testing.

AmFlint: I'm Ok to add it to `make unittest-charts` as a dependency, but the "problem" is the same one I've cited in my previous comment -> the CI step `test` will fail, which leads to confusion.

In my opinion, I prefer to leave it as a separate make command (I've pushed `make validate-commander-airflow-version`) so that there's a different CI job/step running ONLY for this purpose, and it doesn't create confusion "like why are the tests failing" ?

danielhoherd: i think a good middle ground is to have a make target for `make test-commander-version` and have a new target `test: test-commander-version unittest-charts`

danielhoherd (approved):
This LGTM. I think we can merge it as is. We can always circle back on it to improve it if we're not happy with these exact changes, but this does add value and function as is, so I think it's good to go.

.circleci/config.yml.j2
AmFlint: I added a new Make target `test` that uses `validate-commander-airlfow-version` and `unittest-charts` as described here.

danielhoherd (approved):

PR #1450, opened 2022-04-05
Feature/AP-4177 Functional Test - Container running as non-root user

mishah334:
## Description

- Adding functional test that is testing the containers for non-root user.
- Moving the functional test and chart-tests together. This makes many files moving.

## Related Issues

https://github.com/astronomer/issues/issues/4177

## Testing

This PR is about to add the functional test.


tests/functional_tests/test_container_no_root.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+
+import pytest
+import testinfra
+
+from tests.functional_tests.conftest import get_pod_running_containers
+
+container_ignore_list = ["astronomer-kubed"]
```
mishah334, line 8: We have this single container that is currently not supporting execution.

tests/functional_tests/conftest.py
```diff
@@ -90,13 +91,20 @@ def kube_client(request, in_cluster=False):
     Return a kubernetes client. By default, use kube-config. If running in a pod, use k8s service account.
     """
 
+    k8s_clinet = get_kube_client(in_cluster)
+    yield k8s_clinet
+
+
+def get_kube_client(in_cluster=False):
```
mishah334, line 98: This change is required, as we can not call the pytest fixture directly.

tests/functional_tests/conftest.py
```diff
@@ -108,3 +116,21 @@ def get_pod_by_label_selector(kube_client, label_selector, namespace=namespace)
...
+
+def get_pod_running_containers(namespace=namespace):
+    """Return the containers from pods found."""
+    pods = get_kube_client().list_namespaced_pod(namespace).items
+
+    containers = []
+    for pod in pods:
+        pod_name = pod.metadata.name
+        for container_status in pod.status.container_statuses:
+            if container_status.ready:
```
mishah334, line 129: We test those containers, which are in `ready` state.

tests/functional_tests/conftest.py
mishah334: Done

danielhoherd (approved):
Looks good! I love the reorganization.

tests/functional_tests/test_container_no_root.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.mark.parametrize("container", get_pod_running_containers())
```
danielhoherd, line 11: The way the id's work here, we currently see `test_container_non_root[container6]`. It would be great if we could get the name of the container instead of the index. I know this is possible, but I'm not sure what the exact syntax would be. We don't have to fix this now, and can circle back and fix it after this first iteration, if that makes it easier to get forward progress.

danielhoherd: I figured it out. <https://app.circleci.com/pipelines/github/astronomer/astronomer/4270/workflows/45ee2365-42ff-46cb-b71b-4d80b4b92ae3/jobs/28388/steps>

```
=========================== short test summary info ============================
FAILED tests/functional_tests/test_container_no_root.py::test_container_non_root[astronomer-elasticsearch-exporter-59b6fdb7c8-hkvkl]
FAILED tests/functional_tests/test_container_no_root.py::test_container_non_root[astronomer-fluentd-mtvct]
FAILED tests/functional_tests/test_container_no_root.py::test_container_non_root[astronomer-houston-5bb74656bd-prpmf]
FAILED tests/functional_tests/test_container_no_root.py::test_container_non_root[astronomer-kube-state-b75f44c4d-qbh68]
```

```suggestion
containers = get_pod_running_containers()
container_ids = ids = [x["pod_name"] for x in containers]

@pytest.mark.parametrize("container", containers, ids=container_ids)
```

tests/functional_tests/conftest.py
mishah334: Done

tests/functional_tests/conftest.py
```diff
@@ -90,21 +91,48 @@ def kube_client(request, in_cluster=False):
...
+                container_name = container_status.name
+                containers[container_name] = {
+                    "container": vars(container_status).copy(),
+                    "namespace": pod_namespace,
+                    "pod_name": pod_name,
+                }
```
mishah334, line 136: @danielhoherd Now we return dict instead of the list. That will help to get ids in pytest parameters.

tests/functional_tests/test_container_no_root.py
```diff
@@ -0,0 +1,32 @@
...
+
+from tests.functional_tests.conftest import get_pod_running_containers
+
+container_ignore_list = ["astronomer-kubed"]
+
+
+@pytest.mark.parametrize(
+    "container_info",
+    get_pod_running_containers().values(),
+    ids=get_pod_running_containers().keys(),
```
mishah334, line 14: Workaround to get proper ids here. Example build: https://app.circleci.com/pipelines/github/astronomer/astronomer/4273/workflows/2d5e8308-5408-4d25-be48-aa9383635917/jobs/28405

danielhoherd:
I think we're going to have to add a permanent exception in here for fluentd because that DaemonSet requires root privileges in order to read the files it consumes.

tests/functional_tests/test_container_no_root.py
danielhoherd: The way this is written, it calls `get_pod_running_containers()` twice, which causes the kubernetes client to query kube api for the list of pods twice, and the result may be different. It would be more efficient and more accurate to query kube api once and generate two lists from that one query result. We could try to get creative, but i think the easiest way to do this is to just make the variables outside the function call.

mishah334:
> I think we're going to have to add a permanent exception in here for fluentd because that DaemonSet requires root privileges in order to read the files it consumes.

Added those container into ignore list for now.

tests/functional_tests/test_container_no_root.py
```diff
@@ -5,7 +5,7 @@
 
 from tests.functional_tests.conftest import get_pod_running_containers
 
-container_ignore_list = ["astronomer-kubed", "kube-state"]
+container_ignore_list = ["astronomer-kubed", "kube-state", "houston", "flentd", "metrics-exporter"]
```
danielhoherd:
```suggestion
container_ignore_list = ["astronomer-kubed", "kube-state", "houston", "fluentd", "metrics-exporter"]
```

danielhoherd (approved):
Good stuff!

PR #1464, opened 2022-04-13
Remove wasteful, unhelpful test case. Remove docker fixture autouse=true

danielhoherd:
## Description

Removed a wasteful test case that rendered every template in the repo (as of when we first started writing pytest cases) one at a time to test each individual template for the count of how many docs was returned. Now that we have much more broad coverage I think it is time we retired this wasteful and mostly unhelpful test. This cut ~20s (~22%) of test time on my M1 pro. As part of this I refactored `test_default_chart_with_basedomain` to be parametrized, so the test cases are now named like:

```
 tests/test_default_chart.py::TestAllPodSpecContainers.test_default_chart_with_basedomain[StatefulSet/release-name-prometheus] ✓
 tests/test_default_chart.py::TestAllPodSpecContainers.test_default_chart_with_basedomain[StatefulSet/release-name-stan] ✓
 tests/test_default_chart.py::TestAllPodSpecContainers.test_all_default_charts_with_private_registry[DaemonSet/release-name-fluentd] ✓
 tests/test_default_chart.py::TestAllPodSpecContainers.test_all_default_charts_with_private_registry[DaemonSet/release-name-prometheus-node-exporter] ✓
```

I also removed `autouse=True` on the docker fixture because we don't want it to be used everywhere, and having that was breaking tests on my laptop that does not have docker available.

## Related Issues

N/A

## Testing

This is testing, so nothing more needs to be done. This is actually a reduction in testing, though I think a lot of this was redundant.

## Merging

This should be fine to merge into any branches.


mishah334 (approved):
LGTM

PR #1466, opened 2022-04-14
fix loadBalancerIPs to loadBalancerIP

rbankston:
## Description

Update values typo introduced in pull #1160

## Related Issues

closes https://github.com/astronomer/issues/issues/4522

## Testing

Verified that  loadBalancerIP was used in the code but incorrectly identified in the values file

### DIFF
--- a/charts/nginx/values.yaml
+++ b/charts/nginx/values.yaml
@@ -47,7 +47,7 @@ ingressAnnotations: {}
 antiAffinity: "soft"
 
 # String IP address the nginx ingress should bind to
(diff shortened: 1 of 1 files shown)


vishwas-astro (approved):

danielhoherd (approved):
LGTM. The templates were not affected by this so this simple change is good. There are tests that run against `loadBalancerIP` and they have continued to pass <https://github.com/astronomer/astronomer/blob/cba1485d13/tests/test_nginx_service.py#L37-L60>

PR #1470, opened 2022-04-19
Use ap-auth-sidecar

pgvishnuram:
## Description

Replace nginx unprivileged image with ap-authsidecar 

## Related Issues

NA

## Testing

NA

## Merging

merge this change to master

### DIFF
--- a/tests/chart_tests/test_authsidecar.py
+++ b/tests/chart_tests/test_authsidecar.py
@@ -130,8 +130,8 @@ class TestAuthSidecar:
                 "global": {
                     "authSidecar": {
                         "enabled": True,
-                        "repository": "quay.io/astronomer/ap-auth-sidecar",
-                        "tag": "1.21.6-2",
+                        "repository": "someregistry.io/my-custom-image",
+                        "tag": "my-custom-tag",
                     }
                 }
             },
@@ -150,8 +150,8 @@ class TestAuthSidecar:
         prod = yaml.safe_load(doc["data"]["production.yaml"])
         expected_output = {
             "enabled": True,
-            "repository": "quay.io/astronomer/ap-auth-sidecar",
-            "tag": "1.21.6-2",
+            "repository": "someregistry.io/my-custom-image",
+            "tag": "my-custom-tag",
             "port": 8084,
             "pullPolicy": "IfNotPresent",
             "annotations": {},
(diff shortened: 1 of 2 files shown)


andriisoldatenko (approved):

pgvishnuram:
@vishwas-astro

tests/chart_tests/test_authsidecar.py
```diff
@@ -156,7 +164,11 @@ def test_authSidecar_houston_configmap_with_annotation(self, kube_version):
             kube_version=kube_version,
             values={
                 "global": {
-                    "authSidecar": {"enabled": True},
+                    "authSidecar": {
+                        "enabled": True,
+                        "repository": "nginxinc/nginx-unprivileged",
```
danielhoherd, line 169: If we're going to set this then there should be an assertion against it.

pgvishnuram: okay will make those changes

tests/chart_tests/test_authsidecar.py
```diff
@@ -126,7 +126,15 @@ def test_authSidecar_houston_configmap_without_annotation(self, kube_version):
         """Test Houston Configmap with authSidecar."""
         docs = render_chart(
             kube_version=kube_version,
-            values={"global": {"authSidecar": {"enabled": True}}},
+            values={
+                "global": {
+                    "authSidecar": {
+                        "enabled": True,
+                        "repository": "quay.io/astronomer/ap-auth-sidecar",
```
danielhoherd, line 133: Can we make this something that is nothing like what we would actually use so when it fails it's clear that the failure isn't related to the defaults?

danielhoherd, line 133:
```suggestion
                        "repository": "some/custom/repository",
```

tests/chart_tests/test_authsidecar.py
```diff
@@ -126,7 +126,15 @@ def test_authSidecar_houston_configmap_without_annotation(self, kube_version):
...
+                        "tag": "1.21.6-2",
```
danielhoherd, line 134:
```suggestion
                        "tag": "custom-tag",
```

rob-1126 (approved):

rob-1126:
@pgvishnuram I've got a similar PR open at https://github.com/astronomer/astronomer/pull/1337 - I'm OK with this one but that one fixes a bug that will prevent users with custom repositories from over-riding the image name that probably would be useful a number of places in our chart. It certainly won't keep me up at night if you decide to go with this one.

danielhoherd (approved):

PR #1475, opened 2022-04-22
fix kubed performance in large clusters: scope to release ns

AmFlint:
## Description

Use a configuration flag on KubeD Deployment to scope it to Astronomer release namespace. Should fix performance issues and latency before synchronizing secrets.

## Related Issues

https://github.com/astronomer/issues/issues/4546

### DIFF
--- a/charts/kubed/templates/kubed-deployment.yaml
+++ b/charts/kubed/templates/kubed-deployment.yaml
@@ -59,6 +59,7 @@ spec:
         - --tls-private-key-file=/var/serving-cert/tls.key
         - --use-kubeapiserver-fqdn-for-aks={{ .Values.apiserver.useKubeapiserverFqdnForAks }}
         - --enable-analytics={{ .Values.enableAnalytics }}
+        - --config-source-namespace={{ .Release.Namespace }}
         ports:
         - containerPort: 8443
 {{- if .Values.apiserver.healthcheck.enabled }}
--- a/tests/test_kubed.py
+++ b/tests/test_kubed.py
@@ -39,3 +39,16 @@ class TestKubed:
                 values={"global": {"kubedEnabled": False}},
                 show_only=self.show_only,
             )
+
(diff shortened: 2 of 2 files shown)


pgvishnuram:
we may need a test case to validate the release-namespace

AmFlint:
> @AmFlint can you please explain why this change improves performance?
> 
> i don't see any benchmark tests here ... some details [astronomer/issues#4546 (comment)](https://github.com/astronomer/issues/issues/4546#issuecomment-1106621080)

According to the official Helm chart's values for KubeD: https://github.com/kubeops/config-syncer/blob/837527ace59eecf3c38aeb53427e77404394dc4c/charts/kubed/values.yaml#L99, if this parameter is specified, KubeD will look only to the secrets/configmaps from the given namespace, instead of ALL secrets and configmaps in the cluster.

Unfortunately, I can't really demonstrate it to you since I don't own a large enough cluster to run performance tests.

But in my opinion, since KubeD is removed in 0.29 anyway, brining this fix could be a quick win for users in 0.28, if it doesn't completely resolve the problem with this fix, I would say it's not worth spending more time on this since it's removed in 0.29.

However, if it still doesn't fix performance/latency problems in large clusters, we could provide a simple shell script/command for users to "manually" synchronize the necessary secrets from release namespace to airflow deployment namespaces since it's a kubectl get cmd followed by a loop and kubectl apply.

But first, let's see if this fixes issues.

What do you think ?

danielhoherd (approved):
If customers are seeing high utilization, and it is caused by the same things that caused our high utilization, then @AmFlint's suggested fix will fix the problem. Our problems were due to entities being copied back and forth because of incorrect annotations, which were caused by manual interventions to fix other kubed issues. IE: a person manually copies something from the source namespace, including the annotations that set it up as the source for that entity, and then a cycle begins of propagating that entity where both the original and the manually copied namespace are fighting to be the source of truth.

TL;DR: I agree with @AmFlint that this is a good thing to add.

PR #1467, opened 2022-04-15
Use commander config-syncer and remove KubeD

AmFlint:
## Description

We want to remove KubeD.

KubeD is used to synchronize secrets from one namespace to other namespaces. In our case, it's used for Houston JWT signing certificates, and potentially SSL secrets for custom ingress controllers and OpenShift Users.

Multiple problems with KubeD:
- Slow performance in large clusters since it listens to all secrets and configmaps
- In terms of security, it's pretty bad since it uses Cluster-level RBAC permissions and can not be scoped to only certain namespaces.

Moreover, KubeD is only used in very specific use cases:
- Openshift users
- Private CA users

Everyone else doesn't need this config synchronization mechanism.

In order to fix this, we added a sync mechanism in Commander to synchronize necessary secrets from one namespace to another, it's more lightweight and in terms of security we can scope these permissions to namespaces inside a namespace pool.

In this Pull Request, I simply replaced "KubeD" with the new commander config-syncer, and made it compatible with namespace pools:
- If namespace pools disabled, creates ClusterRole + ClusterRoleBinding, since the config syncer needs to upload secrets to auto-generated namespaces (and list these namespaces at the cluster-level).
- For better security, with namespace enabled, permissions are scoped to namespaces in the pool and config-syncer will only synchronize secrets from the astronomer release namespace to these namespaces in the pool.

The config syncer can be completely disabled via a value if the user doesn't need it, as explained in the use cases above.

## Related Issues

https://github.com/astronomer/issues/issues/4451

## Testing

I conducted end to end tests in 2 different use cases:
- Namespace Pools disabled
- Namespace Pools enabled

I'll add e2e testing description for QA folks on the issue.

PR #1478, opened 2022-04-22
4547 remove kubed references

danielhoherd:
## Description

Follow up to @AmFlint's epic replacement/removal of kubed <https://github.com/astronomer/astronomer/pull/1467>

- Remove remaining references to kubed <https://github.com/astronomer/issues/issues/4547>
- Abort `bin/validate_commander_airflow_version` if docker is not found <https://github.com/astronomer/issues/issues/4548>
- Add help text for `make validate-commander-airflow-version`
- Make circleci build-artifact step require check-commander-airflow-version
- Bump some CI images
- Add test to validate that all annotation keys on all k8s entities are valid
- Remove pytest-sugar due to <https://github.com/Teemu/pytest-sugar/pull/234>

## Related Issues

Noted above

## Testing

I didn't perform any testing outside of CI. This change set does include some houston and commander changes though, so we should be aware of that if anything goes wrong while QA is doing their thing.

## Merging

This change is only for 0.29.x, though the annotation test and removal of pytest-sugar can be backported.


charts/astronomer/templates/houston/api/houston-jwt-certificate-secret.yaml
```diff
@@ -49,7 +49,6 @@ metadata:
     heritage: {{ .Release.Service }}
   annotations:
     # Indicate that we want to sync this to airflow namespaces
-    "kubed.appscode.com/sync": {{ (printf "platform-release=%s" .Release.Name) | quote }}
```
AmFlint, line 52: We have to think about what we want to do here.

In order to make the new config-syncer compatible with previous iterations of the platform, the config-syncer is based on this very specific annotation, see following workflow:
- Get secrets in release namespace
- For each secret, check if it contains the annotation `kubed.appscode.com/sync`
- IF it does, it means the secret needs to be synchronized to the namespaces containing the label associated to the previous annotation (e.g. `kubed.appscode.com: platform-release=astronomer)

If you remove this annotation here, config-syncer + commander won't work anymore since it's looking for this annotation.

2 solutions in my opinion:
1. Keep this annotation as is, but it means we keep something related to kubed in the codebase.
2. Rename to annotation to somethhing like `astronomer.io/config-syncer/sync`

If we decide to go for nb 2, I need to submit a 1 line PR on Commander to change the annotation used, and release it again.

What do you think ?

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -112,18 +112,6 @@ data:
         airflow:
           useAstroSecurityManager: true
           webserver:
-            {{ if .Values.global.kubedEnabled }}
```
AmFlint, line 115: Are we sure that we want to delete this part here ?

Maybe we want it in case the `config-syncer` is enabled ?

I didn't change this part in my PR because I didn't know about it. My understanding here is that if kubed isn't enabled, then Houston uses its own jwt Signing certificate instead of the one generated by the Helm release (contained in the secret mounted here).

In my demo and tests, it worked fine since my Airflow deployments have access to Houston and can retrieve it by sending requests to the resolver, but I believe it wouldn't work if Airflow deployment can't talk to Houston (Private CA situations).

I think the safest solution here would be to keep this code but change:
```yaml
{{ if .Values.global.kubedEnalbed }}
# becomes
{{ if .Values.astronomer.configSyncer.enabled }}
```

tests/chart_tests/test_houston_configmap.py
```diff
@@ -80,47 +80,6 @@ def test_houston_configmap_with_azure_enabled():
     assert livenessProbe["periodSeconds"] == 10
 
 
-def test_houston_configmap_with_kubed_enabled():
```
AmFlint, line 83: See my previous comment about the mounted volume if configSyncer is enabled.

I think we should keep these tests cases but check for astronomer.configSyncer.enabled instead of kubed.enabled

AmFlint (commented):
Thanks for this PR, however it seems I forgot to update a few places in my own PR.

charts/astronomer/templates/houston/houston-configmap.yaml
danielhoherd: Good catch. I don't know any of the behaviors around the questions you raised but we should figure that out before choosing a path forward here. I think your suggestion would work. @rob-1126 any ideas here regarding private CA behaviors with airflow?

tests/chart_tests/test_houston_configmap.py
danielhoherd: Yes, I'll put this back if we opt to go with the `.Values.astronomer.configSyncer.enabled` path you proposed.

charts/astronomer/templates/houston/api/houston-jwt-certificate-secret.yaml
AmFlint: Agreed, I kept it now for the transition but let's move to `astronomer.io/config-syncer/sync` now

AmFlint: Agreed, I kept it now for the transition but let's move to `astronomer.io/config-syncer/sync` now

charts/astronomer/templates/houston/houston-configmap.yaml
rob-1126: Yes, we actually had to issue updated versions to all currently maintained versions of Airflow about a month ago to avoid using jwk handling logic that fetched the secret via http because of private CAs. I'd like to see certs in the extraCaCerts bundle be replicated by houston into the airflow namespaces and the work done in airflow to mount them into webserver (and wouldn't be apposed to seeing them mounted in other containers).

rob-1126: I will say that if your replicating extraCaCerts for some installs, I'm very inclined to say "we might as well replicate the jwk in the same way for 100% of installs and not fetch via http at all"

AmFlint (approved):

PR #1495, opened 2022-04-29
configurable global interval for prometheus

pgvishnuram:
## Description

Make scrape_interval and  evaluation_interval values configurable

## Related Issues

## Testing

## Merging


pgvishnuram:
comments from @stuart23 
The only thing is that we wanted to specifically increase the scrape interval for airflow
https://astronomer.slack.com/archives/D02SREQ0RKJ/p1651246176378779
Roblox have 1 deployment with 1000 DAGs and 10 seconds is killing prometheus

pgvishnuram:
@danielhoherd 
in values.yml
```
global:
   job_name:
      airflow:
         scrape_interval: 10s
       node-exporter:
          scrape_interval: 30s
        fluentd:
          scrape_interval: 30s
        kube-state:
         scrape_interval: 10s
```

pgvishnuram:
@stuart23

stuart23:
:kissing_heart::ok_hand:LGTM

(Sorry about those reviewers/assignments - I was trying to add @gil-rosario as a watcher and haven't had my coffee yet!)

charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -15,8 +15,8 @@ data:
   config: |-
     # https://prometheus.io/docs/prometheus/latest/configuration/configuration/
     global:
-      scrape_interval: 30s
-      evaluation_interval: 30s
+      scrape_interval: {{ .Values.global.scrape_interval }}
```
danielhoherd, line 18: Just to circle back on what we talked about on zoom, this "global" is inside the prom config, not a helm namespace. We should probably go with `.Values.prometheus.scrape_interval` for this. Maybe for the ones below we can have EG `.Values.prometheus.velero.scrape_interval` and `.Values.prometheus.velero.scrape_timeout`. Just a guess, there might be an existing data structure that makes more sense.

pgvishnuram: the above values configured inside prometheus and inside configmap it has global config so the placeholder is correct. 

when user make a overrides then --set prometheus.global.scrape_interval

santoshr1016: Un-knowingly i too worked in this ticket..
What @danielhoherd is pointing does makes sense to me, IMO there is no need to mention global values for the local prometheus charts, We should override the values from the main values.yaml in the basedir and use --set flag to replace..
https://github.com/astronomer/astronomer/pull/1578

danielhoherd: Oh dang, yeah, this is basically the same thing we were talking about in https://github.com/astronomer/issues/issues/4782

stuart23: This is in a subchart (prometheus), so when the user is installing the astronomer (main) chart, they will still need to scope to this namespace (eg `--set prometheus.velero.scrape_interval=30s`) right? If we were to template it as `.Values.prometheus.velero.scrape_interval`, wont the user have to use `--set prometheus.prometheus.velero.scrape_interval=30s`? I would much prefer the first option.

rbankston: i much prefer not using global namespace but setting all the values in the prometheus subchart values.yaml for documentation and not using the global astronomer/astronomer key unless absolutely necessary as Daniel mentioned [here](https://github.com/astronomer/astronomer/pull/1495#discussion_r866455077)

santoshr1016: @stuart23 How it works is, if you have provided the values in the main values yaml, https://github.com/astronomer/astronomer/blob/master/values.yaml
this will override the local prometheus chart values otherwise default values are picked up from local chart, if you don't give any values

danielhoherd:
I think we can close this in favor of https://github.com/astronomer/astronomer/pull/1578/

danielhoherd:
Duplicate of https://github.com/astronomer/astronomer/pull/1578

gil-rosario:
😜👍🏽

PR #1499, opened 2022-05-02
airflow chart not added in master branch

pgvishnuram:
## Description

Add airflow chart version to master branch

## Related Issues

NA

## Testing

NA

## Merging

merge to release-0.29 branch


pgvishnuram:
@vishwas-astro

danielhoherd (approved):
RC's are an odd case. I'm not sure if we should be merging RC's to master and then into the release branch, or just keep those in the release branch. I'm not against merging them to master though, I just don't know what the best tactic should be.

PR #1515, opened 2022-05-11
Fix calico in CI

danielhoherd:
## Description

Make calico install from a known source inside the repo instead of a file hosted upstream. CI was broken recently for version 1.19 and 1.20 due to changes in calico, even though they were working fine. Vendoring in these calico configs will make our CI more stable.

## Related Issues

- https://github.com/astronomer/issues/issues/4616

## Testing

This fixes the broken CI, which is probably good enough testing. These changes are only for CI.

## Merging

This should be merged into any supported branch because CI is likely broken in all of them.


andriisoldatenko (approved):

danielhoherd:
This was cherry picked into release branches 0.25-0.29

PR #1523, opened 2022-05-16
Added astronomer-nginx-metrics-policy global flag

shmanu017:
## Description

Disable all the network policies from global configuration.Fixed this in regression testing cc @pgvishnuram 

## Related Issues

https://github.com/astronomer/issues/issues/4386


pgvishnuram (approved):

pgvishnuram:
do we have a test case for this to see all network policy len is 0

shmanu017:
https://github.com/astronomer/astronomer/blob/master/tests/chart_tests/test_global_network_policy_flag.py will check if flag is present, 
Trying to figure out a test case which will test that state.

danielhoherd:
Would've been great to have a test with this change.

PR #1536, opened 2022-05-19
increase-cronjob-frequency for airflow check

shmanu017:
## Description

Run cronjob/astronomer-houston-update-airflow-check more frequently by default from 24hrs to hourly at the 57 min mark in every hour as suggested by @danielhoherd 

## Related Issues

https://github.com/astronomer/issues/issues/4633

## Merging

Can be cherry-picked into any releases. Not mandatory, not high priority.

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -191,7 +191,7 @@ houston:
     url: https://updates.astronomer.io/astronomer-certified
 
     # Default here is to run at midnight every night
-    schedule: "0 0 * * *"
+    schedule: "57 * * * *"
 
   # Check for Astronomer Runtime Updates
   # This runs as a CronJob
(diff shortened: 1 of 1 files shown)


vishwas-astro:
Manu, it is not every 57 mins, but hourly at the 57 min mark in every hour :)

charts/astronomer/values.yaml
```diff
@@ -191,7 +191,7 @@ houston:
     url: https://updates.astronomer.io/astronomer-certified
 
     # Default here is to run at midnight every night
```
danielhoherd, line 193:
```suggestion
    # Run at minute 57 every hour https://crontab.guru/#57_*_*_*_*
```

vishwas-astro (approved):

jedcunningham (commented):
LGMT once @danielhoherd's suggestion is committed.

jedcunningham (approved):

PR #568, opened 2020-02-17
[WIP] Add pgbouncer support

andriisoldatenko:
TODO:
- [x] update prisma
- [x] update connection for knex migrations
- [x] update connection to create airflow deployment dbs
- [x] implement new pgbouncer-bootsrapper https://github.com/astronomer/db-bootstrapper/pull/9

fix https://github.com/astronomer/issues/issues/647

PR #1545, opened 2022-05-27
Add pgbouncer against houston-api

andriisoldatenko:
This PR brings new pgbouncer helm Chart.
It works by default with kerberized pgbouncer.

Some default values are in values.yaml. 

Resolves https://github.com/astronomer/issues/issues/4780

https://github.com/astronomer/astronomer/pull/568
Related PR https://github.com/astronomer/ap-vendor/pull/312


pgvishnuram:
can you add support for extravolumes and volumemounts too, In case if sslmode is set to require this will help us to add extra certs

santoshr1016 (approved):
LGTM..

charts/astronomer/values.yaml
```diff
@@ -38,6 +38,10 @@ images:
     repository: quay.io/astronomer/ap-cli-install
     tag: 0.26.7
     pullPolicy: IfNotPresent
+  pgbouncer:
+    repository: quay.io/astronomer/ap-pgbouncer
```
andriisoldatenko, line 42:
```suggestion
    repository: asoldatenko/ap-pgbouncer-krb
```

charts/pgbouncer/values.yaml
```diff
@@ -0,0 +1,12 @@
+#############################
+#                           #
+#  Pgbouncer Configuration  #
+#                           #
```
danielhoherd, line 4:
```suggestion
#  Pgbouncer Configuration  #
```

charts/pgbouncer/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+#############################
+image:
+  repository: asoldatenko/ap-pgbouncer-krb
```
danielhoherd, line 7: Need to update this before merging.

charts/pgbouncer/templates/pgbouncer-backend-secret.yaml
```diff
@@ -0,0 +1,21 @@
+################################
```
andriisoldatenko, line 1: we can remove it, still don't have any final requirements, so i do assumption that customer can configure everything via ENV variables.

charts/pgbouncer/templates/_helpers.yaml
```diff
@@ -0,0 +1,70 @@
...
+{{- end -}}
+
+{{- define "pgbouncer.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+
+{{ define "pgbouncer_image" -}}
+{{ printf "%s:%s" .Values.image.repository .Values.image.tag }}
+{{- end }}
```
danielhoherd, line 12: This need to be refactored to support private registry. EG: https://github.com/astronomer/astronomer/blob/276a732ee3/charts/astronomer/templates/_helpers.yaml#L36-L42

andriisoldatenko: good call, will do.

charts/pgbouncer/templates/_helpers.yaml
```diff
@@ -0,0 +1,70 @@
...
+{{ .Release.Name }}-pgbouncer-stats
+{{- end }}
+
+{{- define "pgbouncer_environment" }}
+- name: DB_USER
+  value: postgres
+- name: DB_PASSWORD
+  value: postgres
+# - name: DB_NAME
+#   value: postgres
```
danielhoherd, line 37: Seems like if we're leaving all the other defaults in here, like DB_USER = postgres, then we should uncomment this too.

andriisoldatenko: unfortunately it will not work, cause it generates concrete DB_NAME https://github.com/astronomer/ap-vendor/blob/main/pgbouncer-krb/entrypoint.sh#L122 but during db_boostrapwe do connect using `DB_NAME=template1`:

https://github.com/astronomer/db-bootstrapper/blob/master/main.py#L68

but we want:
```
* = host=astronomer-postgresql.astronomer.svc.cluster.local port=5432 user=postgres
```

andriisoldatenko: @danielhoherd best i can do here just remove to not confuse.

charts/pgbouncer/templates/pgbouncer-deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+      release: {{ .Release.Name }}
+      component: pgbouncer
+  template:
+    metadata:
+      labels:
+        app: {{ template "pgbouncer.name" . }}
+        chart: {{ template "pgbouncer.chart" . }}
+        heritage: {{ .Release.Service | quote }}
+        release: {{ .Release.Name }}
+        component: pgbouncer
```
danielhoherd, line 30: `spec.template.metadata.labels` now requires both `app` and `version`. see https://github.com/astronomer/astronomer/pull/1609

charts/pgbouncer/templates/pgbouncer-service.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  type: ClusterIP
+  selector:
+    app: {{ template "pgbouncer.name" . }}
+    release: {{ .Release.Name | quote }}
+  ports:
+    - name: pgbouncer
+      protocol: TCP
+      port: {{ .Values.servicePort }}
+      targetPort: {{ .Values.internalPort }}
```
danielhoherd, line 20: `spec.ports[*]` now require `appProtocol`. See https://github.com/astronomer/astronomer/pull/1609/files

charts/pgbouncer/values.yaml
```diff
@@ -0,0 +1,92 @@
+#############################
+#  Pgbouncer Configuration  #
+#############################
+image:
+  repository: asoldatenko/ap-pgbouncer-krb
+  tag: 1.1.17-wo
```
danielhoherd, line 6: Need to update these before merge

charts/pgbouncer/templates/_helpers.yaml
andriisoldatenko: fixed.

.circleci/config.yml
```diff
@@ -283,6 +283,7 @@ workflows:
           matrix:
             parameters:
               docker_image:
+                - astronomer/ap-pgbouncer-krb:1.1.17
```
andriisoldatenko, line 286: @danielhoherd i think it's wrong image...

andriisoldatenko: fixed in 5bfca2d9

danielhoherd (approved):

charts/postgresql/templates/astronomer-bootstrap-secret.yaml
```diff
@@ -11,5 +11,9 @@ metadata:
     argocd.argoproj.io/sync-wave: "-1"
   {{- end }}
 data:
+  {{- if .Values.global.pgbouncer.enabled }}
+  connection: {{ printf "postgres://postgres:postgres@astronomer-pgbouncer.astronomer.svc.cluster.local:5432" | b64enc | quote }}
```
pgvishnuram, line 15: we need to remove hardcoding here - there are customers who use different release name

andriisoldatenko: it's only for testing, but i'll change it with values.

pgvishnuram: no probs then, i thought we missed

charts/pgbouncer/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+servicePort: 5432
+
+podDisruptionBudget:
+  enabled: false
+  config: {}
+
+networkPolicies:
+  enabled: true
+
+env: []
```
pgvishnuram, line 19: there is no place to add extraenv as per values.yaml  - all are currently hardcoded in helper.yaml

charts/pgbouncer/templates/_helpers.yaml
danielhoherd: it's fine as it is. i didn't expect it to cause a problem.

charts/pgbouncer/values.yaml
danielhoherd: I fixed this in https://github.com/astronomer/astronomer/pull/1545/commits/376ee8b3ed18968db3f107ea4358c1f44ae59b06

danielhoherd: Maybe a better way to do this would be to put the default values into the `env` section.

andriisoldatenko: TBH: it's not a default values. It's required values.

tests/chart_tests/test_pgbouncer.py
```diff
@@ -0,0 +1,75 @@
...
+        assert labels == {
+            "app": "pgbouncer",
+            "chart": "pgbouncer-0.0.1",
+            "component": "pgbouncer",
+            "heritage": "Helm",
+            "release": "release-name",
+            "test_label": "test_label1",
+            "tier": "astronomer",
+            "version": "0.0.1",
+        }
```
danielhoherd, line 75: In order to avoid changing this test every time the version is bumped, we should probably change this to:

```suggestion
        assert {"test_label": "test_label1"} in labels
```

PR #1564, opened 2022-06-08
enhance es exporter service

pgvishnuram:
## Description

ES Exporter is set to run as root it's not needed anymore , current change will run the service a less privileged user.

**changes**
* run as non root service to removed unnecessary root usage

## Related Issues
Yet to define

## Testing

Yet to define

## Merging

Yet to define

### DIFF
--- a/charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
+++ b/charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
@@ -50,7 +50,8 @@ spec:
 {{ toYaml (default .Values.global.platformNodePool.tolerations .Values.tolerations) | indent 8 }}
       restartPolicy: {{ .Values.exporter.restartPolicy }}
       securityContext:
-        runAsUser: 0
+        runAsNonRoot: true
+        runAsUser: {{ .Values.securityContext.runAsUser }}
       containers:
         - name: metrics-exporter
           image: {{ template "exporter.image" . }}
--- a/charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
+++ b/charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
@@ -1,6 +1,6 @@
-################################
-## Elasticsearch Exporter Deployment
-#################################
+#######################################
+## Elasticsearch Exporter Deployment ##
+#######################################
 apiVersion: apps/v1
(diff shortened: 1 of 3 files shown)


pgvishnuram:
Need to write test cases for the same self @pgvishnuram

charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
```diff
@@ -50,7 +50,8 @@ spec:
 {{ toYaml (default .Values.global.platformNodePool.tolerations .Values.tolerations) | indent 8 }}
       restartPolicy: {{ .Values.exporter.restartPolicy }}
       securityContext:
-        runAsUser: 0
+        runAsNonRoot: true
+        runAsUser: {{ .Values.securityContext.runAsUser }}
```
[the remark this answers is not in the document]
pgvishnuram: if you have open policy agent or any equivalent policy enforcement - it wont allow to run as root.
In openshift we have scc privileges to run as anyuid - that allows users to set and run with custom id's.

santoshr1016 (approved):
So you are running as userid 1000, which is safe.. 👍 
Test case would complete it..

tests/chart_tests/test_elasticsearch.py
```diff
@@ -205,3 +205,34 @@ def test_nginx_es_client_network_selector_with_logging_sidecar_enabled(
...
+            ],
+        )
+
+        assert len(docs) == 1
+        doc = docs[0]
+        es_index = doc["data"]["nginx.conf"]
+        assert doc["kind"] == "ConfigMap"
+        assert "vector.$remote_user.*/$1" in es_index
+
+    def test_elasticsearch_exporter_securitycontext_defaults(self, kube_version):
```
danielhoherd, line 225: Can we get one more test here that uses a custom `.Values.securityContext.runAsUser`?

danielhoherd, line 225: Can we get one more test here that uses a custom `.Values.securityContext.runAsUser`?

pgvishnuram: yes sure

pgvishnuram: yes sure

danielhoherd (approved):

PR #1566, opened 2022-06-10
Issue-4443 | Add user configurable suffix in various dashboard domains to avoid conflicts

santoshr1016:
## Description
Add subDomain as global variables

## Related Issues

https://github.com/astronomer/issues/issues/4443

## Testing

## Merging

### DIFF
--- a/charts/alertmanager/templates/_helpers.tpl
+++ b/charts/alertmanager/templates/_helpers.tpl
@@ -43,7 +43,7 @@ Image name.
 {{- end -}}
 
 {{ define "alertmanager.url" -}}
-alertmanager.{{ .Values.global.baseDomain }}
+{{ .Values.global.prefixDomain }}-alertmanager.{{ .Values.global.baseDomain }}
 {{- end }}
 
 {{/*
--- a/charts/alertmanager/templates/_helpers.tpl
+++ b/charts/alertmanager/templates/_helpers.tpl
@@ -43,7 +43,11 @@ Image name.
 {{- end -}}
 
 {{ define "alertmanager.url" -}}
-{{ .Values.global.prefixDomain }}-alertmanager.{{ .Values.global.baseDomain }}
+{{- if .Values.global.suffixDomain -}}
+alertmanager-{{ .Values.global.suffixDomain }}.{{ .Values.global.baseDomain }}
+{{- else -}}
+alertmanager.{{ .Values.global.baseDomain }}
+{{- end }}
 {{- end }}
 
 {{/*
--- a/values.yaml
+++ b/values.yaml
@@ -6,6 +6,7 @@ global:
   baseDomain: ~
 
   # Name of secret containing TLS certificate
+  subDomain: # Name of subdomains to avoid conflicts
   tlsSecret: astronomer-tls
 
   # List of secrets containing private CA certificates
--- a/values.yaml
+++ b/values.yaml
@@ -6,7 +6,7 @@ global:
   baseDomain: ~
 
   # Name of secret containing TLS certificate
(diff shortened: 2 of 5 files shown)


pgvishnuram:
@danielhoherd - Questions below 
1. should prefix domain be optional ( this PR seems makes it mandatoy now)
2. prefix domain configuration currently is applied at platform level , to pass to airflow deployments houston needs a change to consume that config
3. astro-ui url should also get changed based on pre-fix domain

santoshr1016:
Yes, As @pgvishnuram  suggested it makes sense to make the suffix are configurable item..
The reason for adding suffix is that. houston api has a regex for the ingress.
https://github.com/astronomer/houston-api/blob/main/src/routes/v1/authorization/handler.js#L24
While hitting the these endpoints **grafana-silly.santosh-test21.astro-7051.com** they are redirected to astro home page..

It does need the astro ui change too...

It looks like more of a feature, Lot of tweaking is needed to the internal components

`
❯ k -n astronomer get ing
NAME                              CLASS    HOSTS                                                                                                                ADDRESS                                                                  PORTS     AGE
astronomer-alertmanager-ingress   <none>   alertmanager-silly.santosh-test21.astro-7051.com                                                                     a5cead5e1b463480aad90bc0e958df6b-681576468.us-east-1.elb.amazonaws.com   80, 443   3h27m
astronomer-grafana-ingress        <none>   grafana-silly.santosh-test21.astro-7051.com                                                                          a5cead5e1b463480aad90bc0e958df6b-681576468.us-east-1.elb.amazonaws.com   80, 443   3h27m
astronomer-houston-ingress        <none>   houston.santosh-test21.astro-7051.com                                                                                a5cead5e1b463480aad90bc0e958df6b-681576468.us-east-1.elb.amazonaws.com   80, 443   3h27m
astronomer-kibana-ingress         <none>   kibana-silly.santosh-test21.astro-7051.com                                                                           a5cead5e1b463480aad90bc0e958df6b-681576468.us-east-1.elb.amazonaws.com   80, 443   3h27m
astronomer-prometheus-ingress     <none>   prometheus-silly.santosh-test21.astro-7051.com  
`

values.yaml
```diff
@@ -6,6 +6,7 @@ global:
   baseDomain: ~
 
   # Name of secret containing TLS certificate
+  suffixDomain: # Name of subdomains to avoid conflicts
```
pgvishnuram, line 9: comment in wrong place

charts/alertmanager/templates/_helpers.tpl
```diff
@@ -43,7 +43,7 @@ Image name.
 {{- end -}}
 
 {{ define "alertmanager.url" -}}
-alertmanager.{{ .Values.global.baseDomain }}
+{{ .Values.global.prefixDomain }}-alertmanager.{{ .Values.global.baseDomain }}
```
danielhoherd: Why are we forcing users into a prefix mechanism rather than letting them specify an arbitrary domain name?

santoshr1016: Houston as this regex for subdomain to check string that starts with these names
**/^(grafana|kibana|alertmanager|prometheus)$** 
https://github.com/astronomer/houston-api/blob/main/src/routes/v1/authorization/handler.js#L24
Lets discuss this today...and can tell Alex, based on which they can change..

aliotta: Would it make sense for this regex /^(grafana|kibana|alertmanager|prometheus)$/ to be configurable? It could default to the current value but allow someone to modify it so they can use w/e domains they want?

PR #1578, opened 2022-06-28
Issue-4752 | Add configurable scrape interval for prometheus

santoshr1016:
## Description

Addresses the gh issue [4752](https://github.com/astronomer/issues/issues/4782)

## Related Issues

https://github.com/astronomer/issues/issues/4782

## Testing

Add test case

## Merging

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -363,7 +363,8 @@ prometheus:
   podLabels: {}
   # Data retention
   retention: 15d
-
+  # Scrape interval
+  scrape_interval: 30s
   # Persistence configuration
   persistence:
     enabled: true
--- a/values.yaml
+++ b/values.yaml
@@ -358,27 +358,12 @@ grafana:
 
 #################################
 ## Prometheus configuration
+## Scrape interval can be modified by using arg --set in helm install/update operations
 #################################
 prometheus:
   podLabels: {}
   # Data retention
   retention: 15d
-  # Scrape interval
-  scrape_interval: 30s
-  evaluation_interval: 30s
-  velero:
-    scrape_interval: 30s
-  kubeState:
-    scrape_interval: 10s
-  fluentd:
-    scrape_interval: 30s
-  nodeExporter:
-    scrape_interval: 30s
-  airflow:
-    scrape_interval: 10s
-  postgresqlExporter:
-    scrape_interval: 60s
-    scrape_timeout: 30s
 
   # Persistence configuration
   persistence:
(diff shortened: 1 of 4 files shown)


andriisoldatenko (approved):

stuart23:
Personally I don't really like the idea of having a `globals` branch, and then defining associated parameters in a seperate branch. For example, we now need:

```
global:
  veleroEnabled: true
velero:
  scrape_interval: 30s
```

In my experience, the standard (and more user friendly) way to do this would be:

```
velero:
  enabled: true
  scrape_interval: 30s
```

@rbankston do you have any opinion?

values.yaml
```diff
@@ -363,6 +363,22 @@ prometheus:
...
+    scrape_interval: 10s
+  fluentd:
+    scrape_interval: 30s
+  nodeExporter:
+    scrape_interval: 30s
+  airflow:
+    scrape_interval: 10s
+  postgresqlExporter:
+    scrape_interval: 60s
+    scrape_timeout: 30s
```
danielhoherd, line 381: These are probably not worth exposing to both the prometheus chart and the astronomer umbrella chart. I think we can go with just the prometheus values.yaml changes. It'll still work the same.

santoshr1016: Yes, it works the same, only thing is rather than manually doing
```
--set prometheus.scrape_interval=30s
```
for each helm release will be an effort and error prone, giving here some default will be seamless, i am not sure of exposing these time intervals, who is best person to ask or is it our team to decide?

danielhoherd: We would not need to use `--set`, we could still use values in `values.yaml` files. EG: [we set `prometheus.ingressNetworkPolicyExtraSelectors` in a values file](https://github.com/astronomer/google-environments/blob/09da774b52/prod/cloud/app/config.yaml#L107-L108) and that is not exposed to [the astronomer umbrella chart's values file](https://github.com/astronomer/astronomer/blob/ec12fba78c/values.yaml). It is exposed only within [the prometheus subchart's values.yaml](https://github.com/astronomer/astronomer/blob/ec12fba78c/charts/prometheus/values.yaml#L62). Using `--set` gives you no extra features than `-f values.yaml`. Setting those values in the astronomer umbrella chart's values file is just overriding whatever is in the subchart's values file, and in this case it is not helpful. The only reason we'd want to put things into the umbrella chart is to give customers additional hints that those values are configurable, so we should be selective about what we put in there. We do not want to drown out the more important values that most customers will want to tune by including a bunch of values that most customers are not going to modify.

PR #1581, opened 2022-06-29
Feature/4472 Adding Extra Object Capability

mishah334:
## Description

- Adding capability to add any Kubernetes Object from values.
- Adding capability to add extra containers and volumes for `astro-ui` service.
- Improving CircleCI `unittest` execution time by using the larger machine to test.

## Related Issues

[#4472 ](https://github.com/astronomer/issues/issues/4472)

## Testing

- Added test case to test the extra-object input.

## Merging

N/A

### DIFF
--- a/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
@@ -38,6 +38,14 @@ spec:
           imagePullPolicy: {{ .Values.images.astroUI.pullPolicy }}
           resources:
 {{ toYaml .Values.astroUI.resources | indent 12 }}
+          {{- if .Values.astroUI.command }}
+          command:
+            {{- toYaml .Values.astroUI.command | nindent 12 }}
+          {{- end }}
+          {{- if .Values.astroUI.args }}
+          args:
+            {{- toYaml .Values.astroUI.args | nindent 12 }}
+          {{- end }}
           ports:
             - name: astro-ui-http
               containerPort: {{ .Values.ports.astroUIHTTP }}
@@ -64,3 +72,14 @@ spec:
             - name: APP_API_LOC_WSS
               value: "wss://houston.{{ .Values.global.baseDomain }}/ws"
             {{- end }}
+          volumeMounts:
+{{- if .Values.astroUI.volumeMounts }}
+{{- toYaml .Values.astroUI.volumeMounts | nindent 12 }}
+{{- end }}
+{{- if .Values.astroUI.extraContainers }}
+{{- toYaml .Values.astroUI.extraContainers | nindent 8 }}
+{{- end }}
+      volumes:
+{{- if .Values.astroUI.extraVolumes }}
+{{- toYaml .Values.astroUI.extraVolumes | nindent 8 }}
+{{- end }}
--- a/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
@@ -74,12 +74,12 @@ spec:
             {{- end }}
           volumeMounts:
 {{- if .Values.astroUI.volumeMounts }}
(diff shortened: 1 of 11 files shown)


charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
```diff
@@ -38,6 +38,14 @@ spec:
...
           resources:
 {{ toYaml .Values.astroUI.resources | indent 12 }}
+          {{- if .Values.astroUI.command }}
+          command:
+            {{- toYaml .Values.astroUI.command | nindent 12 }}
+          {{- end }}
+          {{- if .Values.astroUI.args }}
+          args:
+            {{- toYaml .Values.astroUI.args | nindent 12 }}
+          {{- end }}
```
danielhoherd, line 48: I like this whole block here. It seems like this should be standard. Good stuff.

vishwas-astro:
@mishah334 is this ready for review?

charts/astronomer/templates/extra-objects.yaml
```diff
@@ -0,0 +1,4 @@
+{{ range .Values.extraObjects }}
```
danielhoherd, line 1: This file should have a header. `helm template --show-only` returns an error if the template has null output. We sometimes do this in tests, which triggers the problem.

charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
```diff
@@ -83,19 +83,23 @@ spec:
         - name: houston
           image: {{ template "houston.image" . }}
           imagePullPolicy: {{ .Values.images.houston.pullPolicy }}
+          {{- if .Values.houston.workerCommand }}
```
danielhoherd, line 86: There is already a `houston.worker` namespace, though there is no `houston.api` namespace. I kinda think this should be `houston.worker.command`, `houston.worker.args`, and the same for `houston.api`. WDYT?

tests/chart_tests/test_astronomer_houston_sidecar.py
```diff
@@ -6,17 +6,17 @@
 chart_values = {
     "astronomer": {
         "houston": {
+            "command": ["/bin/sh"],
             "apiArgs": [
-                "sh",
                 "-c",
-                "yarn serve 1> >( tee -a /var/logs/houston/data.out.log ) 2> >( tee -a /var/logs/houston/data.err.log >&2 )",
+                "yarn serve 1> >( tee -a /var/log/houston/data.out.log ) 2> >( tee -a /var/log/houston/data.err.log >&2 )",
```
danielhoherd, line 12: good catch! [definitely should be `/var/log`](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard#Directory_structure)

danielhoherd (changes requested):
Looks good! I left some feedback, and only the `houston.worker` thing needs to be answered before we merge this. I'd like others to give feedback before we move forward.

charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -114,11 +114,15 @@ spec:
             initialDelaySeconds: {{ .Values.houston.readinessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.houston.readinessProbe.periodSeconds }}
             failureThreshold: {{ .Values.houston.readinessProbe.failureThreshold }}
+          {{- if .Values.houston.command }}
+          command:
+            {{- toYaml .Values.houston.command | nindent 12 }}
+          {{ else }}
+          command: ["yarn", "serve"]
+          {{- end }}
```
pgvishnuram, line 122: @mishah334 - if no command block is mention it will take from defaults , should we still need to override them

charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
bote795: yea i think it should be houston.worker.X

danielhoherd (approved):

PR #1598, opened 2022-07-11
updated node url to http

rishkarajgi:
## Description

Node url check for https is causing external logging with es to fail

## Related Issues

https://github.com/astronomer/issues/issues/4825

## Testing

## Merging

0.29.2

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -350,7 +350,7 @@ data:
     elasticsearch:
       enabled: true
       client:
-        node: {{ printf "https://%s-external-es-proxy:9201" .Release.Name }}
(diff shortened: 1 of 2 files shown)


andriisoldatenko:
If other client needs https? Please make it configurable trough values instead.

pgvishnuram (approved):
updates protocol changes , defaults to http. 

This was http earlier - due to some enhancements it was set to https - This applies only for external es proxy service.

rishkarajgi:
> If other client needs https? Please make it configurable trough values instead.

@andriisoldatenko this was http by default and applies only for external es proxy service.

andriisoldatenko:
> > If other client needs https? Please make it configurable trough values instead.
> 
> 
> 
> @andriisoldatenko this was http by default and applies only for external es proxy service.

Thanks for clarifying!

PR #1607, opened 2022-07-19
Included new parameter regionendpoint under s3 registry backend configuration

pritt20:
## Description

PR change request to include new parameter `regionendpoint` under s3 registry backend configuration to support IBM COS bucket integration via S3 api.

## Related Issues

https://github.com/astronomer/issues/issues/4853

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -169,6 +169,7 @@ s3:
   encrypt: {{ .Values.registry.s3.encrypt }}
   keyid: {{ .Values.registry.s3.keyid }}
   rootdirectory: {{ .Values.registry.s3.rootdirectory }}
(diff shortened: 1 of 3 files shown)


danielhoherd (changes requested):
Looks good since it's so small of a change, but I agree with @pgvishnuram that this would be a great opportunity to add tests for `registry.s3Config` behavior entirely. @pritt20 if you need help with that let me know and I can assist.

pritt20:
Thanks @danielhoherd and @pgvishnuram !

I have updated the test case as well for this change. Please let me know there are any additional changes :)

danielhoherd (approved):
LGTM! Thanks for adding the missing tests @pritt20

pgvishnuram:
@danielhoherd @vishwas-astro this is release as a part of 0.30 can be safely closed

PR #1608, opened 2022-07-21
commander: remove unused pv permissions

AmFlint:
## Description

Remove PV permissions in the Commander role as we don't need these. PVs are obtained through PersistentVolumeClaims.

## Related Issues

https://github.com/astronomer/issues/issues/4679

## Testing

I did not feel the need to add unit tests to cover this case. Otherwhise, we would have to check for every resource that should not be in commander's role.


danielhoherd (approved):
LGTM. I agree that tests aren't needed here, and that if we really want to test for this sort of thing we should be doing blanket tests for all resources that should not have `persistentvolumes` permissions.

PR #1611, opened 2022-07-25
DoNotMerge | cluster proportionate autoscaler

santoshr1016:
## Description

This is implementation if the Cluster proportionate autoscaler with pod preemption
For more info 
[cpa](https://github.com/kubernetes-sigs/cluster-proportional-autoscaler)
[overprovisioning](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler)

## Related Issues

## Testing

## Merging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -283,6 +283,8 @@ workflows:
           matrix:
             parameters:
               docker_image:
+                - k8s.gcr.io/cpa/cluster-proportional-autoscaler:1.8.4
+                - k8s.gcr.io/pause
                 - quay.io/astronomer/ap-alertmanager:0.23.0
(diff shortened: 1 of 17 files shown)


pgvishnuram:
Are we going to ship this service to all customers ? 
The Goal we are seeking here is to reduce costing on our own cloud so this should either be committed into google-environments , for testing and implementation @danielhoherd

danielhoherd:
@santoshr1016 I agree with @pgvishnuram. The goals here are strictly nebula related, not astronomer/astronomer related. We should be making changes to our specific installation, not the core product. IE: this is a prod-ops project, not a product engineering project.

PR #1613, opened 2022-07-26
update nginx ingress controller image

pgvishnuram:
## Description

* CVE fixes and upstream enhancements
**release_notes:** 
* https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v1.3.0

**Points to note:**
* This release removes support for Kubernetes v1.19.0
* permissions on the coordination.k8s.io/leases resource for leaderelection lock added 

## Related Issues

Yet to create

## Testing

Yet to add

## Merging

### DIFF
--- a/charts/nginx/templates/nginx-config-role.yaml
+++ b/charts/nginx/templates/nginx-config-role.yaml
@@ -44,4 +44,26 @@ rules:
       - get
       - create
       - update
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    resourceNames:
+      - ingress-controller-leader-{{ template "nginx.fullname" . }}
+    verbs:
+      - get
+      - update
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - create
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - create
+      - patch
 {{- end }}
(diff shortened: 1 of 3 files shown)


santoshr1016 (approved):
+1

charts/nginx/templates/nginx-config-role.yaml
```diff
@@ -44,4 +44,26 @@ rules:
...
+      - ingress-controller-leader-{{ template "nginx.fullname" . }}
+    verbs:
+      - get
+      - update
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - create
```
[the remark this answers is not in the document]
pgvishnuram: from this nginx version this is the new change they have implemented

charts/nginx/values.yaml
```diff
@@ -9,7 +9,7 @@ tolerations: []
 images:
   nginx:
     repository: quay.io/astronomer/ap-nginx
-    tag: 1.2.1
+    tag: 1.3.0
```
danielhoherd, line 12: Given that nginx 1.3.0 does not support k8s 1.19, we should also remove k8s 1.19 as part of this PR.

pgvishnuram: seems its only  1.19.0 not the whole series

pgvishnuram:
@danielhoherd - we can merge this

charts/nginx/values.yaml
danielhoherd: oh, awesome! then we can completely ignore it because we can expect almost nobody to be running exactly 1.19.0.

vishwas-astro: We should probably call this out in our customer facing documentation. Wdyt @danielhoherd, or is that not necessary?

PR #1616, opened 2022-08-01
Make PodDisruptionBudget optional

danielhoherd:
## Description

Make PDBs optional at a global level

## Related Issues

https://github.com/astronomer/issues/issues/4513

## Testing

Unit tests are passing. It wouldn't be a bad idea to sanity check some other configs.

## Merging

At least 0.30.0, maybe more depending on how the discussion goes in the original ticket.


vishwas-astro (approved):

vishwas-astro:
Oops, it is still in draft state. But you have my 👍 @danielhoherd

vishwas-astro:
@danielhoherd any particular reason PR is in draft state? Are you looking for inputs from @rob-1126 ?

PR #1619, opened 2022-08-02
Feature-3617 Defining Resource Limit for Pods

mishah334:
## Description

Defining Resource Limit for Pods.

## Related Issues

[#3617](https://github.com/astronomer/issues/issues/3617)

## Testing

Added the test case to check resource limit definations.

## Merging

N/A

### DIFF
--- a/charts/nats/values.yaml
+++ b/charts/nats/values.yaml
@@ -190,11 +190,11 @@ exporter:
   # resource configuration for the metrics sidecar
   resources:
     limits:
-     cpu: 150m
-     memory: 150Mi
-    requests:
      cpu: 100m
      memory: 128Mi
+    requests:
+     cpu: 50m
+     memory: 100Mi
 
 # Authentication setup
 auth:
--- a/charts/stan/values.yaml
+++ b/charts/stan/values.yaml
@@ -167,12 +167,11 @@ exporter:
   enabled: true
   resources:
     limits:
-      memory: 150Mi
-      cpu: 150m
+     cpu: 100m
+     memory: 128Mi
     requests:
-      memory: 128Mi
-      cpu: 100m
-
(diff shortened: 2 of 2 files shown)


charts/stan/values.yaml
```diff
@@ -161,13 +165,14 @@ store:
...
-  #   memory: 128Mi
-  #  requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
+    limits:
+     cpu: 100m
+     memory: 100Mi
+    requests:
+     cpu: 100m
```
santoshr1016, line 173: Requests: you can change it to CPU 50m, memory to 80Mi

charts/nats/values.yaml
```diff
@@ -188,13 +188,13 @@ reloader:
...
-  #   memory: 128Mi
-  #  requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
+    limits:
+     cpu: 100m
+     memory: 128Mi
+    requests:
+     cpu: 100m
```
santoshr1016, line 196: Requests: CPU to 50m. memory to 100Mi
idea is not to be same as limits

santoshr1016 (approved):
+1

charts/stan/values.yaml
pgvishnuram: https://github.com/astronomer/astronomer/blob/master/values.yaml#L401-L413

charts/nats/values.yaml
pgvishnuram: https://github.com/astronomer/astronomer/blob/master/values.yaml#L387-L399

danielhoherd (approved):
Looks good!

PR #1628, opened 2022-08-08
fluentd 1.15.1

danielhoherd:
## Description

fluentd version 1.15.1 (which uses ubuntu 22.04 as base image instead of 20.04)

## Related Issues

https://github.com/astronomer/issues/issues/4908

## Testing

None so far. Normal fluentd tests should be fine: just making sure logs are shipping, no unexpected output it seen in the fluentd logs.

## Merging

Merge everywhere.


shmanu017 (approved):

danielhoherd:
Merged and cherry-picked to releases 0.30, 0.29. Have not merged to 0.28 or earlier though because it looks like those may be deliberately held back to fluentd 1.14.x

PR #1680, opened 2022-10-03
es-client customize 

shmanu017:
- Add the ability to customize securityContext for es-client container

https://github.com/astronomer/issues/issues/4997

Can update/remove the values via setting `--set elasticsearch.client.securityContext.capabilities.add={}`


pgvishnuram:
I don't see any difference from moving from es chart to default values

charts/elasticsearch/templates/client/es-client-deployment.yaml
```diff
@@ -87,10 +87,7 @@ spec:
       - name: es-client
         securityContext:
 {{ toYaml .Values.securityContext | indent 10 }}
-          capabilities:
-            add:
-              - IPC_LOCK
-              - SYS_RESOURCE
+{{ toYaml .Values.securityContext_esclient | indent 10 }}
```
danielhoherd, line 90: What is the behavior here if there are conflicting `capabilities` defined in `Values.securityContext` and `Values.securityContext_esclient` ?

danielhoherd: I don't like how it ends up defining the same keys multiple times. We should find a way to merge the second one into the first one.

danielhoherd: @shmanu017 actually i think this is pretty straightforward. we can just do this as an `if ... else` block, saying `if Values.securityContext_esclient` use those, otherwise use `Values.securityContext`. That will keep it from double-defining that block.

charts/elasticsearch/templates/client/es-client-deployment.yaml
```diff
@@ -86,11 +86,11 @@ spec:
...
         securityContext:
+        {{- if .Values.securityContext_esclient }}
+{{ toYaml .Values.securityContext_esclient | indent 10 }}
+        {{- else }}
 {{ toYaml .Values.securityContext | indent 10 }}
-          capabilities:
-            add:
-              - IPC_LOCK
-              - SYS_RESOURCE
+        {{- end }}
```
danielhoherd, line 93: Now that we have an if/else, we should add a test to show that the client config overrides the global config.

tests/chart_tests/test_elasticsearch.py
```diff
@@ -120,6 +120,30 @@ def test_elasticsearch_securitycontext_defaults(self, kube_version):
...
+            },
+            show_only=[
+                "charts/elasticsearch/templates/client/es-client-deployment.yaml",
+            ],
+        )
+        assert len(docs) == 1
+        doc = docs[0]
+        pod_data = doc["spec"]["template"]["spec"]["containers"][0]
+        assert pod_data["securityContext"]["capabilities"]["add"] == ["IPC_LOCK"]
+        assert pod_data["securityContext"]["capabilities"]["add"] != ["SYS_RESOURCE"]
```
danielhoherd, line 145: With the way this test is written, this line is redundant. The prior line checks the whole `pod_data["securityContext"]["capabilities"]["add"]` data structure to make sure it is equal to a list with one item whose value is `IPC_LOCK`. That inherently excludes `SYS_RESOURCE` from the list.
```suggestion
```

danielhoherd (changes requested):
Looks good so far. I'd like to see the same feature imlpemented for es-master and es-data, just so we have consistent mechanics across all parts of the es chart.

charts/elasticsearch/values.yaml
```diff
@@ -65,6 +65,13 @@ common:
 # by the ingest processors and the required resources, it may make sense to
 # have dedicated ingest nodes, that will only perform this specific task.
 client:
+
+  securityContext:
+    capabilities:
+      add:
+        - IPC_LOCK
+        - SYS_RESOURCE
+        -
```
pgvishnuram, line 74: extra - to be removed

pgvishnuram:
Im not sure why for es-data and es-master need seperate securityContext.

charts/elasticsearch/values.yaml
```diff
@@ -65,6 +65,12 @@ common:
 # by the ingest processors and the required resources, it may make sense to
 # have dedicated ingest nodes, that will only perform this specific task.
 client:
+
+  securityContext:
+    capabilities:
+      add:
+        - IPC_LOCK
+        - SYS_RESOURCE
```
pgvishnuram, line 73: @manu / @danielhoherd - We can remove since elastic doesn't need any extra priv. There default is DROP.

shmanu017: @pgvishnuram we need custom permissions for client other than drop that's the reason its present.

danielhoherd: @shmanu017 can we comment it out with a working `drop` example?

shmanu017: Global helm values set `charts/elasticsearch/values.yaml` aren't getting overridden to null values in the test case to disable client security context to accept global security context.Since, we will not be facing this condition not adding test in this case.

tests/chart_tests/test_elasticsearch.py
```diff
@@ -120,6 +120,31 @@ def test_elasticsearch_securitycontext_defaults(self, kube_version):
             assert pod_data["securityContext"]["runAsNonRoot"] is True
             assert pod_data["securityContext"]["runAsUser"] == 1000
 
+    def test_elasticsearch_client_securitycontext_overrides(self, kube_version):
+        """Test ElasticSearch client with securityContext custom values."""
```
danielhoherd, line 124: We need this same sort of test for data and master.

danielhoherd (changes requested):
Looks good so far, but we need more tests to cover data and master the same way we cover client.

charts/elasticsearch/values.yaml
pgvishnuram: # Dont set any security context to es-client since it will need privileged access to run in openshift 
securityContext: []

danielhoherd (approved):
Shipit!

PR #1690, opened 2022-10-12
bug-fix-sidecar-fails-to-terminate

anilsaini1212:
## Description

Sidecar Occasionally Fails To Terminate

## Related Issues

curl query changed to 

`curl -fsSL --retry 5 --max-time 10 --retry-connrefused -XPOST {{ .Values.global.loggingSidecar.terminationEndpoint }}"`

Earlier it was 

`curl -fsSL -XPOST {{ .Values.global.loggingSidecar.terminationEndpoint }}"`

## Testing

There doesn't seem to be an easy way to test the added behavior because we don't have a rich functional test setup. I'd say we will have to leave this as "just make sure sidecar log containers continue to function" for now. (filled in by @danielhoherd)

## Merging

Merge everywhere that sidecar logging exists. (filled in by @danielhoherd)

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -258,7 +258,7 @@ data:
                   if container.args[0:3] == ["airflow", "tasks", "run"]:
                       container.command = ["tini", "--"]
                       new_args = ["bash", "-c"]
-                      command_str = "/entrypoint " + ' '.join([str(arg) for arg in container.args]) + " " + log_cmd + "; curl -fsSL -XPOST {{ .Values.global.loggingSidecar.terminationEndpoint }}"
+                      command_str = "/entrypoint " + ' '.join([str(arg) for arg in container.args]) + " " + log_cmd + "; curl -fsSL --retry 5 --max-time 10 --retry-connrefused -XPOST {{ .Values.global.loggingSidecar.terminationEndpoint }}"
                       new_args.append(command_str)
                       container.args = new_args
 


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -258,7 +258,7 @@ data:
                   if container.args[0:3] == ["airflow", "tasks", "run"]:
                       container.command = ["tini", "--"]
                       new_args = ["bash", "-c"]
-                      command_str = "/entrypoint " + ' '.join([str(arg) for arg in container.args]) + " " + log_cmd + "; curl -fsSL -XPOST {{ .Values.global.loggingSidecar.terminationEndpoint }}"
+                      command_str = "/entrypoint " + ' '.join([str(arg) for arg in container.args]) + " " + log_cmd + "; curl -fsSL --retry 5 --max-time 10 --retry-connrefused -XPOST {{ .Values.global.loggingSidecar.terminationEndpoint }}"
```
pgvishnuram, line 261: @danielhoherd should we make this configurable or let it be defaults

anilsaini1212: @pgvishnuram @danielhoherd 
This works occasionally,not a correct bug fix

pgvishnuram: Okay - let me test this with multiple short lived tasks

danielhoherd: I really like this `--retry 5 --max-time 10 --retry-connrefused` addition. This seems like a fantastic way to handle this problem.

vishwas-astro:
@danielhoherd @pgvishnuram - where are we with respect to getting this merged?

vishwas-astro:
@anilsaini1212 @pgvishnuram please create a GH issue for it so we can review and merge this.

danielhoherd:
https://github.com/astronomer/issues/issues/5116 exists, so i think this is good to merge.

danielhoherd (approved):

PR #1708, opened 2022-10-19
add cron jobs to registry network policy

bote795:
## Description

add cron job network policies for registry, two new cron jobs were unable to communicate with registry we forgot to add the network policies 

## Related Issues

Related https://github.com/astronomer/issues/issues/5118
## Testing

will be tested in dev

## Merging

0.31

### DIFF
--- a/charts/astronomer/templates/registry/registry-networkpolicy.yaml
+++ b/charts/astronomer/templates/registry/registry-networkpolicy.yaml
@@ -72,10 +72,10 @@ spec:
           component: houston-worker
           release: {{ .Release.Name }}
     - podSelector:
-      matchLabels:
-        tier: astronomer
-        release: {{ .Release.Name }}
-        component: houston-populate-daily-task-metrics
+        matchLabels:
+          tier: astronomer
+          release: {{ .Release.Name }}
+          component: houston-populate-daily-task-metrics
     - podSelector:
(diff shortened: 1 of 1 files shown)


danielhoherd (approved):

charts/astronomer/templates/registry/registry-networkpolicy.yaml
```diff
@@ -71,6 +71,16 @@ spec:
           tier: astronomer
           component: houston-worker
           release: {{ .Release.Name }}
+    - podSelector:
+      matchLabels:
+        tier: astronomer
+        release: {{ .Release.Name }}
+        component: houston-populate-daily-task-metrics
```
danielhoherd, line 78:
```suggestion
        matchLabels:
          tier: astronomer
          release: {{ .Release.Name }}
        component: houston-populate-daily-task-metrics
```

danielhoherd: You know what? If that slipped through unit tests, it means we need unit tests to validate that this file is valid.

danielhoherd: I'm going to add tests in a different branch so we can unblock this, and because it's becoming quite messy.

PR #1711, opened 2022-10-19
5137 fix schema validation

danielhoherd:
## Description

Fix schema validation by using strict schema. We ran into some networkPolicy errors that should have been caught by schema validation. Using strict caught them, and another error that existed in our chart.

While debugging the schema validation problem I also added a previously impossible test by adding headers to some templates. And while adding those headers, I lined up some header contents on some other files.

## Related Issues

https://github.com/astronomer/issues/issues/5137

## Testing

These are tests. I don't think any further testing needs to be done.

## Merging

AFAIK these changes should be merged into all supported branches.

### DIFF
--- a/charts/alertmanager/templates/alertmanager-networkpolicy.yaml
+++ b/charts/alertmanager/templates/alertmanager-networkpolicy.yaml
@@ -1,6 +1,6 @@
 ################################
 ## Alertmanager NetworkPolicy
-#################################
+################################
 {{- if .Values.global.networkPolicy.enabled }}
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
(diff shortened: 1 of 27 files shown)


pgvishnuram (approved):
externalIPs - we have not noticed the typo error

danielhoherd:
Merged and cherry-picked the relevant schema validation and externalIPs changes to all supported branches. The header changes and np_disabled stuff didn't apply to all branches so that was left out if it wasn't relevant.

vishwas-astro:
@danielhoherd Will require a 0.30.3-rc3 if we want this included in 0.30.3.

danielhoherd:
@vishwas-astro this is not an urgent change. There was a bug fix, but no customers had complained about it, so it's very likely going to be a no-op for everybody. The real meat of this change is we are getting better k8s schema validation, which would have caught a bug during dev iterations on a recent feature addition.

PR #1715, opened 2022-10-24
Add PoC of migrating airflow db

andriisoldatenko:
This PR is just PoC to demonstrate how to migrate Airflow
 deployment to new PostgreSQL instance.
 
 I've tested in local cluster with 2 DB instances;
 Also it works if you migrate to new DB and click upgrade deployment from UI: e.g. change executor type.

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -66,6 +66,7 @@ data:
     deployments:
       fluentdIndexPrefix: {{ include "fluentd.IndexPattern" .}}
       enableHoustonInternalAuthorization: {{ include "houston.InternalAuthorization" . }}
+      dagOnlyDeployment: {{ .Values.global.dagOnlyDeployment.enabled }}
       namespaceFreeFormEntry: {{ .Values.global.namespaceFreeFormEntry }}
       # Airflow chart settings
       # Static helm configurations for this chart are found below.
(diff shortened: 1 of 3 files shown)


bin/migrate_airflow_db.sh
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env bash
+
```
danielhoherd, line 2: Fail if variables are not defined.

```suggestion
set -u

```

bin/migrate_airflow_db.sh
```diff
@@ -0,0 +1,119 @@
...
+  PG server 2 and airflow deployment.
+
+  Assumption 2: you have access to both DB instanceses so you can run pg_dump pg_restore
+
+  Assumption 3:  Make sure you pg_dump/pg_restore versions matched
+"
+
+
+export PATH="/usr/local/opt/postgresql@11/bin:$PATH"
+ASTRONOMER_AIRFLOW_CHART="astronomer-ee/airflow"
```
danielhoherd, line 20: This assumes that the helm repo is named `astronomer-ee`. We should make this variable.

bin/migrate_airflow_db.sh
```diff
@@ -0,0 +1,119 @@
...
+  Assumption 1: user is responsible of making any networking connectivity between
+  PG server 2 and airflow deployment.
+
+  Assumption 2: you have access to both DB instanceses so you can run pg_dump pg_restore
+
+  Assumption 3:  Make sure you pg_dump/pg_restore versions matched
+"
+
+
+export PATH="/usr/local/opt/postgresql@11/bin:$PATH"
```
danielhoherd, line 19: This postgresql bin path should be variable.

bin/migrate_airflow_db.sh
```diff
@@ -0,0 +1,119 @@
...
+ASTRONOMER_AIRFLOW_CHART="astronomer-ee/airflow"
+RELEASE_NAME="${1:-dynamical-nucleus-9402}"
```
danielhoherd, line 21: We should not have a default value for RELEASE_NAME
```suggestion
RELEASE_NAME="$1"
```

danielhoherd:
I think this is a good start, but I'd like to see a lot of cleanup if we're planning on giving this to customers. It seems like this is the kind of thing that would be used as a reference, so it could likely have a lot of comments in it explaining what is going on, or better yet, be written as a markdown doc. If we're planning on giving customers something that we expect them to actually run, then we should definitely improve the code quality.

PR #1718, opened 2022-10-26
allow updates url to be added as env vars to houston

pgvishnuram:
## Description

Allow updates url for runtime and certified to be passed to env vars.
This resolves an issue where customer/user to add url separately as env vars to houston pods

## Related Issues

https://github.com/astronomer/issues/issues/5127

## Testing

## Merging

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -226,6 +226,16 @@ s3:
   value: {{ template "houston.nats.servers" . }}
 - name: NATS__CLUSTER_ID
   value: {{ .Release.Name }}-stan
+{{- if .Values.houston.updateAirflowCheck.enabled }}
+- name: HOUSTON_SCRIPT_UPDATE_SERVICE_URL
+  value: {{ .Values.houston.updateAirflowCheck.url }}
+{{- end }}
+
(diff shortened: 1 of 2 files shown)


andriisoldatenko (approved):
thanks!

andriisoldatenko:
after thinking a bit more:
better would be to move this code to appropriate job:
https://github.com/astronomer/astronomer/blob/41a3f89484d1f8e333c3cd72adb7c0bb924ab47e/charts/astronomer/templates/houston/cronjobs/houston-check-airflow-version-updates-cronjob.yaml#L52-L53

Instead of adding it to every houston pod.

andriisoldatenko:
> after thinking a bit more: better would be to move this code to appropriate job:
> 
> https://github.com/astronomer/astronomer/blob/41a3f89484d1f8e333c3cd72adb7c0bb924ab47e/charts/astronomer/templates/houston/cronjobs/houston-check-airflow-version-updates-cronjob.yaml#L52-L53
> 
> Instead of adding it to every houston pod.

@pgvishnuram just suggested we use it in houston too not only cron-jobs. So PR LGTM.

danielhoherd (changes requested):
Looks good. Needs tests.

PR #1725, opened 2022-11-03
test against kubernetes 1.24

danielhoherd:
## Description

Run tests against k8s 1.24.

## Related Issues

https://github.com/astronomer/issues/issues/5148

## Testing

No manual testing needed.

## Merging

The 1.24 bits of this should be merged into any branches that support 1.23. The only important changes are the generate_circleci_config.py and tests/__init__.py changes, everything else is auto-generated, so if cherry-picking fails we can just manually make those changes.


shmanu017 (approved):
LGTM

PR #1726, opened 2022-11-04
Skip externalTrafficPolicy when using ClusterIP

danielhoherd:
## Description

In #1052 we merged a change that was mostly good, but had one edge case that was broken, which is that when the serviceType is ClusterIP, there can be no externalTrafficPolicy.

## Related Issues

- Internal ticket describing the core problem being fixed here: <https://github.com/astronomer/issues/issues/4991>
- Original PR where the change that caused this problem was made: <https://github.com/astronomer/astronomer/pull/1052>

## Testing

I've added tests for the new code change as well as previously untested configuration combinations.

## Merging

The original problem showed up in 0.29.2, so this should be merged to 0.29, 0.30, 0.31.

## More details

- <https://kubernetes.io/docs/concepts/services-networking/service/#external-traffic-policy>
- <https://www.asykim.com/blog/deep-dive-into-kubernetes-external-traffic-policies>

### DIFF
--- a/charts/nginx/templates/nginx-service.yaml
+++ b/charts/nginx/templates/nginx-service.yaml
@@ -41,7 +41,7 @@ spec:
   {{- if .Values.loadBalancerSourceRanges }}
   loadBalancerSourceRanges: {{ .Values.loadBalancerSourceRanges | toJson }}
   {{- end }}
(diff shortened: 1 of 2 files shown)


rbankston (approved):
Thank you Daniel. Looks good to me

PR #1463, opened 2022-04-13
Test for duplicate env vars

danielhoherd:
## Description

Test that none of our containers have duplicate env vars. I'm not sure this is actually productive work since k8s has a policy for handling duplicate env vars, so I'm not sure if we should complete this work or leave it as is. More discussion in the linked issue.

The new test class is parametrized in a way that generate test names that are easy to grok:

```
tests/test_default_chart.py::TestDuplicateEnvironment.test_env_vars_have_no_duplicates[DaemonSet/release-name-fluentd] ✓
tests/test_default_chart.py::TestDuplicateEnvironment.test_env_vars_have_no_duplicates[DaemonSet/release-name-prometheus-node-exporter] ✓
tests/test_default_chart.py::TestDuplicateEnvironment.test_env_vars_have_no_duplicates[Deployment/release-name-astro-ui] ✓
tests/test_default_chart.py::TestDuplicateEnvironment.test_env_vars_have_no_duplicates[Deployment/release-name-cli-install] ✓
tests/test_default_chart.py::TestDuplicateEnvironment.test_env_vars_have_no_duplicates[Deployment/release-name-commander] ✓
```

Assertions also identify if the problem was in the container or initContainer:

```
E               AssertionError: container has duplicate env vars
E               assert ['DATABASE_URL'] == []
E                 Left contains one more item: 'DATABASE_URL'
...
E               AssertionError: initContainer has duplicate env vars
E               assert ['DATABASE_URL'] == []
E                 Left contains one more item: 'DATABASE_URL'
```

## Related Issues

<https://github.com/astronomer/issues/issues/4393>

## Testing

This is all testing so nothing more is needed.

## Merging

If accepted, this wouldn't be a bad thing to merge into any other supported branches. The test is structured in a way where it should work fine without breaking in other branches.

PR #1747, opened 2022-11-18
Revert changes made in #1463 in order to avoid triggering kubernetes bug

danielhoherd:
## Description

Revert changes made in #1463, which unfortunately triggered a kubernetes bug that would delete all instances of the env var instead of just the duplicate. The original PR was just cleanup, and didn't have a functional advantage, so this should be a no-op. We should put this check back when the kubernetes bug has been fixed.

## Links

- Upstream issue describing the bug we triggered: https://github.com/kubernetes/kubernetes/issues/58477
- Some slack discussion: <https://astronomer.slack.com/archives/C03JM4MV8LF/p1667500806445899?thread_ts=1667422605.740839&cid=C03JM4MV8LF>

## Related Issues

- https://github.com/astronomer/issues/issues/5217

## Testing

These changes and the duplicate all use the same logic, so there's really nothing to test. The same exact env var will be defined twice.

## Merging

This needs to go into 0.29, 0.30 and 0.31. 0.29.2 is where the bug was introduced.


andriisoldatenko (approved):

andriisoldatenko:
yeah may be add a comment, since it second attempt ?

danielhoherd:
@andriisoldatenko good call. I'm going to put the test back and have it skip.

PR #1776, opened 2022-12-28
Adding timeoutSeconds option in the prometheus block #5324

shubh-am-kumar:
## Description

Adding `timeoutSeconds` option in liveness and readiness for the prometheus block in the helm chart to make the value persistent.

Customers are not able to set timeoutSeconds in the prometheus block through the helm.

## Related Issues

https://github.com/astronomer/issues/issues/5324

## Testing

This can be tested by providing any numeric value to `timeoutSeconds` in values.YAML file and verifying the same in the Prometheus STS.

## Merging

"This should be safe to merge to all supported branches." - DanielH

### DIFF
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -37,13 +37,13 @@ resources: {}
 livenessProbe:
   initialDelaySeconds: 10
   periodSeconds: 5
-  failureThreshold: 3
   timeoutSeconds: 1
+  failureThreshold: 3
 readinessProbe:
   initialDelaySeconds: 10
   periodSeconds: 5
-  failureThreshold: 3
   timeoutSeconds: 1
+  failureThreshold: 3
 
 configMapReloader:
   resources: {}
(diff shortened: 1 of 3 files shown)


danielhoherd (changes requested):
Looking good. Thank you for taking the effort to update the tests! Just need the PR description to be completed and this will be good to merge.

We likely don't need any human testing on this because it will all get caught by unittests and functional tests in CI. This should be safe to merge to all supported branches.

charts/prometheus/values.yaml
```diff
@@ -38,10 +38,12 @@ livenessProbe:
   initialDelaySeconds: 10
   periodSeconds: 5
   failureThreshold: 3
+  timeoutSeconds: 1
 readinessProbe:
   initialDelaySeconds: 10
   periodSeconds: 5
   failureThreshold: 3
+  timeoutSeconds: 1
```
danielhoherd, line 46:
```suggestion
  initialDelaySeconds: 10
  periodSeconds: 5
  timeoutSeconds: 1
  failureThreshold: 3
readinessProbe:
  initialDelaySeconds: 10
  periodSeconds: 5
  timeoutSeconds: 1
  failureThreshold: 3
```

pgvishnuram (approved):

danielhoherd (approved):

PR #1778, opened 2023-01-04
Add delay for nats.ackWait, upgradeTimeout. Improve some test syntaxes.

danielhoherd:
## Description

Increase timeouts related to airflow upgrades. This is due to a new dag parsing behavior in Airflow that makes upgrades take longer when there are lots of dags present.

## Related Issues

- Issue: https://github.com/astronomer/issues/issues/5301
- These changes were already applied to Nebula in https://github.com/astronomer/google-environments/pull/1031 

## Testing

Tests are added, but it wouldn't be bad to double-check them in QA testing.

## Merging

This should be merged to all supported releases because it addresses a core airflow behavior.

### DIFF
--- a/tests/chart_tests/test_astronomer_commander.py
+++ b/tests/chart_tests/test_astronomer_commander.py
@@ -31,7 +31,7 @@ class TestAstronomerCommander:
         assert c_by_name["commander"]["resources"]["limits"]["memory"] == "2Gi"
         assert c_by_name["commander"]["resources"]["requests"]["memory"] == "1Gi"
         env_vars = {x["name"]: x["value"] for x in c_by_name["commander"]["env"]}
-        assert env_vars["COMMANDER_UPGRADE_TIMEOUT"] == "300"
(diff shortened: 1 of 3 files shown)


tests/chart_tests/test_astronomer_commander.py
```diff
@@ -41,7 +41,7 @@ def test_astronomer_commander_deployment_upgrade_timeout(self, kube_version):
         """
         docs = render_chart(
             kube_version=kube_version,
-            values={"astronomer": {"commander": {"upgradeTimeout": 600}}},
+            values={"astronomer": {"commander": {"upgradeTimeout": 997}}},
```
aliotta, line 44: Im surprised to see 997 here thats a curious number

danielhoherd: The prior value for the non-default test was 600, which is what the new default value is, so I set the non-default test value to something conspicuous.

PR #1795, opened 2023-01-26
Test for pod securityContexts running as non-root

shmanu017:
## Description

Test for pod securityContexts running as non-root

## Related Issues

https://github.com/astronomer/issues/issues/5339


charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
```diff
@@ -39,6 +39,8 @@ spec:
         - name: astro-ui
           image: {{- include "astroUI.image" . | indent 1 }}
           imagePullPolicy: {{ .Values.images.astroUI.pullPolicy }}
+          securityContext:
+{{ toYaml (.Values.securityContext)| indent 12 }}
```
danielhoherd, line 43:
```suggestion
          securityContext:{{ toYaml (.Values.securityContext)| nindent 12 }}
```

charts/nats/values.yaml
```diff
@@ -13,6 +13,7 @@ images:
     tag: 0.10.0-3
     pullPolicy: IfNotPresent
 
+
```
danielhoherd, line 16:
```suggestion

```

charts/postgresql/templates/statefulset-slaves.yaml
```diff
@@ -91,7 +92,8 @@ spec:
         {{- if .Values.securityContext.enabled }}
         securityContext:
           runAsUser: {{ .Values.securityContext.runAsUser }}
-        {{- end }}
+          runAsNonRoot: { { .Values.securityContext.runAsNonRoot } }
```
danielhoherd, line 95:
```suggestion
          runAsNonRoot: {{ .Values.securityContext.runAsNonRoot }}
```

charts/prometheus-blackbox-exporter/values.yaml
```diff
@@ -11,6 +11,7 @@ strategy:
     maxUnavailable: 1
   type: RollingUpdate
 
+
```
danielhoherd, line 14: Are you adding these blank lines intentionally? I made a suggestion above to remove them, but if you're adding them on purpose then disregard the suggestions.
```suggestion
```

danielhoherd (commented):
Looking good so far. I left some optional suggestions, but I think the `prettier` pre-commit hook should take care of them.

charts/postgresql/templates/statefulset-slaves.yaml
```diff
@@ -91,7 +92,8 @@ spec:
         {{- if .Values.securityContext.enabled }}
         securityContext:
           runAsUser: {{ .Values.securityContext.runAsUser }}
-        {{- end }}
+          runAsNonRoot: {{ .Values.securityContext.runAsNonRoot }}
+  {{- end }}
```
danielhoherd, line 96: This should be aligned with the `if` block, which I think is:
```suggestion
        {{- end }}
```

tests/chart_tests/test_non_root_user.py
```diff
@@ -0,0 +1,40 @@
...
+ignore_kind_list = ["Job"]
+ignore_list = [
+    "fluentd_fluentd",
+    "elasticsearch-exporter_metrics-exporter",
+    "elasticsearch-nginx_nginx",
+]
+
+
+def init_test_non_root_user():
+    chart_values = chart_tests.get_all_features()
```
danielhoherd, line 16: Please add a docstring and type annotation here so it's easier to understand.

tests/chart_tests/test_non_root_user.py
```diff
@@ -0,0 +1,40 @@
...
+
+    containers = {}
+    for k8s_version in supported_k8s_versions:
+        k8s_version_containers = chart_tests.get_chart_containers(
+            k8s_version, chart_values, ignore_kind_list
+        )
+        containers = {**containers, **k8s_version_containers}
```
danielhoherd, line 23: This might be better understood as a dict comprehension, and would be shorter.

tests/chart_tests/test_non_root_user.py
```diff
@@ -0,0 +1,40 @@
...
+
+    return containers
+
+
+class TestProbes:
+    chart_containers = init_test_non_root_user()
+
+    @pytest.mark.parametrize(
+        "container", chart_containers.values(), ids=chart_containers.keys()
+    )
```
danielhoherd, line 33: Would it make more sense to have this parametrize on the class definition?

tests/chart_tests/test_non_root_user.py
```diff
@@ -0,0 +1,40 @@
...
+    def test_container_runasnonroot(self, container):
+        """Ensure all containers have runAsNonRoot."""
+
+        if re.split("(?<=name-)(.*$)", container["key"])[1] in ignore_list:
```
danielhoherd, line 37: We should probably use some form of `yaml.load` here instead of regex. Doing so would be more specific, more legible, and potentially use fewer resources.

danielhoherd (approved):
LGTM

shmanu017:
Changes are working as expecting on a Test cluster, need to update image tags before merging to master.

charts/alertmanager/values.yaml
```diff
@@ -9,11 +9,12 @@ tolerations: []
 images:
   alertmanager:
     repository: quay.io/astronomer/ap-alertmanager
-    tag: 0.25.0
+    tag: 0.25.0-1
     pullPolicy: IfNotPresent
 
 securityContext:
     runAsNonRoot: true
+    runAsUser: 65534
```
danielhoherd, line 17: Is this needed? Does the docker image already run as 65534?

charts/astronomer/values.yaml
```diff
@@ -15,16 +15,16 @@ tolerations: []
 images:
   commander:
     repository: quay.io/astronomer/ap-commander
-    tag: 0.31.4
+    tag: master
```
danielhoherd, line 18: This will need to be reverted.

charts/astronomer/values.yaml
```diff
@@ -15,16 +15,16 @@ tolerations: []
...
   registry:
     repository: quay.io/astronomer/ap-registry
-    tag: 3.16.3
+    tag: 3.16.3-1
     pullPolicy: IfNotPresent
     # httpSecret: ~
   houston:
     repository: quay.io/astronomer/ap-houston-api
-    tag: 0.31.15
+    tag: master
```
danielhoherd: This will also need to be reverted.

charts/kube-state/values.yaml
```diff
@@ -9,11 +9,12 @@ tolerations: []
 images:
   kubeState:
     repository: quay.io/astronomer/ap-kube-state
-    tag: 2.7.0
+    tag: 2.7.0-1
     pullPolicy: IfNotPresent
 
 securityContext:
     runAsNonRoot: true
+    runAsUser:
```
danielhoherd, line 17: This needs a value.

shmanu017:
`commander` , `houston-api` and `db-bootstraper` are currently pointing to master or main , as it will either way fail when running  test on cluster enlivenment.
Waiting on merge the change until the release tag are released.

charts/astronomer/values.yaml
```diff
@@ -55,7 +58,7 @@ astroUI:
 houston:
   prismaConnectionLimit: 5
   replicas: 2
-  # This only applies when replicas > 3
+  # This only applies when replicas >
```
danielhoherd, line 61: I assume this deletion was a mistake
```suggestion
  # This only applies when replicas > 3
```

danielhoherd (approved):
LGTM. I like the two additions of PodSecurityContext if those are the only two places that differentiation is needed.

PR #1804, opened 2023-02-02
allow fluentd to gather gitsync relay pod logs

pgvishnuram:
## Description

currently we don't capture gitsync relay pod logs . In case after deployment it goes into error state ( like clone error or branch error) , customer should get access to k8s cluster for that release and verify those errors. This way it can send them to elasticsearch and can be used for further processing

## Related Issues

https://github.com/astronomer/issues/issues/5246

## Testing

## Merging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
 
   run_pre_commit:
     docker:
-      - image: quay.io/astronomer/ci-pre-commit:2023-01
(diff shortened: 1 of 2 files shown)


pgvishnuram:
@rishkarajgi @danielhoherd - This is a initial work 
we should do the same for vector as well

rishkarajgi (approved):

PR #1808, opened 2023-02-06
add alertmanager extra volumes and volume mounts

dintorf:
## Description

Add `extraVolumes` and `extraVolumeMounts` to allow Kubernetes secrets to be mounted to `altertmanager`. Mounting secrets enables webhook alerts to be configured with password files instead of plain text passwords in helm values.

## Related Issues

Resolves: https://github.com/astronomer/issues/issues/5396

## Testing

Add the following to the helm values for the astronomer chart:

```yaml
alertmanager:
  extraVolumes:
    - name: webhook-alert-secret
       secret:
         secretName: webhook-alert-secret
  extraVolumeMounts:
    - mountPath: "/var/webhook-alert-secret"
       name: webhook-alert-secret
       readOnly: true
```

Create a secret in Kubernetes called `webhook-alert-secret` in the astronomer namespace and add a password to the data. Helm upgrade astronomer, then exec into alertmanager pod looking for the password file in `/var/webhook-alert-secret`.

## Merging

If this is approved, I'd like to have this backported to 0.30 if possible :)


dintorf:
If this is approved, I'd like to have this backported to 0.30 if possible :)

tests/chart_tests/test_alertmanager.py
```diff
@@ -88,3 +92,39 @@ def test_alertmanager_customReceiver():
...
+
+    assert "webhook-alert-secret" in [
+        x["name"] for x in jmespath.search("spec.template.spec.volumes", docs[0])
+    ]
+    assert "webhook-alert-secret" in [
+        x["name"]
+        for x in jmespath.search(
+            "spec.template.spec.containers[0].volumeMounts", docs[0]
+        )
+    ]
```
danielhoherd, line 130: We're actually trying to get away from using jmespath due to some syntactical sharp edges. 

```suggestion
    doc = render_chart(
        kube_version=kube_version,
        values=values,
        show_only=["charts/alertmanager/templates/alertmanager-statefulset.yaml"],
    )[0]

    c_by_name = get_containers_by_name(doc, include_init_containers=False)

    assert "webhook-alert-secret" in [
        x["name"] for x in doc["spec"]["template"]["spec"]["volumes"]
    ]
    assert "webhook-alert-secret" in [
        x["name"] for x in c_by_name["alertmanager"]["volumeMounts"]
    ]
```

requirements/chart-tests.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.9
```
danielhoherd, line 2: Is this change needed as part of this PR? It seems like it's not. It would be great to take this requirements change out of the PR.

dintorf: reverted 👍

tests/chart_tests/test_alertmanager.py
dintorf: ACK - made the suggested changes, thanks!

dintorf:
Thanks for the feedback, changes are in 👍

requirements/chart-tests.txt
dintorf: wasn't required for the change, just got caught up in my updates so I could run tests locally

danielhoherd (approved):

PR #1813, opened 2023-02-16
Confine user actions to the index of the user performing the action

danielhoherd:
## Description

Add nginx rewrite rules that lock all user operations into their own indices.

## Related Issues

- https://github.com/astronomer/issues/issues/5417
- https://astronomer.slack.com/archives/C04P3TRUV3L

## Testing

This should be tested with both sidecar and vector.

@jedcunningham and I have personally tested these changes in a dedicated AWS cluster and they all look good to us.

## Merging

These changes should be good to merge everywhere.

### DIFF
--- a/charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml
+++ b/charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml
@@ -36,6 +36,24 @@ data:
         proxy_set_header Connection "Keep-Alive";
         proxy_set_header Proxy-Connection "Keep-Alive";
 
+        location ~* /_count$ {
+          {{- if .Values.global.loggingSidecar.enabled }}
+          rewrite /_count(.*) /vector.$remote_user.*/_count$1 break;
+          {{- else }}
+          rewrite /_count(.*) /fluentd.$remote_user.*/_count$1 break;
+          {{- end }}
+          proxy_pass http://elasticsearch;
+        }
+
+        location ~* /_bulk$ {
+          {{- if .Values.global.loggingSidecar.enabled }}
+          rewrite /_bulk(.*) /vector.$remote_user.*/_bulk$1 break;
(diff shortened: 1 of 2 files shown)


tests/chart_tests/test_elasticsearch.py
```diff
@@ -304,11 +304,20 @@ def test_nginx_es_index_pattern_defaults(self, kube_version):
...
-        es_index = doc["data"]["nginx.conf"]
         assert doc["kind"] == "ConfigMap"
-        assert "fluentd.$remote_user.*/$1" in es_index
 
-    def test_nginx_es_index_pattern_with_sidecar_logging_enabled(self, kube_version):
+        nginx_config = " ".join(doc["data"]["nginx.conf"].split())
+        assert all(
+            x in nginx_config
+            for x in [
+                "location ~* /_search$ { rewrite /_search(.*) /fluentd.$remote_user.*/_search$1 break;",
```
jedcunningham, line 313: Should we have a check for bulk and count also?

charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml
```diff
@@ -57,5 +69,8 @@ data:
           proxy_set_header Content-Length "";
           proxy_set_header X-Original-URI $request_uri;
         }
+        location ~ ^/ {
+          deny all;
```
jedcunningham, line 73: Should we add a comment (above somewhere, maybe not right here) stating that any additional proxy passes must be scoped to the deployment index only? Might still be missed, but at least its better than nothing?

tests/chart_tests/test_elasticsearch.py
danielhoherd: Yes! I just committed them.

jedcunningham (approved):

PR #1832, opened 2023-03-10
Add /_cluster/health, anchor other regexes

danielhoherd:
## Description

- Allow /_cluster/health
- Refactor tests to not use jmespath

## Related Issues

- <https://github.com/astronomer/issues/issues/5473>

## Testing

Normal test cycle should cover this. I have tested this change in aws.

## Merging

This should be merged everywhere.


charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml
```diff
@@ -38,29 +38,29 @@ data:
 
         # The following "location" rules limit airflow interactions to only their indices. Any further
         # additions should follow this pattern.
-        location ~* /_count$ {
+        location ~* ^/_count$ {
```
jedcunningham, line 41: I don't think this will work, as we need to redirect the `/_all/_count` form as well, that is used in provider versions [4.3.2+](https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/index.html#id3).

danielhoherd: What about doing `^/(_all/)?_count` ?

jedcunningham: If you also cover `/*/_count` (literal star), should be fine. `_all` and `*` both mean all indices.

The good news is if it doesn't match, it just hits the deny block anyways.

danielhoherd: I tried to tackle this today but ran into the error `invalid number of arguments in "rewrite" directive in /etc/nginx/nginx.conf:25` when trying out:

```
        location ~* ^/(_all/|\*/)?_count$ {
          {{- if .Values.global.loggingSidecar.enabled }}
          rewrite ^/(_all/|\*/)?_count(.*) /vector.$remote_user.*/_count$1 break;
          {{- else }}
          rewrite ^/(_all/|\*/)?_count(.*) /fluentd.$remote_user.*/_count$1 break;
          {{- end }}
          proxy_pass http://elasticsearch;
        }
```

danielhoherd: I'm going to take this change out and maybe merge it in a next iteration.

jedcunningham: Sorry, I didn't mean to imply this is limited to _count, it applies so the other endpoints as well.

jedcunningham: You'll want to take out all of the anchoring changes.

jedcunningham (approved):

PR #1834, opened 2023-03-14
List of verison

shmanu017:
## Description

Creating a sample table to list out all the images used in astronomer/astronomer branch for individual versions.

## Related issue 

https://github.com/astronomer/issues/issues/5479

### DIFF
--- a/image.md
+++ b/image.md
@@ -1,0 +1,37 @@
+## Images used to astronomer Software Installation
+
+
+
+| IMAGE                | Version     |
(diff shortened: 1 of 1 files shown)


danielhoherd (changes requested):
PR description needs a ticket link, slack link, etc., and details about why this is needed. EG: why is this not sufficient? https://github.com/astronomer/astronomer/blob/902ff8d5ccad/.circleci/config.yml#L301-L333

PR #1842, opened 2023-03-30
Adding Trigger to call feature_stack deployment

mishah334:
## Description

Adding a trigger to call the `feature_stack` workflow that performs the deployment on the cluster. This will add the capability to refresh the cluster with the latest build immediately.

## Related Issues

https://github.com/astronomer/issues/issues/5507

## Merging

- release-0.30
- release-0.31

### DIFF
--- a/bin/trigger_feature_stack_update.py
+++ b/bin/trigger_feature_stack_update.py
@@ -99,15 +99,16 @@ def main(circleci_token: str, astro_path: str, branch: str):
     print("INFO: Waiting until pipeline starts running. It will wait for 5 min.")
     pipeline_state = "pending"
     counter = 0
-    while "pending" != pipeline_state:
-        time.sleep(60)
+
+    while "pending" == pipeline_state:
+        time.sleep(10)
         job_state_resp = get_job_state(
             circleci_token=circleci_token, pipeline_id=pipeline_id
         )
         pipeline_state = json.loads(job_state_resp)["items"][0]["status"]
         counter = counter + 1
 
-        if counter == 5:
+        if counter == 6:
             break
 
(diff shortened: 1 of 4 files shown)


bin/trigger_feature_stack_update.py
```diff
@@ -0,0 +1,135 @@
...
+
+    if "success" != pipeline_state and "running" != pipeline_state:
+        print(f"Error: Failed to run pipeline. Last Status: {pipeline_state}")
+        raise SystemError(1)
+    else:
+        raise SystemExit(0)
+
+
+if __name__ == "__main__":
+    arg_parser = argparse.ArgumentParser(description="Optional app description")
```
danielhoherd, line 122: Need to update "Optional app description".

bin/trigger_feature_stack_update.py
```diff
@@ -0,0 +1,135 @@
...
+    circle_ci_conn.request(
+        method="POST", url=api_endpoint, body=json.dumps(payload), headers=headers
+    )
+    resp = circle_ci_conn.getresponse().read().decode("utf-8")
+    circle_ci_conn.close()
+    return resp
+
+
+def get_job_state(circleci_token: str, pipeline_id: str):
+    circle_ci_conn = http.client.HTTPSConnection(CIRCLECI_URL)
```
danielhoherd, line 44: Need a timeout here.

danielhoherd (approved):
Seems fine to me. Lots of deja vu. I'm not sure if we're using these mechanics lots of places or if I'm just reviewing the same code over and over, but it seems like we could put this into a reusable module so we don't have to copy/paste/review code in so many places.

PR #1844, opened 2023-03-30
Updating Cronjob Naming Convention to Limit Name Length to 52 Characters

prasadrayudu:
## Description

Platform version 0.31.3
Updating Cronjob Naming Convention to Limit Name Length to 52 Characters.

There is a corn job - charts/astronomer/templates/houston/cronjobs/houston-update-root-admin-password.yaml

Cronjob takes the following as name - 

```
metadata:
  name: {{ .Release.Name }}-update-root-admin-password-cronjob
```

Changed to "upd-root-adm-pwd-cron"

## Related Issues

Few of the customers when the release name is added it is exceeding 52 char limit leading to the failure of upgrade.
Example: if a customer has release name "prod-data-astronomer" then "cronjob is prod-data-astronomer--update-root-admin-password-cronjob" and when they upgrade from 0.30.2 or any lower version to 0.31.3 then it fails.

[gh-issue](https://github.com/astronomer/issues/issues/5522)
[zendesk](https://astronomer.zendesk.com/agent/tickets/17584)

### DIFF
--- a/charts/astronomer/templates/houston/cronjobs/houston-update-root-admin-password.yaml
+++ b/charts/astronomer/templates/houston/cronjobs/houston-update-root-admin-password.yaml
@@ -4,7 +4,7 @@
 apiVersion: {{ include "apiVersion.batch.cronjob" . }}
 kind: CronJob
 metadata:
-  name: {{ .Release.Name }}-update-root-admin-password-cronjob
+  name: {{ .Release.Name }}-upd-root-adm-pwd-cron
   labels:
     tier: astronomer
     component: houston-update-root-admin-password-cronjob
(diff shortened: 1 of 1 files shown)


danielhoherd (approved):
I don't see a problem with this rename, but it's not going to solve the core issue in all cases. We recommend that customers limit their deployment name to 12 characters in order to avoid this case (though it would currently work up to 16 characters) https://docs.astronomer.io/software/install-gcp#step-9-install-astronomer We test against 12 characters https://github.com/astronomer/astronomer/blob/f086eb92f6/tests/chart_tests/test_default_chart.py#L18-L29 which is why we recommend that they use no more than 12 characters.

danielhoherd:
I'm going to close this because we already have guidelines for length limitations, and tests that check against those guidelines.

PR #1847, opened 2023-03-31
Adding Trigger to call feature_stack deployment (#1842)

mishah334:
## Description

Adding a trigger to call the `feature_stack` workflow that performs the deployment on the cluster. This will add the capability to refresh the cluster with the latest build immediately.

## Related Issues

https://github.com/astronomer/issues/issues/5507

## Master PR

https://github.com/astronomer/astronomer/pull/1842

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,7 @@ jobs:
             bin/release-helm-chart /tmp/workspace/astronomer-*.tgz
 
   trigger-rc-test:
-    executor: python/default
+    executor: docker-executor
     resource_class: small
     steps:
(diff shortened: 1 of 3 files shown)


mishah334:
> @mishah334 is this the same code that was merged elsewhere? It would be great to know if it is so I don't have to go line by line when reviewing.

Yes @danielhoherd, it is same as the master branch PR: https://github.com/astronomer/astronomer/pull/1842

danielhoherd (commented):
I've already reviewed these changes in another branch, so I'm approving them here.

danielhoherd (approved):

PR #1887, opened 2023-05-16
Relabeling to release for precreatednamespaces and also namespaceFreeFormEntry

rujhan-arora-astronomer:
## Description

For pre-created namespaces pods, Relabeling and storing the release names in `release` tag, so that they can be fetched by the query made from houston, under metrics tab

## Related Issues

[4650](https://github.com/astronomer/issues/issues/4650)

## Testing

Kind cluster locally. Also, spinning up a separate cluster.

## Merging

0.32.1

### DIFF
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -278,6 +278,16 @@ data:
             replacement: "$1"
             {{- end}}
             target_label: release
+          {{- if .Values.global.features.namespacePools.enabled }}
+          - source_labels: [pod]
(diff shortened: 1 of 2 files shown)


danielhoherd (changes requested):
Seems fine. I discussed with with Vishnu yesterday and it seems like this would be a good solution.

We need tests against prometheus-config-configmap.yaml for this new optional block.

danielhoherd (changes requested):
@rujhan-arora-astronomer please `pip install pre-commit` and then run `pre-commit install && pre-commit run --all-files` from within the astronomer/astronomer repository and commit the changes.

rujhan-arora-astronomer:
> pre-commit install && pre-commit run --all-files

Done!

danielhoherd (approved):
LGTM

rishkarajgi (approved):

PR #1910, opened 2023-06-06
add extraenv support for registry backend service

pgvishnuram:
## Description

Allow referencing custom env vars and secret ref in registry env vars to remove sensitive information passed as direct values in config. 

## Related Issues

https://github.com/astronomer/issues/issues/4559

## Testing

QA should able to verify by passing env vars for registry configuration

## Merging

cherry-pick to release-0.30 and release-0.32.

### DIFF
--- a/tests/chart_tests/test_astronomer_registry.py
+++ b/tests/chart_tests/test_astronomer_registry.py
@@ -30,8 +30,8 @@ class TestRegistryStatefulset:
         )
 
     def test_astronomer_registry_statefulset_with_customenv(self, kube_version):
-        """Test that helm renders a good statefulset template for astronomer
-        registry."""
+        """Test that helm renders statefulset template for astronomer
+        registry with custom env values."""
         docs = render_chart(
             kube_version=kube_version,
             values={
--- a/tests/chart_tests/test_astronomer_registry.py
+++ b/tests/chart_tests/test_astronomer_registry.py
@@ -32,16 +32,10 @@ class TestRegistryStatefulset:
     def test_astronomer_registry_statefulset_with_custom_env(self, kube_version):
         """Test that helm renders statefulset template for astronomer
         registry with custom env values."""
+        extra_env = {"name": "TEST_ENV_VAR_876", "value": "test"}
         docs = render_chart(
             kube_version=kube_version,
-            extra_env = {"name": "TEST_ENV_VAR_876", "value": "test"}
-            values={
-                "astronomer": {
-                    "registry": {
-                        "extraEnv": [extra_env]
-                    }
-                }
-            },
+            values={"astronomer": {"registry": {"extraEnv": [extra_env]}}},
             show_only=[
                 "charts/astronomer/templates/registry/registry-statefulset.yaml"
             ],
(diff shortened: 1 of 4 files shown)


pgvishnuram:
@danielhoherd

tests/chart_tests/test_astronomer_registry.py
```diff
@@ -8,20 +8,49 @@
...
 )
-def test_astronomer_registry_statefulset(kube_version):
-    """Test that helm renders a good statefulset template for astronomer
-    registry."""
-    docs = render_chart(
-        kube_version=kube_version,
-        show_only=["charts/astronomer/templates/registry/registry-statefulset.yaml"],
-    )
+class TestRegistryStatefulset:
+    def test_astronomer_registry_statefulset(self, kube_version):
```
danielhoherd, line 12:
```suggestion
    def test_astronomer_registry_statefulset_defaults(self, kube_version):
```

tests/chart_tests/test_astronomer_registry.py
```diff
@@ -8,20 +8,49 @@
...
+        doc = docs[0]
+        assert doc["kind"] == "StatefulSet"
+        assert doc["apiVersion"] == "apps/v1"
+        assert doc["metadata"]["name"] == "release-name-registry"
+        assert any(
+            "quay.io/astronomer/ap-registry:" in item
+            for item in jmespath.search("spec.template.spec.containers[*].image", doc)
+        )
+
+    def test_astronomer_registry_statefulset_with_customenv(self, kube_version):
```
danielhoherd, line 32:
```suggestion
    def test_astronomer_registry_statefulset_with_custom_env(self, kube_version):
```

tests/chart_tests/test_astronomer_registry.py
```diff
@@ -8,20 +8,49 @@
...
+        )
+
+        assert len(docs) == 1
+        doc = docs[0]
+        assert doc["kind"] == "StatefulSet"
+        assert doc["apiVersion"] == "apps/v1"
+        assert doc["metadata"]["name"] == "release-name-registry"
+        assert {"name": "REGISTRY_ENV_STORAGE", "value": "test"} in doc["spec"][
+            "template"
+        ]["spec"]["containers"][0]["env"]
```
danielhoherd, line 56: I think something like this is better. I like this pattern we've used elsewhere where we define the object, then include it in the values, and then check for it. This makes it easy to change the object, and ensures it's the same input and output. Also using an env var name that is obviously a test env var will make it easy to spot if we need to debug it.

```suggestion
            extra_env = {"name": "TEST_ENV_VAR_876", "value": "test"}
            values={
                "astronomer": {
                    "registry": {
                        "extraEnv": [extra_env]
                    }
                }
            },
            show_only=[
                "charts/astronomer/templates/registry/registry-statefulset.yaml"
            ],
        )

        assert len(docs) == 1
        doc = docs[0]
        assert doc["kind"] == "StatefulSet"
        assert doc["apiVersion"] == "apps/v1"
        assert doc["metadata"]["name"] == "release-name-registry"
        assert extra_env in doc["spec"]["template"]["spec"]["containers"][0]["env"]
```

danielhoherd (changes requested):
Code looks great. PR description needs to be completed.

Simpcyclassy:
can you add the branch it will be cherry-picked to in the description?

PR #1917, opened 2023-06-21
Handle empty URL parameter values

pgvishnuram:
## Description

Resolves an issue when a empty query param is passed or sent its being handled properly rather ending with 500 internal server error

## Related Issues

https://github.com/astronomer/issues/issues/5683

## Merging

cherry-pick to release-0.30, 0.32

### DIFF
--- a/charts/nginx/templates/nginx-configmap.yaml
+++ b/charts/nginx/templates/nginx-configmap.yaml
@@ -87,7 +87,7 @@ data:
 
         -- Split the query string into individual parameters
         for param in string.gmatch(queryString, "([^&]+)") do
-          local paramName, paramValue = string.match(param, "(.+)=(.*)") -- Allow empty paramValue
+          local paramName, paramValue = string.match(param, "([^=]+)=(.*)") -- Allow empty paramValue
 
           -- Check if the parameter should be masked
           if paramsToMask[paramName] then
--- a/charts/nginx/templates/nginx-configmap.yaml
+++ b/charts/nginx/templates/nginx-configmap.yaml
@@ -48,7 +48,7 @@ data:
 
         -- Split the query string into individual parameters
         for param in string.gmatch(queryString, "([^&]+)") do
-          local paramName, paramValue = string.match(param, "(.+)=(.*)") -- Allow empty paramValue
+          local paramName, paramValue = string.match(param, "([^=]+)=(.*)") -- Allow empty paramValue
 
           -- Check if the parameter should be masked
           if paramsToMask[paramName] then
--- a/charts/nginx/templates/nginx-configmap.yaml
+++ b/charts/nginx/templates/nginx-configmap.yaml
@@ -55,7 +55,7 @@ data:
               paramValue = "##REDACTED##" -- Replace the value with REDACTED
           end
 
-          if paramName and paramName ~= '' and paramValue then -- Check if paramName and paramValue exist
+          if paramName and paramValue then -- Check if paramName and paramValue exist
              table.insert(queryParams, paramName .. "=" .. paramValue)
           end
         end
@@ -94,7 +94,7 @@ data:
               paramValue = "##REDACTED##" -- Replace sensitive data with REDACTED
           end
 
-          if paramName and paramName ~= '' and paramValue then -- Check if paramName and paramValue exist
+          if paramName and paramValue then -- Check if paramName and paramValue exist
              table.insert(queryParams, paramName .. "=" .. paramValue)
           end
         end
--- a/charts/nginx/templates/nginx-configmap.yaml
+++ b/charts/nginx/templates/nginx-configmap.yaml
@@ -48,14 +48,16 @@ data:
 
         -- Split the query string into individual parameters
         for param in string.gmatch(queryString, "([^&]+)") do
-          local paramName, paramValue = string.match(param, "(.+)=(.+)")
+          local paramName, paramValue = string.match(param, "(.+)=(.*)") -- Allow empty paramValue
 
           -- Check if the parameter should be masked
           if paramsToMask[paramName] then
-              paramValue = "REDACTED" -- Replace the value with REDACTED
+              paramValue = "##REDACTED##" -- Replace the value with REDACTED
           end
 
-          table.insert(queryParams, paramName .. "=" .. paramValue)
+          if paramName and paramName ~= '' and paramValue then -- Check if paramName and paramValue exist
+             table.insert(queryParams, paramName .. "=" .. paramValue)
+          end
         end
 
         -- Reconstruct the masked query string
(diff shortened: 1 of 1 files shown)


charts/nginx/templates/nginx-configmap.yaml
```diff
@@ -48,14 +48,16 @@ data:
...
+          local paramName, paramValue = string.match(param, "(.+)=(.*)") -- Allow empty paramValue
 
           -- Check if the parameter should be masked
           if paramsToMask[paramName] then
-              paramValue = "REDACTED" -- Replace the value with REDACTED
+              paramValue = "##REDACTED##" -- Replace the value with REDACTED
           end
 
-          table.insert(queryParams, paramName .. "=" .. paramValue)
+          if paramName and paramName ~= '' and paramValue then -- Check if paramName and paramValue exist
```
danielhoherd, line 58: wouldn't `paramName ~= ''` only match a blank `paramName`?

pgvishnuram: if paramName and paramValue then -- this should be enough we dont need to check paramName to be empty is that okay @danielhoherd

danielhoherd: it seems like that would work. the regex specifies that the paramName must be at least one character.

charts/nginx/templates/nginx-configmap.yaml
```diff
@@ -48,14 +48,16 @@ data:
 
         -- Split the query string into individual parameters
         for param in string.gmatch(queryString, "([^&]+)") do
-          local paramName, paramValue = string.match(param, "(.+)=(.+)")
+          local paramName, paramValue = string.match(param, "(.+)=(.*)") -- Allow empty paramValue
```
danielhoherd, line 51: We should tighten this regex up a bit. Here's at least one example:
```suggestion
          local paramName, paramValue = string.match(param, "([^=]+)=(.*)") -- Allow empty paramValue
```
I wonder if there is a function we can use to do this, because the regex to reliably capture the parameters could be complex.

pgvishnuram: this seems to add %3D at the end with the new change

danielhoherd: I think we should not use a regex to do this since there's the built-in function.

pgvishnuram: let me try to optimize the existing code and update the PR tomorrow

danielhoherd: I'm going to bang on this and try to figure it out. It's not quite working yet. The blank values are still causing a problem.

danielhoherd:
@pgvishnuram i don't think this fix is going to solve the problem. i have completely deleted both of these blocks and i still see problems in the UI caused by 500 errors when i specify an empty parameter value like `?token=thisisthetoken&foo=&bar=`

charts/nginx/templates/nginx-configmap.yaml
danielhoherd: I've been banging on this and I've run into a bunch of edge cases and can't solve them all. I've never done lua before, so that's part of my problem.

One case we should cover is `&foo`

`([^?&]+)=([^&]*)` solves the `&root=` case

danielhoherd: After digging banging on the code and digging through the API reference <https://github.com/openresty/lua-nginx-module#nginx-api-for-lua> I don't see any good way to do the referer other than regex because there is no function to parse the arguments from URL string. All of the URL parsing functions do not take a URL, they imply the URL that is part of the current request.

pgvishnuram:
```
set_by_lua_block $request_uri_masked {
      local args = ngx.req.get_uri_args()
      local paramsToMask = {
        token = true
      }
      local maskedArgs = {}
      for paramName, paramValue in pairs(args) do
        -- Check if the parameter should be masked
        if paramsToMask[paramName] then
          paramValue = "##REDACTED##" -- Replace the value with REDACTED
        end

        if paramName and paramValue then -- Check if paramName and paramValue exist
          table.insert(maskedArgs, paramName .. "=" .. paramValue)
        end
      end
      local maskedQueryString = table.concat(maskedArgs, "&")
      local maskedUrl = ngx.var.uri
      if maskedQueryString ~= "" then
        maskedUrl = maskedUrl .. "?" .. maskedQueryString
      end
      
      return maskedUrl
   
    }
  ```

File: charts/nginx/templates/nginx-configmap.yaml [hunk 3]
```diff
@@ -48,14 +48,16 @@ data:
 
         -- Split the query string into individual parameters
         for param in string.gmatch(queryString, "([^&]+)") do
-          local paramName, paramValue = string.match(param, "(.+)=(.+)")
+          local paramName, paramValue = string.match(param, "([^=]+)=(.*)") -- Allow empty paramValue
 
           -- Check if the parameter should be masked
           if paramsToMask[paramName] then
-              paramValue = "REDACTED" -- Replace the value with REDACTED
+              paramValue = "##REDACTED##" -- Replace the value with REDACTED
```
danielhoherd, line 55: Same as below
```suggestion
              paramValue = "**REDACTED**" -- Replace the value with REDACTED
```

charts/nginx/templates/nginx-configmap.yaml
```diff
@@ -85,14 +87,16 @@ data:
 
         -- Split the query string into individual parameters
         for param in string.gmatch(queryString, "([^&]+)") do
-          local paramName, paramValue = string.match(param, "(.+)=(.+)")
+          local paramName, paramValue = string.match(param, "(.+)=(.*)") -- Allow empty paramValue
 
           -- Check if the parameter should be masked
           if paramsToMask[paramName] then
-              paramValue = "REDACTED" -- Replace sensitive data with REDACTED
+              paramValue = "##REDACTED##" -- Replace sensitive data with REDACTED
```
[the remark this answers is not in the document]
pgvishnuram: yes this is fine i added ## since in astro-ui we were using same convention

charts/nginx/templates/nginx-configmap.yaml
pgvishnuram: yes this is fine i added ## since in astro-ui we were using same convention

danielhoherd: ok, let's keep it consistent for now. we can change this in a future release.

charts/nginx/templates/nginx-configmap.yaml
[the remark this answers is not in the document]
danielhoherd: ok, let's keep it consistent for now. we can change this in a future release.

charts/nginx/templates/nginx-configmap.yaml
```diff
@@ -85,14 +87,16 @@ data:
 
         -- Split the query string into individual parameters
         for param in string.gmatch(queryString, "([^&]+)") do
-          local paramName, paramValue = string.match(param, "(.+)=(.+)")
+          local paramName, paramValue = string.match(param, "(.+)=(.*)") -- Allow empty paramValue
```
danielhoherd, line 90: Shouldn't this be consistent between the two blocks?
```suggestion
          local paramName, paramValue = string.match(param, "([^=]+)=(.*)") -- Allow empty paramValue
```

danielhoherd (approved):
Looks good. We've done a lot of back and forth and debugging on this and these changes are good for now. We have more things to tweak in the future if we choose to do so.

PR #1920, opened 2023-07-03
Update image template to use ap-image instead of ap-base.

rob-1126:
Update astronomer stan to use the correct image name when a private registry is specified.

## Description

Until now stan always incorrectly used the ap-base name but the ap-image tag when private registry was enabled. Tags were closely enoogh version that it just happened to work.

## Related Issues

[Do not merge this PR until this text is replaced with links to related issues.
](https://github.com/astronomer/issues/issues/5698)

## Testing

Visual inspection of helm -n astronomer template ~/astronomer --set global.privateRegistry.enabled=false |grep "ap-init" -B only

## Merging

All currently supported release branches.

### DIFF
--- a/charts/stan/templates/_helpers.tpl
+++ b/charts/stan/templates/_helpers.tpl
@@ -41,7 +41,7 @@ Return the list of peers in a NATS Streaming cluster.
 
 {{ define "stan.init.image" -}}
 {{- if .Values.global.privateRegistry.enabled -}}
-{{ .Values.global.privateRegistry.repository }}/ap-base:{{ .Values.images.init.tag }}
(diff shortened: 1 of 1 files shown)


rob-1126:
@pgvishnuram Can you take ownership of this?

pgvishnuram:
@rob-1126 yes sure we will cover-up in upcomming releases and backport them to 0.30 as well

danielhoherd:
@rob-1126 @pgvishnuram i'll take this. i want to add tests that will catch all cases like this.

danielhoherd:
Thanks for the PR @rob-1126 but I'm closing this and will include this fix in a more broad PR that checks the whole chart for this condition, and broadens other private registry tests to include init containers.

PR #1921, opened 2023-07-04
sync sidecar indexPattern from main config

pgvishnuram:
## Description

This PR allows user to configure es index pattern  added in common config and is synced to airflow deployments and curator config without any additional effort

## Related Issues

https://github.com/astronomer/issues/issues/5594

## Related PRs

https://github.com/astronomer/houston-api/pull/1521

## Testing

## Merging


pgvishnuram:
@rishkarajgi and @danielhoherd

charts/elasticsearch/templates/_helpers.tpl
```diff
@@ -153,3 +153,12 @@ imagePullSecrets:
 {{ . }},
 {{- end -}}
 {{- end -}}
+
+
+{{- define "curator.indexPattern" -}}
+{{ if and .Values.global.loggingSidecar.enabled  .Values.global.loggingSidecar.indexPattern }}
+{{- .Values.global.loggingSidecar.indexPattern | squote }}
+{{ else }}
+{{- .Values.curator.age.timestring | squote}}
```
danielhoherd, line 162: Why do we need "squote" here? We'll definitely need to make test cases with various quoting possibilities. https://helm.sh/docs/chart_template_guide/function_list/#printf may be needed.

danielhoherd (commented):
Looks good. I think that having tests will make it all more clear.

charts/elasticsearch/templates/curator/es-curator-configmap.yaml
```diff
@@ -34,7 +34,7 @@ data:
         - filtertype: age
           source: name
           direction: older
-          timestring: '{{ .Values.curator.age.timestring }}'
+          timestring: {{ template "curator.indexPattern" . }}
```
pgvishnuram, line 37: @danielhoherd it was with single quote thats why i added the same in the helper templates

rishkarajgi (approved):
lgtm, please address @danielhoherd's comments

tests/chart_tests/test_houston_configmap.py
```diff
@@ -232,6 +232,42 @@ def test_houston_configmap_with_loggingsidecar_enabled_with_overrides():
...
+    doc = docs[0]
+    prod_yaml = yaml.safe_load(doc["data"]["production.yaml"])
+    log_cmd = 'log_cmd = " 1> >( tee -a /var/log/sidecar-log-consumer/out.log ) 2> >( tee -a /var/log/sidecar-log-consumer/err.log >&2 ) ; "'
+    assert (
+        log_cmd in prod_yaml["deployments"]["helm"]["airflow"]["airflowLocalSettings"]
+    )
+    assert prod_yaml["deployments"]["loggingSidecar"] == {
+        "enabled": True,
+        "name": sidecar_container_name,
+        "image": "quay.io/astronomer/ap-vector:0.22.3",
```
danielhoherd, line 265: For test cases we should always set the test to something that is obviously not real data unless parsing is necessary. This makes debugging errors easier.

```suggestion
    image_name = "example.com/some-repo/test-image-name:test-tag-foo"
    indexPattern = "%Y.%m"
    docs = render_chart(
        values={
            "global": {
                "loggingSidecar": {
                    "enabled": True,
                    "name": sidecar_container_name,
                    "image": image_name,
                    "indexPattern": indexPattern,
                }
            }
        },
        show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
    )

    common_test_cases(docs)
    doc = docs[0]
    prod_yaml = yaml.safe_load(doc["data"]["production.yaml"])
    log_cmd = 'log_cmd = " 1> >( tee -a /var/log/sidecar-log-consumer/out.log ) 2> >( tee -a /var/log/sidecar-log-consumer/err.log >&2 ) ; "'
    assert (
        log_cmd in prod_yaml["deployments"]["helm"]["airflow"]["airflowLocalSettings"]
    )
    assert prod_yaml["deployments"]["loggingSidecar"] == {
        "enabled": True,
        "name": sidecar_container_name,
        "image": image_name,
```

danielhoherd (changes requested):
Looks mostly good. I left an optional suggestion to improve a test case. I am also requesting that we parametrize the indexPattern test to make sure `squote` is compatible with index patters that have single quotes. This is highly unlikely to be a real world configuration, but we need to make sure we are properly quoting our data and not making it possible to produce invalid yaml.

charts/elasticsearch/templates/curator/es-curator-configmap.yaml
pgvishnuram: pytest doen;t seems to like % in a string that gets interpolated

tests/chart_tests/test_houston_configmap.py
```diff
@@ -232,6 +232,42 @@ def test_houston_configmap_with_loggingsidecar_enabled_with_overrides():
     assert "vector" in prod_yaml["deployments"]["loggingSidecar"]["image"]
 
 
+def test_houston_configmap_with_loggingsidecar_enabled_with_indexPattern():
+    """Validate the houston configmap and its embedded data with
+    loggingSidecar."""
+    sidecar_container_name = "sidecar-log-test"
+    image_name = "example.com/some-repo/test-image-name:test-tag-foo"
+    indexPattern = "%Y.%m"
```
danielhoherd, line 240: This parametrize block works and shows that the single-quote val breaks the helm mechanic.
```suggestion
@pytest.mark.parametrize(
    "indexPattern",
    [
        "%Y.%m",
        "%Y.%m.%d",
        "a-single-quote-'-in-the-indexPattern",
        'a-double-quote-"-in-the-indexPattern',
    ],
)
def test_houston_configmap_with_loggingsidecar_enabled_with_indexPattern():
    """Validate the houston configmap and its embedded data with
    loggingSidecar."""
    sidecar_container_name = "sidecar-log-test"
    image_name = "example.com/some-repo/test-image-name:test-tag-foo"
```

danielhoherd: Another way we could work around this is to add a comment that single-quotes are not allowed. That would be a really easy non-code way to handle this potential problem.

danielhoherd: I tried to find a way to handle both single and double quotes and I don't see an easy way to do it without making really ugly code. Helm printf does not appear to have any mechanism to produce the right string. I think we can skip merging the parametrize change and just put a comment in the values.yaml that quote characters are not supported in the indexPattern.

pgvishnuram:
it will require a houston update to pass the indexPattern config

PR #1927, opened 2023-07-15
Enforce expected tags.monitoring=false behavior

danielhoherd:
## Description

`tags.monitoring=false` was not disabling all components when it was expected to do so. This was caused by two different implementations of the conditions used in the tags. Components whose conditions were set to `true` would always be on, and components whose conditions were not used elsewhere would work as expected. This is because helm values override the tags behavior, so if a condition is set to true in values.yaml, disabling the tag has lower precedence and is ignored.

The fix I have in this branch may not be the right fix. It does what is expected, but I am not sure if it is the best way to solve the problem. It may be acceptable for our use case.

While working on debugging this I also made the following changes:
- Delete a bunch of cruft related to the Prometheus Operator ServiceMonitor custom resource. AFAIK these have never been supported. The are certainly not supported now, and were breaking some broad tests that I added.
- Re-order some data in Chart.yaml, improve comments, fix conditions so tags work
- Make `DEBUG=1` show additional test data and not delete temp helm values files
- Rename some templates so they have consistent, singluar names that match their k8s kind. EG: secrets.yaml -> secret.yaml
- Add a few more enable_all_features.yaml flags
- Add some missing k8s schema files that cover templates that had not previously been tested

## Related Issues

https://github.com/astronomer/issues/issues/5714

## Testing

I have done a lot of testing on this, and most of the changes are additional tests. Some template files have been renamed though, which may cause undesirable behaviors in operation, such as a deletion of a thing followed by the addition of a thing, which could cause things to shuffle around or potentially break. This is an undesirable behavior, so we should test this and consider the stability of existing deployments compared with the value of this additional test coverage.

## Merging

These changes fix a bug, so they should be merged everywhere.


pgvishnuram:
@danielhoherd since we have enabled nodeExporter as true in https://github.com/astronomer/astronomer/blob/master/values.yaml#L75

danielhoherd:
> @danielhoherd since we have enabled nodeExporter as true in https://github.com/astronomer/astronomer/blob/master/values.yaml#L75

Yeah, that's the source of the problem. Anything in values.yaml will override the condition in Chart.yaml. Because we set global.nodeExporterEnabled, the condition in Chart.yaml can never be false, which means tags.monitoring=false could never work the way it was intended. Changing the condition in Chart.yaml to something that is not explicitly set in values.yaml allows both tags.monitoring to work and global.nodeExporterEnabled to work as expected, which is that tags.monitoring takes precedence.

danielhoherd:
After digging a bit, it looks like our Astronomer chart is not using tags and conditions best practices, and this is not the right time to rip that out and fix it. After my last comment I was hoping to simplify the conditions, but I am not going to do so. I think this PR is good to go. CC @pgvishnuram

Chart.yaml
```diff
@@ -16,35 +19,37 @@ dependencies:
...
-  - name: prometheus-node-exporter
-    condition: global.nodeExporterEnabled
+  - name: prometheus-postgres-exporter
+    condition: global.prometheusPostgresExporter.Enabled
     tags:
       - monitoring
-  - name: prometheus-blackbox-exporter
-    condition: global.blackboxExporterEnabled
+  - name: prometheus
+    condition: global.prometheusEnabled
```
Simpcyclassy, line 49: I like the alphabetical order improvement, easier to find a depency

charts/prometheus-node-exporter/templates/monitor.yaml
```diff
@@ -1,25 +0,0 @@
...
-  endpoints:
-    - port: metrics
-      {{- if .Values.prometheus.monitor.scrapeTimeout }}
-      scrapeTimeout: {{ .Values.prometheus.monitor.scrapeTimeout }}
-      {{- end }}
-{{- if .Values.prometheus.monitor.relabelings }}
-      relabelings:
-{{ toYaml .Values.prometheus.monitor.relabelings | indent 6 }}
-{{- end }}
-{{- end }}
```
Simpcyclassy, line 25: Where are these default values now set?

charts/prometheus-postgres-exporter/values.yaml
```diff
@@ -21,20 +21,6 @@ service:
   labels: {}
   annotations: {}
 
-serviceMonitor:
-  # When set true then use a ServiceMonitor to configure scraping
-  enabled: false
```
Simpcyclassy, line 26: Where is this value now set?

rishkarajgi (approved):

PR #1941, opened 2023-07-26
Refactor houston config's handling of cron splay

danielhoherd:
## Description

Refactor houston configmap's handling of cleanup cron splay.

## Related Issues

https://github.com/astronomer/issues/issues/5670

## Testing

We have unit tests for the changes here, but seeing the cron splay in e2e tests will be required for us to be confident that everything is in place.

## Merging

This should only be merged to branches that support airflowChartVersion 1.9.0

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -359,8 +359,10 @@ data:
             Generate a deterministic random-ish decimal digit from the .Release.Name, which is used to generate
             the minute field of the cron schedule. This is effectively a splay seeded by the release name.
             This is used to spread the load of this job away from high-use minutes 0,15,30,45.
+            This template is passed to the airflow chart as a string where the template is interpreted in the
+            context of the airflow-chart's .Release.Name.
             */}}
-            schedule: "{{- add 3 (regexFind ".$" (adler32sum .Release.Name)) -}}-59/15 * * * *"
+            schedule: "'{{"{{"}}- add 3 (regexFind \".$\" (adler32sum .Release.Name)) -}}-59/15 * * * *'"
             command:
             - airflow-cleanup-pods
             - --namespace={{`{{ .Release.Namespace }}`}}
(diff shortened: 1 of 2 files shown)


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -359,8 +359,10 @@ data:
             Generate a deterministic random-ish decimal digit from the .Release.Name, which is used to generate
             the minute field of the cron schedule. This is effectively a splay seeded by the release name.
             This is used to spread the load of this job away from high-use minutes 0,15,30,45.
+            This template is passed to the airflow chart as a string where the template is interpreted in that
+            context, including the airflow-chart's .Release.Name.
             */}}
-            schedule: "{{- add 3 (regexFind ".$" (adler32sum .Release.Name)) -}}-59/15 * * * *"
+            schedule: "'{{"{{"}}- add 3 (regexFind \".$\" (adler32sum .Release.Name)) -}}-59/15 * * * *'"
```
rishkarajgi, line 365: Can you explain what the difference here is exactly? (just for my knowledge)

rishkarajgi (approved):
lgtm

charts/astronomer/templates/houston/houston-configmap.yaml
danielhoherd: Absolutely, that syntax is pretty confusing. The `{{"{{"}}` is the helm template way to render `{{` into the template. This is needed because we want to render the value shown in the tests, which is `'{{- add 3 (regexFind ".$" (adler32sum .Release.Name)) -}}-59/15 * * * *'`. That is single-quoted because it is sent as a string from houston to airflow-chart values, NOT interpreted by helm when we install astronomer. That value will be rendered as a teamplate by the airflow-chart <https://github.com/apache/airflow/blob/main/chart/templates/cleanup/cleanup-cronjob.yaml#L46>. Using this technique, we can define helm template logic in the astronomer chart that is interpreted by the airflow chart.

PR #1943, opened 2023-07-27
allow disabling default platform and airflow alerts

danielhoherd:
## Description

- Add the ability to disable each of airflow and platform default alerts.
- Refactor the render_chart function to make DEBUG env var do useful things like do not delete the temp file and print up debug info.

## Related Issues

https://github.com/astronomer/issues/issues/4915

## Testing

I have unit tests for all of the changes. It would be good to have a critical eye go over them and maybe test things out, especially if I forgot some edge case.

## Merging

Merge only to 0.33 since this is a new feature.

### DIFF
--- a/charts/prometheus/templates/prometheus-alerts-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-alerts-configmap.yaml
@@ -23,15 +23,16 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   alerts: |-
-    {{- if .Values.customAlertConfig.enabled }}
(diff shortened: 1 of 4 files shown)


rishkarajgi (approved):

pgvishnuram:
with newer condition customer no way can specifically select components and had to compromise all the alerting to be added from config can we add some flexibility like

disableDefaultAirflowAlerts
disableDefaultPlatformAlerts

pgvishnuram (approved):

rishkarajgi (approved):

PR #1946, opened 2023-08-03
remove uid changes from main config

pgvishnuram:
## Description

runAsUser in grafana  securityContext creates a constraint in openshift not to run the deployment since the UID range is not in specified range of openshift approved  ones.

## Related Issues

https://github.com/astronomer/issues/issues/5737

## Testing

QA should able to see grafana pod in running status

## Merging

cherry-pick to release-0.32, 0.33

### DIFF
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -18,7 +18,6 @@ images:
 
 securityContext:
     runAsNonRoot: true
-    runAsUser: 472
 
 backendSecretName: ~
 backendConnection:
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -17,7 +17,7 @@ images:
     pullPolicy: IfNotPresent
 
 securityContext:
-    runAsNonRoot: true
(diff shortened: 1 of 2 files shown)


pgvishnuram:
@danielhoherd @rob-1126

danielhoherd (changes requested):
Seems fine. Needs tests since this is used in a few places https://github.com/astronomer/astronomer/blob/b2049e2e56/charts/grafana/templates/grafana-deployment.yaml#L50 and there are no securityContext tests in https://github.com/astronomer/astronomer/blob/b2049e2e56/tests/chart_tests/test_grafana.py

pgvishnuram:
adding test cases

danielhoherd (approved):
Good improvements on the test with moving them to a class. This all looks good to me.

danielhoherd:
@pgvishnuram do we want to backport this to 0.30?

pgvishnuram:
Not needed for 0.30 since this uid change were added from 0.32

PR #1975, opened 2023-09-04
fix logging flag in stan config

pgvishnuram:
## Description

## Related Issues

## Testing

## Merging


danielhoherd (changes requested):
Changes look good. We need to add tests though. For instance we need to check for the presence and absence of `sd: true` and `sv: true` when helm is given different values.

PR #1990, opened 2023-09-11
Enforce k8s version constraints

danielhoherd:
## Description

- Create a `metadata.yaml` file that holds the single source of truth of which versions of k8s are supported. This file should be used for other cases where having a single source of data that is not specific to helm is needed.
- Enforce k8s version constraints by erroring helm when the k8s version is too new or too old (Major.Minor versions that have not been tested)
- Include an override for the version constraints that can be used to get out of sticky situations.
- Update chart_tests to use metadata.yaml as its source of what k8s versions to test.
- Update generate_circleci_config.py to use metadata.yaml as its source of what k8s versions to generate configs for.
- remove need for GitPython. use pathlib instead.
- Add `validate_objects` arg to `render_chart()`. This is required when we support the latest k8s version available because the schema files of the next version would not yet be available to validate. This also avoids caching schema files in the repo for versions that we do not support.
- Unrelated change: i am sneaking in the removal the ability to release non-release branches into this PR. we never do that, and it's not a good idea, and most of the time it would fail anyway.

## Related Issues

https://github.com/astronomer/issues/issues/5824

See the above ticket for a link to relevant slack discussion about why this change is needed.

## Testing

I think I have covered all cases with unit tests. QA should be on the lookout for helm errors that indicate that the version is too new or old. These errors will be obvious because helm will fail and not install the chart.

## Merging

This change set needs to be merged to all supported branches.


tests/chart_tests/test_default_chart.py
```diff
@@ -12,6 +12,38 @@
...
+        "version",
+        [
+            (k8s_version_too_old),
+            (k8s_version_too_new),
+        ],
+    )
+    def test_k8s_version_out_of_bounds_override(self, version):
+        """Test that no error is returned for versions that are too old or new when forceIncompatibleKubernetes is used."""
+        render_chart(
+            values={"forceIncompatibleKubernetes": True},
```
pgvishnuram, line 41: @danielhoherd @rishkarajgi option forceIncompatibleKubernetes is not in our config.yml we should add it in defaults

danielhoherd: @pgvishnuram I guess best practice would be to add it in values.yaml, but it's supposed to be something customers don't use, so I'd rather not bring attention to it.

pgvishnuram (approved):

PR #1993, opened 2023-09-13
add default index config kibana

pgvishnuram:
## Description

Creates Default index for kibana service 

## Related Issues

https://github.com/astronomer/issues/issues/3253

## Testing

## Merging


charts/kibana/templates/kibana-default-index-cronjob.yaml
```diff
@@ -0,0 +1,48 @@
+#######################################################
+##           Kibana Default Index CronJob        ######
+#######################################################
```
danielhoherd, line 3:
```suggestion
########################################
# Kibana default index helm hook
########################################
```

charts/kibana/templates/kibana-default-index-cronjob.yaml
```diff
@@ -0,0 +1,48 @@
...
+          - -c
+          - |
+            http_status=$(curl -s -o /dev/null -w "%{http_code}" -XGET http://{{ .Release.Name }}-kibana:5601/api/data_views/data_view/fluentd.*)
+            if [ "$http_status" -eq 200 ]; then
+              echo "Index Patterns Already Exists. Skipping Creation"
+            else
+              echo "Creating Index Pattern"
+              curl -XPOST -H 'Content-Type: application/json' -H 'kbn-xsrf: astronomer' --data '{"attributes":{"title":"fluentd.*","timeFieldName":"@timestamp"}}' http://{{ .Release.Name }}-kibana:5601/api/saved_objects/index-pattern/fluentd.*?overwrite=false
+            fi
+        restartPolicy: OnFailure
```
danielhoherd, line 48: I'm not sure that `restartPolicy: OnFailure` is the right thing here. If we run into a failure scenario, this will restart endlessly. I think that is less desirable than just not having the index be created. Do we expect this to restart a few times during normal operation? If so maybe we can move restart logic into the script.
```suggestion
        restartPolicy: OnFailure
```

charts/kibana/templates/kibana-default-index-cronjob.yaml
```diff
@@ -0,0 +1,48 @@
...
+          command:
+          - "/bin/bash"
+          - -c
+          - |
+            http_status=$(curl -s -o /dev/null -w "%{http_code}" -XGET http://{{ .Release.Name }}-kibana:5601/api/data_views/data_view/fluentd.*)
+            if [ "$http_status" -eq 200 ]; then
+              echo "Index Patterns Already Exists. Skipping Creation"
+            else
+              echo "Creating Index Pattern"
+              curl -XPOST -H 'Content-Type: application/json' -H 'kbn-xsrf: astronomer' --data '{"attributes":{"title":"fluentd.*","timeFieldName":"@timestamp"}}' http://{{ .Release.Name }}-kibana:5601/api/saved_objects/index-pattern/fluentd.*?overwrite=false
```
danielhoherd: Maybe we can add this retry logic to avoid having the job itself restart on failure.
```suggestion
              curl --retry 10 --retry-delay 30 --retry-all-errors -XPOST -H 'Content-Type: application/json' -H 'kbn-xsrf: astronomer' --data '{"attributes":{"title":"fluentd.*","timeFieldName":"@timestamp"}}' http://{{ .Release.Name }}-kibana:5601/api/saved_objects/index-pattern/fluentd.*?overwrite=false
```

pgvishnuram: we may need to add in both places

charts/kibana/templates/kibana-default-index-cronjob.yaml
```diff
@@ -0,0 +1,48 @@
...
+          chart: {{ template "kibana.chart" . }}
+          release: {{ .Release.Name }}
+          heritage: {{ .Release.Service }}
+          app: {{ template "kibana.name" . }}-default-index
+          version: {{ .Chart.Version }}
+      spec:
+{{- include "kibana.imagePullSecrets" . | indent 6 }}
+        containers:
+        - name: kibana-default-index
+          image: {{ template "kibana.image" . }}
```
danielhoherd, line 35: ap-kibana is ~325mb. Can we use ap-init (~10mb) or some other lightweight container for this? This job will most likely not run on the same node that kibana will end up running on, so pulling the kibana image is not an advantage.

pgvishnuram: does ap-init has curl checking

danielhoherd (changes requested):
Looks good so far. Chart tests are failing on some docs count check that probably needs to be increased by 1 for this new doc. Also it would be great to have a functional test to verify the presence of the index in kibana.

danielhoherd (approved):
Looks good!

PR #1999, opened 2023-09-20
blackbox-exporter: respect .global.networkPolicy.enabled

pgvishnuram:
## Description

Black box exporter service applies network policies while global.networkPolicy  is set to false. This PR fixes the behaviour and respects the  global.networkPolicy flag

## Related Issues

https://github.com/astronomer/issues/issues/5843

## Testing

QA should able to validate with no network policies when global.networkPolicy  set to false

## Merging

cherry-pick to release-0.30,0.32,0.33


danielhoherd (approved):
LGTM! good catch

PR #2027, opened 2023-10-17
add support for openshift security context

pgvishnuram:
## Description

This PR adds support to native openshift installation by allowing customer to disable securityContext and podSecurityContext
Components affected
- alertmanager
- elasticsearch (es-client-deployment, es-data-statefulset, es-master-statefulset)
- external-es-proxy-deployment
- kibana
- kube-state
- nats (statefulset)
- prometheus-node-exporter
- prometheus
- stan

## Related Issues

https://github.com/astronomer/issues/issues/5919

## Testing

## Merging


vishwas-astro:
@pgvishnuram can we create an issue for this with the description of the problem we are trying to solve here. And also tag it with the releases we are planning to ship this to.

charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -24,8 +24,8 @@ spec:
 {{- if and .Values.rbac.create .Values.serviceAccount.create }}
       serviceAccountName: {{ template "prometheus-node-exporter.serviceAccountName" . }}
 {{- end }}
-{{- if .Values.securityContext }}
-      securityContext: {{ toYaml .Values.PodsecurityContext | nindent 8 }}
+{{- with .Values.podSecurityContext }}
```
danielhoherd, line 27: What is the value of switching from `if` to `with` ? Is it just cleaner? Is the functionality the same and the produced output is the same?

pgvishnuram: with allows to completely turn off the config by passing podSecurityContext: [] this means an empy values

charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
```diff
@@ -41,7 +41,7 @@ spec:
     spec:
       serviceAccountName: {{ include "external-es-proxy.serviceAccountName" . }}
       securityContext:
-        {{- toYaml .Values.PodSecurityContext | nindent 8 }}
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
```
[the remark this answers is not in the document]
pgvishnuram: the place where we have not used with is not using fsgroup or runAsUser by default

charts/prometheus-node-exporter/templates/daemonset.yaml
pgvishnuram: with the existing if condition it will work like below 
alertmanager: 
  PodSecurityContext: {}
  
 i can revert the code and re-apply the similar config  to cross verify the same

danielhoherd (changes requested):
Looks good so far. We need to ensure these are all covered by tests.

rishkarajgi (commented):
lgtm. We should add/update the tests

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -159,6 +159,11 @@ data:
                   # to the airflow namespaces.
                   secretName: {{ template "houston.jwtCertificateSecret" . }}
             {{ end  }}
+            {{ if .Values.global.openshiftEnabled }}
+            securityContexts:
+              pod:
+                runAsNonRoot: true
+            {{ end  }}
```
danielhoherd, line 166: If this works for openshift, can we just use it as the standard config? Is there any good reason to run it without this setting when we are running in a non-openshift cluster?

pgvishnuram:
> pre-commit failing

Its fixed

charts/astronomer/templates/houston/houston-configmap.yaml
rishkarajgi: I was thinking on the same lines. We can keep changes agnostic to environoment as the standard config and openshift deployment specific config under the flag

nsAstro: Agreed on Daniels point. I think the biggest shift is switching logging from FluentD to Vector as a default which would probably cause issues to existing customers.

danielhoherd: We should reduce the variety between openshift and non-openshift as much as possible. The more we reduce that diff, the less code we have to maintain, and the less things there are that can break, and the less QA testing needs to be done.

rishkarajgi (approved):

PR #2036, opened 2023-10-19
fix tags naming in chart config

pgvishnuram:
## Description

This PR reverts back to old naming to sync with the default values 

## Related Issues

https://github.com/astronomer/issues/issues/5928

## Testing

## Merging


danielhoherd (approved):

danielhoherd (changes requested):
Please add tests to validate the behavior. Whatever tests I had added when making this change were clearly not sufficient. We may need to parametrize a test for each sub-chart that has a tags.monitoring entry. The tests should verify that the components are disabled when `tags.monitoring` is set to false, and enabled when it is true.

rishkarajgi (approved):

PR #2066, opened 2023-11-23
add internal auth url support for astronomer components

pgvishnuram:
## Description

Add internal auth support for grafana, kibana and prometheus

## Related Issues

https://github.com/astronomer/issues/issues/5963

## Testing

## Merging


templates/_helpers.tpl
```diff
@@ -9,3 +9,12 @@ fluentd
...
 {{- end -}}
+
+
+{{ define "houston.internalauthurl" -}}
+{{- if .Values.global.enableHoustonInternalAuthorization  }}
+nginx.ingress.kubernetes.io/auth-url: http://{{ .Release.Name }}-houston.{{ .Release.Namespace }}.svc.cluster.local:8871/v1/authorization
+{{- else }}
+nginx.ingress.kubernetes.io/auth-url: https://houston.{{ .Values.global.baseDomain }}/v1/authorization
+{{- end }}
+{{- end }}
```
pgvishnuram, line 20: should i keep only the value here or current format is fine @danielhoherd

danielhoherd: We should return the same type of data for both options, so if we're going to return the full `key: value` on one option we should do it on both.

danielhoherd (approved):
Looks good!

PR #2072, opened 2023-12-01
6015 make cli install service optionable

pgvishnuram:
## Description

## Related Issues

## Testing

## Merging


charts/astronomer/values.yaml
```diff
@@ -423,6 +423,7 @@ registry:
   logLevel: info
 
 install:
+  enabled: true
```
rishkarajgi, line 426: `.Values.install.enabled` is not conveying the reference to CLI 
We should rename the flag to `cliEnabled` for better clarity

pgvishnuram: this might affect existing deployment config since customers are using this for a long time we can add both configs and we can state old config will be removed in 0.34

danielhoherd: i agree with @rishkarajgi that having cliInstall would have been better, but i also agree with @pgvishnuram that changing the parent namespace of the new `enabled` parameter is outside of the scope of this ticket and could introduce problems. if we want to improve the `install` value namespace, we should do that on a minor version bump and address that in another ticket.

danielhoherd (approved):
This all looks great to me. If we want to address the `install` -> `cliInstall` change, we should open another ticket and schedule that work.

charts/astronomer/values.yaml
danielhoherd: i experienced this ambiguity again today when looking at [the missing probes issue](https://github.com/astronomer/issues/issues/6606) and yeah, it's too ambiguous. however, i think we should just get rid of cli-install altogether. <https://github.com/astronomer/issues/issues/6611>

PR #2073, opened 2023-12-04
argocd-improvements

luismacosta:
## Description
argocd-improvements:

- Secrets: db_connection_string/smtp/certificate
- Templates with function randAlphaNum

## Related Issues

## Testing

## Merging


danielhoherd:
@luismacosta Thanks for the contribution! This broke a lot of our tests though. If you can fix the tests that would be great (`make unittest-charts`), otherwise I'm not sure when we'll get to this. Generally I like the idea though.

danielhoherd:
Without user stories and understanding what this is solving, I don't see these changes as being valuable. @luismacosta if you still would like us to make changes to help out your argocd challenges, can you add more details about what exactly the problem is? Otherwise we're going to close this stale PR soon.

danielhoherd (changes requested):
Need more details about what problem this is solving. User stories would be a great way to communicate those. Reproduction steps of the problem would also be great.

PR #2079, opened 2024-01-02
add privateCA support for containerd images

pgvishnuram:
## Description

Adds the ability to support privateCA for containerd images

## Related Issues

https://github.com/astronomer/issues/issues/6022

## Testing

tested on AWS, GCP, Azure, local clusters

## Merging

release-0.34


rishkarajgi (commented):
Please add tests to this change

templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
```diff
@@ -0,0 +1,95 @@
...
+  labels:
+    component: houston
+    tier: astronomer
+    release: {{ .Release.Name }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    heritage: {{ .Release.Service }}
+data:
+  update-containerd-certs.sh: |
+    #!/usr/bin/env sh
+    if [ ! -f /hostcontainerd/config.toml ]; then
```
rob-1126, line 18: theoretically I think this file could be totally absent and need to be created, but if we have tested on Azure, AWS, and GCP I'm fine with crossing that bridge when we come to it. In reality, I would expect it always exists.

values.yaml
```diff
@@ -19,6 +19,10 @@ global:
   privateCaCertsAddToHost:
     enabled: false
     hostDirectory: /etc/docker/certs.d
+    addToContainerd: false
+    containerdConfigPath: /etc/containerd/certs.d
```
rob-1126, line 23: key name should be something else as that sounds like the path that should put to containerd.toml itself. Maybe rename to containerdCertsConfigPath or similar?

values.yaml
```diff
@@ -19,6 +19,10 @@ global:
...
+    containerdnodeAffinitys: []
```
rob-1126, line 24: Are these actually templated into the daemonset anywhere?

rob-1126 (changes requested):

templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
pgvishnuram: yes this file always exists in all cloud provider only difference is the content

rob-1126 (approved):
Needs to be tested in QA on all three major cloud-providers as part of the release process for both new install and upgrade path.

rishkarajgi:
> Needs to be tested in QA on all three major cloud-providers as part of the release process for both new install and upgrade path.

cc: @himabindu07 @Priyanka13Astro

PR #2099, opened 2024-02-01
Fix broken scan-docker-images CI job

danielhoherd:
## Description

- Use cimg/base:current-22.04
- use /usr/bin/bash for shell
- move shell options to their own step

## Related Issues

https://github.com/astronomer/issues/issues/6125

## Testing

This should be manually tested. It only affects dev workflows, not the product itself.

## Merging

This should be merged everywhere

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
       - run:
           name: Install trivy
           command: |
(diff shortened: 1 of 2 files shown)


rishkarajgi (approved):

pgvishnuram:
bin/trivy-scan.sh: line 22: trivy: command not found

PR #2116, opened 2024-02-07
Replace cert copier image

pgvishnuram:
## Description

move cert copier image alpine:3.18 to astronomer managed quay repo 
alpine:3.18 -> quay.io/astronomer/ap-base:3.18 

## Related Issues

https://github.com/astronomer/issues/issues/6102

## Testing

when global private registry it should auto switch to private registry

## Merging


danielhoherd (approved):
Very nice. Using templates for image:tag is so clean. I like this pattern we've been using lately. I also like the addition of this image to enable_all_features.yaml. Tests are failing, but once they are passing I say Shipit!

rishkarajgi (approved):

PR #2123, opened 2024-02-08
Make houston-db-migrations run both pre-upgrade and post-upgrade helm hooks

danielhoherd:
## Description

Workaround to an order-of-operations bug that we discovered. More details in the issue linked below.

## Related Issues

https://github.com/astronomer/issues/issues/6152

## Testing

This solves a problem where houston requires deploys to happen twice before picking up the AU vs resources configuration setting.

## Merging

AFAIK this would only be needed in release-0.34, but the root cause may be happening elsewhere and we haven't noticed.


pgvishnuram (approved):
right now we dont have an option to update configmap first and initiate hook job to take latest changes of config - by running post-upgrade job it will resolve the issue  or we should run that particular job as a post upgrade job

PR #2147, opened 2024-03-08
Improve check for duplicate tag versions of all image components

danielhoherd:
## Description

- Add `--check-tags` feature to `bin/show-docker-images.py` and delete `bin/verify_image_tags.py`. This is a better implementation that covers more edge cases while reducing the overall LOC.
- Improve existing pre-commit hook configs

The new output looks like this:

```
$ pre-commit run check-image-tags --all-files
Ensure unique tags per image.............................................Failed
- hook id: check-image-tags
- exit code: 1

ALERT: example.com/the-private-registry/ap-init has 2 different tags: 1.2.3, 3.18.6
```

You can also run this manually with `bin/show-docker-images.py --private-registry --with-houston -c` and the help text looks like this:

```
$ bin/show-docker-images.py --help
usage: show-docker-images.py [-h] [--private-registry] [-v] [--with-houston] [-c]

options:
  -h, --help          show this help message and exit
  --private-registry  show images when using private registry
  -v, --verbose       be verbose
  --with-houston      include images from houston configmap
  -c, --check-tags    Only check for multiple tags for the same image, do not print the images
```

## Related Issues

- https://github.com/astronomer/issues/issues/6023
- https://github.com/astronomer/issues/issues/6115

## Testing

As long as CI passes we're good. QA does not need to do anything here.

## Merging

Merge everywhere.


rishkarajgi (approved):

danielhoherd:
merged/cherry-picked to all release branches.

PR #2154, opened 2024-03-26
update houston worker command args

pgvishnuram:
## Description

Add update-ca-certificates as default for worker

## Related Issues

https://github.com/astronomer/issues/issues/6229

## Testing

updated on ticket

## Merging

cherry-pick to all valid releases

### DIFF
--- a/charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
+++ b/charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
@@ -94,7 +94,7 @@ spec:
           command:
             {{- toYaml .Values.houston.worker.command | nindent 12 }}
           {{ else }}
-          command: ["yarn", "worker"]
+          command: ["/bin/sh", "-c", "update-ca-certificates;yarn worker"]
           {{- end }}
           {{- if .Values.houston.worker.args }}
           args:
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -40,7 +40,7 @@ images:
     pullPolicy: IfNotPresent
 
 securityContext:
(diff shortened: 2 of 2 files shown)


charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
```diff
@@ -94,7 +94,7 @@ spec:
           command:
             {{- toYaml .Values.houston.worker.command | nindent 12 }}
           {{ else }}
-          command: ["yarn", "worker"]
+          command: ["/bin/sh", "-c", "update-ca-certificates;yarn worker"]
```
pgvishnuram, line 97: @danielhoherd @rishkarajgi should this be added behind the flag or lets run them as bydefault

pgvishnuram: @danielhoherd

danielhoherd: @pgvishnuram if we run this by default we need to think about how it can fail and what will happen when it does fail.

pgvishnuram: since we have fixed the perm issue at max it wont fail and also houston always runs this command by default via entrypoint

pgvishnuram: or i can add this when global.privatecaenabled

danielhoherd: i'm fine with us running this every time as long as the pod will not fail if this command exits non-zero.

PR #2159, opened 2024-03-27
add additional logic for es restriction from airflow

pgvishnuram:
## Description

Updates to external es nginx config with restricted location block.

## Related Issues

https://github.com/astronomer/issues/issues/6241

## Testing

Requires external elastic search to test this change

## Merging

Yet to update


charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
```diff
@@ -31,7 +31,21 @@ data:
...
+          proxy_pass http://localhost:{{ .Values.service.awsproxy }};
+          {{- else }}
+          access_by_lua_file  /usr/local/openresty/nginx/conf/setenv.lua;
+          proxy_pass {{.Values.global.customLogging.scheme}}://{{.Values.global.customLogging.host}}:{{.Values.global.customLogging.port}};
+          {{- include "external-es-proxy-trustcerts" . | indent 8 }}
+          {{- end }}
+        }
+
+        location ~* /_bulk$ {
+          rewrite /_bulk(.*) /{{ include "logging.indexNamePrefix" . }}.$remote_user.*/_bulk$1 break;
```
danielhoherd, line 48: if this is just copied code, can we put this whole 'nginx.conf` or `http` block into a template and include it in both external and non-external es configmaps ? that would be DRY, and as long as they're going to be the same then that would be helpful.

pgvishnuram: only parts of the changes are copied like rewrite parts

pgvishnuram: then we need to move the the common changes to global helpers.tpl to share  the common code

charts/external-es-proxy/templates/_helpers.tpl
```diff
@@ -101,3 +101,15 @@ imagePullSecrets:
 {{ .Values.images.awsproxy.repository }}:{{ .Values.images.awsproxy.tag }}
 {{- end }}
 {{- end }}
+
+
+
```
danielhoherd, line 106: The space between blocks keeps getting longer and longer. It's 1 line in some places, 2 lines in others, but I think 3 lines is a bit too much.
```suggestion

```

danielhoherd: `external-es-proxy-nginx-common` is an ambiguous name. Maybe we should rename it to `external-es-proxy-nginx-location-common` ? 

Also the templating I was talking about earlier was to reduce duplication between charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml and charts/external-es-proxy/templates/external-es-proxy-configmap.yaml

danielhoherd (changes requested):
Needs tests. Unfortunately I haven't been able to find a maintained nginx configuration parser for python, so we'll have to test the contents of the nginx config as a string.

charts/external-es-proxy/templates/_helpers.tpl
```diff
@@ -65,12 +65,12 @@ Create the name of the service account to use
 {{- define "external-es-proxy-trustcerts" -}}
 {{- if .Values.global.customLogging.trustCaCerts  }}
 {{- $secret_name := .Values.global.customLogging.trustCaCerts }}
-  proxy_ssl_trusted_certificate /etc/ssl/certs/{{ $secret_name }}.pem;
-  proxy_ssl_verify              on;
-  proxy_ssl_verify_depth        2;
-  proxy_ssl_session_reuse on;
+proxy_ssl_trusted_certificate /etc/ssl/certs/{{ $secret_name }}.pem;
```
Simpcyclassy, line 68: Adding comments for easier review

Eg
# Specify the trusted SSL certificate for proxy connections
`proxy_ssl_trusted_certificate /etc/ssl/certs/{{ $secret_name }}.pem;`

charts/external-es-proxy/templates/_helpers.tpl
```diff
@@ -65,12 +65,12 @@ Create the name of the service account to use
...
+proxy_ssl_verify              on;
+proxy_ssl_verify_depth        2;
+proxy_ssl_session_reuse on;
```
Simpcyclassy, line 71: > Enable SSL certificate verification`

charts/external-es-proxy/templates/_helpers.tpl
```diff
@@ -65,12 +65,12 @@ Create the name of the service account to use
...
 {{- else }}
-  proxy_ssl_verify              off;
+proxy_ssl_verify              off;
```
Simpcyclassy, line 73: > If no trusted certificate is provided, disable SSL verification

charts/external-es-proxy/templates/_helpers.tpl
```diff
@@ -101,3 +101,13 @@ imagePullSecrets:
 {{ .Values.images.awsproxy.repository }}:{{ .Values.images.awsproxy.tag }}
 {{- end }}
 {{- end }}
+
+{{- define "external-es-proxy-nginx-location-common" -}}
+{{- if  or .Values.global.customLogging.awsSecretName  .Values.global.customLogging.awsServiceAccountAnnotation .Values.global.customLogging.awsIAMRole }}
+proxy_pass http://localhost:{{ .Values.service.awsproxy }};
```
Simpcyclassy, line 107: > If AWS-related parameters are defined, use the local AWS proxy.

Simpcyclassy:
Does this PR have a gh issue?

PR #2162, opened 2024-03-27
Removed server-snippet which runs on every ingress request which is being logged

rujhan-arora-astronomer:
## Description

Removed server-snippet which runs on every ingress request which is being logged as it is no longer needed because the Houston token is no longer being passed in the URL.

## Related Issues

- <https://github.com/astronomer/issues/issues/6235>

## Testing

Locally

## Merging

0.34.1


danielhoherd (approved):
Looks great! I love that we're no longer passing the token in the URL.

PR #2171, opened 2024-04-04
Add openshift support for nginx default backend service

pgvishnuram:
## Description

Add openshift support for nginx default backend service

## Related Issues

https://github.com/astronomer/issues/issues/5927

## Testing

when global.openshiftEnabled: true default backend service should come as it is 

## Merging

Yet to update


pgvishnuram:
@danielhoherd - this is a small improvement for openshift based installations

charts/nginx/templates/_helpers.tpl
```diff
@@ -68,3 +68,11 @@ imagePullSecrets:
 {{- define "defaultBackend.fullname" -}}
 {{ printf "%s-default-backend" (include "nginx.fullname" .)}}
 {{- end -}}
+
+{{- define "nginx.securityContext" -}}
+{{- if or (eq ( toString ( .Values.securityContext.runAsUser )) "auto") ( .Values.global.openshiftEnabled ) -}}
+{{- omit  .Values.securityContext "runAsUser" | toYaml | nindent 10 }}
+{{- else -}}
+{{- .Values.securityContext | toYaml | nindent 10 }}
```
danielhoherd, line 76: I think we should remove the nindent stuff here since we have to nindent elsewhere anyways. Having an arbitrary indentation in the template just makes it confusing. I think it'd be better with no indentation in the template and set the precise indentation that is needed where the template is used.

> The behaviour of a unit of code should be as obvious as possible by looking only at that unit of code - <https://htmx.org/essays/locality-of-behaviour>

Removing indentation from here will improve the readability where the template is used, and make the usage of the template more intuitive.

danielhoherd (approved):
Looks mostly good. I would remove the indentation from within the template though. (See my comment about this)

danielhoherd:
Parent ticket has been closed, so I am closing this PR.

PR #2177, opened 2024-04-08
Added support for MySQL in astronomer/astronomer

rujhan-arora-astronomer:
## Description

Added support for MySQL in astronomer/astronomer
## Related Issues

[Do not merge this PR until this text is replaced with links to related issues.
](https://github.com/astronomer/issues/issues/6281)
## Testing

Tested locally against Houston only, for now.

## Merging

0.35.0


charts/astronomer/values.yaml
```diff
@@ -24,15 +24,15 @@ images:
     # httpSecret: ~
   houston:
     repository: quay.io/astronomer/ap-houston-api
-    tag: 0.34.5
+    tag: 6012
```
danielhoherd, line 27: This needs to be updated before merging.

charts/mysql/charts/common/README.md
```diff
@@ -0,0 +1,235 @@
+# Bitnami Common Library Chart
```
danielhoherd, line 1: I wonder if we should remove all the Bitnami chart README files.

values.yaml
```diff
@@ -289,6 +289,14 @@ global:
     extraLabels: []
       # do-funky-injection: maybe
 
+  # Enable default mysql database.
+  # This is not recommended for production.
+  mysql:
+    enabled: false
+    servicePort: "3306"
+    username: "root"
+    password: "root"
```
danielhoherd, line 298: Currently this PR does not have any tests in tests/chart_tests. I think we should make some before merging the PR. This bit of code would be a good place to start.

charts/mysql/values.schema.json
```diff
@@ -0,0 +1,195 @@
+{
```
danielhoherd, line 1: I think we should remove this file. We have not done a good job of maintaining our own jsonschema file, and I doubt we would do a good job of maintaining this one.

charts/mysql/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,46 @@
+{{- /*
```
danielhoherd, line 1: Are we going to support this serviceMonitor? We do not use serviceMonitors anywhere else in our chart. I feel like we should remove all mentions of serviceMonitor.

danielhoherd (changes requested):
I think we should take some time to reduce the number of additions in this PR. I think we should remove anything we don't need, and make this a hard fork of the bitnami chart that is highly tailored to our use case. This would include removing components and documentation that we have no plans to use.

We also need to add tests for this chart. https://github.com/astronomer/astronomer/tree/master/tests/chart_tests

We could do this work in branches off this branch, and merge back into this branch, then when we're happy with all of those reductions, merge this one. Or we could do the work here (though that change set may get pretty long).

I would be happy to take on this work.

Also https://github.com/astronomer/issues/issues/6293 should be part of that work.

PR #2183, opened 2024-04-25
add individual ingress per service

pgvishnuram:
## Description

This PR segreagates the common ingress rules to individual rules per service to avoid policy issues

## Related Issues

https://github.com/astronomer/issues/issues/6323

## Testing

Yet to update

## Merging

NA


tests/chart_tests/test_ingress.py
```diff
@@ -32,10 +32,48 @@ def test_basic_ingress(self, kube_version):
...
+        )
+        assert len(docs) == 1
+        expected_rules_v1 = json.loads(
+            """
+        [
+        {"host":"registry.example.com","http":{"paths":[{"path":"/","pathType":"Prefix","backend":{"service":{"name":"release-name-registry","port":{"name":"registry-http"}}}}]}}]
+        """
+        )
+        assert docs[0]["apiVersion"] == "networking.k8s.io/v1"
+        assert docs[0]["spec"]["rules"] == expected_rules_v1
```
danielhoherd, line 79: What we really need in this file is a test case that renders the full chart, finds all of the ingress objects, and ensures that each of them only has one host.

pgvishnuram: Yes and also now we have seperate template included per service - currently asserting the templates

values.yaml
```diff
@@ -122,6 +122,8 @@ global:
 
   enableHoustonInternalAuthorization: false
 
+  enablePerHostIngress: false
+
```
danielhoherd, line 126: I think this per-host ingress configuration should be the only way to do it. I don't think we should allow the old way of doing it. If we switch to this new method then we reduce the amount of code we have to maintain and test while still providing the same features, just in a different way that is compatible with more environments.

Is there some reason we should keep the old way as an option?

danielhoherd: Slack thread <https://astronomer.slack.com/archives/C02J8JT58CC/p1715623166453439>

We agreed to keep the old way in, behind a feature flag, and run it through a deprecation phase.

tests/chart_tests/test_ingress.py
```diff
@@ -32,10 +32,66 @@ def test_basic_ingress(self, kube_version):
...
+
+    def test_single_ingress_per_host(self, kube_version):
+        default_docs = render_chart(values={"global": {"enablePerHostIngress": True}})
+        ingresses = [
+            doc for doc in default_docs if doc["kind"].lower() == "Ingress".lower()
+        ]
+        assert all(len(i["spec"]["rules"]) == 1 for i in ingresses)
+        assert all(len(i["spec"]["tls"][0]["hosts"]) == 1 for i in ingresses)
```
danielhoherd, line 97: I think that rather than having several tests functions per-ingress, we could have one test or one parametrized test that pulls all ingresses from the chart and tests them. That's an optimization though, so if we want to not do it and spend that time on something else then that's fine.

danielhoherd (approved):

values.yaml
rishkarajgi: We can default this to true @pgvishnuram

PR #2188, opened 2024-05-06
fix external es search configuration

pgvishnuram:
## Description

Fixes log search in search endpoint to its own indexes

## Related Issues

https://github.com/astronomer/issues/issues/6339

## Testing

QA should able to see task logs in webserver

## Merging

cherry-pick to release-0.34, release-0.35

### DIFF
--- a/charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
+++ b/charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
@@ -43,10 +43,10 @@ data:
           {{- include "external-es-proxy-nginx-location-common" . | indent 10 }}
         }
 
-        location = /_search {
+        location ~* /_search$ {
           # This combined with disabling explicit index searching downstream
           # prevents any deployment from being able to query any other indexes.
-          rewrite ^/(.*) /{{ include "logging.indexNamePrefix" . }}.$remote_user.*/$1 break;
+          rewrite /_search(.*) /{{ include "logging.indexNamePrefix" . }}.$remote_user.*/_search$1 break;
           {{- include "external-es-proxy-nginx-location-common" . | indent 10 }}
         }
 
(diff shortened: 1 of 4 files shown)


charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
```diff
@@ -43,10 +43,10 @@ data:
           {{- include "external-es-proxy-nginx-location-common" . | indent 10 }}
         }
 
-        location = /_search {
+        location ~* /_search$ {
           # This combined with disabling explicit index searching downstream
           # prevents any deployment from being able to query any other indexes.
-          rewrite ^/(.*) /{{ include "logging.indexNamePrefix" . }}.$remote_user.*/$1 break;
+          rewrite /_search(.*) /{{ include "logging.indexNamePrefix" . }}.$remote_user.*/_search$1 break;
```
danielhoherd, line 49: Does this actually work? It seems like an anchored regex `_search$` would not match `_search(.*)` unless the contents of `(.*)` were `$`

pgvishnuram: this is as per our existing design , i have validated them in test cluster.

danielhoherd: Are you sure it works?

Screenshot 2024-05-07 at 9 17 27 AM

https://nginx.viraptor.info

danielhoherd: Oooh, it does work with `_search?foo=bar`. Is that what we're matching?

danielhoherd (approved):
Seems fine to me. The changes here don't reduce my confidence that we're doing the right thing. It would be great if we had a better way to test these nginx configs, but without that this is as good as we can expect.

PR #2218, opened 2024-06-15
add es security context for missing components

pgvishnuram:
## Description

removed hardcoded security context from elasticsearch exported and reworked it support both pod and container security context and tweaked default privilges

## Related Issues

https://github.com/astronomer/issues/issues/6363

## Testing

elasticsearch exporter should able to expose metrics

## Merging

cherry-pick to releaase-0.35

### DIFF
--- a/charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
+++ b/charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
@@ -52,7 +52,7 @@ spec:
 {{ toYaml (default .Values.global.platformNodePool.tolerations .Values.tolerations) | indent 8 }}
       restartPolicy: {{ .Values.exporter.restartPolicy }}
       securityContext:
-{{ toYaml .Values.exporter.securityContext | indent 8 }}
+{{ toYaml .Values.exporter.podSecurityContext | indent 8 }}
 {{- include "elasticsearch.imagePullSecrets" . | indent 6 }}
       containers:
         - name: metrics-exporter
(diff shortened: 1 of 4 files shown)


charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
```diff
@@ -66,23 +66,7 @@ spec:
                     "--web.listen-address=:{{ .Values.exporter.service.httpPort }}",
                     "--web.telemetry-path={{ .Values.exporter.web.path }}"]
           securityContext:
-            capabilities:
```
rishkarajgi, line 69: where have we moved this for podSecurity capabilities?

charts/elasticsearch/values.yaml
```diff
@@ -206,6 +206,13 @@ exporter:
   # Restart policy for all containers
   restartPolicy: Always
 
+  podSecurityContext: {}
+  securityContext:
```
rishkarajgi, line 210: What is the difference between `podSecurityContext` and `securityContext`?

charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
pgvishnuram: we have completely removed the hardcoding and moved to its own values dropping ALL privs by default

charts/elasticsearch/values.yaml
pgvishnuram: podSecurityContext applies to all pods and containers
containerSecurityContext applies only to containers

charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
pgvishnuram: below changes are added to es chart values

```
capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
```

charts/elasticsearch/values.yaml
pgvishnuram: i need to write up a seperate doc on what policies applies to which service

rishkarajgi: Do we need to update `securityContext` to `containerSecurityContext` in that case?

pgvishnuram: may be on next PR since for this one service if we make change that might create confusion

charts/elasticsearch/templates/exporter/es-exporter-deployment.yaml
```diff
@@ -52,7 +52,7 @@ spec:
 {{ toYaml (default .Values.global.platformNodePool.tolerations .Values.tolerations) | indent 8 }}
       restartPolicy: {{ .Values.exporter.restartPolicy }}
       securityContext:
-{{ toYaml .Values.exporter.securityContext | indent 8 }}
```
rishkarajgi, line 55: Will this break the customer environment?

rishkarajgi (approved):

PR #2241, opened 2024-07-02
add extraEnv support for fluentd

pgvishnuram:
## Description

add extraEnv support for fluentd

## Related Issues

https://github.com/astronomer/issues/issues/6465

## Testing

Yet to update

## Merging

cherry-pick to release-0.34,release-0.35

### DIFF
--- a/charts/fluentd/values.yaml
+++ b/charts/fluentd/values.yaml
@@ -53,3 +53,5 @@ s3:
   use_server_side_encryption: ~
 
 priorityClassName: ~
+
+extraEnv: ~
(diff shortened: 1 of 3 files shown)


charts/fluentd/values.yaml
```diff
@@ -53,3 +53,5 @@ s3:
   use_server_side_encryption: ~
 
 priorityClassName: ~
+
+extraEnv: ~
```
pgvishnuram, line 57: @danielhoherd is this to be extraEnv or env is enough

danielhoherd: I like extraEnv because it indicates that we are not defining the env entirely, we are adding to it.

PR #2244, opened 2024-07-09
rework  kube state metrics roles and collectors configuration

pgvishnuram:
## Description

This PR reworks the existing role and collector definition for kube state metrics and only generates roles for the enabled collectors 

## Related Issues

astronomer/issues#6389

## Testing

QA should now able to enable or disable or pass known list of collectors and should verify roles to see only those are created and applied

## Merging

cherry-pick to release-0.34 and release-0.35


charts/kube-state/values.yaml
```diff
@@ -32,3 +32,35 @@ replicas: 1
 
 extraArgs:
 - "--metric-labels-allowlist=namespaces=[*],pods=[*]"
+
+priorityClassName: ~
+
+collectors:
```
rishkarajgi, line 38: TODO: Remove unused collectors and test metrics working without disruption

pgvishnuram: i have commented out the unused collectors

charts/kube-state/templates/kube-state-deployment.yaml
```diff
@@ -28,6 +28,9 @@ spec:
         app: {{ template "kube-state.name" . }}
         version: {{ .Chart.Version }}
     spec:
+    {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
+    {{- end }}
```
danielhoherd, line 33: There are no tests for this.

charts/kube-state/templates/kube-state-deployment.yaml
```diff
@@ -42,6 +45,7 @@ spec:
       {{- if .Values.global.singleNamespace }}
         - --namespaces={{ .Release.Namespace }}
       {{- end }}
+        - --resources={{ .Values.collectors | join "," }}
```
danielhoherd, line 48: There are no tests for this.

tests/chart_tests/test_kube_state_deployment.py
```diff
@@ -83,3 +83,30 @@ def test_kube_state_deployment_singleNamespace(self, kube_version):
...
+        assert len(docs[0]["rules"]) == 23
+
+    def test_kube_state_specific_collector_enabled(self, kube_version):
+        docs = render_chart(
+            kube_version=kube_version,
+            values={"kube-state": {"collectors": ["certificatesigningrequests"]}},
+            show_only=["charts/kube-state/templates/kube-state-role.yaml"],
+        )
+
+        assert len(docs[0]["rules"]) == 1
```
danielhoherd, line 103: We can do better here since we know that `certificatesigningrequests` will be in the rules.

danielhoherd (changes requested):
I'd like to see better tests. I left some comments.

charts/kube-state/templates/kube-state-deployment.yaml
pgvishnuram: test cases are added in here

tests/chart_tests/test_kube_state_deployment.py
danielhoherd:
```suggestion
        assert len(docs[0]["rules"]) == 1
        assert "certificatesigningrequests" in docs[0]["rules"][0]["resources"]
```

danielhoherd (approved):
All of these changes look good to me. I think the feature is good and the reduction in prometheus metrics will be great. I left one optional suggestion to address the one remaining comment about improving a test case.

PR #2252, opened 2024-07-23
remove expire deployment feature

pgvishnuram:
## Description

removed depricated feature expire deployment which is no longer needed

## Related Issues

https://github.com/astronomer/issues/issues/6510

## Testing

QA should not see expire deployment cronjob or template in our helm chart

## Merging

cherry-pick to release-0.35, release-0.36

PR #2253, opened 2024-07-23
Remove commander expire-deployments networkPolicies

danielhoherd:
## Description

Remove commander expire-deployments networkPolicies. This is a follow-up to https://github.com/astronomer/astronomer/pull/2252

## Related Issues

https://github.com/astronomer/issues/issues/6510

## Testing

No extra human testing needed. However, if we do encounter any failures related to this, it only points to an additional bug. It is unlikely we have such a bug, but if we do find one, we should fix it instead of worrying about this change. With that in mind, all we need to do is be aware that this change has been made.

## Merging

Merge everywhere that we are merging https://github.com/astronomer/astronomer/pull/2252

### DIFF
--- a/charts/astronomer/templates/commander/commander-networkpolicy.yaml
+++ b/charts/astronomer/templates/commander/commander-networkpolicy.yaml
@@ -55,11 +55,6 @@ spec:
         matchLabels:
           tier: astronomer
           release: {{ .Release.Name }}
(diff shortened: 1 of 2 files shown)


rishkarajgi (approved):

pgvishnuram:
good to merge

PR #2254, opened 2024-07-24
remove tiller config from helm chart

pgvishnuram:
## Description

remove tiller related configuration from helm chart which is no longer used

## Related Issues

https://github.com/astronomer/issues/issues/6155

## Testing

NA

## Merging


danielhoherd (approved):
Haha, I just saw this last week and did a double-take. I swore we had removed this all like 2 years ago! Good cleanup work.

Edit: I just checked. This should have been removed way back with 0.26, in https://github.com/astronomer/astronomer/pull/1233 or probably even earlier.

PR #2263, opened 2024-07-26
enable dag server persistance spec

pgvishnuram:
## Description

enable dag server persistance spec 

## Related Issues

https://github.com/astronomer/issues/issues/6436

## Testing

```
global:
  dagOnlyDeployment:
    enabled: true
    persistence:
      persistentVolumeClaimRetentionPolicy:
        whenDeleted: Delete
        whenScaled: Retain
```

Action explanation:

whenDeleted: Delete - this means once the statefulset template is deleted kubernetes will automatically delete the pvc as well
whenScaled: Retain. - his means once the statefulset scaled up or down pvc is retained

## Merging

cherry-pick to release-0.35

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -165,7 +165,7 @@ data:
           resources: {{- .Values.global.dagOnlyDeployment.resources | toYaml | nindent 12 }}
         {{- end }}
         {{- if .Values.global.dagOnlyDeployment.persistence }}
-        persistence: {{- .Values.global.dagOnlyDeployment.persistence | toYaml | nindent 12 }}
+        persistence: {{- .Values.global.dagOnlyDeployment.persistence | toYaml | nindent 10 }}
         {{- end }}
       {{- end }}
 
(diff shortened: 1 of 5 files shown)


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -164,6 +164,9 @@ data:
         client:
           resources: {{- .Values.global.dagOnlyDeployment.resources | toYaml | nindent 12 }}
         {{- end }}
+        {{- if .Values.global.dagOnlyDeployment.persistence }}
+        persistence: {{- .Values.global.dagOnlyDeployment.persistence | toYaml | nindent 10 }}
```
danielhoherd, line 168: Do we want to add a default value for this? Even something like `[]` or `{}` ?

pgvishnuram: persistence: {} should be fine

pgvishnuram: with this change https://github.com/astronomer/airflow-chart/blob/master/values.yaml#L555-L560 these sections are now overridable

danielhoherd (approved):
Looks even better with tests and default values 😎

PR #2290, opened 2024-08-20
Add debug port for nginx-ingress

stuart23:
## Description

Exposes the debug port on the nginx-ingress controller. This is already running in the container but is not currently exposed to be able to port-forward and see the heap dump from a browser

## Related Issues

Fixes: https://github.com/astronomer/issues/issues/6579

## Testing

Manually edited the deployment configuration to add it. You can then port forward the port from each pod and run `go tool pprof -http=:8082 localhost:10245/debug/pprof` to view the heap

## Merging

Ship in next release

### DIFF
--- a/charts/nginx/templates/nginx-deployment.yaml
+++ b/charts/nginx/templates/nginx-deployment.yaml
@@ -73,6 +73,8 @@ spec:
               containerPort: {{ .Values.ports.https }}
             - name: metrics
               containerPort: {{ .Values.ports.metrics }}
+            - name: debug
+              containerPort: 10245
           livenessProbe:
             httpGet:
               path: /healthz


pgvishnuram:
@stuart23 we should put this behind a flag

stuart23:
Can the flag have a default as "on"? This is not running anything extra in the container/pod, it is just exposing an endpoint that already exists.

danielhoherd:
@stuart23 @pgvishnuram we can't open a TCP port by default without getting buy-in from security team and from major customer account managers. I expect that they would not want this enabled by default. To get started on that discussion, we'll need a ticket in astronomer/issues that describes the use case: 

<https://github.com/astronomer/issues/issues/new?assignees=&labels=kind%2Ffeature&projects=&template=feature_request.yaml>

pgvishnuram:
@danielhoherd that what i mentioned we either need a custom flag to enable this and not as default and also previously we have metrics endpoint exposed that caused some customer not to expose it and removed it and also we need to restrict this endpoint not to expose via public url and only available via internal url

charts/nginx/templates/nginx-deployment.yaml
```diff
@@ -73,6 +73,8 @@ spec:
               containerPort: {{ .Values.ports.https }}
             - name: metrics
               containerPort: {{ .Values.ports.metrics }}
+            - name: debug
```
Simpcyclassy, line 76: Use `.Values.ports.debug` here and add a default to `charts/nginx/values.yaml` to keep this consistent with the other port definitions in the template.

Simpcyclassy (changes requested):
Every other port in this template is templated via `.Values.ports.*` (http, https, metrics), but `debug` hardcodes `10245` directly. That is inconsistent with the existing pattern and means operators cannot change the port without editing the template. Please add `ports.debug: 10245` to `charts/nginx/values.yaml` and reference it as `{{ .Values.ports.debug }}`. No chart test covers the new port; a basic rendering assertion should be added to the chart test suite.

PR #2298, opened 2024-09-04
Obliterate all traces of podSecurityPolicies

danielhoherd:
## Description

Finalize removal of podSecurityPolicies, which were removed in k8s 1.25, and nginx-ingress 1.10

## Related Issues

https://github.com/astronomer/issues/issues/6595

## Testing

Normal tests should cover these changes. We are finally removing code for a feature that we have not been testing and have not supported for a while now.

## Merging

This should be merged everywhere since psp will not function once we upgrade nginx to >1.9


rishkarajgi (approved):

pgvishnuram:
we already have a exisiting helm template validation that only applies psp on clusters with version less than 1.25 so even with nginx 1.10 this doesnt make any template renderings

danielhoherd:
@pgvishnuram right, but the code would not work if people turned it on.

PR #2300, opened 2024-09-06
6606 fix test probes

danielhoherd:
## Description

Completely refactor the way we do all liveness and readiness probes, and make them optional in every ds, sts, and deployment, while maintaining existing default probes.

**This is a breaking change.** Customers who have configured probe values will be bitten by the breaking changes in this PR if they do not adjust their values. Users now need to include the entire probe when making a customization, not just customize the one property they want to tweak. This is a compromise that was necessary to make all probes fully configurable without making the code too complicated.

Other included changes:
- Improve some helm template syntaxes
- I think fix some bugs with some esoteric configs like pullSecrets
- Reduce the CircleCI unit test resource_class from xlarge to large, and run with local docker, which combines to reduce costs for this step quite a bit.
- Update some in-repo documentation.
- Add some debug info to CI runs, both successful and failing.
- Add readiness port in ES application config.
- Stop using jmespath in tests/chart_tests/__init__.py because python jmespath syntax is tricky to get right.
- Add INTERNET_ACCESS env var to run tests with no internet access and avoid making outbound connections
- Use dedent on some fluentd test config
- Wrote some big test cases in test_probes.py that ensure existing `master` branch probes still work, and show how to configure probes for every other ds, sts, and deployment. This may be a good reference for engineers and customers.

## Related Issues

https://github.com/astronomer/issues/issues/6606

## Testing

Only probes and test code is changed here, so we should be on the lookout for restarting containers that should not be restarting.

## Merging

This should be merged everywhere. These probes make our application more resilient and customizable, and do not change core functionality.


lzdanski:
Hey @danielhoherd - Docs got tagged as a reviewer for this PR. It doesn't look like any docs strings have been changed. Does this affect any behavior we should update user documentation for?

danielhoherd:
@lzdanski sorry, this is a draft. I have converted it. There are some risks related to this change, but we haven't finished the implementation so we're not yet sure what needs to be documented. I'll ping you again about this as we get closer to completion.

tests/chart_tests/test_probes.py
```diff
@@ -1,40 +1,82 @@
...
+    "charts/nginx/templates/nginx-deployment.yaml": {"nginx": default_probes},
+    "charts/pgbouncer/templates/pgbouncer-deployment.yaml": {"pgbouncer": default_probes},
+    "charts/postgresql/templates/statefulset-slaves.yaml": {"postgres": default_probes},
+    "charts/postgresql/templates/statefulset.yaml": {"postgres": default_probes},
+    "charts/prometheus/templates/prometheus-statefulset.yaml": {"prometheus": default_probes},
+    "charts/prometheus-blackbox-exporter/templates/deployment.yaml": {"prometheus-blackbox-exporter": default_probes},
+    "charts/prometheus-node-exporter/templates/daemonset.yaml": {"prometheus-node-exporter": default_probes},
+    "charts/prometheus-postgres-exporter/templates/deployment.yaml": {"prometheus-postgres-exporter": default_probes},
+    "charts/stan/templates/statefulset.yaml": {"stan": default_probes},
+}
```
danielhoherd, line 50: @pgvishnuram @shubh-am-kumar what do you two think of this approach? This seems like the only way we'll be able to get a comprehensive test of all the different ways we will need to configure probes, due to the inconsistencies in our chart. This seems to be working really well, but I'm curious if you think it's a bad idea.

charts/alertmanager/templates/alertmanager-statefulset.yaml
```diff
@@ -107,20 +103,8 @@ spec:
...
-            periodSeconds: 10
-          livenessProbe:
-            httpGet:
-              path: /healthz
-              port: {{ .Values.global.authSidecar.port }}
-              scheme: HTTP
-            initialDelaySeconds: 10
-            periodSeconds: 10
+          readinessProbe: {{- tpl (toYaml .Values.readinessProbe) . | nindent 12 }}
+          livenessProbe: {{ tpl (toYaml .Values.livenessProbe) . | nindent 12 }}
```
danielhoherd, line 107: This `tpl` technique lets us embed helm logic in values.yaml so we can access global values.

charts/alertmanager/values.yaml
```diff
@@ -29,7 +29,32 @@ replicas: 1
...
+      scheme: HTTP
+    initialDelaySeconds: 10
+    periodSeconds: 10
+  readinessProbe:
+    httpGet:
+      path: /healthz
+      port: "{{ .Values.global.authSidecar.port }}"
+      scheme: HTTP
+    initialDelaySeconds: 10
+    periodSeconds: 10
```
danielhoherd, line 56: This section is run through a `{{ tpl (toYaml ...) . | nindent 12 }}` section so that the helm values from global are rendered in the k8s manifest output.

tests/chart_tests/test_probes.py
rishkarajgi: @danielhoherd you've tagged the wrong Shubham :)
cc: @Shubham-astro

Shubham-astro (approved):

rishkarajgi (approved):

PR #2301, opened 2024-09-09
6537 BYO Service Account

Shubham-astro:
## Description

This PR Disables Astronomer from creating Roles, Rolebindings, and SAs and Allows users to specify a pre-created SAs.

## Related Issues

Related:
- https://github.com/astronomer/issues/issues/6537
- https://github.com/astronomer/issues/issues/6583

## Testing

Tested locally:

Screenshot 2024-09-09 at 9 56 40 AM

## Merging


charts/astronomer/values.yaml
```diff
@@ -100,6 +100,17 @@ houston:
...
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: false
+    # Annotations to add to the service account
+    annotations: {}
+    # The name of the service account to use.
+    # If not set and create is true, a name is generated using the fullname template
+    name: ""
+    # Specifies whether a scc privilege should be created for custom sa
+    sccEnabled: false
```
pgvishnuram, line 112: sccEnabled is particularly for openshift and we required that for registry for other services its handled in a different way so this is not required here.

charts/astronomer/values.yaml
```diff
@@ -337,6 +358,17 @@ commander:
...
```
pgvishnuram, line 370: same here

charts/astronomer/values.yaml
```diff
@@ -318,6 +329,16 @@ configSyncer:
...
```
pgvishnuram, line 341: same here

charts/astronomer/values.yaml
```diff
@@ -318,6 +329,16 @@ configSyncer:
   # If not provided, will generate a random hour and minutes to spread cronjob workloads.
   schedule: ~
   securityContext: {}
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: false
+    # Annotations to add to the service account
+    annotations: {}
```
pgvishnuram, line 336: if we are going to add annotation support then we need to extend that to our serviceAccount template as well currently its hardcoded for argo usage

charts/astronomer/values.yaml
```diff
@@ -337,6 +358,17 @@ commander:
     enabled: false
   upgradeTimeout: 600
 
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: false
+    # Annotations to add to the service account
+    annotations: {}
```
pgvishnuram, line 365: same here

charts/astronomer/values.yaml
```diff
@@ -100,6 +100,17 @@ houston:
   jwtSigningKeySecretName: ~
   jwtSigningCertificateSecretName: ~
 
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: false
+    # Annotations to add to the service account
+    annotations: {}
```
pgvishnuram, line 107: same here

tests/chart_tests/test_BYO_SAs.py
```diff
@@ -0,0 +1,44 @@
...
+        assert len(roles) == 0
+        assert len(rolebindings) == 0
+     
+
+
+        
+
+
+
+
```
pgvishnuram, line 44: fix pre-commit `pre-commit run --all-files`

charts/astronomer/templates/commander/commander-serviceaccount.yaml
```diff
@@ -12,7 +12,12 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     heritage: {{ .Release.Service }}
   {{ if .Values.global.enableArgoCDAnnotation }}
```
pgvishnuram, line 14: this can be removed since we have moved the annotation inside service account

charts/astronomer/templates/config-syncer/config-syncer-serviceaccount.yaml
```diff
@@ -12,7 +12,12 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     heritage: {{ .Release.Service }}
   {{ if .Values.global.enableArgoCDAnnotation }}
+  {{- with .Values.configSyncer.serviceAccount.annotations }}
```
pgvishnuram, line 15: these too can be removed

charts/astronomer/values.yaml
danielhoherd: @pgvishnuram can you make a code suggestion to better indicate what section is not required here?

danielhoherd (commented):
Looks great so far! I added one code suggestion that tests the same thing but uses significantly different code structure, and fixes a significant bug in the test case.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -127,18 +127,57 @@
 {{ default (printf "%s-registry-auth" .Release.Name) .Values.registry.auth.secretName }}
 {{- end }}
 
+{{ define "registry.ServiceAccount" -}}
+{{- if .Values.global.rbacEnabled }}
+  {{- if .Values.registry.serviceAccount.name -}}
+    {{ printf "%s-%s" .Release.Name .Values.registry.serviceAccount.name }}
+  {{- else -}}
+    {{ printf "%s-registry" .Release.Name }}
```
pgvishnuram, line 135: {{ default "default" .Values.registry.serviceAccount.name }}

charts/astronomer/templates/_helpers.yaml
```diff
@@ -127,18 +127,57 @@
...
+  {{- end }}
+{{- else -}}
+  default
+{{- end }}
+{{- end }}
```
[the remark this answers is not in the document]
danielhoherd: This behavior is all over the _helpers.yaml file. Do we need to remove all the other instances of that?

charts/astronomer/templates/_helpers.yaml
```diff
@@ -127,18 +127,49 @@
...
+
 {{ define "houston.bootstrapperServiceAccount" -}}
 {{ printf "%s-houston-bootstrapper" .Release.Name }}
 {{- end }}
 
-{{ define "registry.ServiceAccount" -}}
-{{- if .Values.registry.serviceAccount.name -}}
-{{ printf "%s-%s" .Release.Name .Values.registry.serviceAccount.name }}
+
+
```
danielhoherd, line 147: Remove excessive blank lines.

```suggestion


```

danielhoherd (changes requested):
The code seems to be in good shape. I have a suggested code snip that fixes a test case that currently gives 100% success even when there are failures.

charts/astronomer/templates/_helpers.yaml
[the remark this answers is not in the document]
Shubham-astro: this is done

pgvishnuram:
naming convention change test_BYO_SAs.py to test_byo_sa.py should be fine

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,36 @@
...
+from tests import supported_k8s_versions
+from tests.chart_tests.helm_template_generator import render_chart
+
+
+@pytest.mark.parametrize(
+    "kube_version",
+    supported_k8s_versions,
+)
+class TestServiceAccounts:
+    def test_SAs_created(self, kube_version):
```
danielhoherd, line 12: function definition should describe what is being tested, like `test_serviceaccount_rbac_disabled`

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,36 @@
...
+        docs = render_chart(kube_version=kube_version, values={"global": {"rbacEnabled": False}})
+
+        # Check that no ServiceAccount resources are created
+        """
+        service_accounts = [
+            doc["metadata"]["name"] for doc in docs if isinstance(doc, dict) and doc.get("kind") == "ServiceAccount"
+        ]
+        assert len(service_accounts) == 0, "No ServiceAccounts should be created when rbacEnabled is False"
+        """
+        # Check that the Deployment or StatefulSet is using the default ServiceAccount
```
danielhoherd, line 23: This whole block needs to be cleaned up before we merge.

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,36 @@
...
+        for doc in docs:
+            if isinstance(doc, dict) and doc.get("kind") in ["Deployment", "StatefulSet"]:
```
danielhoherd, line 25: No need to test if it's a dict. It is. If it's not we need to know that, so we should fail.

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,36 @@
...
+                spec = doc.get("spec", {}).get("template", {}).get("spec", {})
+                sa_name = spec.get("serviceAccountName", "default")
+                assert sa_name == "default", f"Expected default ServiceAccount, but got {sa_name}"
```
danielhoherd, line 28: These `.get()` are not going to prevent the assertion from failing, so we don't really need them. We can just assume that the data structure is what we hope to find, because we'll encounter some kind of failure if it's not.

Also we don't want to insert a default value into the tests. We want to see whatever comes out of the helm chart. If we need a default, that belongs in the helm template, not the test case.

```suggestion
                assert (
                    sa_name == doc["spec"]["template]"]["spec"]["serviceAccountName"]
                ), f"Expected default ServiceAccount, but got {sa_name}"
```

danielhoherd (changes requested):
I left some comments I'd like to see addressed. I'd be happy to jump on zoom and talk about them if that would be helpful.

charts/astronomer/values.yaml
Shubham-astro: this is removed.

tests/chart_tests/test_byo_sa.py
Shubham-astro: done

tests/chart_tests/test_byo_sa.py
Shubham-astro: done

charts/astronomer/templates/_helpers.yaml
```diff
@@ -139,6 +135,26 @@
...
+{{- end }}
+{{- end }}
+
+{{ define "commander.ServiceAccount" -}}
+{{- if .Values.commander.serviceAccount.name -}}
+{{ printf "%s-%s" .Release.Name .Values.commander.serviceAccount.name }}
+{{- else -}}
+{{ printf "%s-commander" .Release.Name }}
+{{- end }}
+{{- end }}
```
danielhoherd, line 156: Would it be better to have the boostrapperServiceAccount use the exact same logic as the other ones, if for no other reason than to be consistent? CC @pgvishnuram

charts/grafana/templates/_helpers.yaml
```diff
@@ -46,8 +46,9 @@ grafana.{{ .Values.global.baseDomain }}
 {{ default (printf "%s-grafana-backend" .Release.Name) .Values.backendSecretName }}
 {{- end }}
 
+
 {{ define "grafana.bootstrapper.serviceaccount" -}}
-{{ printf "%s-grafana-bootstrapper" .Release.Name }}
+  {{ printf "%s-grafana-bootstrapper" .Release.Name }}
```
danielhoherd, line 51: This leading space is inconsistent with similar templates.
```suggestion
{{ printf "%s-grafana-bootstrapper" .Release.Name }}
```

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,40 @@
...
+from tests.chart_tests.helm_template_generator import render_chart
+
+
+@pytest.mark.parametrize(
+    "kube_version",
+    supported_k8s_versions,
+)
+class TestServiceAccounts:
+    def test_serviceaccount_rbac_disabled(self, kube_version):
+        # Render the chart with rbacEnabled set to False
```
danielhoherd, line 13: Unnecessary comment, we can see in the next two lines that rbacEnabled is False.
```suggestion
```

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,40 @@
...
+        # Check that no ServiceAccount resources are created
+
+        service_accounts = [
+            doc["metadata"]["name"] for doc in docs if isinstance(doc, dict) and doc.get("kind") == "ServiceAccount"
+        ]
+        assert len(service_accounts) == 0, "No ServiceAccounts should be created when rbacEnabled is False"
+
+        # Check that the Deployment or StatefulSet is using the default ServiceAccount
+        sa_name = ""
+        # print(docs)
```
danielhoherd, line 27:
```suggestion
```

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,40 @@
...
+        # Render the chart with rbacEnabled set to False
+        docs = render_chart(
+            kube_version=kube_version, values={"global": {"rbacEnabled": False}, "nats": {"nats": {"createJetStreamJob": False}}}
+        )
+
+        # Check that no ServiceAccount resources are created
+
+        service_accounts = [
+            doc["metadata"]["name"] for doc in docs if isinstance(doc, dict) and doc.get("kind") == "ServiceAccount"
+        ]
```
danielhoherd, line 22:
```suggestion
        service_accounts = [doc["metadata"]["name"] for doc in docs if doc["kind"] == "ServiceAccount"]
```

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,40 @@
...
+        assert len(service_accounts) == 0, "No ServiceAccounts should be created when rbacEnabled is False"
+
+        # Check that the Deployment or StatefulSet is using the default ServiceAccount
+        sa_name = ""
+        # print(docs)
+        for doc in docs:
+            if doc.get("kind") in ["Deployment", "StatefulSet"]:
+                assert (
+                    sa_name == doc["metadata"]["name"] for doc in docs if doc.get("kind") == "ServiceAccount"
+                ), f"Expected default ServiceAccount, but got {sa_name}"
```
danielhoherd, line 32: This logic doesn't make sense. We're iterating docs twice, once on line 28, once on line 31.

charts/prometheus-node-exporter/values.yaml
```diff
@@ -40,7 +40,7 @@ resources:
 
 serviceAccount:
   # Specifies whether a ServiceAccount should be created
-  create: true
+  create: false
```
danielhoherd, line 43: Why was this changed to false?

charts/grafana/templates/_helpers.yaml
```diff
@@ -46,6 +46,7 @@ grafana.{{ .Values.global.baseDomain }}
 {{ default (printf "%s-grafana-backend" .Release.Name) .Values.backendSecretName }}
 {{- end }}
 
+
```
danielhoherd, line 49: Inconsistent extra line
```suggestion
```

charts/astronomer/templates/_helpers.yaml
```diff
@@ -139,6 +135,30 @@
 {{- end }}
 {{- end }}
 
+{{ define "houston.bootstrapperServiceAccount" -}}
+{{- if .Values.houston.serviceAccount.create -}}
+{{ default (printf "%s-houston-bootstrapper" .Release.Name ) .Values.houston.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.houston.serviceAccount.name }}
+{{- end }}
+{{- end }}
```
pgvishnuram, line 144: @danielhoherd @Shubham-astro i made this change should we adopt to all other templates as well . please add your views on the comment

Shubham-astro: So, if block is fine, else block means if houston.serviceAccount.name is not defined it defaults to "default" and the template will use the existing "default" service account in the namespace, I think we can use this for other SA's.

tests/chart_tests/test_byo_sa.py
Shubham-astro: removed the extra iteration.

charts/astronomer/templates/_helpers.yaml
Shubham-astro: this was changed to 
`{{ define "houston.bootstrapperServiceAccount" -}}
{{- if .Values.houston.serviceAccount.create -}}
{{ default (printf "%s-houston-bootstrapper" .Release.Name ) .Values.houston.serviceAccount.name }}
{{- else -}}
    {{ default "default" .Values.houston.serviceAccount.name }}
{{- end }}
{{- end }}` by @pgvishnuram

charts/astronomer/templates/_helpers.yaml
danielhoherd: I definitely think we should be consistent wherever possible.

I also like how this template keeps the logic in the helm template and avoids template logic in the kubernetes manifest.

I say we should adopt this wherever possible.

What may be even better is if we can create a single function that would handle all of these duplicate template definitions.

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,31 @@
...
+from tests.chart_tests.helm_template_generator import render_chart
+
+
+@pytest.mark.parametrize(
+    "kube_version",
+    supported_k8s_versions,
+)
+class TestServiceAccounts:
+    def test_serviceaccount_rbac_disabled(self, kube_version):
+        docs = render_chart(
```
danielhoherd, line 13:
```suggestion
        """Test that no ServiceAccounts are rendered when rbac is disabled."""
        docs = render_chart(
```

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,31 @@
...
+            kube_version=kube_version, values={"global": {"rbacEnabled": False}, "nats": {"nats": {"createJetStreamJob": False}}}
+        )
+
+        # Check that the Deployment or StatefulSet is using the default ServiceAccount
```
danielhoherd, line 17: This comment is incorrect, right?

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,31 @@
...
+        sa_name = ""
+        for doc in docs:
+            if doc.get("kind") == "ServiceAccount":
+                assert (
+                    sa_name == doc["metadata"]["name"]
+                ), f"No ServiceAccounts were expected when rbacEnabled is False, but got {sa_name}"
```
danielhoherd, line 23: The assertion in this block is misleading. It's asserting about the contents of `doc["metadata"]["name"], but that is not the problem. The problem is that there is any service account at all.

IMHO the better way to handle this would be to check that there are no ServiceAccount entities, and report about which ones are found.

```py
service_account_names = [doc['metadata']['name'] for doc in docs if doc['kind'] == 'ServiceAccount']
assert not doc, f'Expected no ServiceAccounts but found {service_account_names}'
```

danielhoherd (changes requested):
We have these new ServiceAccount helm templates:

```
{{ define "commander.ServiceAccount" -}}
{{ define "configSyncer.ServiceAccount" -}}
{{ define "houston.bootstrapperServiceAccount" -}}
{{ define "pgbouncer.ServiceAccount" -}}
```

But `commander.ServiceAccount` and `configSyncer.ServiceAccount` are never used in the chart. Why is that?

pgvishnuram:
@Shubham-astro we asked to call as helper templates but it didnt referenced in the serviceaccount

tests/chart_tests/test_byo_sa.py
Shubham-astro: thanks, removed

tests/chart_tests/test_byo_sa.py
Shubham-astro: makes sense

rishkarajgi (approved):

tests/chart_tests/test_byo_sa.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+        service_account_names = [doc["metadata"]["name"] for doc in docs if doc["kind"] == "ServiceAccount"]
+        assert not service_account_names, f"Expected no ServiceAccounts but found {service_account_names}"
+
+    def test_role_created(self, kube_version):
+        """Test that no roles or rolebindings are created when rbac is disabled."""
+        values = {"global": {"rbacEnabled": False}, "nats": {"nats": {"createJetStreamJob": False}}}
+
+        docs = [doc for doc in render_chart(kube_version=kube_version, values=values) if doc["kind"] in ["RoleBinding", "Role"]]
+        print(docs)
```
danielhoherd, line 25:
```suggestion
```

danielhoherd (approved):
LGTM. I left one optional `print()` removal.

PR #2308, opened 2024-09-13
Make helm values for docker images more consistent

Shubham-astro:
## Description

This PR adds images section to a few components like loggingSidecar, certCopier, authSidecar and dagOnlyDeployment

## Related Issues

Related astronomer/issues#6267

## Testing

## Merging


values.yaml
```diff
@@ -126,7 +127,9 @@ global:
 
   dagOnlyDeployment:
     enabled: false
-    image: quay.io/astronomer/ap-dag-deploy:0.5.3
+    images:
+      repository: quay.io/astronomer/ap-dag-deploy
+      tag: 0.5.3
```
pgvishnuram, line 132: we actually dont need images we can just keep it simple like
```
dagOnlyDeployment:
    enabled: false
    repository: quay.io/astronomer/ap-dag-deploy
    tag: 0.5.3
```

pgvishnuram: @danielhoherd what do you think if you are okay with above format i'm fine

Shubham-astro: Sure, will do the same, I thought that we need to keep it similar to what we have at other places like: https://github.com/astronomer/astronomer/blob/3b3a9481295fe95d7074d3529a2d1c7a3d5835de/charts/alertmanager/values.yaml#L9C1-L13C29

values.yaml
```diff
@@ -139,7 +143,10 @@ global:
   loggingSidecar:
     enabled: false
     name: sidecar-log-consumer
-    image: quay.io/astronomer/ap-vector:0.40.1
+    images:
+      repository: quay.io/astronomer/ap-vector
+      tag: 0.40.1
+      pullPolicy: IfNotPresent
```
pgvishnuram, line 149: let not add this pullPolicy in this PR this requires a code change

Shubham-astro: ok, will revert this.

values.yaml
pgvishnuram: https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/houston/houston-configmap.yaml#L152-L158

we are internally parsing the format from our configmap i feel like global can be real simple initial ask was to have repository and tag as seperate just a thought here

pgvishnuram: you need to fix in the configmap as well since i have added a split list directly in the config

values.yaml
```diff
@@ -155,9 +162,10 @@ global:
   # Deploy auth sidecar to use openshift native features
   authSidecar:
     enabled: false
-    repository: quay.io/astronomer/ap-auth-sidecar
-    tag: 1.25.5
-    pullPolicy: IfNotPresent
+    images:
+      repository: quay.io/astronomer/ap-auth-sidecar
+      tag: 1.25.5
+      pullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
Shubham-astro: done

values.yaml
```diff
@@ -27,9 +27,10 @@ global:
     containerdnodeAffinitys: []
     containerdTolerations: []
     certCopier:
-      repository: quay.io/astronomer/ap-base
-      tag: 3.18.8
-      pullPolicy: IfNotPresent
+      images:
+        repository: quay.io/astronomer/ap-base
+        tag: 3.18.8
+        pullPolicy: IfNotPresent
```
pgvishnuram, line 33: https://github.com/astronomer/astronomer/blob/fix/6267/templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml#L58 - requires changes here too

Shubham-astro: done

values.yaml
danielhoherd: FWIW there is an issue to standardize the way we configure images so that they are all the same data structure. It might be worth thinking about that when we think about what to do here.

danielhoherd: How could we make this so that changing one value enables an airgap? For instance, how could we make the `quay.io/astronomer` part a global value, and we only have `repo: ap-vector`, `tag: 1.2.3` ?

pgvishnuram:
@danielhoherd we didnt reference global.<services>.images.repository because underlying helm config referenced in houston configmap impose that users who wants to overrides them will have to pass them in deployments.helm.<services>.images but not in globals to make it very minimal and make user representation simpler

values.yaml
Shubham-astro: @pgvishnuram config map for fixed, can you check and let me know if there is anything else.?

pgvishnuram: I have fixed failing ones i have some doubt on formats imposed at global level we can discuss on the standup to finalize and merge the changes

values.yaml
```diff
@@ -139,7 +143,9 @@ global:
   loggingSidecar:
     enabled: false
     name: sidecar-log-consumer
-    image: quay.io/astronomer/ap-vector:0.40.2-1
+    images:
+      repository: quay.io/astronomer/ap-vector
+      tag: 0.40.1
```
danielhoherd, line 148: Why are we not going with 0.40.2-1 ?

pgvishnuram: that we have not tested yet and we are not combining with this change

values.yaml
```diff
@@ -30,6 +30,7 @@ global:
       repository: quay.io/astronomer/ap-base
       tag: 3.18.9
       pullPolicy: IfNotPresent
+
```
pgvishnuram, line 33: is this space introduced by pre-commit

Shubham-astro: removed

.circleci/config.yml
```diff
@@ -389,7 +389,7 @@ workflows:
                 - quay.io/astronomer/ap-postgresql:15.8.0
                 - quay.io/astronomer/ap-prometheus:2.53.1
                 - quay.io/astronomer/ap-registry:3.18.9
-                - quay.io/astronomer/ap-vector:0.40.2-1
```
rishkarajgi, line 392: why have we reverted this version?

values.yaml
```diff
@@ -139,7 +140,8 @@ global:
   loggingSidecar:
     enabled: false
     name: sidecar-log-consumer
-    image: quay.io/astronomer/ap-vector:0.40.2-1
+    repository: quay.io/astronomer/ap-vector
+    tag: 0.40.1
```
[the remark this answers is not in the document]
pgvishnuram: it should be 0.40.2-1 as per master branch

rishkarajgi (approved):

PR #2311, opened 2024-09-16
Add a few helper scripts

danielhoherd:
## Description

These are helper scripts that we can use during development or customer engagements.

I suppose these scripts could go somewhere else, but I don't know where that would be. [Neso](https://github.com/astronomer/neso) mabye? But that repo is not public.

These scripts are currently not called anywhere, so there is no need to merge master and waste CircleCI credits re-running CI.

## Related Issues

<https://github.com/astronomer/issues/issues/6632>

## Testing

No testing needed, these are just helper scripts for dev workflows.

## Merging

We might as well merge these everywhere. They are never called during operation of our app or in CI.


danielhoherd:
@pgvishnuram @rishkarajgi:

> These scripts are currently not called anywhere, so there is no need to merging master and let CI re-run or anything like that.

Are we going to approve this PR, because if not then we should close it.

rishkarajgi (approved):

pgvishnuram:
This is fine to be in master branch - im merging to master

PR #2318, opened 2024-09-24
Fix 6651 disable default jetstream

Shubham-astro:
## Description

This PR removes the default usage of the jetstream component. 

## Related Issues

Related astronomer/issues#6651

## Testing

Tested locally with the modified testcases

Screenshot 2024-09-24 at 6 26 46 PM

## Merging

### DIFF
--- a/tests/chart_tests/test_nats_jetstream.py
+++ b/tests/chart_tests/test_nats_jetstream.py
@@ -13,6 +13,7 @@ class TestNatsJetstream:
         """Test that nats statefulset is good with defaults."""
         values = {
             "global": {"nats": {"jetStream": {"enabled": True}}},
+            "nats": {"nats": {"createJetStreamJob": True}},
         }
         docs = render_chart(
             kube_version=kube_version,
@@ -36,6 +37,7 @@ class TestNatsJetstream:
         """Test Nats with jetstream config."""
         values = {
             "global": {"nats": {"jetStream": {"enabled": True, "tls": True}}},
+            "nats": {"nats": {"createJetStreamJob": True}},
         }
         docs = render_chart(
             kube_version=kube_version,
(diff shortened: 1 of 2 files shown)


tests/chart_tests/test_nats_jetstream.py
```diff
@@ -13,6 +13,7 @@ def test_nats_statefulset_with_jetstream(self, kube_version):
         """Test that nats statefulset is good with defaults."""
         values = {
             "global": {"nats": {"jetStream": {"enabled": True}}},
+            "nats": {"nats": {"createJetStreamJob": True}},
```
pgvishnuram, line 16: this is a default test cases we are ensuring templates are rendered or not , so its disabled by default it should just asserted with len(docs)

Shubham-astro: Ok, so the assert len(docs) should be 3, but then there is a condition in the next lines, should that be removed?:
`assert "jetstream" in nats_cm
  assert {"runAsUser": 1000, "runAsNonRoot": True} == docs[2]["spec"]["template"]["spec"]["containers"][0]["securityContext"]`

Shubham-astro: done.

danielhoherd:
@pgvishnuram is this the right approach? Looking at the original ticket, it seems like the right approach would be to move away from labeling the namespace to labeling some other namespace-local component.

What is the value of these jetstream jobs, and do are we really OK disabling them by default?

rishkarajgi:
> @pgvishnuram is this the right approach? Looking at the original ticket, it seems like the right approach would be to move away from labeling the namespace to labeling some other namespace-local component.
> 
> What is the value of these jetstream jobs, and do are we really OK disabling them by default?

@danielhoherd this will be a temporary fix and the purpose of this job is to allow switchback from jetstream -> STAN and vice versa in an automated manner (creation of new nats stateful set with jetstream support)
There is a manual override to this via a kubectl command `kubectl delete sts astronomer-nats —cascade=background`
This will delete the stateful set template but the pod will continue to run and on a helm upgrade it will recreate the new template (less destructive approach)

rishkarajgi (approved):

PR #2320, opened 2024-09-26
Add release_bom CI job

Shubham-astro:
## Description

This PR adds capability to CircleCI script to run the bom_generator script.

## Related Issues

Related astronomer/issues#6117

## Testing

Not needed

## Merging

Everywhere


pgvishnuram:
@danielhoherd should this not be linked to public release flow every time a public release we publich a bom or im missing anything here

.circleci/config.yml
```diff
@@ -514,7 +530,15 @@ workflows:
...
-
+      - release_bom:
+          context:
+            - github-repo
+            - gcp-astronomer-prod
+          filters:
+              branches:
+                only:
+                  - '/^release-0\.\d+$/'
+                  - fix-6117-adding-ci-job
```
danielhoherd, line 541: This need to be removed before merge.

.circleci/config.yml
```diff
@@ -514,7 +530,15 @@ workflows:
...
                 - '/^release-0\.\d+$/'
-
+      - release_bom:
+          context:
+            - github-repo
+            - gcp-astronomer-prod
+          filters:
+              branches:
+                only:
+                  - '/^release-0\.\d+$/'
```
danielhoherd, line 540: We should do this on tag, not branch. Here's an example of how it works with tags: https://github.com/astronomer/astro/blob/e071bd49ae/.circleci/config.yml#L427-L431

We need to find a regex that matches our tag pattern, which would be `v<semver-compatible-regex>`

.circleci/config.yml.j2
```diff
@@ -378,7 +394,15 @@ workflows:
...
-
+      - release_bom:
+          context:
+            - github-repo
+            - gcp-astronomer-prod
+          filters:
+              branches:
+                only:
+                  - '/^release-0\.\d+$/'
+                  - fix-6117-adding-ci-job
```
danielhoherd, line 405: We should do this on tag, not branch. Here's an example of how it works with tags: https://github.com/astronomer/astro/blob/e071bd49ae/.circleci/config.yml#L427-L431

We need to find a regex that matches our tag pattern, which would be `v<semver-compatible-regex>`

.circleci/config.yml
Shubham-astro: done

.circleci/config.yml
```diff
@@ -108,6 +108,22 @@ jobs:
...
+      - run:
+          name: Install dependencies
+          command: |
+            python -m venv venv
+            venv/bin/pip install git+https://${GITHUB_TOKEN}@github.com/astronomer/astronomer-bom-generator.git
+      - run:
+          name: Run Release BOM Script
+          command: |
+            source venv/bin/activate
+            bin/release-bom
```
danielhoherd, line 125:
```suggestion
            venv/bin/release-bom
```

.circleci/config.yml.j2
```diff
@@ -106,6 +106,22 @@ jobs:
...
+      - run:
+          name: Install dependencies
+          command: |
+            python -m venv venv
+            venv/bin/pip install git+https://${GITHUB_TOKEN}@github.com/astronomer/astronomer-bom-generator.git
+      - run:
+          name: Run Release BOM Script
+          command: |
+            source venv/bin/activate
+            bin/release-bom
```
danielhoherd, line 123:
```suggestion
          command: |
            venv/bin/release-bom
```

.circleci/config.yml
```diff
@@ -108,6 +108,26 @@ jobs:
...
 
+  release_bom:
+    docker:
+      - image: cimg/python:3.12
+    steps:
+      - checkout
+      - run:
+          name: Get Version from Git Tag
+          command: |
+            VERSION=$(git describe --tags --abbrev=0)
```
danielhoherd, line 119: This syntax doesn't work as expected. It is currently showing `v0.32.0` even though the newest tag is `v0.35.4`

.circleci/config.yml
```diff
@@ -501,7 +521,15 @@ workflows:
...
-
+      - release_bom:
+          context:
+            - github-repo
+            - gcp-astronomer-prod
+          filters:
+              branches:
+                only:
+                  - '/^release-0\.\d+$/'
+                  - 'fix-6117-adding-ci-job'
```
danielhoherd, line 532: We should change this to only run on tags, not on branches. If we do that, then we can use the CIRCLECI env var that shows which tag triggered the run, `CIRCLE_TAG` <https://circleci.com/docs/variables/>

bin/release-bom
```diff
@@ -28,32 +28,45 @@ fi
...
 fi
 
+echo "Installing gcloud"
+sudo apt-get update
+sudo apt-get install apt-transport-https ca-certificates gnupg curl
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo gpg --dearmor -o /usr/share/keyrings/cloud.google.gpg
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+sudo apt-get update && sudo apt-get install google-cloud-cli
```
danielhoherd, line 55: We should not do this in this script. Instead, we should ensure that this CI step runs within a container that already has the necessary scripts set up, like quay.io/astronomer/ci-helm-release <https://github.com/astronomer/ci-images/blob/f8db5ecb69/helm-release/Dockerfile#L43>

.circleci/config.yml
Shubham-astro: So it is working for the current branch it is triggered from, for example currently it is showing:
```
shubhamsingh@Shubhams-MacBook-Pro astro % git describe --tags --abbrev=0
v0.32.0
```

.circleci/config.yml.j2
Shubham-astro: done

.circleci/config.yml
Shubham-astro: sure, added the tag for now i am keeping the branches as well for testing, @danielhoherd can you explain how we can use CIRCLE_TAG.? Should it be put as an echo $CIRCLE_TAG .?

bin/release-bom
Shubham-astro: @danielhoherd I have added these libraries in circleci config, is that fine.?

.circleci/config.yml
Shubham-astro: sure, added the tag for now i am keeping the branches as well for testing, @danielhoherd

bin/release-bom
danielhoherd: no, that's what i'm saying. we don't want to install these in the circleci pipeline. we want to run in an image that has these already installed.

Shubham-astro: done

.circleci/config.yml
Shubham-astro: changed it to use CIRCLE_TAG env variable

.circleci/config.yml
Shubham-astro: CIRCLE_TAG didn't work but awk '/^version: / {print $2}' Chart.yaml works fine @danielhoherd

danielhoherd: CIRCLE_TAG is not going to work unless it's running from a tag, so it won't work in this branch.

bin/release-bom
```diff
@@ -28,13 +31,18 @@ fi
...
 max_age="300"
+index_file="index.html"
+
+# Step 0: Installing uv package manager
+curl -LsSf https://astral.sh/uv/install.sh | sh
 
 # Step 1: Runs the Python extractor script to generate the BOM JSON file
 echo "Running the extractor script to generate ${json_file_name}..."
-pip install "git+https://${GITHUB_TOKEN}@github.com/astronomer/astronomer-bom-generator.git@${BOM_GENERATOR_VERSION}"
+uv tool run --from git+https://"${GITHUB_TOKEN}"@github.com/astronomer/astronomer-bom-generator.git"${BOM_GENERATOR_VERSION}" bom-generator --version "${version}"
```
danielhoherd, line 45: Let's not install `uv` here, let's just go with `pip install`

.circleci/config.yml.j2
```diff
@@ -106,6 +106,27 @@ jobs:
...
+      - image: quay.io/astronomer/ci-helm-release:{{ ci_runner_version }}
+    steps:
+      - checkout
+      - run:
+          name: Get current Version from the chart
+          command: |
+            VERSION=$(awk '/^version: / {print $2}' Chart.yaml)
+            VERSION=${VERSION#v}
+            pwd
+            echo "VERSION=${VERSION}" >> $BASH_ENV
```
danielhoherd, line 120: Let's just use `${CIRCLE_TAG}` for the version number.

.circleci/config.yml.j2
```diff
@@ -367,7 +388,14 @@ workflows:
...
                 - '/^release-0\.\d+$/'
-
+      - release_bom:
+          context:
+            - github-repo
+            - gcp-astronomer-prod
+          filters:
+              tags:
+                only:
+                  - '/^v\d+\.\d+\.\d+(-?(rc(\.\d+)?|alpha(\.\d+)?)?)?$/'
```
danielhoherd, line 398:
```suggestion
            tags:
              only:
                - '/^v\d+\.\d+\.\d+(-?(rc(\.\d+)?|alpha(\.\d+)?)?)?$/'
```

.circleci/config.yml
```diff
@@ -501,7 +522,14 @@ workflows:
...
                 - '/^release-0\.\d+$/'
-
+      - release_bom:
+          context:
+            - github-repo
+            - gcp-astronomer-prod
+          filters:
+              tags:
+                only:
+                  - '/^v\d+\.\d+\.\d+(-?(rc(\.\d+)?|alpha(\.\d+)?)?)?$/'
```
danielhoherd, line 525: Let's put this in its own workflow.

.circleci/config.yml.j2
```diff
@@ -379,6 +392,16 @@ workflows:
               only:
                 - '/^release-0\.\d+$/'
 
+  release-bom-worflow:
```
danielhoherd, line 395:
```suggestion
  release-bom-workflow:
```

danielhoherd (approved):
This looks good. We'll keep an eye on this during the next release and see if we need to tweak anything.

danielhoherd:
cherry-picked to release-0.34,35,36

PR #2322, opened 2024-09-27
Fix 6582 ns resources

Shubham-astro:
## Description

This PR introduces a keep flag so the already create role and rolebindings are not deleted and synchronize on a helm update/upgrade. Disable management of namespace resources - patching labels/annotations etc.

## Related Issues

## Testing

## Merging

### DIFF
--- a/charts/astronomer/templates/commander/commander-role.yaml
+++ b/charts/astronomer/templates/commander/commander-role.yaml
@@ -4,10 +4,18 @@
 # Here, we either:
 # 1. Do not create any resource if rbacEnabled is disabled
 # 2. Create a Cluster Role if clusterRoles and rbacEnabled are enabled and namespacePools is disabled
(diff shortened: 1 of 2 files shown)


charts/astronomer/templates/commander/commander-role.yaml
```diff
@@ -24,15 +25,20 @@ metadata:
   namespace: {{ $namespaceName }}
   {{- end }}
   name: {{ $.Release.Name }}-commander
+  {{- if not $.Values.global.disableManageClusterScopedResources }}
```
pgvishnuram, line 28: can we know why this condition is added here

pgvishnuram: there are created by the installer service account which is provided at the time of installation of platform

PR #2333, opened 2024-10-01
Disable commander namespace creation with MANUAL_NAMESPACES_MODE enabled

Shubham-astro:
## Description

This PR introduces a keep flag so the already create role and rolebindings are not deleted and synchronize on a helm update/upgrade. Disable management of namespace resources - patching labels/annotations etc

## Related Issues

## Testing

## Merging


charts/astronomer/templates/commander/commander-role.yaml
```diff
@@ -29,20 +29,26 @@ metadata:
     release: {{ $.Release.Name }}
     chart: "{{ $.Chart.Name }}-{{ $.Chart.Version }}"
     heritage: {{ $.Release.Service }}
-  {{- if $.Values.global.enableArgoCDAnnotation }}
+
   annotations:
+    {{- if not $.Values.global.features.namespacePools.namespaces.create }}
+    helm.sh/resource-policy: keep
```
danielhoherd, line 35: Why are we adding this? Won't this orphan this object? https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

charts/astronomer/templates/commander/commander-role.yaml
```diff
@@ -29,20 +29,26 @@ metadata:
...
-  {{- end }}
+    {{- end }}
 rules:
 - apiGroups: [""]
   resources: ["configmaps"]
   verbs: ["create", "delete", "deletecollection", "get", "list", "patch", "update", "watch"]
 - apiGroups: [""]
   resources: ["secrets"]
   verbs: ["create", "delete", "deletecollection", "get", "list", "patch", "update", "watch"]
+{{- if not $.Values.global.disableManageClusterScopedResources }}
```
danielhoherd, line 47: This double negative is unfortunate. Is this something we can change before release, or do we have to live with this?

<https://cleankotlin.nl/blog/double-negations>

danielhoherd (changes requested):
I have two comments I'd like to see responses for before approving this.

charts/astronomer/templates/commander/commander-role.yaml
Shubham-astro: So the variable is about disableManageClusterScopedResources, and it is set to false by default, once this is enabled we do not manage the namespace in commander that is why the double negative.

charts/astronomer/templates/commander/commander-role.yaml
Shubham-astro: So this is done to make sure already created roles and role bindings are not deleted that is why this keep flag is introduced. @danielhoherd

danielhoherd: When we do this, we lose all control of helm managing these resources. Is that really what we want? CC @pgvishnuram 

> However, this resource becomes orphaned. Helm will no longer manage it in any way. This can lead to problems if using helm install --replace on a release that has already been uninstalled, but has kept resources.

charts/astronomer/templates/commander/commander-role.yaml
pgvishnuram: feature flag MANUAL_NAMESPACE_NAMES is activated in combination with  namespace pools that wont use clusterrole and  clusterrolebinding

pgvishnuram: can you assess the impact if we remove these permissions with namespacepools with some deployments

danielhoherd: We don't need to solve this double negative. It's just a nuance of how we are configuring this. `disableManageClusterScopedResources` already exists in master branch.

PR #2340, opened 2024-10-04
remove rbac conditional changes

pgvishnuram:
## Description

remove rbac conditional changes

This is a continutation from the PR https://github.com/astronomer/astronomer/pull/2301

our original intention was to mix it up with global.rbacEnabled so that we will be removing the serviceAccount section from all the templates, with the redefined approach we configured our helper templates to use default service account when serviceAccount create is set to false and no name is defined it will fallback the serviceAccountName to default 

## Related Issues

https://github.com/astronomer/issues/issues/6537
https://github.com/astronomer/issues/issues/6583

## Testing

refer notion doc

## Merging

cherry-pick to release-0.36

### DIFF
--- a/charts/astronomer/templates/houston/helm-hooks/houston-au-strategy-job.yaml
+++ b/charts/astronomer/templates/houston/helm-hooks/houston-au-strategy-job.yaml
@@ -35,9 +35,7 @@ spec:
         sidecar.istio.io/inject: "false"
       {{- end }}
     spec:
-      {{- if .Values.global.rbacEnabled }}
       serviceAccountName: {{ template "houston.bootstrapperServiceAccount" . }}
-      {{- end}}
       nodeSelector:
 {{ toYaml (default .Values.global.platformNodePool.nodeSelector .Values.nodeSelector) | indent 8 }}
       affinity:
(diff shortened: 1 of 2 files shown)


danielhoherd:
@pgvishnuram what is the context here? why is this change needed?

rishkarajgi (changes requested):
Update PR description on what is the issue you are addressing, why these changes are needed and if any breaking changes. Add link to Notion doc

pgvishnuram:
This is an update from previous PR where we missed to remove the rbac conditions

rishkarajgi (changes requested):
please update the PR description. Add details on why are doing this change and what is the problem we are trying to solve

pgvishnuram:
@rishkarajgi this is a missed commit from the list of changes we introduced in PR https://github.com/astronomer/astronomer/pull/2301

rishkarajgi (approved):

rishkarajgi (changes requested):
Still don't understand the context, what is the need to remove `if .Values.global.rbacEnabled` wrt BYO service accounts?

rishkarajgi (approved):

PR #2342, opened 2024-10-07
Introduce rbac flag to optionate role and rolebindings for namespace pools

Shubham-astro:
## Description

This PR introduces a keep flag so the already create role and rolebindings are not deleted and synchronize on a helm update/upgrade. Disable management of commander namespace resources.

## Related Issues
Related https://github.com/astronomer/issues/issues/6582

## Testing

## Merging

### DIFF
--- a/charts/astronomer/templates/commander/commander-role.yaml
+++ b/charts/astronomer/templates/commander/commander-role.yaml
@@ -5,7 +5,7 @@
 # 1. Do not create any resource if rbacEnabled is disabled
 # 2. Create a Cluster Role if clusterRoles and rbacEnabled are enabled and namespacePools is disabled
 # 3. Create roles for each namespaces in the namespacePool (+ astronomer namespace) if enabled.
-{{- $namespacepoolsRBACresources:= and .Values.global.features.namespacePools.rbacCreate .Values.global.features.namespacePools.enabled}}
+{{- $namespacepoolsRBACresources:= and .Values.global.features.namespacePools.createRbac .Values.global.features.namespacePools.enabled}}
 {{- $useClusterRoles := and .Values.global.rbacEnabled .Values.global.clusterRoles (not .Values.global.features.namespacePools.enabled)}}
 {{- $shouldCreateResources := and .Values.global.rbacEnabled (or .Values.global.clusterRoles $namespacepoolsRBACresources) }}
 {{- $namespaces := .Values.global.features.namespacePools.namespaces.names }}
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -41,6 +41,7 @@ global:
     # namespace pools: define k8s namespaces in which Astronomer will manage Airflow deployments
     namespacePools:
       enabled: false
+      keeproleandrolebindings: false
```
pgvishnuram, line 44: it can just be role: true or false it should be default to true

pgvishnuram: here we are not keeping the existing roles we are creating an option to either allow creation of roles by platform or by customer and our naming

Shubham-astro: @pgvishnuram one of the acceptance criteria says: 
Introduce a keep flag so already create role and rolebindings are not deleted and synchronize on a helm update/upgrade.

pgvishnuram: why do we need a keep flag here rbac: true is default unless customer sets them to false so this is fully of customer responsibility to make revoke resources and the naming ```keeproleandrolebindings``` does make any for the customers who already has those rbac applied

Shubham-astro: done, since by default we are keeping the value to true so customer when needs to recreate role can mark it as false

Shubham-astro:
@pgvishnuram can you re-review.?

charts/astronomer/templates/commander/commander-role.yaml
```diff
@@ -5,6 +5,7 @@
 # 1. Do not create any resource if rbacEnabled is disabled
 # 2. Create a Cluster Role if clusterRoles and rbacEnabled are enabled and namespacePools is disabled
 # 3. Create roles for each namespaces in the namespacePool (+ astronomer namespace) if enabled.
+{{- $commanderrole:= and .Values.global.features.namespacePools.keeproleandrolebindings (not $.Values.global.features.namespacePools.namespaces.create)}}
 {{- $useClusterRoles := and .Values.global.rbacEnabled .Values.global.clusterRoles (not .Values.global.features.namespacePools.enabled)}}
 {{- $shouldCreateResources := and .Values.global.rbacEnabled (or .Values.global.clusterRoles .Values.global.features.namespacePools.enabled) }}
```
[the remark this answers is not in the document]
danielhoherd: @pgvishnuram 

> both should be replaced by above condition

both what? can you be more specific?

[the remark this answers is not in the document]
Shubham-astro: this is done now.

values.yaml
```diff
@@ -41,6 +41,8 @@ global:
     # namespace pools: define k8s namespaces in which Astronomer will manage Airflow deployments
     namespacePools:
       enabled: false
+      # if rbacCreate is set to false then platform does not create roles and rolebinding.
+      rbacCreate: true
```
danielhoherd, line 45: This should probably be createRbac so it follows the same form as other variables we use, like `createJetStreamJob`, `createSecret`, `createDefaultIndex`, etc..

Shubham-astro: done, changes to createRbac

danielhoherd (approved):
Seems good

pgvishnuram (approved):

PR #2343, opened 2024-10-08
add support for runtime release config overrides from config

pgvishnuram:
## Description

Allow users to pass runtime versions list  overrides from astronomer helm chart 

## Related Issues

https://github.com/astronomer/issues/issues/6680

## Testing

QA should able to pass custom runtime versions config from helm values

## Merging

cherry-pick to all valid releases

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -311,6 +311,19 @@ houston:
     #   - 2.0.0-alpine3.10-onbuild
     #   channel: stable
 
+  RuntimeReleasesConfig:
+    # example of usage:
+    # available_releases:
+    #runtimeVersions:
+    #  4.2.6:
+    #    metadata:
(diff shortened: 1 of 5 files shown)


charts/astronomer/values.yaml
```diff
@@ -311,6 +311,19 @@ houston:
...
+    # example of usage:
+    # available_releases:
+    #runtimeVersions:
+    #  4.2.6:
+    #    metadata:
+    #      airflowVersion: 2.2.5
+    #      channel: stable
+    #      releaseDate: '2022-04-19'
+    #    migrations:
+    #      airflowDatabase: false
```
danielhoherd, line 324: What does this field do? Users will want to know when it needs to be included and if it can be omitted.

danielhoherd (approved):
LGTM. I left some optional suggestions that would probably help our users.

PR #2351, opened 2024-10-09
Add missing resources definition for astronomer components

pgvishnuram:
## Description

Add missing resources definition for astronomer components

## Related Issues

https://github.com/astronomer/issues/issues/6615

## Testing

QA should able to pass resources to above referenced commented services

## Merging

cherry-pick to master, release-0.36

### DIFF
--- a/tests/chart_tests/test_resources.py
+++ b/tests/chart_tests/test_resources.py
@@ -32,12 +32,13 @@ def init_test_pod_resources():
 test_pod_resources_configs_data = init_test_pod_resources()
 
 
-@pytest.mark.parametrize(
-    "pod_resources",
-    test_pod_resources_configs_data.values(),
-    ids=test_pod_resources_configs_data.keys(),
-)
-def test_pod_resources_configs(pod_resources):
-    """Labels check for definition."""
-    for container in pod_resources:
-        assert "resources" in container
+class TestPodResources:
+    @pytest.mark.parametrize(
+        "pod_resources",
+        test_pod_resources_configs_data.values(),
+        ids=test_pod_resources_configs_data.keys(),
+    )
+    def test_pod_resources_configs(pod_resources):
+        """Labels check for definition."""
+        for container in pod_resources:
+            assert "resources" in container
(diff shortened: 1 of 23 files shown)


tests/chart_tests/test_resources.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class TestPodResources:
+    @pytest.mark.parametrize(
+        "pod_resources",
+        test_pod_resources_configs_data.values(),
+        ids=test_pod_resources_configs_data.keys(),
+    )
+    def test_pod_resources_configs(self, pod_resources):
+        """Labels check for definition."""
```
Shubham-astro, line 42: I think this comment would be more helpful:
`
"""Test that all pod containers have a 'resources' section defined."""
`

tests/chart_tests/test_resources.py
```diff
@@ -0,0 +1,44 @@
...
+    for key, val in kubernetes_objects.items():
+        pod_docs += jmespath.search(
+            f"[?kind == `{key}`].{{name: metadata.name, chart: metadata.labels.chart, kind: kind, container: {val}}}",
+            docs,
+        )
+
+    return {f'{doc["chart"]}_{doc["kind"]}_{doc["name"]}': doc["container"] for doc in pod_docs}
+
+
+test_pod_resources_configs_data = init_test_pod_resources()
```
Shubham-astro, line 32: The function init_test_pod_resources  is run during import time since it's outside any function. This may unnecessarily slow down test initialization.

My suggestion would be to  move test_pod_resources_configs_data = init_test_pod_resources() inside the test class TestPodResources as part of a setup function.

danielhoherd: We can't move it inside the class because the result of this function is used to parametrize the class.

danielhoherd (approved):
Looks good to me. I really dislike that we use jmespath in python because it's so easy to make syntax errors with that combination, but we can optimize that later.

tests/chart_tests/test_resources.py
```diff
@@ -0,0 +1,44 @@
+import jmespath
```
danielhoherd, line 1: Can we rename this file to `test_container_resources.py` since the word "resources" is used in other contexts even within kubernetes manifests, for instance within roles. https://github.com/astronomer/astronomer/blob/59ad90435a/charts/astronomer/templates/commander/commander-role.yaml#L41

Shubham-astro (approved):
Looks good

PR #2363, opened 2024-10-17
Added rbac condition for porometheus-postgres

Shubham-astro:
## Description

This PR intends to add BYO SA condition to prometheus-postgres-exporter.

## Related Issues

Related astronomer/issues#6704

## Testing

Upon setting global. rbacEnabled to False., kubectl get as should show only default SA created by k8s namespace. 

## Merging

Needs to be cherry picked to 0.36.0

### DIFF
--- a/charts/prometheus-postgres-exporter/templates/serviceaccount.yaml
+++ b/charts/prometheus-postgres-exporter/templates/serviceaccount.yaml
@@ -1,3 +1,4 @@
+{{ if .Values.global.rbacEnabled }}
 {{- if .Values.serviceAccount.create -}}
 apiVersion: v1
 kind: ServiceAccount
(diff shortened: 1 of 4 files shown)


charts/prometheus-postgres-exporter/templates/serviceaccount.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.serviceAccount.create -}}
+{{- if and .Values.serviceAccount.create .Values.global.rbacEnabled -}}
```
pgvishnuram, line 1: we are handling create inside helper templates if .Values.serviceAccount.create is false then default will be mapped
{{- if and .Values.serviceAccount.create .Values.global.rbacEnabled -}} -> {{- if .Values.global.rbacEnabled -}}

should be fine

Shubham-astro: done

PR #2369, opened 2024-10-22
adding commander ns conditions

Shubham-astro:
## Description

This PR intends to put condition around commander's namespace permissions. 

## Related Issues

## Testing

Commander should not be able to do any namespace related operations when .Values.global.features.namespacePools.createRbac is set to false 

## Merging

cherry pick to 0.35, 0.36


danielhoherd (changes requested):
Seems fine. Needs tests.

PR #2393, opened 2024-11-05
add extra namespace values support

pgvishnuram:
## Description

This PR adds support for  custom namespace list in order to work with namespace pools feature

## Related Issues

- https://github.com/astronomer/issues/issues/6580

## Testing

QA should able to see the status indicator working properly for non namespace pools based deployments

## Merging

cherry-pick to release-0.36


rishkarajgi:
please add tests

charts/kube-state/values.yaml
```diff
@@ -43,6 +43,9 @@ extraArgs:
 
 priorityClassName: ~
 
+# works only with namespace pools
+extraNamespaceList: []
```
rishkarajgi, line 47: let's keep this under namespace pools feature flag itself?
https://github.com/astronomer/astronomer/blob/20b78234218db8a06c741c655d6425873bb16606/values.yaml#L42

rishkarajgi: agree but if this is contextual only to namespace pools, we might as well nest the flag for ease of discoverability

danielhoherd:
The linked issue here is labeled for 0.36.1. What's the status?

Simpcyclassy (commented):
Adds an `extraNamespaces` list to support namespace pools. The intent is clear and the linked issue gives context.

Before merge: the testing section says it verifies the status indicator works for non-namespace-pool deployments, but the actual test described (checking the status indicator) does not seem directly related to the `extraNamespaces` list feature. How was the namespace pool path specifically tested? Also, the cherry-pick target is listed as release-0.36 — does 1.x need this as well? The namespace pools feature was part of the 1.x roadmap.

Left as a comment pending clarification on test coverage and branch targets.

PR #2394, opened 2024-11-05
allow prometheus to disable rbac independently

pgvishnuram:
## Description

Allow prometheus to indepently disable rbac and service account when global.rbacEnabled is set to true 

## Related Issues

- https://github.com/astronomer/issues/issues/6537
- https://github.com/astronomer/issues/issues/6742

## Testing

QA should able to disable prometheus service account and rbac below config
```
prometheus:
  serviceAccount:
      create: false 
      name: platform-prometheus
```
## Merging

NA


danielhoherd (approved):
This looks good.

PR #2396, opened 2024-11-06
adding prometheusblackbox enabled condition

Shubham-astro:
## Description

This PR intends to add flag for controlling prometheus blackbox exporter. 

## Related Issues

Related astronomer/issues#6447

## Testing

with global.prometheusBlackboxExporterEnabled set to false prometheus blackbox exporter should not be present by default. 

## Merging

0.36

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -353,6 +353,7 @@ workflows:
                 - quay.io/astronomer/ap-auth-sidecar:1.27.1
                 - quay.io/astronomer/ap-awsesproxy:1.5.0-11
                 - quay.io/astronomer/ap-base:3.18.9
+                - quay.io/astronomer/ap-blackbox-exporter:0.25.0-3
                 - quay.io/astronomer/ap-cli-install:0.26.26
                 - quay.io/astronomer/ap-commander:0.36.7
                 - quay.io/astronomer/ap-configmap-reloader:0.13.1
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -353,7 +353,6 @@ workflows:
                 - quay.io/astronomer/ap-auth-sidecar:1.27.1
                 - quay.io/astronomer/ap-awsesproxy:1.5.0-11
                 - quay.io/astronomer/ap-base:3.18.9
-                - quay.io/astronomer/ap-blackbox-exporter:0.25.0-3
```
rishkarajgi, line 356: why did we remove this?
Double entry?

Shubham-astro: @rishkarajgi it looks like this was removed due to pre-commit run, the pre-commit uses bin/show-docker-images.py script which is currently unable to list quay.io/astronomer/ap-blackbox-exporter image due to which this has been removed, I will try to check this and update.

pgvishnuram: global.blackboxExporterEnabled we already have this flag is that not working

values.yaml
```diff
@@ -82,6 +82,7 @@ global:
   #nodeExporterEnabled: true
 
   prometheusPostgresExporterEnabled: false
+  prometheusBlackboxExporterEnabled: false
```
rishkarajgi, line 85: There is an already existing flag
https://github.com/astronomer/astronomer/blob/ea401718d2393651e7cddff3011daf1565891cca/values.yaml#L87

.circleci/config.yml
Shubham-astro: @pgvishnuram it is commented, LEt me use that only.

values.yaml
Shubham-astro: using this flag.

danielhoherd (approved):
LGTM

PR #2399, opened 2024-11-06
deploy prometheus without a cluster-role

rob-1126:
## Description

Allows running without a cluster-role when in namespace pools mode in select configurations.  When enabled creates a role/role-binding instead of a cluster-role and adds individual namespaces to the list checked by prometheus. Because namespaces are a cluster-scoped resource, customers not willing to grant cluster-roles need this enumeration because they can't grant list namespaces permission.

Descended from work of Chris Abella.

## Related Issues

https://github.com/astronomer/issues/issues/6342

## Testing

QA should see status indicator similar to defaults with cluster role disabled

## Merging

cherry-pick to release-0.37


rob-1126:
@rishkarajgi Would like to see something similar in v0.36.1. IMO not important for 0.36.0.

rob-1126:
@pgvishnuram

rob-1126:
Very minimally tested but yes, I get a green status light! This is on a byo
ingress install and installed from a persona that cannot create
clusterroles ornclusterrolebindings at all.

pgvishnuram:
@rob-1126 - i will go ahead and merge this change and later will revist on more flags to be included

pgvishnuram:
@danielhoherd @rishkarajgi seems i have missed to add test cases will fix them today since the functionality has already been implemented

charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -56,6 +57,8 @@ data:
           - targets: ["localhost:9090"]
       {{ end }}
 
+      # TODO add a proper config flag option for disabling kubernetes-apiservers scraping instead of using useClusterRoles
```
danielhoherd, line 60: When does this `TODO` need to be done? before 0.37.0 or after?

charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -68,7 +71,9 @@ data:
           - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
             action: keep
             regex: default;kubernetes;https
-
+      {{- end }}
+      # TODO add a proper config flag option for disabling core-dns scraping instead of using useClusterRoles
```
danielhoherd, line 75: When does this TODO need to be done?[ We probably need a ticket for this.](url)

charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -305,7 +310,8 @@ data:
             replacement: "$1"
             target_label: release
           {{- end}}
-
+      # TODO add a proper config flag option for disabling fluentd scraping instead of using useClusterRoles
```
danielhoherd, line 313: We probably need a ticket for this.

charts/prometheus/templates/prometheus-config-configmap.yaml
pgvishnuram: we are currently going to give one flag option to disable clusterrole altogether

charts/prometheus/templates/prometheus-config-configmap.yaml
pgvishnuram: we are currently going to give one flag option to disable clusterrole altogether

danielhoherd (approved):
Looks good

PR #2400, opened 2024-11-07
Fix 6380

Shubham-astro:
## Description

This PR intends to prioritize component.storageclass over global.storageclass

## Related Issues
https://github.com/astronomer/issues/issues/6380

## Testing

When both global.storageclass and component.storageclass are added, the priority should be component.storageclass, for example:
```
(venv) shubhamsingh@Shubhams-MacBook-Pro astro % helm template astronomer --show-only charts/elasticsearch/templates/master/es-master-statefulset.yaml . --set elasticsearch.common.persistence.storageClassName=gp2 --set global.storageClass=gp1 --debug | grep -i storageclass
install.go:222: [debug] Original chart version: ""
install.go:239: [debug] CHART PATH: /Users/shubhamsingh/Desktop/git/astro

      storageClassName: gp2                    <<<< expected value
(venv) shubhamsingh@Shubhams-MacBook-Pro astro % 

```

## Merging

0.36.1

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -376,20 +376,21 @@ Return  the proper Storage Class
 {{/*
 Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
 but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else logic.
+This version prioritizes component-specific registry.persistence.storageClassName over global.storageClass.
 */}}
-{{- if .Values.global.storageClass -}}
-    {{- if (eq "-" .Values.global.storageClass) -}}
+{{- if .Values.registry.persistence.storageClassName -}}
+    {{- if (eq "~" .Values.registry.persistence.storageClassName) -}}
         {{- printf "storageClassName: \"\"" -}}
     {{- else }}
-        {{- printf "storageClassName: %s" .Values.global.storageClass -}}
+        {{- printf "storageClassName: %s" .Values.registry.persistence.storageClassName -}}
     {{- end -}}
 {{- else -}}
-    {{- if .Values.registry.persistence.storageClassName -}}
-          {{- if (eq "~" .Values.registry.persistence.storageClassName) -}}
-              {{- printf "storageClassName: \"\"" -}}
-          {{- else }}
-              {{- printf "storageClassName: %s" .Values.registry.persistence.storageClassName -}}
-          {{- end -}}
+    {{- if .Values.global.storageClass -}}
+        {{- if (eq "-" .Values.global.storageClass) -}}
+            {{- printf "storageClassName: \"\"" -}}
+        {{- else }}
+            {{- printf "storageClassName: %s" .Values.global.storageClass -}}
+        {{- end -}}
     {{- end -}}
 {{- end -}}
 {{- end -}}
--- a/charts/elasticsearch/templates/_helpers.tpl
+++ b/charts/elasticsearch/templates/_helpers.tpl
@@ -117,20 +117,21 @@ Return  the proper Storage Class
 {{/*
 Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
 but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else logic.
+This version prioritizes component-specific common.persistence.storageClassName over global.storageClass.
 */}}
-{{- if .Values.global.storageClass -}}
-    {{- if (eq "-" .Values.global.storageClass) -}}
+{{- if .Values.common.persistence.storageClassName -}}
+    {{- if (eq "-" .Values.common.persistence.storageClassName) -}}
         {{- printf "storageClassName: \"\"" -}}
     {{- else }}
-        {{- printf "storageClassName: %s" .Values.global.storageClass -}}
+        {{- printf "storageClassName: %s" .Values.common.persistence.storageClassName -}}
     {{- end -}}
 {{- else -}}
-    {{- if .Values.common.persistence.storageClassName -}}
-          {{- if (eq "-" .Values.common.persistence.storageClassName) -}}
-              {{- printf "storageClassName: \"\"" -}}
-          {{- else }}
-              {{- printf "storageClassName: %s" .Values.common.persistence.storageClassName -}}
-          {{- end -}}
+    {{- if .Values.global.storageClass -}}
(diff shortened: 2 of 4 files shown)


charts/alertmanager/templates/_helpers.tpl
```diff
@@ -53,24 +53,26 @@ Return  the proper Storage Class
 {{/*
 Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
 but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else logic.
+This refactoring prioritizes persistence.storageClassName over global.storageClass.
```
danielhoherd, line 56: This comment will need to be removed, or updated for context outside of this PR.

charts/alertmanager/templates/_helpers.tpl
```diff
@@ -53,24 +53,26 @@ Return  the proper Storage Class
...
 */}}
-{{- if .Values.global.storageClass -}}
-    {{- if (eq "-" .Values.global.storageClass) -}}
+{{- if .Values.persistence.storageClassName -}}
+    {{- if (eq "~" .Values.persistence.storageClassName) -}}
```
danielhoherd, line 59: In YAML, `~` is `null` <https://yaml.org/spec/1.2-old/spec.html#id2805071> so I think we can get away with `if not .Values.persistence.storageClassName`

charts/astronomer/templates/_helpers.yaml
```diff
@@ -376,20 +376,21 @@ Return  the proper Storage Class
...
-              {{- printf "storageClassName: \"\"" -}}
-          {{- else }}
-              {{- printf "storageClassName: %s" .Values.registry.persistence.storageClassName -}}
-          {{- end -}}
+    {{- if .Values.global.storageClass -}}
+        {{- if (eq "-" .Values.global.storageClass) -}}
+            {{- printf "storageClassName: \"\"" -}}
+        {{- else }}
+            {{- printf "storageClassName: %s" .Values.global.storageClass -}}
+        {{- end -}}
```
danielhoherd, line 393: This logic can be drastically simplified using `or`

```
$ cat templates/serviceaccount.yaml
{{- if .Values.serviceAccount.create -}}
apiVersion: v1
kind: ServiceAccount
metadata:
  name: {{ or .Values.foo .Values.bar }}
{{- end }}
$ helm template . | grep name:
  name:
$ helm template . --set foo=first --set bar=second | grep name:
  name: first
$ helm template . --set bar=second | grep name:
  name: second
```

charts/alertmanager/templates/_helpers.tpl
Shubham-astro: done.

charts/alertmanager/templates/_helpers.tpl
Shubham-astro: done

charts/astronomer/templates/_helpers.yaml
Shubham-astro: done

charts/elasticsearch/templates/_helpers.tpl
```diff
@@ -114,25 +114,7 @@ Elasticsearch NGINX variable definitions
...
-{{- else -}}
-    {{- if .Values.common.persistence.storageClassName -}}
-          {{- if (eq "-" .Values.common.persistence.storageClassName) -}}
-              {{- printf "storageClassName: \"\"" -}}
-          {{- else }}
-              {{- printf "storageClassName: %s" .Values.common.persistence.storageClassName -}}
-          {{- end -}}
-    {{- end -}}
-{{- end -}}
+storageClassName: {{ or .Values.common.persistence.storageClassName .Values.global.storageClass | default "" }}
```
danielhoherd, line 117: Fantastic improvement!

charts/postgresql/templates/_helpers.tpl
```diff
@@ -303,31 +303,22 @@ Return  the proper Storage Class
 {{- define "postgresql.storageClass" -}}
 {{/*
 Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
-but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else logic.
+but Helm 2.9 and 2.10 do not support it, so we need to implement this if-else logic.
+This version prioritizes component-specific persistence.storageClass over global.storageClass.
```
danielhoherd, line 307: We don't support helm 2 at all. Can we remove these if-else blocks?

tests/chart_tests/test_global_storageclass.py
```diff
@@ -28,5 +28,23 @@ def test_global_storageclass(supported_types, expected_output):
 
     assert all(
         expected_output in storageClassNames
-        for storageClassNames in jmespath.search("[*].spec.volumeClaimTemplates[*].spec.storageClassName", docs)
+        for storageClassNames in jmespath.search("[].spec.volumeClaimTemplates[].spec.storageClassName", docs)
```
danielhoherd, line 31: Ugh. jmespath is so unreliable in python that any time we modify it, I can't trust that what I understand what's happening.

danielhoherd: I think we should rework this test to not use jmespath, and instead use comprehensions or some other way to extract data where we know what we're getting. We can work on this together if you want. I have a datastructure in mind that will make this a lot easier to understand and be confident with, which is like:

```
input_data = (
    ("charts/alertmanager/templates/alertmanager-statefulset.yaml", "astrosc"),
    ("charts/elasticsearch/templates/master/es-master-statefulset.yaml", "astrosc"),
    ("charts/elasticsearch/templates/data/es-data-statefulset.yaml", "astrosc"),
    ("charts/prometheus/templates/prometheus-statefulset.yaml", "astrosc"),
    ("charts/astronomer/templates/registry/registry-statefulset.yaml", None),
)
```

Where the parameters are "show_only" and "expected_sc_name".

Also that "expected_types" var was probably copied and pasted from somewhere else and needs to be changed to something that makes sense here.

charts/postgresql/templates/_helpers.tpl
pgvishnuram: we can remove this altogether

Shubham-astro: done

pgvishnuram (approved):

PR #2410, opened 2024-11-13
add conditional csp flag

pgvishnuram:
## Description

This PR allows external src to allow conditionally with astronomer platform  

## Related Issues

- https://github.com/astronomer/issues/issues/6749
- https://github.com/astronomer/issues/issues/6709

## Testing

QA should able to open the playground without any errors in webbrowser console logs by default

## Merging

cherry-pick to release-0.36

### DIFF
--- a/charts/nginx/values.yaml
+++ b/charts/nginx/values.yaml
@@ -130,4 +130,3 @@ cspPolicy:
   connectsrc: "cdn.jsdelivr.net"
   fontsrc: "fonts.gstatic.com"
   scriptsrc: "fonts.googleapis.com cdn.jsdelivr.net"
-  stylesrc: "cdn.jsdelivr.net fonts.googleapis.com"
--- a/charts/nginx/values.yaml
+++ b/charts/nginx/values.yaml
@@ -124,3 +124,9 @@ defaultBackend:
   resources: {}
   readinessProbe: {}
   livenessProbe: {}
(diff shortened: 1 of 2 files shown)


charts/nginx/values.yaml
```diff
@@ -124,3 +124,10 @@ defaultBackend:
   resources: {}
   readinessProbe: {}
   livenessProbe: {}
+
+cspPolicy:
```
rishkarajgi, line 128: cdnPolicy

rishkarajgi, line 128:
```suggestion
cdnPolicy:
```

pgvishnuram: this change comes under csp policy

rishkarajgi: then let's change `enabled` to `cdnEnabled`

rishkarajgi: `cspPolicy.cdnEnabled` looks more legible

charts/nginx/values.yaml
```diff
@@ -124,3 +124,9 @@ defaultBackend:
...
+  enabled: true
```
rishkarajgi, line 129:
```suggestion
  cdnEnabled: true
```

rishkarajgi (approved):

PR #2413, opened 2024-11-18
add pod annotation support for astronomer components

pgvishnuram:
## Description

This PR adds supports for per component pod annotation and also globally defined pod annotations to all the components

## Related Issues

- https://github.com/astronomer/issues/issues/6752

## Merging

cherry-pick to release-0.36


charts/alertmanager/templates/alertmanager-statefulset.yaml
```diff
@@ -32,6 +32,9 @@ spec:
         version: {{ .Chart.Version }}
       annotations:
         checksum/configmap: {{ include (print $.Template.BasePath "/alertmanager-configmap.yaml") . | sha256sum }}
+{{- if .Values.podAnnotations }}
+{{ toYaml .Values.podAnnotations | indent 8 }}
+{{- end }}
```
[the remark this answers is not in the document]
danielhoherd: Yeah, I think that would be the same. Write a test for this syntax, then change it and see if the test still passes.

Shubham-astro (approved):

tests/chart_tests/test_pods_annotation.py
```diff
@@ -0,0 +1,45 @@
...
+
+import tests.chart_tests as chart_tests
+from tests.chart_tests.helm_template_generator import render_chart
+
+
+def init_test_pod_annotation_configs():
+    chart_values = chart_tests.get_all_features()
+    chart_values["global"]["podAnnotations"] = {"app.cloud.io": "astronomer"}
+
+    print(chart_values)
```
rishkarajgi, line 12: please omit

rishkarajgi (commented):
TODO: per component pod override testing

charts/fluentd/values.yaml
```diff
@@ -71,3 +71,6 @@ serviceAccount:
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
   name: ~
+
```
danielhoherd, line 74:
```suggestion
```

charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -15,10 +15,15 @@ spec:
   template:
     metadata:
       labels: {{ include "prometheus-node-exporter.labels" . | indent 8 }}
-      {{- if .Values.podAnnotations }}
+      {{- if or  .Values.global.podAnnotations .Values.podAnnotations }}
       annotations:
+      {{- if .Values.global.podAnnotations }}
+      {{- toYaml .Values.global.podAnnotations | nindent 8 }}
```
danielhoherd, line 21: This syntax is inconsistent with all the other modifications above here in the PR. It indents the whole golang template block, and uses nindent and `{{-`. I prefer this syntax. In any case, we should be consistent with our syntaxes. IMHO we should modify all these changes to have the indented template  + nindent + `{{-`

tests/chart_tests/test_pods_annotation.py
danielhoherd, line 12:
```suggestion
```

tests/chart_tests/test_pods_annotation.py
```diff
@@ -0,0 +1,45 @@
...
+
+    kubernetes_objects = {
+        "StatefulSet": "spec.template.metadata.annotations",
+        "Deployment": "spec.template.metadata.annotations",
+        "CronJob": "spec.jobTemplate.spec.template.metadata.annotations",
+        "Job": "spec.template.metadata.annotations",
+        "DaemonSet": "spec.template.metadata.annotations",
+        "Pod": "metadata.annotations",
+    }
+
```
danielhoherd, line 22: Why is this here?

danielhoherd (approved):
This is all good. The only thing I'd change is consistent syntax, which I left a comment about, and removal of the `print()` statement.

PR #2422, opened 2024-11-21
default serviceAccountName tests

danielhoherd:
## Description

- Add tests for all default serviceAccountNames
- Improve serviceAccountName feature
- ~Fold `get_cronjob_containerspec_by_name` functions into `get_containers_by_name()`~ This was added to its own PR that should be merged first. https://github.com/astronomer/astronomer/pull/2425

This PR will have a follow-up that tests customization of all of these serviceAccountNames

## Related Issues

https://github.com/astronomer/issues/issues/6747

## Testing

QA can pay special attention to anything that customizes serviceAccountNames. There are extensive unit tests here though, and there will be a follow-up, so I don't think much human-effort attention is needed here.

## Merging

This is meant for release-0.37, nothing earlier.


pgvishnuram:
@danielhoherd i have fixed default service account changes for nginx default backed template and service account.

pgvishnuram (approved):

charts/nginx/templates/nginx-rolebinding.yaml
```diff
@@ -18,6 +18,6 @@ roleRef:
   name: {{ template "nginx.fullname" . }}
 subjects:
   - kind: ServiceAccount
-    name: {{ template "nginx.serviceAccount" . }}
+    name: {{ template "nginx.ServiceAccount" . }}
```
danielhoherd, line 21: This inconsistency between serviceAccount and ServiceAccount is bad. It looks like we have a lot of it in our chart already, but we should try as hard as possible to stick to convention, which is `camelCase`, not `PascalCase`.

PR #2425, opened 2024-11-22
Fold features of get_cronjob_containerspec_by_name into get_containers_by_name

danielhoherd:
## Description

Fold `get_cronjob_containerspec_by_name()` features into `get_containers_by_name()`. Having one function that does the same thing to all pod manager docs is easier than having to handle Jobs separately from other pod managers.

## Related Issues

This was created as part of https://github.com/astronomer/issues/issues/6747 but is not directly related to that work. I'm just breaking this out into its own PR so it's easy to backport.

## Testing

These are only test changes, so no human QA is needed. CI passing is enough.

## Merging

This should be merged to all branches to make our test functions consistent.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -12,24 +12,22 @@ k8s_version_too_old = f'1.{int(supported_k8s_versions[0].split(".")[1]) - 1!s}.0
 k8s_version_too_new = f'1.{int(supported_k8s_versions[-1].split(".")[1]) + 1!s}.0'
 
 
-def get_containers_by_name(doc, *, include_init_containers=False) -> dict:
+def get_containers_by_name(doc: dict, *, include_init_containers=False) -> dict:
     """Given a single doc, return all the containers by name.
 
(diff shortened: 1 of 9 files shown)


Simpcyclassy:
@danielhoherd What do you think about maintaining a modular approach but having a parent function for the two conditions? i.e. separation of concerns

danielhoherd:
@Simpcyclassy the function returns the containers for a single doc, and optionally includes initContainers. It seems to me that the separation of concerns is outside the function with the data that is given to it. For instance:

```py
pod_managers = [
    "CronJob",
    "DaemonSet",
    "Deployment",
    "Job",
    "StatefulSet",
]
docs = render_chart()
pm_docs = [doc for doc in docs if doc["kind"] in pod_managers]
containers = [get_containers_by_name(doc) for doc in pm_docs]
```

If the user does not want `Job` to be included, they can not feed Job docs to `get_containers_by_name()`

Simpcyclassy (approved):
LGTM

PR #2445, opened 2024-12-02
6747 custom serviceAccountName tests

danielhoherd:
## Description

Add tests for custom serviceAccountName.

- Lots of whitespace changes to be more consistent and remove unneeded whitespace.
- Add required labels to postgresql slave sts.
- Sort labels on postgresql sts master and slave
- Fix indentation bug in postgresql slave sts.
- Fix cache poisoning bug by removing `functools.lru_cache`
- Add `deepmerge` python package for use in tests where we want to merge a base set of values with per-template values.
- Remove indirect `exceptiongroup` package dependency now that it is not needed to satisfy the hash requirement.
- Sort `tests/enable_all_features.yaml` contents
- Add `houston.cleanupAirflowDb.enabled=True` and `postgresql.replication.enabled=True` to `tests/enable_all_features.yaml`

## Related Issues

https://github.com/astronomer/issues/issues/6747

## Testing

This change set is *mostly* tests, but there are some changes to template files. Those changes were made to fix bugs or improve code quality. I think we should not need any manual testing on this, and that normal testing should cover it.

## Merging

This is only meant for 0.37.


charts/nats/templates/jetstream-job.yaml
```diff
@@ -86,7 +86,7 @@ spec:
       {{- end }}
     spec:
       restartPolicy: Never
-      serviceAccountName: {{ .Release.Name }}-jetstream-sa
+      serviceAccountName: {{ template "nats.serviceAccountName" . }}
```
pgvishnuram, line 89: we need to test since  we have added this as a hook job it even comes before nats component is installed it might not even come up

danielhoherd: On a call just now Vishnu described this change potentially creating an order-of-operations problem if the serviceAccount is not created when this job is run. I'll manually test this because it's not something we can unit test. This would also be tested passively during any functional tests where jestream job is enabled.

danielhoherd: This did break. We need to figure out how to handle this jetstream-job serviceAccountName. Discussion here: <https://astronomer.slack.com/archives/C02J8JT58CC/p1733239688054509>

danielhoherd: @pgvishnuram this should fix it. let me know if this works or if it breaks more stuff https://github.com/astronomer/astronomer/commit/3dbfbc15c924dd8d1420681541c628d9befe898c

pgvishnuram (commented):
this also need to be changes to {{ template "nats.jetStream.serviceAccountName" . }}

charts/nats/templates/jetstream-job.yaml
```diff
@@ -46,7 +46,7 @@ metadata:
   annotations:
     "helm.sh/hook": "pre-install,pre-upgrade"
     "helm.sh/hook-weight": "3"
-  name: {{ .Release.Name }}-jetstream-sa
+  name: {{ template "nats.jetStream.serviceAccountName" . }}
```
danielhoherd, line 49: We need a test for this change before we merge this PR.

danielhoherd: I verified that this is covered by the tests that exist in this PR.

charts/nats/templates/_helpers.tpl
```diff
@@ -77,10 +76,19 @@ imagePullSecrets:
...
+
 {{ define "nats.serviceAccountName" -}}
 {{- if and .Values.nats.serviceAccount.create .Values.global.rbacEnabled -}}
 {{ default (printf "%s" (include "nats.name" . )) .Values.nats.serviceAccount.name }}
 {{- else -}}
     {{ default "default" .Values.nats.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{ define "nats.jetStream.serviceAccountName" -}}
```
pgvishnuram, line 88: is it fine to have  jetStream.serviceAccountName

pgvishnuram (approved):
except the small naming clarification rest all fine

PR #2452, opened 2024-12-17
disable default platform check job

pgvishnuram:
## Description

disable default platform check job

## Related Issues

- https://github.com/astronomer/issues/issues/6680

## Testing

QA should see update-platform-job check 

## Merging

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -249,7 +249,7 @@ houston:
   # This runs as a CronJob
   updateCheck:
     # Enable check updates CronJob
-    enabled: true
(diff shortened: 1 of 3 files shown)


charts/astronomer/values.yaml
```diff
@@ -249,7 +249,7 @@ houston:
   # This runs as a CronJob
   updateCheck:
     # Enable check updates CronJob
-    enabled: true
```
rishkarajgi, line 252: please add a test for `false` condition

Shubham-astro (approved):

PR #2453, opened 2024-12-17
Add cert manager as default when operator chart is enabled

Shubham-astro:
## Description

This PR intends to add cert manager by default when the operator is enabled. 

## Related Issues

Related astronomer/issues#6822
astronomer/issues#7050

## Testing

NA

## Merging

0.37


values.yaml
```diff
@@ -36,6 +36,11 @@ global:
   networkPolicy:
     enabled: true
 
+  #Global flag to enable Operator based deployment
+  operator:
+    enabled: false
+    cert-manager:
+      enabled: true
```
pgvishnuram, line 43: it can be seperate flag i suppose

Shubham-astro: done

values.yaml
```diff
@@ -36,6 +36,11 @@ global:
   networkPolicy:
     enabled: true
 
+  #Global flag to enable Operator based deployment
+  operator:
+    enabled: false
+  cert-manager:
+    enabled: true
```
danielhoherd, line 43: Should this be default enabled? Are we going to be using this in all installations?

Shubham-astro: done

PR #2461, opened 2025-01-10
add airflow extra ingress annotation support

pgvishnuram:
## Description

Add airflow extra ingress annotation support

## Related Issues

- https://github.com/astronomer/issues/issues/6744

## Testing
QA should able to see custom ingress annotation passed from global.extraAnnotations to airflow ingress objects as well

## Merging

cherry-pick to release-0.37

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -210,8 +210,7 @@ data:
 
       {{- if and .Values.global.extraAnnotations ( not .Values.global.authSidecar.enabled ) }}
         ingress:
-          extraIngressAnnotations:
-{{- toYaml .Values.global.extraAnnotations | nindent 12 }}
+          extraIngressAnnotations: {{- toYaml .Values.global.extraAnnotations | nindent 12 }}
       {{- end }}
 
       {{- if .Values.global.networkNSLabels }}
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -208,6 +208,12 @@ data:
         sccEnabled: true
       {{- end }}
 
+      {{- if and .Values.global.extraAnnotations ( not .Values.global.authSidecar.enabled ) }}
+        ingress:
+          extraIngressAnnotations:
+{{- toYaml .Values.global.extraAnnotations | nindent 12 }}
+      {{- end }}
+
       {{- if .Values.global.networkNSLabels }}
         networkNSLabels: true
       {{- end }}
--- a/tests/chart_tests/test_houston_configmap.py
+++ b/tests/chart_tests/test_houston_configmap.py
@@ -705,10 +705,7 @@ class TestHoustonConfigmap:
             show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
         )
 
-        # Ensure the chart was rendered
-        assert len(docs) > 0
-
-        # Parse the rendered configmap
+        assert len(docs) == 1
         doc = docs[0]
         prod_yaml = yaml.safe_load(doc["data"]["production.yaml"])
 
@@ -739,10 +736,7 @@ class TestHoustonConfigmap:
             show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
         )
 
-        # Ensure the chart was rendered
-        assert len(docs) > 0
-
-        # Parse the rendered configmap
+        assert len(docs) == 1
         doc = docs[0]
         prod_yaml = yaml.safe_load(doc["data"]["production.yaml"])
 
@@ -778,10 +772,7 @@ class TestHoustonConfigmap:
             show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
         )
 
-        # Ensure the chart was rendered
-        assert len(docs) > 0
-
-        # Parse the rendered configmap
+        assert len(docs) == 1
         doc = docs[0]
         prod_yaml = yaml.safe_load(doc["data"]["production.yaml"])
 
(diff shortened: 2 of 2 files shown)


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -208,6 +208,12 @@ data:
         sccEnabled: true
       {{- end }}
 
+      {{- if and .Values.global.extraAnnotations ( not .Values.global.authSidecar.enabled ) }}
+        ingress:
+          extraIngressAnnotations:
+{{- toYaml .Values.global.extraAnnotations | nindent 12 }}
```
danielhoherd, line 214:
```suggestion
          extraIngressAnnotations: {{- toYaml .Values.global.extraAnnotations | nindent 12 }}
```

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -208,6 +208,12 @@ data:
         sccEnabled: true
       {{- end }}
 
+      {{- if and .Values.global.extraAnnotations ( not .Values.global.authSidecar.enabled ) }}
```
danielhoherd, line 211: why does `not .Values.global.authSidecar.enabled` have to be here?

tests/chart_tests/test_houston_configmap.py
```diff
@@ -36,827 +36,846 @@ def common_test_cases(docs):
...
-    """Validate the houston configmap and its embedded data with
-    loggingSidecar."""
-    docs = render_chart(
-        values={
-            "global": {
-                "loggingSidecar": {
-                    "enabled": True,
-                    "repository": "quay.io/astronomer/ap-vector",
-                    "tag": "0.22.3",
+class TestHoustonConfigmap:
```
danielhoherd, line 39: Why do we need this class? We're not parametrizing it. Is there some other benefit to having it?

tests/chart_tests/test_houston_configmap.py
```diff
@@ -36,827 +36,846 @@ def common_test_cases(docs):
...
-        show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
-    )
+            },
+            show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
+        )
 
-    # Ensure the chart was rendered
-    assert len(docs) > 0
+        # Ensure the chart was rendered
+        assert len(docs) > 0
```
danielhoherd, line 873: Would we be ok with len(docs) > 1 ? I don't think so.

charts/astronomer/templates/houston/houston-configmap.yaml
pgvishnuram: when authsidecar is enabled passing extra ingress annotation back to houston code does get respected thats why we added this additional condition

tests/chart_tests/test_houston_configmap.py
pgvishnuram: if you are okay with individual functions i can revert them back

charts/astronomer/templates/houston/houston-configmap.yaml
danielhoherd: That seems like a bug. The effect is that if a customer configures extraAnnotations, they expect them to be applied everywhere, but they will not be.

tests/chart_tests/test_houston_configmap.py
danielhoherd: If there's a reason to use a class I'm all for it. What I observe here though is that adding the class makes the code more complicated for no benefit, and having them all as separate functions is cleaner code that is more readable.

tests/chart_tests/test_houston_configmap.py
```diff
@@ -4,859 +4,865 @@
...
+        docs = render_chart(
+            values={
+                "global": {"authSidecar": {"enabled": True}, "extraAnnotations": {"route.openshift.io/termination": "passthrough"}}
+            },
+            show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
+        )
 
-    # Ensure the chart was rendered
-    assert len(docs) > 0
+        # Ensure the chart was rendered
```
danielhoherd, line 859: We usually don't have comments like this in our tests, which seem unnecessary because what the code is doing is obvious. If we want to start adding comments that's ok, but IMHO these comments are not helpful. If others think they are helpful we can leave them. Personally I would remove them.

tests/chart_tests/test_houston_configmap.py
```diff
@@ -4,859 +4,865 @@
...
+        # Parse the rendered configmap
+        doc = docs[0]
+        prod_yaml = yaml.safe_load(doc["data"]["production.yaml"])
 
-    # Validate readinessProbe
-    assert "readinessProbe" in prod_yaml["deployments"]["loggingSidecar"]
-    assert prod_yaml["deployments"]["loggingSidecar"]["readinessProbe"] == readiness_probe
+        # Validate extra ingress annotation
+        helm = prod_yaml["deployments"]["helm"]
+        assert "ingress" not in helm
```
danielhoherd, line 868:
```suggestion
        assert not prod_yaml["deployments"]["helm"].get("ingress")
```

danielhoherd (approved):
This code seems fine. I left some optional feedback. My biggest concern here is that extraAnnotations is incompatible with authSidecar. I'll bring this up in slack though, and we can solve it in another ticket if we think it's a problem, so it shouldn't prevent merging this PR.

tests/chart_tests/test_houston_configmap.py
pgvishnuram: let me revert back those change back to previous ones

pgvishnuram:
Im going to merge this change and file a new ticket for the optimization

PR #2469, opened 2025-01-29
dag server config optimization 

pgvishnuram:
## Description

This PR allows the config to be expaned only when either of resources or liveliness or readiness probes are defined

## Related Issues

- Related https://github.com/astronomer/issues/issues/6745
- Related https://github.com/astronomer/issues/issues/6928
- Related https://github.com/astronomer/issues/issues/6932

## Testing

QA should able to create dag only deployments with below scenario
specifiy only liveliness and readiness probes and create deployment
specifiy only resources and create deployment

## Merging

cherry-pick to release-0.37


Shubham-astro (changes requested):
Changes look good, Please add unittests for this.

pgvishnuram:
I have migrated all the test to dag only deploy specific test files to assert and validate new changes

Shubham-astro (approved):
looks good

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -178,6 +178,7 @@ data:
         {{- if .Values.global.dagOnlyDeployment.securityContexts }}
         securityContexts: {{ template "dagOnlyDeployment.securityContexts" . }}
         {{- end }}
+        {{- if or .Values.global.dagOnlyDeployment.resources .Values.global.dagOnlyDeployment.server.livenessProbe .Values.global.dagOnlyDeployment.server.readinessProbe }}
```
danielhoherd, line 181: Nit: I wonder if wide helm template code like would be more readable if we we spanned it across multiple lines... 🤔

```
        {{- if or
          .Values.global.dagOnlyDeployment.resources
          .Values.global.dagOnlyDeployment.server.livenessProbe
          .Values.global.dagOnlyDeployment.server.readinessProbe
        }}
```

What do you think? I'm not sure it makes it more readable.

pgvishnuram: Anything is okay as long as formatting does put back to earlier state

tests/chart_tests/test_astronomer_dag_only_deploy.py
```diff
@@ -193,3 +208,71 @@ def test_dagonlydeploy_config_enabled_with_persistence_delete(self, kube_version
...
+            },
+            show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
+        )
+
+        assert len(docs) == 1
+        doc = docs[0]
+        prod_yaml = yaml.safe_load(doc["data"]["production.yaml"])
+
+        assert "readinessProbe" in prod_yaml["deployments"]["dagDeploy"]["server"]
+        assert "readinessProbe" in prod_yaml["deployments"]["dagDeploy"]["client"]
```
danielhoherd, line 276: If these do not have a default value then we need to a test to make sure they do not show up when no value is specified.

tests/chart_tests/test_astronomer_dag_only_deploy.py
```diff
@@ -193,3 +208,71 @@ def test_dagonlydeploy_config_enabled_with_persistence_delete(self, kube_version
         prod = yaml.safe_load(doc["data"]["production.yaml"])
         assert prod["deployments"]["dagOnlyDeployment"] is True
         assert persistenceRetain == prod["deployments"]["dagDeploy"]["persistence"]
+
+    def test_houston_configmap_with_dagonlydeployment_liveness_probe(self, kube_version):
```
danielhoherd, line 212: I think it would be find to merge these two tests into `test_houston_configmap_with_dagonlydeployment_probes` and test customizing them both in one test case.

If we're not going to do that, then we should add checks that readinessProbe does not show up when only livenessProbe is specified, and vice versa.

PR #2470, opened 2025-01-30
Fix a simple shell whitespace breaking bug

danielhoherd:
## Description

This code fixes a whitespace breaking bug by using `shlex` to properly detect and quote shell arguments that have spaces in them.

## Related Issues

- <https://github.com/astronomer/issues/issues/6834>

## Testing

Unfortunately this code is not something we can test within this repo, because it is python code that is embedded in yaml, and only gets deployed to airflow pods when sidecar logging is enabled for an airflow deployment.

QA can test this by:
1. enable sidecar logging
2. deploy dags that have a space in the name
3. Ensure that there are no `airflow command error: unrecognized arguments` entries in the logs.

## Merging

This is a bugfix and should be merge into all supported versions.


pgvishnuram (approved):
This will require a small regression testing for validation

Simpcyclassy (approved):
LGTM

danielhoherd:
Merged to master and cherry-picked to all supported branches.

PR #2474, opened 2025-02-05
ap-base 3.20.5 updates

danielhoherd:
## Description

Update all ap-base derived images to use ap-base:3.20.5

## Related Issues

https://github.com/astronomer/issues/issues/6962

## Testing

No testing was done. These tests will all be covered by normal QA tests.

## Merging

This is just for 0.37. I'm not sure that all of these images will work on 0.36 so we can hold off on that for now.


himabindu07 (approved):
LGTM

PR #2473, opened 2025-02-05
Enable Prometheus to use both service label and service annotation when Airflow operator is enabled

Shubham-astro:
## Description

This PR enabled Prometheus to use both service label and service annotation when Airflow operator is enabled

## Related Issues

Related astronomer/issues#6954

## Testing

Tested on Stage and with unittests

## Merging

0.37

PR #2475, opened 2025-02-05
Add k8s spec files used in airflow-operator testing

danielhoherd:
## Description

Just adding k8s schema files, which are used in tests. This is purely a test environment optimization.

## Related Issues

https://github.com/astronomer/issues/issues/6740

## Testing

No testing needed.

## Merging

This is only for 0.37.


danielhoherd:
@Shubham-astro maybe you should just merge this into https://github.com/astronomer/astronomer/pull/2473

Shubham-astro:
> @Shubham-astro maybe you should just merge this into #2473

Thank you, If it's okay with you I can generate these files and add in the same PR.

PR #2485, opened 2025-02-17
fix operator metrics port

pgvishnuram:
## Description

fix operator metrics port

## Related Issues

- https://github.com/astronomer/issues/issues/7011
- https://github.com/astronomer/issues/issues/7012

## Testing

QA should able to deploy depricated image version controlled by api and exclusion from af operator

## Merging

NA

### DIFF
--- a/charts/airflow-operator/templates/manager/controller-manager-metrics-service.yaml
+++ b/charts/airflow-operator/templates/manager/controller-manager-metrics-service.yaml
@@ -13,6 +13,7 @@ metadata:
   name: {{ .Release.Name }}-aocm-metrics-service
   namespace: '{{ .Release.Namespace }}'
 spec:
+  type: ClusterIP
   ports:
   - name: https
     port: {{ .Values.ports.managerContainerPort }}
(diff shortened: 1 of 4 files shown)


Shubham-astro (approved):

charts/airflow-operator/templates/manager/controller-manager-metrics-service.yaml
```diff
@@ -13,10 +13,11 @@ metadata:
   name: {{ .Release.Name }}-aocm-metrics-service
   namespace: '{{ .Release.Namespace }}'
 spec:
+  type: ClusterIP
   ports:
   - name: https
     port: {{ .Values.ports.managerContainerPort }}
-    targetPort: https
+    targetPort: {{ ternary "https" .Values.ports.managerUpstreamPort  .Values.manager.metrics.useSecuredEndpoint }}
```
rujhan-arora-astronomer, line 20: @pgvishnuram Do we have a container port named https? Can you point me to its remote URL?

pgvishnuram: when useSecuredEndpoint is enabled we expose container port name as https its in helpers.tpl under airflow-operator

PR #2490, opened 2025-02-20
Added a default images section under deployments.helm config so that those images are picked up in the operator flow

rujhan-arora-astronomer:
## Description

Added a default images section under deployments.helm config so that those images are picked up in the operator flow
Currently, other components in Airflow (redis, pgb etc) use the tags specified in the operator version and they don't use the tags specified under the airflow chart. This PR changes the behaviour by adding the default image tags under the houston values section.

## Related Issues
Related astronomer/issues#7014

## Testing

Will test on stage after rc2 is released.

## Merging

0.37.0

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -149,6 +149,25 @@ houston:
   config:
     nats:
       ackWait: 600000
+    deployments:
+      helm:
+        airflow:
+          images:
+            statsd:
+              repository: quay.io/astronomer/ap-statsd-exporter
+              tag: 0.27.2
+            redis:
+              repository: quay.io/astronomer/ap-redis
+              tag: 7.2.6
+            pgbouncer:
+              repository: quay.io/astronomer/ap-pgbouncer
+              tag: 1.23.1-1
+            pgbouncerExporter:
+              repository: quay.io/astronomer/ap-pgbouncer-exporter
+              tag: 0.18.0
+            gitSync:
+              repository: quay.io/astronomer/ap-git-sync
+              tag: 4.2.3-1
 
   # Worker to connect to NATS
   worker:
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -149,25 +149,6 @@ houston:
   config:
     nats:
       ackWait: 600000
-    deployments:
-      helm:
-        airflow:
-          images:
-            statsd:
-              repository: quay.io/astronomer/ap-statsd-exporter
-              tag: 0.27.2
-            redis:
-              repository: quay.io/astronomer/ap-redis
-              tag: 7.2.6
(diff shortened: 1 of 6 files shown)


charts/astronomer/values.yaml
```diff
@@ -149,6 +149,25 @@ houston:
...
+              tag: 7.2.6
+            pgbouncer:
+              repository: quay.io/astronomer/ap-pgbouncer
+              tag: 1.23.1-1
+            pgbouncerExporter:
+              repository: quay.io/astronomer/ap-pgbouncer-exporter
+              tag: 0.18.0
+            gitSync:
+              repository: quay.io/astronomer/ap-git-sync
+              tag: 4.2.3-1
```
pgvishnuram, line 170: can we move this to houston configmap

tests/chart_tests/test_houston_configmap.py
```diff
@@ -50,6 +50,24 @@ def test_houston_configmap():
     # Ensure elasticsearch client param is at the correct location and contains http://
     assert "node" in prod["elasticsearch"]["client"]
     assert prod["elasticsearch"]["client"]["node"].startswith("http://")
+
+    # Assert that the configMap contains required component tags
+    assert prod["deployments"]["helm"]["airflow"]["images"]["statsd"]["tag"] == "0.27.2"
+    assert prod["deployments"]["helm"]["airflow"]["images"]["redis"]["tag"] == "7.2.6"
+    assert prod["deployments"]["helm"]["airflow"]["images"]["pgbouncer"]["tag"] == "1.23.1-1"
+    assert prod["deployments"]["helm"]["airflow"]["images"]["pgbouncerExporter"]["tag"] == "0.18.0"
+    assert prod["deployments"]["helm"]["airflow"]["images"]["gitSync"]["tag"] == "4.2.3-1"
```
danielhoherd, line 59: We should not make assertions about specific tag version because we do not want to update the tests when the tag changes. The only exception is if we customized the tag to a test value, and are asserting against that customization.

Maybe we could change this to assert that the image is a valid semver, or matches `^[0-9]+\.[0-9]+\.[0-9]+\`, or is not `latest` or something like that?

danielhoherd (changes requested):
https://github.com/astronomer/astronomer/blob/0cf3c44b67/bin/show-docker-images.py#L62 needs to be updated to show these images

tests/chart_tests/test_houston_configmap.py
danielhoherd: Personally, I'd remove all of these assertions about the tag value. I don't think that it's valuable enough to assert anything here. We don't make assertions about tag values anywhere else, and I don't think it's something we should start doing.

danielhoherd (approved):
LGTM.

PR #2520, opened 2025-03-11
KubernetesPodOperator with xcom_push enabled automatically uses an Alpine image from Docker Hub

Shubham-astro:
## Description

This PR fixes a bug where KubernetesPodOperator with xcom_push enabled automatically uses an Alpine image from Docker Hub for the xcom-sidecar container. This behavior violates our promise to customers that we won't contact external sources without explicit permission.

## Related Issues

Related astronomer/issues#6947

## Technical Details
The issue occurs because Airflow's KubernetesPodOperator automatically adds a sidecar container when xcom_push is enabled. This sidecar is used to harvest xcom values, but by default it uses an Alpine image from Docker Hub.
Our solution:

1. Adds a configuration snippet to airflowLocalSettings in houston-configmap.yaml
2. Uses an existing Astronomer image that's suitable for the xcom-sidecar purpose
3. Leverages Airflow's configuration capabilities ([similar to apache/airflow#10605](https://github.com/apache/airflow/issues/10605))

## Testing

- Confirmed the change with helm template locally:
Screenshot 2025-03-11 at 7 11 11 PM

- Deployed the changes to a test environment

- Creating a DAG that uses KubernetesPodOperator with xcom_push enabled and verified  that the pod uses our approved image instead of pulling from Docker Hub.
Screenshot 2025-03-11 at 6 54 15 PM

- Need to confirm that xcom functionality continues to work as expected with our image

## Merging

0.36.1. 0.37.1


charts/astronomer/templates/_helpers.yaml
```diff
@@ -9,6 +9,14 @@
   {{ $globals | toJson | print }}
 {{- end }}
 
+{{ define "defaultXcomISidecarImage" -}}
```
danielhoherd, line 12: Shouldn't this be `defaultXcomSidecarImage` ?
```suggestion
{{ define "defaultXcomSidecarImage" -}}
```

values.yaml
```diff
@@ -316,6 +316,9 @@ global:
       gitSync:
         repository: quay.io/astronomer/ap-git-sync
         tag: 4.2.3-1
+      xcomm:
```
danielhoherd, line 319: Why is this xcomm instead of xcom? <https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/xcoms.html>

pgvishnuram (approved):

PR #2522, opened 2025-03-11
Add internal houston authorization support for authsidecar

pgvishnuram:
## Description

This PR moves all the houston authorization url to internal  endpoints to make all authsidecar requests to be connected  internally using k8s service  allowing services to be up and running  without dns reflection.

## Related Issues

- https://github.com/astronomer/issues/issues/7101

## Testing

Yet to update

## Merging

Yet to update

### DIFF
--- a/charts/grafana/templates/grafana-auth-sidecar-configmap.yaml
+++ b/charts/grafana/templates/grafana-auth-sidecar-configmap.yaml
@@ -25,8 +25,8 @@ data:
       server_tokens off;
 
       location  = /auth {
-        proxy_set_header Host houston.{{ .Values.global.baseDomain }};
-        proxy_pass  https://houston.{{ .Values.global.baseDomain }}/v1/authorization;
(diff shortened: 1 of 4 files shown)


Simpcyclassy (commented):
Routing Houston auth-sidecar requests to the internal k8s service endpoint instead of the external DNS name — a legitimate fix for environments where DNS reflection is unavailable. The approach (adding an `internalhouston` URL and pointing authsidecar to it) is the right direction.

However, the PR description says "Yet to update" for both testing and merging. Before this can merge:

Verify that the internal service endpoint format is correct for all supported deployment topologies (unified, control, data plane split). In CP/DP mode, the auth sidecar and Houston are on different planes, so the internal service URL may not be resolvable across planes.

Add chart test coverage for the authsidecar annotation/config change to confirm the internal URL is rendered correctly in each plane mode.

Fill in the merge targets. This affects 1.0 and potentially 0.37.x depending on whether CP/DP was backported.

PR #2428, opened 2024-11-25
rework serviceAccountName with standard naming conventions

pgvishnuram:
## Description

rework serviceAccountName with standard naming conventions

## Related Issues

- https://github.com/astronomer/issues/issues/6747

## Testing

NA

## Merging

NA

PR #2544, opened 2025-03-18
Remove Duplicate serviceAccountName Key in Helm Chart

Vasu-Madaan:
**Description:**
This PR fixes a linting issue caused by a duplicate serviceAccountName key in the Helm chart configuration.

**Fix applied:**
The duplicate key was located at line 95 of file "es-client-deployment.yaml" and has been removed.
Keeping only one instance ensures the chart passes the linting process and prevents potential conflicts during deployment.

Error was reproducable with following steps:

```
make lint
touch helmfile.yaml
helmfile lint \
       --environment nonprod && \
touch .build/lint.done
Adding repo astronomer https://artifactory2.westpac.co.nz/helm-astronomer-virtual
"astronomer" has been added to your repositories

in ./helmfile.yaml: [exit status 1

COMMAND:
 kustomize build /tmp/chartify3710637956/airflow-astro-nonprod/astronomer/astronomer --output /tmp/chartify3710637956/airflow-astro-nonprod/astronomer/astronomer/all.patched.yaml

OUTPUT:
 Error: map[string]interface {}(nil): yaml: unmarshal errors:
   line 50: mapping key "serviceAccountName" already defined at line 36]
make: *** [Makefile:66: .build/lint.done] Error 1

```
Error Details:

```
Error: map[string]interface {}(nil): yaml: unmarshal errors:  
   line 50: mapping key "serviceAccountName" already defined at line 36  
```

### DIFF
--- a/charts/elasticsearch/templates/client/es-client-deployment.yaml
+++ b/charts/elasticsearch/templates/client/es-client-deployment.yaml
@@ -92,7 +92,6 @@ spec:
         readinessProbe: {{ tpl (toYaml .Values.sysctlInitContainer.readinessProbe) . | nindent 12 }}
         {{- end }}
   {{- end }}
-      serviceAccountName: {{ template "elasticsearch.serviceAccountName" . }}
       containers:
       - name: es-client
         securityContext:


rob-1126:
Fixing this may break things beyond what the reasonable person should expect.  

Check the upgrade path here and test it and report back. I'm expecting to to erroneously remove both copies of the service account token on the first apply and be broken. 

Unless something had changed since we last tried this, when we remove a duplicate the kubernetes logic that applies the chart sees the key has been removed and then decides to remove it from the kubernetes resource itself but because the kubernetes resource deduplicates it at creation it removes the only remaining copy and breaks stuff.

It's possible we could test and verify that a "just let it break" path is actually not disruptive and prescribe a "apply it twice" regimine so the second apply recreates it. If we do so we should try to do ALL our duplicate references in one swoop and I suspect most of them only appear in certain configurations.

Vasu-Madaan:
Hi Rob,

I have tested this scenario locally with a test helm chart, but it didn't remove the service account from the existing deployment.

Steps taken to verify:
1. Create the Helm Chart
Create a new Helm chart named test-chart:

```
helm create test-chart
```


2. Modify the Chart to Add a Duplicate Key
Edit the templates/deployment.yaml file and add a duplicate serviceAccountName key.
```

apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ .Release.Name }}
  labels:
    app: {{ .Release.Name }}
spec:
  replicas: 1
  selector:
    matchLabels:
      app: {{ .Release.Name }}
  template:
    metadata:
      labels:
        app: {{ .Release.Name }}
    spec:
      serviceAccountName: test-sa   # First instance
      containers:
        - name: test-container
          image: nginx:latest
          ports:
            - containerPort: 80
      serviceAccountName: test-sa   # Duplicate instance

```

3. Create the ServiceAccount
In the templates folder, add a new file called serviceaccount.yaml with the following content:

```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test-sa
```


4. Deploy the Chart with the Duplicate Key
Install the chart in your cluster:

`helm install test-release test-chart --namespace test-ns --create-namespace
`

Verify the ServiceAccount is Applied Correctly:

`kubectl describe pod -n test-ns | grep -i service
`

5. Simulate the Key Removal Issue
Now, edit the templates/deployment.yaml file to remove the second occurrence of the serviceAccountName:
```

apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ .Release.Name }}
  labels:
    app: {{ .Release.Name }}
spec:
  replicas: 1
  selector:
    matchLabels:
      app: {{ .Release.Name }}
  template:
    metadata:
      labels:
        app: {{ .Release.Name }}
    spec:
      serviceAccountName: test-sa   # Only one instance now
      containers:
        - name: test-container
          image: nginx:latest
          ports:
            - containerPort: 80
```


6. Upgrade the Chart with the Duplicate Removed
Run the upgrade:

`helm upgrade test-release test-chart --namespace test-ns
`

Verified the ServiceAccount Again:
`kubectl describe pod -n test-ns | grep -i service
`


Reference screenshot:
image

rob-1126:
Excellent, thank you! Maybe since it's not a list it is not subject to the same issues we see with duplicate env vars.

Upstream at kubernetes for 7 years now.

https://github.com/kubernetes/kubernetes/issues/58477

Vasu-Madaan:
Hi Rob,

Shall we proceed with this PR? The tests show that it didn't affect the existing deployment.
Actually, we tested this in software also, and the results are positive.

Reference testing doc: https://docs.google.com/document/d/1-GLNrcvBQC5BzDEnDUWctXbIsnBG3lVuPrgReeGVnZ4/edit?tab=t.0

rob-1126:
Yes, thank you!

danielhoherd (approved):
LGTM. Both occurrences of `serviceAccountName` in this file were introduced in https://github.com/astronomer/astronomer/pull/2428/files#diff-7efc2adb130709e5d46c8d396d03530e644ad6b74499bc6dda3cf48885c5ee9f

These changes should not cause any problems like duplicate ENV vars do because they get squashed when they are applied to the k8s cluster. Good catch @Vasu-Madaan!

danielhoherd:
I've created a local branch so CI can be run against this change, and will merge that branch instead of this, since CI will not run tests against repo forks.

PR #2565, opened 2025-04-01
update 0.37.1 rc1

pgvishnuram:
## Description

update 0.37.1 rc1

## Related Issues

NA, ktlo

## Testing

NA

## Merging

NA

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: astronomer
-version: 0.37.0-rc1
-appVersion: 0.37.0-rc1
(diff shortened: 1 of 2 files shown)


rujhan-arora-astronomer (approved):
What's ktlo? it's there in the description

danielhoherd:
> What's ktlo? it's there in the description

KTLO is "keeping the lights on", which is usually ongoing maintenance that is needed to keep things from breaking. IMHO this isn't strictly KTLO work, and should be linked to an umbrella "release 0.37.1" ticket, but we do not currently do that, so for now KTLO makes sense here.

PR #2566, opened 2025-04-02
Introduces logic in NetworkPolicy specs to allow traffic from different namespaces

karankhanchandani:
## Description

This PR adds a mechanism by which we can override the current `network.openshift.io/policy-group: ingress`
- Takes namespaces as input in the `ingressAllowedNamespaces` field. All the traffic coming from these namespaces will be whitelisted by the authSidecar.

The namespaceSelectors will be **ORed**, which means, if a namespace has the above label along with existing label, its workloads can communicate with the authSidecar.

## Related Issues
https://github.com/astronomer/issues/issues/7060

## Testing

Wrote unit test cases for the charts. Tests are passing. Also did manual testing by printing out the yaml spec for `NetworkPolicy`.

Output: 
```
Results (215.52s):
      2209 passed
        58 skipped
       420 warnings
```

Did the actual testing of the PR by deploying the changes in the GKE cluster `gke_astronomer-cloud-dev-236021_us-east4-a_kktest-cluster` and upgrading the helm chart.

The changes are reflected properly in network policies and houston config map.

I created a new image based deployment with celery executor and the [Airflow UI](https://deployments.kktest.astrodev.io/native-elliptical-1526/airflow/home) and [Flower UI](https://deployments.kktest.astrodev.io/native-elliptical-1526/flower) is accessible.

Screenshot 2025-04-10 at 6 50 51 AM

## Merging

0.37.1 , 0.36.2

### DIFF
--- a/charts/alertmanager/templates/alertmanager-networkpolicy.yaml
+++ b/charts/alertmanager/templates/alertmanager-networkpolicy.yaml
@@ -31,6 +31,14 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
     {{- else }}
     - podSelector:
         matchLabels:
--- a/charts/alertmanager/templates/alertmanager-networkpolicy.yaml
+++ b/charts/alertmanager/templates/alertmanager-networkpolicy.yaml
@@ -31,9 +31,6 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
-    - namespaceSelector:
-        matchLabels:
-          app.astronomer.io/allow-authsidecar-traffic: "true"
     {{- if .Values.global.authSidecar.ingressAllowedNamespaces }}
     - namespaceSelector:
         matchExpressions:
--- a/charts/alertmanager/templates/alertmanager-networkpolicy.yaml
+++ b/charts/alertmanager/templates/alertmanager-networkpolicy.yaml
@@ -33,11 +33,13 @@ spec:
           network.openshift.io/policy-group: ingress
     - namespaceSelector:
         matchLabels:
-          config.astronomer.io/allow-authsidecar-traffic: "true"
-    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+          app.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- if .Values.global.authSidecar.ingressAllowedNamespaces }}
     - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: {{ . }}
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values: {{ .Values.global.authSidecar.ingressAllowedNamespaces | toJson }}
     {{- end }}
     {{- else }}
     - podSelector:
--- a/charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml
@@ -26,9 +26,6 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
-    - namespaceSelector:
-        matchLabels:
-          app.astronomer.io/allow-authsidecar-traffic: "true"
     {{- if .Values.global.authSidecar.ingressAllowedNamespaces }}
     - namespaceSelector:
         matchExpressions:
--- a/charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml
@@ -28,11 +28,13 @@ spec:
           network.openshift.io/policy-group: ingress
     - namespaceSelector:
         matchLabels:
-          config.astronomer.io/allow-authsidecar-traffic: "true"
-    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+          app.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- if .Values.global.authSidecar.ingressAllowedNamespaces }}
     - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: {{ . }}
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values: {{ .Values.global.authSidecar.ingressAllowedNamespaces | toJson }}
     {{- end }}
     {{- else }}
     - podSelector:
--- a/charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml
@@ -26,6 +26,14 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
     {{- else }}
     - podSelector:
         matchLabels:
--- a/charts/astronomer/templates/houston/api/houston-networkpolicy.yaml
+++ b/charts/astronomer/templates/houston/api/houston-networkpolicy.yaml
@@ -26,9 +26,6 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
-    - namespaceSelector:
-        matchLabels:
-          app.astronomer.io/allow-authsidecar-traffic: "true"
     {{- if .Values.global.authSidecar.ingressAllowedNamespaces }}
     - namespaceSelector:
         matchExpressions:
--- a/charts/astronomer/templates/houston/api/houston-networkpolicy.yaml
+++ b/charts/astronomer/templates/houston/api/houston-networkpolicy.yaml
@@ -28,11 +28,13 @@ spec:
           network.openshift.io/policy-group: ingress
     - namespaceSelector:
         matchLabels:
-          config.astronomer.io/allow-authsidecar-traffic: "true"
-    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+          app.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- if .Values.global.authSidecar.ingressAllowedNamespaces }}
     - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: {{ . }}
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values: {{ .Values.global.authSidecar.ingressAllowedNamespaces | toJson }}
     {{- end }}
     {{- else }}
     - podSelector:
--- a/charts/astronomer/templates/houston/api/houston-networkpolicy.yaml
+++ b/charts/astronomer/templates/houston/api/houston-networkpolicy.yaml
@@ -26,6 +26,14 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
     {{- else }}
     - podSelector:
         matchLabels:
--- a/charts/astronomer/templates/registry/registry-networkpolicy.yaml
+++ b/charts/astronomer/templates/registry/registry-networkpolicy.yaml
@@ -86,6 +86,14 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
     {{- else }}
     - podSelector:
         matchLabels:
--- a/charts/astronomer/templates/registry/registry-networkpolicy.yaml
+++ b/charts/astronomer/templates/registry/registry-networkpolicy.yaml
@@ -88,11 +88,13 @@ spec:
           network.openshift.io/policy-group: ingress
     - namespaceSelector:
         matchLabels:
-          config.astronomer.io/allow-authsidecar-traffic: "true"
-    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+          app.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- if .Values.global.authSidecar.ingressAllowedNamespaces }}
     - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: {{ . }}
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values: {{ .Values.global.authSidecar.ingressAllowedNamespaces | toJson }}
     {{- end }}
     {{- else }}
     - podSelector:
--- a/charts/astronomer/templates/registry/registry-networkpolicy.yaml
+++ b/charts/astronomer/templates/registry/registry-networkpolicy.yaml
@@ -86,9 +86,6 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
-    - namespaceSelector:
-        matchLabels:
-          app.astronomer.io/allow-authsidecar-traffic: "true"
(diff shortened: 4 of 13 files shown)


charts/alertmanager/templates/alertmanager-networkpolicy.yaml
```diff
@@ -31,6 +31,14 @@ spec:
...
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
```
pgvishnuram, line 41: why do we need a range function here

charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml
```diff
@@ -26,6 +26,14 @@ spec:
...
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
```
pgvishnuram, line 36: why do we need a range function here

charts/astronomer/templates/houston/api/houston-networkpolicy.yaml
```diff
@@ -26,6 +26,14 @@ spec:
...
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
```
pgvishnuram, line 36: why do we need a range function here

charts/astronomer/templates/registry/registry-networkpolicy.yaml
```diff
@@ -86,6 +86,14 @@ spec:
...
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
```
pgvishnuram, line 96: why do we need a range function here

charts/grafana/templates/grafana-networkpolicy.yaml
```diff
@@ -26,6 +26,14 @@ spec:
...
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
+    {{- range .Values.global.authSidecar.ingressAllowedNamespaces }}
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ . }}
+    {{- end }}
```
pgvishnuram, line 36: why do we need a range function here

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
karankhanchandani: I have added the `ingressAllowedNamespaces` as an array. That way it is extensible for any future usecases.

charts/astronomer/templates/registry/registry-networkpolicy.yaml
```diff
@@ -86,6 +86,14 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
```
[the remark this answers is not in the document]
karankhanchandani: It was a config hence I set it to `config.astronomer.io`. Is `app.astronomer.io` a convention? What are we following in Astro?

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
pgvishnuram: we wont be having multiple ingress installation to the same platform from different base domains

charts/astronomer/templates/registry/registry-networkpolicy.yaml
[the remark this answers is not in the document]
pgvishnuram: we already have a naming used within our app as astronomer.io/commander-sync for secret syncing

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
```diff
@@ -31,6 +31,14 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          config.astronomer.io/allow-authsidecar-traffic: "true"
```
pgvishnuram, line 36: if we are going to rely on kubernetes.io/metadata.name for the ingress source we might not need this particular segment of the code in here

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
rujhan-arora-astronomer: Yeah, I agree with @pgvishnuram . Since we have a single ingress as a source, there is no need to have a list of labels and use a range function.

rujhan-arora-astronomer:
@karankhanchandani There's an option in the PR itself, to mark it as DRAFT. No need to add it in the title :)

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
karankhanchandani: @rujhan-arora-astronomer @pgvishnuram 

Ingress installation and Ingress traffic are two different things.

Ingress installation refers to setting up the necessary components, like an Ingress controller, to manage incoming traffic into a Kubernetes cluster.

"Ingress traffic" is the actual network traffic entering the cluster that is routed to specific services.

This authSidecar may receive traffic from any of the namespaces, which is why I made it configurable. As of now it is only receiving it from Kubernetes ingress. I don't see any issues with keeping it as an array. I don't see any drawback for keeping it as an array vs keeping a single value.

If you see a drawback for this approach, please let me know.

charts/astronomer/templates/registry/registry-networkpolicy.yaml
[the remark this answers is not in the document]
karankhanchandani: ok will change the naming.

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
karankhanchandani: I kept this as fallback mechanism if customers don't specify any value in helm, they can label their namespace with this.

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
pgvishnuram: ingress controller src -> authSidecar( via vhosts)

tests/chart_tests/test_authsidecar.py
```diff
@@ -161,9 +161,93 @@ def test_authSidecar_kibana(self, kube_version):
...
+
+        assert "Service" == jmespath.search("kind", docs[2])
+        assert "release-name-kibana" == jmespath.search("metadata.name", docs[2])
+        assert "ClusterIP" == jmespath.search("spec.type", docs[2])
+        assert {
+            "name": "auth-proxy",
+            "protocol": "TCP",
+            "port": 8084,
+            "appProtocol": "tcp",
+        } in jmespath.search("spec.ports", docs[2])
```
danielhoherd, line 199: Yikes. We really need to stop using jmespath. I was going to suggest an alternate form here that does not use jmespath, but since this is the pattern we use elsewhere then I'll go ahead and approve this as is. We should probably rework our tests to not use jmespath wherever we don't get a huge benefit from it. Why? <https://danielhoherd.com/tech-notes/jmespath/#quotes-matter>

tests/chart_tests/test_authsidecar.py
```diff
@@ -161,9 +161,93 @@ def test_authSidecar_kibana(self, kube_version):
...
+        assert len(docs) == 5
+        doc = docs[0]
+        assert doc["kind"] == "Deployment"
+        assert doc["apiVersion"] == "apps/v1"
+        assert doc["metadata"]["name"] == "release-name-kibana"
+        assert "auth-proxy" == doc["spec"]["template"]["spec"]["containers"][1]["name"]
+
+        assert "Service" == jmespath.search("kind", docs[2])
+        assert "release-name-kibana" == jmespath.search("metadata.name", docs[2])
+        assert "ClusterIP" == jmespath.search("spec.type", docs[2])
```
danielhoherd, line 193: These three uses of jmespath are especially ridiculous. These could all be simple value comparisons instead of searches, like `docs[2]['spec']['type'] == "ClusterIP"`

tests/chart_tests/test_authsidecar.py
```diff
@@ -271,3 +355,62 @@ def test_authSidecar_houston_configmap_with_securityContext(self, kube_version):
...
+            kube_version=kube_version,
+            values={"global": {"authSidecar": {"enabled": True, "ingressAllowedNamespaces": ["astro", "ingress-namespace"]}}},
+            show_only=[
+                "charts/alertmanager/templates/alertmanager-networkpolicy.yaml",
+                "charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml",
+                "charts/astronomer/templates/houston/api/houston-networkpolicy.yaml",
+                "charts/astronomer/templates/registry/registry-networkpolicy.yaml",
+                "charts/grafana/templates/grafana-networkpolicy.yaml",
+                "charts/kibana/templates/kibana-networkpolicy.yaml",
+                "charts/prometheus/templates/prometheus-networkpolicy.yaml",
```
danielhoherd, line 397: Same comment as above, urging this list to be made dynamically.

tests/chart_tests/test_authsidecar.py
karankhanchandani: I copy pasted from the previously written test cases. Do you think we should open a new PR for the changes that you are suggesting or do it in the same PR? I think it (code cleanup) is beyond the scope of this PR .

tests/chart_tests/test_authsidecar.py
```diff
@@ -271,3 +355,62 @@ def test_authSidecar_houston_configmap_with_securityContext(self, kube_version):
...
+            kube_version=kube_version,
+            values={"global": {"authSidecar": {"enabled": True, "ingressAllowedNamespaces": []}}},
+            show_only=[
+                "charts/alertmanager/templates/alertmanager-networkpolicy.yaml",
+                "charts/astronomer/templates/astro-ui/astro-ui-networkpolicy.yaml",
+                "charts/astronomer/templates/houston/api/houston-networkpolicy.yaml",
+                "charts/astronomer/templates/registry/registry-networkpolicy.yaml",
+                "charts/grafana/templates/grafana-networkpolicy.yaml",
+                "charts/kibana/templates/kibana-networkpolicy.yaml",
+                "charts/prometheus/templates/prometheus-networkpolicy.yaml",
```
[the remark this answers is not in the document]
karankhanchandani: Let me check this and get back to you. I did a search of usage of old label `network.openshift.io/policy-group: ingress` and made changes where ever they were being used. I need to check if the other network policies require my change or not.

tests/chart_tests/test_authsidecar.py
karankhanchandani: Thanks @danielhoherd for pointing this out. Let's create a ticket for it?

tests/chart_tests/test_authsidecar.py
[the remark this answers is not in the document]
karankhanchandani: I don't think all of the network policy files require this change @danielhoherd

tests/chart_tests/test_authsidecar.py
danielhoherd: If you're copy/pasting, it might be a good idea to do this one test case right, merge it, then we can circle back and refactor those other test cases. I don't think we should fix the other test cases in this PR.

If you want to leave it as is then that's fine too, since it was copy/pasted. I think it's more important for us to be consistent and have the same implementation for similar test cases than have different implementations for test cases that are doing similar things. So if we're not going to take the time to circle back and fix the other test cases, it's probably best to just leave it as is.

tests/chart_tests/test_authsidecar.py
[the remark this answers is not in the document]
danielhoherd: Can we put a comment in that says what is special about these? Like "only include networkpolicies that have the network.openshift.io/policy-group: ingress label" ?

tests/chart_tests/test_authsidecar.py
danielhoherd: Good call. I thought we already had one, but we don't. I created one. https://github.com/astronomer/issues/issues/7156

charts/astronomer/templates/registry/registry-networkpolicy.yaml
karankhanchandani: Removed the range functions and used IN operator and matchExpressions in namespaceSelectors. Refer the new change.

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
```diff
@@ -31,6 +31,16 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
+    - namespaceSelector:
```
rujhan-arora-astronomer, line 34: Why do we have 2 different namespaceSelectors for this? Both app.astronomer.io/allow-authsidecar-traffic and ingressAllowedNamespaces

karankhanchandani: @rujhan-arora-astronomer 
This was already discussed and resolved here
https://github.com/astronomer/astronomer/pull/2566#discussion_r2024460873
https://github.com/astronomer/astronomer/pull/2566#discussion_r2024100764

rujhan-arora-astronomer: I still think it's un-needed and more code which we need to maintain. Sorry, I didn't pick it up in the draft PR but this is what I think.

karankhanchandani: @rujhan-arora-astronomer If we go with the single label approach `astronomer.io/allow-authsidecar-traffic` we have to **ASK** the customer to label their namespaces with this label. There might be some resistance (approvals, delays, etc) vs a CRE person (or us) updating the helm chart and adding customer's namespace in a field WE own.

I felt it would be wise to keep both approaches in urgent situations.

rishkarajgi (approved):

charts/alertmanager/templates/alertmanager-networkpolicy.yaml
rujhan-arora-astronomer: I am suggesting only to use this - ingressAllowedNamespaces

tests/chart_tests/test_authsidecar.py
```diff
@@ -271,3 +396,65 @@ def test_authSidecar_houston_configmap_with_securityContext(self, kube_version):
...
+        )
+
+        assert len(docs) == 7
+
+        for doc in docs:
+            assert "NetworkPolicy" == doc["kind"]
+            namespaceSelectors = doc["spec"]["ingress"][0]["from"]
+            assert {"namespaceSelector": {"matchLabels": {"network.openshift.io/policy-group": "ingress"}}} in namespaceSelectors
+            assert {
+                "namespaceSelector": {"matchLabels": {"app.astronomer.io/allow-authsidecar-traffic": "true"}}
```
danielhoherd, line 452: We should remove this `app.astronomer.io/allow-authsidecar-traffic` label here since we decided to go with `ingressAllowedNamespaces`. Sometimes people use tests as indicators of features that can be enabled, and it would be unfortunate if somebody saw this and went looking for this feature.

danielhoherd (changes requested):
During a meeting today we decided to remove `app.astronomer.io/allow-authsidecar-traffic` implementation and stick with only `ingressAllowedNamespaces` implementation. Those changes will need to be made before merging this branch.

pgvishnuram:
i still feel ingressAllowedNamespaces vs ingressNamespace is quite different 

```
authSidecar:
   enabled: true
   # this context is to define where is src installed ingress
   ingressAllowedNamespaces: ingressns1, ingressns1
```

```
authSidecar:
   enabled: true
   # this context is to define where is src installed ingress
   ingressNamespaces: ingressns1, ingressns1
```

tests/chart_tests/test_authsidecar.py
```diff
@@ -150,20 +151,101 @@ def test_authSidecar_kibana(self, kube_version):
...
-        } in jmespath.search("spec.ports", docs[2])
+        } in docs[2]["spec"]["ports"]
 
         assert "NetworkPolicy" == docs[3]["kind"]
-        assert [{"namespaceSelector": {"matchLabels": {"network.openshift.io/policy-group": "ingress"}}}] == jmespath.search(
-            "spec.ingress[0].from", docs[3]
+        assert [
+            {"namespaceSelector": {"matchLabels": {"network.openshift.io/policy-group": "ingress"}}},
+        ] == docs[3]["spec"]["ingress"][0]["from"]
+        assert {"port": 8084, "protocol": "TCP"} in jmespath.search("spec.ingress[*].ports[0]", docs[3])
```
danielhoherd, line 168: Let switch this type of check to
```suggestion
        assert any([x['ports'] == [{'protocol': 'TCP', 'port': 8084}] for x in docs[3]['spec']['ingress']])
```
then we can get rid of jmespath in this test.

rujhan-arora-astronomer (approved):

danielhoherd (approved):
Looks good! Great job on the tests!

PR #2567, opened 2025-04-03
enable es scc tpl to allow custom service account

pgvishnuram:
## Description

enable es scc tpl to allow custom service account

## Related Issues

- https://github.com/astronomer/issues/issues/7131

## Testing

QA should able to bring up elastic components without any custom config changes

## Merging
merge to all valid releases

### DIFF
--- a/charts/elasticsearch/templates/es-scc.yaml
+++ b/charts/elasticsearch/templates/es-scc.yaml
@@ -1,6 +1,6 @@
-#################################
-### Astronomer Registry Scc   ###
-#################################
+####################################
+### Astronomer Elasticsearch Scc ###
+####################################
 {{- if and .Values.common.serviceAccount.create .Values.common.serviceAccount.sccEnabled .Values.global.sccEnabled }}
 apiVersion: security.openshift.io/v1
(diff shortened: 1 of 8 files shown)


charts/elasticsearch/templates/es-scc.yaml
```diff
@@ -0,0 +1,42 @@
+####################################
+### Astronomer Elasticsearch Scc ###
+####################################
+{{- if and .Values.common.serviceAccount.create .Values.common.serviceAccount.sccEnabled .Values.global.sccEnabled }}
```
Shubham-astro, line 4: @pgvishnuram is there a reason we are referring sccEnabled twice.? using:
.Values.common.serviceAccount.sccEnabled and .Values.global.sccEnabled.?

pgvishnuram: just a thought lets say if globally scc is Enabled  but customer wants to disable them for specific component.

do you think `{{- if and .Values.common.serviceAccount.create .Values.common.serviceAccount.sccEnabled  }}` this is more than enough

Shubham-astro (approved):

PR #2572, opened 2025-04-07
Add Docker Image Signing With Cosign

Shubham-astro:
## Description

This PR adds image signing capabilities to our CI/CD pipeline using Cosign. It enables us to cryptographically sign our release images to provide stronger security guarantees and verify the authenticity of our published container images.

## Changes

 - Add a new Python script (bin/sign-images.py) that fetches the BOM file for a specified release and signs all Astronomer images with Cosign
 - Update the CircleCI configuration to include a sign-released-image job that runs after release
 - Configure the job to use environment variables from the software-cosign-keys context
 - The script handles base64-encoded keys from environment variables

## Related Issues

Related astronomer/issues#7082

## How It Works

- The script reads COSIGN_PRIVATE_KEY, COSIGN_PUBLIC_KEY, and COSIGN_PASSWORD from environment variables
- It fetches the BOM file for the specific version from updates.astronomer.io
- It signs each Astronomer container image listed in the BOM file using the supplied keys
- It uses temporary files to securely handle the key material
- It verifies each image hasn't already been signed before attempting to sign

## Testing

- Tested the script locally with sample BOM data
- Verified proper handling of base64-encoded keys
- Confirmed the workflow integration with CircleCI

## Merging

Can we merged everywhere.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,12 +35,8 @@ jobs:
       - setup_remote_docker:
           docker_layer_caching: true
           version: docker23
-      - run:
-          name: Install cosign
-          command: |
-            curl -LO https://github.com/sigstore/cosign/releases/download/v2.2.2/cosign-linux-amd64
-            chmod +x cosign-linux-amd64
-            sudo mv cosign-linux-amd64 /usr/local/bin/cosign
+      - cosign/install:
+          version: "2.2.2"
       - run:
           name: Login to docker registry
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,7 @@ executors:
 orbs:
   slack: circleci/slack@4.10.1
   python: circleci/python@2.1.1
+  cosign: cpanato/cosign-orb@2.0.0
 
 parameters:
   scan-docker-images:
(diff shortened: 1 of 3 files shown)


rujhan-arora-astronomer (approved):

.circleci/config.yml
```diff
@@ -27,6 +28,48 @@ parameters:
...
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+          version: docker23
+      - run:
+          name: Install cosign
+          command: |
+            curl -LO https://github.com/sigstore/cosign/releases/download/v2.2.2/cosign-linux-amd64
+            chmod +x cosign-linux-amd64
+            sudo mv cosign-linux-amd64 /usr/local/bin/cosign
```
pgvishnuram, line 43: we have a orb and also manuall cosign install ??

bin/sign-images.py
```diff
@@ -0,0 +1,175 @@
...
+        dockerhub_repo = repository.replace("quay.io/astronomer/", "astronomerinc/")
+        # Check if the image exists on Docker Hub
+        try:
+            # Use Docker Hub API to check if the image exists
+            dockerhub_api_url = f"https://hub.docker.com/v2/repositories/{dockerhub_repo}/tags/{tag}"
+            response = requests.head(dockerhub_api_url, timeout=100)
+            return response.status_code == 200
+        except requests.RequestException:
+            return False
+    return False
```
danielhoherd, line 69: Why are we only checking docker hub? We don't use docker hub as our primary image location, and AFAIK we don't know if we use it for anything. Quay is the main registry we should be checking.

FYI we do have another project that was intended to be an Astronomer "Software" Engineering utilities tool that validates that tags exist. It's not perfect, but it does have reference code for how to validate both docker hub and quay.io https://github.com/astronomer/neso/blob/main/doc/docker.md#validate-tags / https://github.com/astronomer/neso/blob/main/neso/cli/docker_operations.py#L85

bin/sign-images.py
```diff
@@ -0,0 +1,175 @@
...
+            print(f"✓ Signed {full_image}")
+        except subprocess.CalledProcessError as e:
+            print(f"Error signing {full_image}: {e}")
+            # Continue with other images instead of exiting
+            continue
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Sign container images in an Astronomer release.")
+    parser.add_argument("--version", "-v", help="Version tag to sign (e.g., 0.36.0)", required=False)
```
danielhoherd, line 113: Can we change this to `--tag`? `--version` is usually used to check the version of the CLI tool that is being run.

danielhoherd: Oh, wait, `--tag` doesn't make sense in the context of the Astronomer "Software" version. I recently implemented this in `software-vuln-exporter` as `--release`, but I honestly don't know if that is any better than `--version` https://github.com/astronomer/software-vuln-exporter/blob/main/src/software_vuln_exporter/cli.py#L20-L24

We can leave this as `--version` if we want. I don't think there's a clear better option.

danielhoherd (changes requested):
This looks great other than using docker hub as the source of truth for the images.

bin/sign-images.py
Shubham-astro: I thought our customers use it, but yeah, if they don't, I can remove this.

Shubham-astro:
> This looks great other than using docker hub as the source of truth for the images.

It is removed.

danielhoherd (approved):
LGTM

PR #2573, opened 2025-04-08
Adding CP/DP flags to features

Shubham-astro:
## Description

This PR implements the Control Plane and Data Plane feature flags, These flags allow users to selectively enable/disable specific components based on whether they belong to the control plane or data plane

## Key changes:

- Added global.controlplane.enabled and global.dataplane.enabled configuration flags in values.yaml
- Added `plane: controlplane` or `plane: dataplane` labels to relevant resources for identification.

## Related Issues

- astronomer/issues#7134
-  astronomer/issues#7135

## Testing

- Added comprehensive unit tests that verify:
  - Control Plane components render only when CP flag is enabled
  - Data Plane components render only when DP flag is enabled
  - Both component types render when both flags are enabled
  - Neither component type renders when both flags are disabled

- All pods are coming up fine:
Screenshot 2025-04-17 at 12 02 05 PM

- All CP and DP pods are coming up fine with their labels:
Screenshot 2025-04-17 at 12 56 41 PM

- Checked for other components like cronjobs, services, replicasets, statefulsets: 
Screenshot 2025-04-17 at 1 04 14 PM

## Merging

1.0


pgvishnuram:
prometheus will sit on both control plane and dataplane as well

charts/astronomer/templates/astro-ui/astro-ui-service.yaml
```diff
@@ -1,6 +1,7 @@
 ###########################
 ## Astronomer UI Service ##
 ###########################
+{{- if .Values.global.controlplane.enabled}}
```
pgvishnuram, line 4:
```suggestion
{{- if .Values.global.controlplane.enabled }}
```

tests/chart_tests/test_astronomer_cp_dp_feature.py
```diff
@@ -0,0 +1,62 @@
...
+    "kube_version",
+    supported_k8s_versions,
+)
+class TestAstronomerCpDpFeature:
+    @staticmethod
+    def filter_charts_by_component(charts, component):
+        return [chart for chart in charts if chart.get("metadata", {}).get("labels", {}).get("plane") == component]
+
+    def test_astronomer_cp_feature_flag(self, kube_version):
+        """Test that helm renders templates only for astronomer CP features."""
```
danielhoherd, line 17:
```suggestion
    def test_astronomer_cp_only(self, kube_version):
        """Test that helm renders the correct templates when only the controlplane is enabled."""
```

danielhoherd (approved):
This looks great as a first iteration.

Let's follow up with a second iteration of test tweaks that add a map of what components exist in which plane, and then verify that the templates for all components have the correct labels, and not the incorrect labels.

PR #2576, opened 2025-04-09
Add configurable client_max_body_size to NGINX Elasticsearch ConfigMap

Shubham-astro:
## Description

This pull request adds the client_max_body_size directive to the NGINX Elasticsearch ConfigMap, defaulting to 100M and making it configurable through Helm values.

## Motivation

- The client_max_body_size directive in NGINX controls the maximum allowed size of client request bodies. 
- Setting this value appropriately helps prevent issues with large requests being rejected while still providing protection against excessive resource consumption.

## Changes

- Added client_max_body_size directive to the HTTP block in the NGINX Elasticsearch ConfigMap
- Set the default value to 100M
- Made the value configurable via Helm values
- Added unit tests to verify the directive is present with both default and custom values

## Related Issues

Related astronomer/issues#7164

## Testing

- Unit tests have been added to verify that:
     - The directive is present in the default configuration
     - Custom values are properly applied when specified

## Merging

0.37.1

### DIFF
--- a/charts/elasticsearch/values.yaml
+++ b/charts/elasticsearch/values.yaml
@@ -302,8 +302,7 @@ nginx:
   ingressClass: ~
   resources: {}
   securityContext: {}
-  proxy:
-    maxBodySize: "100M"
+  maxBodySize: "100M"
 
 # sysctl init container
 # Warning: sysctl changes affect the kernel, and thus, all containers running on the same node.
--- a/charts/elasticsearch/values.yaml
+++ b/charts/elasticsearch/values.yaml
@@ -302,6 +302,8 @@ nginx:
   ingressClass: ~
   resources: {}
   securityContext: {}
+  proxy:
+    maxBodySize: "100M"
 
 # sysctl init container
 # Warning: sysctl changes affect the kernel, and thus, all containers running on the same node.
--- a/tests/chart_tests/test_elasticsearch.py
+++ b/tests/chart_tests/test_elasticsearch.py
@@ -285,7 +285,7 @@ class TestElasticSearch:
         """Test that custom max body size is properly set."""
         docs = render_chart(
             kube_version=kube_version,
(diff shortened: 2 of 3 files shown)


charts/elasticsearch/values.yaml
```diff
@@ -302,6 +302,8 @@ nginx:
   ingressClass: ~
   resources: {}
   securityContext: {}
+  proxy:
+    maxBodySize: "100M"
```
pgvishnuram, line 306: do we need proxy as sub flag can we straight away pass maxBodySize: "100M"

tests/chart_tests/test_elasticsearch.py
```diff
@@ -279,6 +279,22 @@ def test_elastic_nginx_config_pattern_defaults(self, kube_version):
...
+            values={"elasticsearch": {"nginx": {"maxBodySize": "200M"}}},
+            show_only=[
+                "charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml",
+            ],
+        )
+
+        assert len(docs) == 1
+        doc = docs[0]
+        nginx_config = " ".join(doc["data"]["nginx.conf"].split())
+        assert "client_max_body_size 200M" in nginx_config
```
danielhoherd, line 297: Let's make this something very conspicuous, something we would never actually choose, so it stands out as a test value instead of something that we might actually set it to some day.
```suggestion
            values={"elasticsearch": {"nginx": {"maxBodySize": "123456789M"}}},
            show_only=[
                "charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml",
            ],
        )

        assert len(docs) == 1
        doc = docs[0]
        nginx_config = " ".join(doc["data"]["nginx.conf"].split())
        assert "client_max_body_size 123456789M" in nginx_config
```

danielhoherd (approved):
Looks great, I love the tests.

pgvishnuram (approved):

danielhoherd (approved):

PR #2586, opened 2025-04-11
7137 remove grafana

danielhoherd:
## Description

Remove grafana.

## Related Issues

https://github.com/astronomer/issues/issues/7137

## Testing

Unit tests are passing. We should keep an eye out for grafana related errors in the astronomer logs though.

## Merging

Merge only to the 1.x branch.

**DO NOT MERGE UNTIL AFTER 0.37.1 IS RELEASED**


pgvishnuram:
removing grafana will need an astroUI and houston change since we are having a UI view to open grafana and a flag to control that and in houston api level we have domain validation as well - we also need to cover them as well

karankhanchandani:
@pgvishnuram Thanks for the comment regarding where all the changes are required. 

@danielhoherd @pgvishnuram I think it would be better if you add it in the issue https://github.com/astronomer/issues/issues/7137 so that we can track all the changes needed in a single place. Maybe create a separate issue for grafana link it to above issue, mention what changes are needed there.

pgvishnuram (approved):

PR #2604, opened 2025-04-21
Enhanced Component Plane Verification in Tests

Shubham-astro:
## Description

This PR enhances our testing to verify that all Helm chart components are correctly labeled with their appropriate plane. It introduces a component-to-plane mapping and additional test methods to validate label consistency.

## Related Issues

- Related astronomer/issues#7134
- Related astronomer/issues#7135

## Changes
- Added a centralized `COMPONENT_PLANE_MAP` dictionary that defines which components belong to which plane
- Added helper methods to extract component information from charts
- Added two new test methods:
  - `test_components_have_correct_plane_labels`: Verifies known components have correct plane labels
  - `test_no_components_with_incorrect_plane_labels`: Ensures no resources have mismatched component/plane labels

## Motivation

We need to ensure that our Helm charts consistently apply the correct plane labels to all components, Incorrect plane labels could lead to:

- Components being deployed when they shouldn't be
- Components being omitted when they should be deployed
- Issues with selective deployment of control plane vs. data plane features

## Testing
- This PR is itself a test enhancement, so it directly improves our test coverage. The tests have been run against our current Helm charts to verify correctness.
- The test suite continues to run with our existing parameterized Kubernetes version tests, ensuring compatibility across supported K8s versions.

## Merging

1.0.0

### DIFF
--- a/tests/chart_tests/test_astronomer_cp_dp_feature.py
+++ b/tests/chart_tests/test_astronomer_cp_dp_feature.py
@@ -10,16 +10,19 @@ COMPONENT_PLANE_MAP = {
     "astronomer-astro-ui": "controlplane",
     "astronomer-elasticsearch": "controlplane",
     "astronomer-commander": "dataplane",
-    "astronomer-registry": "dataplane",
+    "astronomer-registry": "controlplane",
     "astronomer-prometheus": "controlplane",
-    "astronomer-prometheus": "dataplane"
 }
+
+
 def filter_docs_by_component(docs, component):
     return [doc for doc in docs if doc.get("metadata", {}).get("labels", {}).get("plane") == component]
+
 
 def get_component_name(doc):
     """Extract component name from chart metadata."""
     return doc.get("metadata", {}).get("labels", {}).get("component")
+
 
 def create_doc_identifier(doc):
     """Create a unique identifier for a k8s doc."""
@@ -27,6 +30,7 @@ def create_doc_identifier(doc):
     name = doc.get("metadata", {}).get("name", "unnamed")
     namespace = doc.get("metadata", {}).get("namespace", "default")
     return f"{kind}/{namespace}/{name}"
+
 
 @pytest.mark.parametrize(
     "kube_version",
@@ -39,10 +43,10 @@ class TestAstronomerCpDpFeature:
             kube_version=kube_version,
             values={"global": {"controlplane": {"enabled": True}, "dataplane": {"enabled": False}}},
         )
-        cp_resources = filter_charts_by_component(docs, "controlplane")
+        cp_resources = filter_docs_by_component(docs, "controlplane")
         assert len(cp_resources) > 0
 
-        dp_resources = filter_charts_by_component(docs, "dataplane")
+        dp_resources = filter_docs_by_component(docs, "dataplane")
         assert len(dp_resources) == 0
 
     def test_astronomer_dp_only(self, kube_version):
@@ -51,10 +55,10 @@ class TestAstronomerCpDpFeature:
             kube_version=kube_version,
             values={"global": {"controlplane": {"enabled": False}, "dataplane": {"enabled": True}}},
         )
-        cp_resources = filter_charts_by_component(docs, "controlplane")
+        cp_resources = filter_docs_by_component(docs, "controlplane")
         assert len(cp_resources) == 0
 
-        dp_resources = filter_charts_by_component(docs, "dataplane")
+        dp_resources = filter_docs_by_component(docs, "dataplane")
         assert len(dp_resources) > 0
 
     def test_astronomer_both_cp_dp_enabled(self, kube_version):
@@ -63,10 +67,10 @@ class TestAstronomerCpDpFeature:
             kube_version=kube_version,
             values={"global": {"controlplane": {"enabled": True}, "dataplane": {"enabled": True}}},
         )
-        cp_resources = filter_charts_by_component(docs, "controlplane")
+        cp_resources = filter_docs_by_component(docs, "controlplane")
         assert len(cp_resources) > 0
 
-        dp_resources = filter_charts_by_component(docs, "dataplane")
+        dp_resources = filter_docs_by_component(docs, "dataplane")
         assert len(dp_resources) > 0
 
     def test_astronomer_both_cp_dp_disabled(self, kube_version):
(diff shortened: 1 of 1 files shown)


tests/chart_tests/test_astronomer_cp_dp_feature.py
```diff
@@ -3,6 +3,19 @@
 from tests import supported_k8s_versions
 from tests.chart_tests.helm_template_generator import render_chart
 
+# Some Values would be added later to these maps
+# "astronomer-jetstream": "dataplane",
+# "astronomer-prometheus": "controlplane",
+COMPONENT_PLANE_MAP = {
```
danielhoherd, line 9: This data structure doesn't appear to handle components that will exist in both planes, like prometheus.

tests/chart_tests/test_astronomer_cp_dp_feature.py
```diff
@@ -13,6 +26,19 @@ class TestAstronomerCpDpFeature:
...
+        """Extract component name from chart metadata."""
+        return chart.get("metadata", {}).get("labels", {}).get("component")
+
+    @staticmethod
+    def get_chart_identifier(chart):
+        """Get a readable identifier for a chart for error messages."""
+        kind = chart.get("kind", "Unknown")
+        name = chart.get("metadata", {}).get("name", "unnamed")
+        namespace = chart.get("metadata", {}).get("namespace", "default")
+        return f"{kind}/{namespace}/{name}"
```
danielhoherd, line 40: I like this, and I think it's generally useful enough that we should put it somewhere more widely available, probably in chart_tests/__init__.py

I dislike the name. A chart is a collection of templates, and each template can render one or more kubernetes objects. Looking at the way this is used, this is working on individual kubernetes spec documents, or as we normally refer to them in this test framework, a `doc`. as in `docs = render_chart()`

tests/chart_tests/test_astronomer_cp_dp_feature.py
```diff
@@ -60,3 +86,56 @@ def test_astronomer_both_cp_dp_disabled(self, kube_version):
 
         dp_resources = self.filter_charts_by_component(charts, "dataplane")
         assert len(dp_resources) == 0
+
+    def test_components_have_correct_plane_labels(self, kube_version):
+        """Test that all components have the correct plane labels according to the component map."""
+        charts = render_chart(
```
danielhoherd, line 92: We should not change terminology from what we use everywhere else. This should be `docs = render_chart()`

tests/chart_tests/test_astronomer_cp_dp_feature.py
```diff
@@ -60,3 +86,56 @@ def test_astronomer_both_cp_dp_disabled(self, kube_version):
...
+            actual_plane = chart.get("metadata", {}).get("labels", {}).get("plane")
+
+            assert actual_plane == expected_plane, (
+                f"Component '{component_name}' ({chart_identifier}) has incorrect plane label. "
+                f"Expected: '{expected_plane}', Actual: '{actual_plane}'"
+            )
+
+    def test_no_components_with_incorrect_plane_labels(self, kube_version):
+        """Test that there are no components with plane labels that don't match the component map."""
+        charts = render_chart(
```
danielhoherd, line 114: This should also be `docs = render_chart()`

tests/chart_tests/test_astronomer_cp_dp_feature.py
```diff
@@ -13,6 +26,19 @@ class TestAstronomerCpDpFeature:
     def filter_charts_by_component(charts, component):
         return [chart for chart in charts if chart.get("metadata", {}).get("labels", {}).get("plane") == component]
 
+    @staticmethod
```
danielhoherd, line 29: I feel like we should move all these helper functions outside of the class. What do you think @Shubham-astro? IMHO it would make the code cleaner.

danielhoherd (changes requested):
I'd like this whole test file to change from `charts = render_chart()` to `docs = render_chart()` and all of the other "charts" and "chart" terminology to be renamed to "doc", because this is consistent with what we use everywhere else in our tests. Having the consistent terminology makes this easier to reason about just due to the lower overhead from not having to juggle semantic meanings.

tests/chart_tests/test_astronomer_cp_dp_feature.py
Shubham-astro: changed chart to docs

tests/chart_tests/test_astronomer_cp_dp_feature.py
danielhoherd: what i mean is by having this be a dict, we can't put one component in two planes. EG: this is invalid:
```py
COMPONENT_PLANE_MAP = {
  "astronomer-prometheus": "controlplane",
  "astronomer-prometheus": "dataplane",
}
```

tests/chart_tests/test_astronomer_cp_dp_feature.py
```diff
@@ -4,59 +4,132 @@
...
+    "astronomer-houston": "controlplane",
+    "astronomer-nginx": "controlplane",
+    "astronomer-astro-ui": "controlplane",
+    "astronomer-elasticsearch": "controlplane",
+    "astronomer-commander": "dataplane",
+    "astronomer-registry": "dataplane",
+    "astronomer-prometheus": "controlplane",
+    "astronomer-prometheus": "dataplane"
+}
+def filter_charts_by_component(docs, component):
```
danielhoherd, line 17:
```suggestion
def filter_docs_by_component(docs, component):
```

tests/chart_tests/test_astronomer_cp_dp_feature.py
```diff
@@ -4,59 +4,132 @@
...
+    return [doc for doc in docs if doc.get("metadata", {}).get("labels", {}).get("plane") == component]
+
+def get_component_name(doc):
+    """Extract component name from chart metadata."""
+    return doc.get("metadata", {}).get("labels", {}).get("component")
+
+def get_chart_identifier(doc):
+    """Get a readable identifier for a chart for error messages."""
```
danielhoherd, line 25:
```suggestion
def create_doc_identifier(doc):
    """Create a unique identifier for a k8s doc."""
```

tests/chart_tests/test_astronomer_cp_dp_feature.py
danielhoherd: This also exposes that having the "plane" label be a single field is insufficient.

danielhoherd (changes requested):
The current code looks mostly good, but we need to deal with the fact that the component / plane relationship is many to many.

Also I left some suggestions for improving function names and docstrings.

PR #2609, opened 2025-04-23
Add ConfigMap for Commander Metadata

Shubham-astro:
## Description

This PR adds a ConfigMap for storing Commander metadata information and mounts it into the Commander pod.

## Changes
- Created a new ConfigMap template for Commander metadata
- Added helper functions for consistent volume mounts and volumes
- Added unit tests to verify the ConfigMap and volume mounts

## Related Issues

Related astronomer/issues#7218

## Testing

- Added comprehensive unit tests that verify:
  - The ConfigMap is created with the correct format and values
  - The deployment includes the proper volume mounts and volumes
  - Values from Helm are correctly templated into the ConfigMap

## Merging

1.0.0


charts/astronomer/templates/commander/commander-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    component: commander
+    tier: astronomer
+    plane: dataplane
+    release: {{ .Release.Name }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    heritage: {{ .Release.Service }}
+data:
+  metadata.yaml: |
+    dataplane_id: xxxxxxx
```
danielhoherd, line 15: Shouldn't this also be an empty string by default?

pgvishnuram: we should have a default name either dynami by randam rand from helm function

danielhoherd: > we should have a default name either dynami by randam rand from helm function

@pgvishnuram i don't understand.

Shubham-astro: @danielhoherd  @pgvishnuram  added randAlphaNum now while rendering template we get dep id something like:
 
is this good.? 

```
# Source: astronomer/charts/astronomer/templates/commander/commander-config.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: release-name-commander-config
  labels:
    component: commander
    tier: astronomer
    plane: dataplane
    release: release-name
    chart: "astronomer-0.37.1"
    heritage: Helm
data:
  metadata.yaml: |
    dataplane_id: "dwIs2NZ1"
    commander_url: ""
    kubernetes_version: ""
    dataplane_url: ""
    commander_version: ""
    helm_chart_version: ""
```

charts/astronomer/templates/commander/commander-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+  labels:
+    component: commander
+    tier: astronomer
+    plane: dataplane
+    release: {{ .Release.Name }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    heritage: {{ .Release.Service }}
+data:
+  metadata.yaml: |
+    dataplane_id: {{ randAlphaNum 8 | quote }}
```
karankhanchandani, line 15: This will be regenerated to a different value if the configmap is deleted and recreated right? Will there be any complications in our systems if this value change?

PR #2614, opened 2025-04-29
add commander metadata ingress for dataplane

pgvishnuram:
## Description

This PR adds commander ingress object to support metadata and grpc endpoint over https for cross communication

## Related Issues

## Testing

## Merging

### DIFF
--- a/charts/astronomer/templates/commander/commander-grpc.yaml
+++ b/charts/astronomer/templates/commander/commander-grpc.yaml
@@ -1,6 +1,6 @@
-#####################
-## Houston Ingress ##
-#####################
+############################
+## Commander GRPC Ingress ##
+############################
 {{- if and .Values.global.dataplane.baseDomain .Values.global.dataplane.enabled }}
 kind: Ingress
 apiVersion: {{ template "apiVersion.Ingress" . }}
--- a/charts/astronomer/templates/commander/commander-grpc.yaml
+++ b/charts/astronomer/templates/commander/commander-grpc.yaml
@@ -5,9 +5,9 @@
 kind: Ingress
 apiVersion: {{ template "apiVersion.Ingress" . }}
 metadata:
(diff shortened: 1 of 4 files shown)


charts/astronomer/templates/commander/commander-grpc.yaml
```diff
@@ -0,0 +1,68 @@
+############################
+## Commander GRPC Ingress ##
+############################
+{{- if and .Values.global.dataplane.baseDomain .Values.global.dataplane.enabled }}
+kind: Ingress
+apiVersion: {{ template "apiVersion.Ingress" . }}
+metadata:
+  name: {{ .Release.Name }}-commander-grpc-ingress
```
pgvishnuram, line 8: should we get rid of commander name just have grpc-ingress

charts/astronomer/templates/commander/commander-grpc.yaml
```diff
@@ -0,0 +1,68 @@
...
+  labels:
+    component: grpc-ingress
```
Shubham-astro, line 10: This is something we can check with the team, whether it should be commander-grpc-ingress or just grpc-ingress.?

Shubham-astro: or i think we could let this be grpc-ingress and we would have a hierarchical structure in our labeling scheme:
- tier: astronomer (high-level system)
- component: metadata-ingress (type of component)
- release/name: specific ns details

charts/astronomer/templates/commander/commander-grpc.yaml
danielhoherd: i'd say if we're going to get rid of any part of it, we should get rid of `grpc`. calling it that will be bad if we decide to use this ingress for additional protocols.

PR #2690, opened 2025-05-15
6744 annotation tests and improvements

danielhoherd:
## Description

- Add annotation tests to verify that all ingresses support extraAnnotations with per-host ingress and without
- Fix one ingress extraAnnotations
- Remove k8s 1.19 ingress block

## Related Issues

Related astronomer/issues#6744

## Testing

Unit tests are covering all cases here, so no extra testing is needed.

## Merging

This is intended for 0.37, but should work with other release branches.

### DIFF
--- a/charts/astronomer/templates/astro-ui/astro-ui-ingress.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-ingress.yaml
@@ -80,7 +80,6 @@ metadata:
       if ($host = '{{ .Values.global.baseDomain }}' ) {
         rewrite ^ https://app.{{ .Values.global.baseDomain }}$request_uri permanent;
       }
-    {{- toYaml .Values.global.extraAnnotations | nindent 4 }}
     {{- end }}
 spec:
   {{- if or .Values.global.tlsSecret .Values.global.acme }}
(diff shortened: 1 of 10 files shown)


charts/astronomer/templates/astro-ui/astro-ui-ingress.yaml
```diff
@@ -86,8 +68,11 @@ metadata:
     heritage: {{ .Release.Service }}
   annotations:
     {{- if .Values.global.authSidecar.enabled  }}
-{{- toYaml .Values.global.extraAnnotations | nindent 4 }}
+      {{- toYaml .Values.global.extraAnnotations | nindent 4 }}
     {{- else }}
+    {{- if .Values.global.extraAnnotations }}
+{{- toYaml .Values.global.extraAnnotations | nindent 4 }}
+    {{- end }}
```
danielhoherd, line 75: @pgvishnuram was this omitted deliberately? This was the only ingress block that did not contain this clause.

pgvishnuram: I think we forgot looking for perhost ingress logic and totally missed to include there

pgvishnuram (approved):

PR #2691, opened 2025-05-16
add probes support for cronjob and jobs

pgvishnuram:
## Description

add probes support for cronjob and jobs

## Related Issues

- https://github.com/astronomer/issues/issues/7289

## Testing

QA should pass liveliness and readiness config for listed services to get covered on the testing

## Merging

merge to master and cherry-pick to release-0.37

PR #2696, opened 2025-05-21
Add podSecurityContext support across all Astronomer components

Shubham-astro:
## Description

This PR adds podSecurityContext support for all components in the Astronomer platform.

## Changes

- Added podSecurityContext support to all pod-based resources
- Added comprehensive unit tests to verify podSecurityContext implementation across all components

## Related Issues

Related astronomer/issues#7066

## Testing

- Covered with unittests
Screenshot 2025-05-21 at 5 39 20 PM

## Merging

0.37

### DIFF
--- a/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
@@ -35,6 +35,7 @@ spec:
 {{- end }}
     spec:
       serviceAccountName: {{ template "astroUI.serviceAccountName" . }}
+      securityContext: {{ toYaml (.Values.astroUI.podSecurityContext)| nindent 8 }}
       nodeSelector:
 {{ toYaml (default .Values.global.platformNodePool.nodeSelector .Values.nodeSelector) | indent 8 }}
       affinity:
@@ -47,7 +48,6 @@ spec:
         - name: astro-ui
           image: {{- include "astroUI.image" . | indent 1 }}
           imagePullPolicy: {{ .Values.images.astroUI.pullPolicy }}
-          securityContext: {{ toYaml (.Values.astroUI.podSecurityContext)| nindent 12 }}
           resources:
 {{ toYaml .Values.astroUI.resources | indent 12 }}
           {{- if .Values.astroUI.command }}
--- a/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
@@ -48,6 +48,7 @@ spec:
         - name: astro-ui
           image: {{- include "astroUI.image" . | indent 1 }}
           imagePullPolicy: {{ .Values.images.astroUI.pullPolicy }}
+          securityContext: {{ toYaml (.Values.securityContext)| nindent 12 }}
           resources:
 {{ toYaml .Values.astroUI.resources | indent 12 }}
           {{- if .Values.astroUI.command }}
--- a/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
+++ b/charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
@@ -47,7 +47,7 @@ spec:
         - name: astro-ui
           image: {{- include "astroUI.image" . | indent 1 }}
           imagePullPolicy: {{ .Values.images.astroUI.pullPolicy }}
-          securityContext: {{ toYaml (.Values.securityContext)| nindent 12 }}
+          securityContext: {{ toYaml (.Values.astroUI.podSecurityContext)| nindent 12 }}
           resources:
 {{ toYaml .Values.astroUI.resources | indent 12 }}
           {{- if .Values.astroUI.command }}
--- a/charts/astronomer/templates/commander/commander-deployment.yaml
+++ b/charts/astronomer/templates/commander/commander-deployment.yaml
@@ -53,6 +53,7 @@ spec:
         - name: commander
           image: {{ template "commander.image" . }}
           imagePullPolicy: {{ .Values.images.commander.pullPolicy }}
+          securityContext: {{ toYaml (.Values.securityContext)| nindent 12 }}
           resources:
 {{ toYaml .Values.commander.resources | indent 12 }}
           {{- if .Values.commander.command }}
--- a/charts/astronomer/templates/commander/commander-deployment.yaml
+++ b/charts/astronomer/templates/commander/commander-deployment.yaml
@@ -52,7 +52,7 @@ spec:
         - name: commander
           image: {{ template "commander.image" . }}
           imagePullPolicy: {{ .Values.images.commander.pullPolicy }}
-          securityContext: {{ toYaml .Values.securityContext| nindent 12 }}
+          securityContext: {{ toYaml .Values.commander.podSecurityContext| nindent 12 }}
           resources:
 {{ toYaml .Values.commander.resources | indent 12 }}
           {{- if .Values.commander.command }}
--- a/charts/astronomer/templates/commander/commander-deployment.yaml
+++ b/charts/astronomer/templates/commander/commander-deployment.yaml
@@ -41,6 +41,7 @@ spec:
     spec:
       nodeSelector:
 {{ toYaml (default .Values.global.platformNodePool.nodeSelector .Values.nodeSelector) | indent 8 }}
+      securityContext: {{ toYaml .Values.commander.podSecurityContext| nindent 8 }}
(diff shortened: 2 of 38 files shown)


charts/astronomer/templates/astro-ui/astro-ui-deployment.yaml
```diff
@@ -35,6 +35,7 @@ spec:
 {{- end }}
     spec:
       serviceAccountName: {{ template "astroUI.serviceAccountName" . }}
+      securityContext: {{ toYaml (.Values.astroUI.podSecurityContext)| nindent 8 }}
```
danielhoherd, line 38:
```suggestion
      securityContext: {{ toYaml (.Values.astroUI.podSecurityContext) | nindent 8 }}
```

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -41,6 +41,7 @@ spec:
     spec:
       nodeSelector:
 {{ toYaml (default .Values.global.platformNodePool.nodeSelector .Values.nodeSelector) | indent 8 }}
+      securityContext: {{ toYaml .Values.commander.podSecurityContext| nindent 8 }}
```
danielhoherd, line 44:
```suggestion
      securityContext: {{ toYaml .Values.commander.podSecurityContext | nindent 8 }}
```

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -52,7 +53,7 @@ spec:
         - name: commander
           image: {{ template "commander.image" . }}
           imagePullPolicy: {{ .Values.images.commander.pullPolicy }}
-          securityContext: {{ toYaml .Values.securityContext| nindent 12 }}
+          securityContext: {{ toYaml (.Values.securityContext)| nindent 12 }}
```
danielhoherd, line 56:
```suggestion
          securityContext: {{ toYaml (.Values.securityContext) | nindent 12 }}
```

charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -38,6 +38,7 @@ spec:
     spec:
       nodeSelector:
 {{ toYaml (default .Values.global.platformNodePool.nodeSelector .Values.nodeSelector) | indent 8 }}
+      securityContext: {{ toYaml .Values.houston.podSecurityContext| nindent 8 }}
```
danielhoherd, line 41:
```suggestion
      securityContext: {{ toYaml .Values.houston.podSecurityContext | nindent 8 }}
```

charts/astronomer/templates/houston/cronjobs/houston-cleanup-deploy-revisions-cronjob.yaml
```diff
@@ -50,6 +51,7 @@ spec:
             - name: cleanup
               image: {{ template "houston.image" . }}
               imagePullPolicy: {{ .Values.images.houston.pullPolicy }}
+              securityContext: {{ toYaml .Values.securityContext| nindent 16 }}
```
danielhoherd, line 54:
```suggestion
              securityContext: {{ toYaml .Values.securityContext | nindent 16 }}
```

charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
```diff
@@ -103,7 +104,7 @@ spec:
         - name: houston
           image: {{ template "houston.image" . }}
           imagePullPolicy: {{ .Values.images.houston.pullPolicy }}
-          securityContext: {{ toYaml (.Values.securityContext)| nindent 12 }}
+          securityContext: {{ toYaml (.Values.houston.podSecurityContext)| nindent 12 }}
```
danielhoherd, line 107:
```suggestion
          securityContext: {{ toYaml (.Values.houston.podSecurityContext) | nindent 12 }}
```

tests/chart_tests/test_astronomer_pod_security_contexts.py
```diff
@@ -0,0 +1,184 @@
...
+from tests import git_root_dir
+from tests.chart_tests.helm_template_generator import render_chart
+
+include_kind_list = ["Deployment", "DaemonSet", "StatefulSet", "ReplicaSet"]
+
+default_podsecuritycontext = {
+    "fsGroup": 1000,
+    "runAsGroup": 1000,
+    "runAsUser": 1000,
+}
```
danielhoherd, line 12: We should not use the same values as the defaults. We should use something obviously custom, like 998877.

tests/chart_tests/test_astronomer_pod_security_contexts.py
```diff
@@ -0,0 +1,184 @@
...
+
+pod_manager_data = {
```
danielhoherd, line 14: This whole block looks like it's copying the old test_probes.py setup, which was problematic. I recently merged a change that moves all of this to a yaml file which is loaded and used for values. I think we should switch this to that same pattern because it's easier to work with. It's not exactly the same test, and does not allow per-template testing, but I think overall it's easier to work with and serves as an easy-to-read reference of how to configure each component.

The new YAML test data https://github.com/astronomer/astronomer/pull/2691/commits/8b280f475f675d6f3c29be0d752bf2ce73047bdd

Commit that introduced the test changes that use the YAML test data https://github.com/astronomer/astronomer/pull/2691/commits/404910b1e6b2eda38b62474846c3b444de0cfd12

danielhoherd: I might be wrong. Maybe this was modeled after tests/chart_tests/test_byo_sa.py ? If we don't need to use an external yaml because the current implementation is better that's fine, and you can ignore this suggestion thread.

Shubham-astro: Sure @danielhoherd Let me check this new test, I did derive this fro existing test_probes file because that seemed to be the way to test this effectively.

pgvishnuram (approved):

pgvishnuram:
astro-ui is crashing in tests

charts/astronomer/values.yaml
```diff
@@ -125,6 +129,10 @@ houston:
 
   podAnnotations: {}
 
+  podSecurityContext:
+    fsGroup: 1000
+    runAsGroup: 1000
+    runAsUser: 1000
```
pgvishnuram, line 135: why are explicitly setting these values now.

charts/astronomer/values.yaml
```diff
@@ -363,6 +371,12 @@ configSyncer:
     # The name of the service account to use.
     # If not set and create is true, a name is generated using the fullname template
     name: ""
+
+  podSecurityContext:
+    fsGroup: 1000
+    runAsGroup: 1000
+    runAsUser: 1000
```
pgvishnuram, line 378: why are explicitly setting these values now.

charts/astronomer/values.yaml
```diff
@@ -395,6 +409,10 @@ commander:
     # initialDelaySeconds: 10
     # periodSeconds: 10
   podAnnotations: {}
+  podSecurityContext:
+    fsGroup: 1000
+    runAsGroup: 1000
+    runAsUser: 1000
```
pgvishnuram, line 415: why are explicitly setting these values now.

charts/astronomer/values.yaml
```diff
@@ -62,6 +62,10 @@ astroUI:
     # If not set and create is true, a name is generated using the fullname template
     name: ""
   podAnnotations: {}
+  podSecurityContext:
+    fsGroup: 1000
+    runAsGroup: 1000
+    runAsUser: 1000
```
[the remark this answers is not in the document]
pgvishnuram: I dont see astroUI has this before verified with master can we recheck pls - if you add this then we have to add custom condition to handle work with openshift

charts/astronomer/values.yaml
pgvishnuram: same for houston as well we didnt had this values before - if we add this we have to handle them for openshift as well

charts/astronomer/values.yaml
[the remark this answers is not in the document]
Shubham-astro: yeah, makes sense

pgvishnuram:
@danielhoherd i have approved this PR since my conversation are addressed and resolved. Can you recheck and approve this change.

danielhoherd (changes requested):
Great looking tests!

Need to add tests for all pods, including jobs and cronjob specs.

Need to add tests for container security contexts.

PR #2716, opened 2025-06-04
Updating Cosign Scripts.

Shubham-astro:
## Description

This PR intends to add a flag to skip artifact signature upload to Rekor and also verify without the signature present in Rekor which is used by default in cosign to store signature. 

## Changes Made
 - Added `--tlog-upload=false` flag to skip artifact signatures upload to Rekor. 
 - Added `--insecure-ignore-tlog=true` flag to verify the signature that was not uploaded to Rekor.

## Related Issues

Related astronomer/issues#7334

## Testing

- All Astronomer images for release 0.37.3 have been manually signed using these updated scripts to ensure:
  - Proper cosign functionality with the new version
  - Successful authentication with Quay.io registry
- Tested that the new flags work as expected. 
Screenshot 2025-06-09 at 3 10 23 PM
- Image mentioned in the screenshot is signed:
Screenshot 2025-06-09 at 3 10 55 PM

## Benefits of these changes:
- No third-party dependencies
- Works in air-gapped environments
- Signatures stored only in container registry
- Faster signing/verification (no network calls)

## Cosign Doc referred:
- https://blog.sigstore.dev/cosign-2-0-released/

## Merging

master and 0.37

### DIFF
--- a/bin/sign-images.py
+++ b/bin/sign-images.py
@@ -138,7 +138,9 @@ def main():
             except (FileNotFoundError, json.JSONDecodeError):
                 print(f"Error: Could not find or parse local file {json_file}")
                 sys.exit(1)
-
+        subprocess.run(
+            ["docker", "login", "quay.io", "-u", os.environ["QUAY_USERNAME"], "-p", os.environ["QUAY_PASSWORD"]], check=True
+        )
         print("Signing Astronomer images...")
         for image_data in data["astronomer"]["images"].values():
             sign_image(image_data["repository"], image_data["tag"], image_data["sha256"], private_key_path, password)
--- a/bin/sign-images.py
+++ b/bin/sign-images.py
@@ -75,7 +75,7 @@ def sign_image(repo, tag, sha, key_path, password=None):
     except subprocess.CalledProcessError:
         pass  # Image is not yet signed
 
-    sign_cmd = ["cosign", "sign", "--key", key_path, digest_reference]
+    sign_cmd = ["cosign", "sign", "--yes", "--key", key_path, digest_reference]
     try:
         subprocess.run(sign_cmd, env=env, check=True)
         print(f"✓ Signed {full_image}")
--- a/bin/sign-images.py
+++ b/bin/sign-images.py
@@ -66,7 +66,7 @@ def sign_image(repo, tag, sha, key_path, password=None):
 
     try:
         subprocess.run(
-            ["cosign", "verify", "--insecure-ignore-tlog", "--key", f"{key_path}.pub", digest_reference],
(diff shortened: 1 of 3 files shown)


bin/sign-images.py
```diff
@@ -138,7 +138,9 @@ def main():
             except (FileNotFoundError, json.JSONDecodeError):
                 print(f"Error: Could not find or parse local file {json_file}")
                 sys.exit(1)
-
+        subprocess.run(
+            ["docker", "login", "quay.io", "-u", os.environ["QUAY_USERNAME"], "-p", os.environ["QUAY_PASSWORD"]], check=True
+        )
```
danielhoherd: This should be a prerequisite step, not included in this script. This script signs the images and publishes, regardless of what registry that image is held in.

bin/sign-images.py
```diff
@@ -75,7 +75,7 @@ def sign_image(repo, tag, sha, key_path, password=None):
     except subprocess.CalledProcessError:
         pass  # Image is not yet signed
 
-    sign_cmd = ["cosign", "sign", "--key", key_path, digest_reference]
+    sign_cmd = ["cosign", "sign", "--yes", "--key", key_path, digest_reference]
```
danielhoherd: I'm not sure this is the right solution. I was under the impressions that we were using a solution that did not involve us storing keys on a third party server, which this does. (Requiring a third party server is not helpful for our airgapped customers anyway.) AFAIK there is a solution that avoids storing keys on a third party server, `--rekor-url=""`, but I'm not sure if it works or what the implication is. I think we should ask the security team for guidance here.

bin/sign-images.py
Shubham-astro: yeah, this is not something that I had to do when i was trying to sign manually, it would not be needed here.

bin/sign-images.py
Shubham-astro: agreed, i have found another way to do this for now, will update the PR.

danielhoherd (approved):
This looks good. I've manually validated all of this. The solution opts out of the transparency log run by the rekor project. <https://docs.sigstore.dev/logging/overview/>

While I do like the idea of the transparency log, the transparency log is surprisingly opaque. I do not see any easy way to view the log, or interact with it. Opting out of the log seems like the best solution for now. We can always revisit this in the future.

danielhoherd:
@Shubham-astro before merging this can you update the downloaded cosign executable version to the latest and put a link to `https://github.com/sigstore/cosign/releases` next to it in the circleci config?

https://github.com/astronomer/astronomer/blob/2bbb62616b4/.circleci/config.yml#L39

PR #2719, opened 2025-06-09
Add COMMANDER_WAIT_ENABLED Environment Variable to Houston Deployment

Shubham-astro:
## Description

This PR adds the COMMANDER_WAIT_ENABLED environment variable to both the wait-for-db init container and the main houston container in the Houston deployment, with conditional logic based on the plane mode configuration, so that houston doesn't wait for commander when `control` mode is set. 
This is related to: https://github.com/astronomer/houston-api/commit/8490d76f2cbe6e2efe02d6bc7e3d81403c65dc38

### Houston Deployment

- Implemented conditional logic based on .Values.global.plane.mode:
  - unified mode → COMMANDER_WAIT_ENABLED = "true"
  - control mode → COMMANDER_WAIT_ENABLED = "false"

This ensures that the commander wait functionality is only enabled in unified plane deployments where houston and commander are in same cluster, while remaining disabled in control plane and other deployment modes

### Commander Deployment

- Fixed COMMANDER_DATAPLANE_MODE environment variable to use new plane mode system and give out mode name as output for commander to use:
```
- name: COMMANDER_DATAPLANE_MODE
   value: {{ .Values.global.plane.mode | quote }}
```

## Related Issues

Related astronomer/issues#7136

## Testing

- Tested the manual rendering of the flag:

  - In Control mode `helm template --set global.plane.mode="control"  .  --show-only charts/astronomer/templates/houston/api/houston-deployment.yaml` :
Screenshot 2025-06-09 at 4 28 28 PM

  - In Unified mode `helm template --set global.plane.mode="unified"  .  --show-only charts/astronomer/templates/houston/api/houston-deployment.yaml`
Screenshot 2025-06-09 at 4 35 22 PM

## Merging

1.0

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -209,8 +209,6 @@ s3:
 {{- end }}
 
 {{- define "houston_environment" }}
-- name: COMMANDER_WAIT_ENABLED
-  value: {{ if eq .Values.global.plane.mode "unified" }}"true"{{ else }}"false"{{ end }}
 {{- /* Dynamically created envs */ -}}
 {{- range $i, $config := .Values.houston.env }}
 - name: {{ $config.name }}
@@ -270,6 +268,8 @@ s3:
 - name: HOUSTON_SCRIPT_UPDATE_RUNTIME_SERVICE_URL
   value: {{ .Values.houston.updateRuntimeCheck.url }}
 {{- end }}
+- name: COMMANDER_WAIT_ENABLED
+  value: "false"
 
 {{- end }}
 
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -264,13 +264,12 @@ s3:
   value: {{ template "houston.nats.servers" . }}
 - name: NATS__CLUSTER_ID
   value: {{ .Release.Name }}-stan
+- name: COMMANDER_WAIT_ENABLED
+  value: {{ if eq .Values.global.plane.mode "unified" }}"true"{{ else }}"false"{{ end }}
 {{- if .Values.houston.updateRuntimeCheck.enabled }}
 - name: HOUSTON_SCRIPT_UPDATE_RUNTIME_SERVICE_URL
   value: {{ .Values.houston.updateRuntimeCheck.url }}
 {{- end }}
-- name: COMMANDER_WAIT_ENABLED
-  value: {{ if eq .Values.global.plane.mode "unified" }}"true"{{ else }}"false"{{ end }}
-
 {{- end }}
 
 {{- define "houston_volume_mounts" }}
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -269,7 +269,7 @@ s3:
   value: {{ .Values.houston.updateRuntimeCheck.url }}
 {{- end }}
 - name: COMMANDER_WAIT_ENABLED
-  value: "false"
+  value: {{ if eq .Values.global.plane.mode "unified" }}"true"{{ else }}"false"{{ end }}
 
 {{- end }}
 
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -209,6 +209,8 @@ s3:
 {{- end }}
 
 {{- define "houston_environment" }}
+- name: COMMANDER_WAIT_ENABLED
+  value: {{ if eq .Values.global.plane.mode "unified" }}"true"{{ else }}"false"{{ end }}
 {{- /* Dynamically created envs */ -}}
 {{- range $i, $config := .Values.houston.env }}
 - name: {{ $config.name }}
--- a/charts/astronomer/templates/commander/commander-deployment.yaml
+++ b/charts/astronomer/templates/commander/commander-deployment.yaml
@@ -143,7 +143,7 @@ spec:
             - name: COMMANDER_BASE_DOMAIN
               value: {{ .Values.global.baseDomain | default "" | quote }}
             - name: COMMANDER_DATAPLANE_MODE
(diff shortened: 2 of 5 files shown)


tests/chart_tests/test_houston_api_deployment.py
```diff
@@ -50,6 +50,14 @@ def test_houston_api_deployment(self, kube_version):
...
         deployments_database_connection_env = next(x for x in houston_env if x["name"] == "DEPLOYMENTS__DATABASE__CONNECTION")
         assert deployments_database_connection_env is not None
+        commander_wait_enabled_env = next(x for x in houston_env if x["name"] == "COMMANDER_WAIT_ENABLED")
+        assert commander_wait_enabled_env is not None
+        assert commander_wait_enabled_env["value"] == "true"
+
+        wait_for_db_env = c_by_name["wait-for-db"]["env"]
+        commander_wait_enabled_init_env = next(x for x in wait_for_db_env if x["name"] == "COMMANDER_WAIT_ENABLED")
+        assert commander_wait_enabled_init_env is not None
+        assert commander_wait_enabled_init_env["value"] == "true"
```
danielhoherd: We have a better pattern for checking env vars, which is to create a dict just like you'd expect an env var to be in a shell. the only difference is that the `valueFrom` is a dict instead of a string, but it still functions well for tests:

```pycon
>>> env_vars = {x["name"]: x["value"] if x.get("value") else x["valueFrom"] for x in c_by_name["wait-for-db"]["env"]}
>>> pp(env_vars)
{'COMMANDER_WAIT_ENABLED': 'true',
 'COMMANDER__HOST': 'release-name-commander',
 'COMMANDER__PORT': '50051',
 'DATABASE_URL': {'secretKeyRef': {'key': 'connection',
                                   'name': 'release-name-houston-backend'}},
 'DATABASE__CONNECTION': {'secretKeyRef': {'key': 'connection',
                                           'name': 'release-name-houston-backend'}},
 'DEPLOYMENTS__DATABASE__CONNECTION': {'secretKeyRef': {'key': 'connection',
                                                        'name': 'release-name-houston-backend'}},
 'HOUSTON_SCRIPT_UPDATE_RUNTIME_SERVICE_URL': 'https://updates.astronomer.io/astronomer-runtime',
 'HOUSTON__HOST': 'release-name-houston',
 'JWT__CERT_PATH': '/etc/houston/tls/self',
 'JWT__REGISTRY__CERT_PATH': '/etc/houston/tls/registry',
 'JWT__REGISTRY__ISSUER': 'houston',
 'JWT__REGISTRY__SERVICE': 'docker-registry',
 'NATS__CLUSTER_ID': 'release-name-stan',
 'NATS__SERVERS': '["release-name-nats-0.release-name-nats.default.svc:4222","release-name-nats-1.release-name-nats.default.svc:4222","release-name-nats-2.release-name-nats.default.svc:4222"]',
 'NODE_ENV': 'production',
 'REGISTRY__HOST': 'release-name-registry',
 'REGISTRY__PORT': '5000'}
```

```suggestion
        env_vars = {x["name"]: x["value"] if x.get("value") else x["valueFrom"] for x in c_by_name["houston"]["env"]}
        assert env_vars["DEPLOYMENTS__DATABASE__CONNECTION"]["secretKeyRef"]
        assert env_vars["COMMANDER_WAIT_ENABLED"] == "true"

        env_vars = {x["name"]: x["value"] if x.get("value") else x["valueFrom"] for x in c_by_name["wait-for-db"]["env"]}
        assert env_vars["COMMANDER_WAIT_ENABLED"] == "true"
```

Shubham-astro:
> Everything here looks good. I added one suggestion for cleaner tests.
> 
> I'd like to see a test that sets `COMMANDER_WAIT_ENABLED` to false and ensures that the expected output is expected. This could be included in an existing test case if we want to do it that way.

Added a new test: `test_houston_api_deployment_control_mode`

charts/astronomer/templates/_helpers.yaml
```diff
@@ -268,6 +268,8 @@ s3:
 - name: HOUSTON_SCRIPT_UPDATE_RUNTIME_SERVICE_URL
   value: {{ .Values.houston.updateRuntimeCheck.url }}
 {{- end }}
+- name: COMMANDER_WAIT_ENABLED
+  value: "false"
```
pgvishnuram: for dataplane mode and control plane mode it can be false
for unified mode this check is needed since commander will be running on control plane that wait for host check will ensure commander becomes active

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -143,7 +143,7 @@ spec:
             - name: COMMANDER_BASE_DOMAIN
               value: {{ .Values.global.baseDomain | default "" | quote }}
             - name: COMMANDER_DATAPLANE_MODE
-              value: {{ .Values.global.dataplane.enabled | default false | quote }}
+              value: {{ if eq .Values.global.plane.mode "data" }}"true"{{ else }}"false"{{ end }}
```
pgvishnuram: we no longer support boolean - we can directly send the dataplane mode types like control, data, or unified

pgvishnuram (approved):

tests/functional_tests/test_container_no_root.py
```diff
@@ -3,11 +3,7 @@
 
 from tests.functional_tests.conftest import get_pod_running_containers
 
-container_ignore_list = [
-    "kube-state",
-    "houston",
-    "fluentd",
-]
+container_ignore_list = ["kube-state", "houston", "fluentd", "commander"]
```
danielhoherd: nit: If you have a trailing comma on the last list item, ruff will format the list as one item per line, otherwise it tries to put it all on one line. Let's leave it as it was and just add commander here so we have a cleaner `git blame`, and because it's easier to read.
```suggestion
container_ignore_list = [
    "commander",
    "kube-state",
    "houston",
    "fluentd",
 ]
```

PR #2726, opened 2025-06-12
7186 refactor functional tests part 2

danielhoherd:
## Description

- Delete old functional test suite
- Migrate all old tests to new functional test suite
- Make the unified scenario tests run in CI instead of the old functional tests
- Remove kubent
- Refactor fixtures to work with new test code
- Refactor the network security tests to be a bit less complicated

## Related Issues

- https://github.com/astronomer/issues/issues/7186
- https://github.com/astronomer/issues/issues/7343

## Testing

These are tests, so there's no extra testing that needs to be done.

## Merging

1.0


tests/data_files/scenario-dataplane.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+tags:
+  logging: true
+  monitoring: true
+  platform: true
+
+astronomer:
+  houston:
+    upgradeDeployments:
+      enabled: true
```
pgvishnuram: this config is wrong for dataplane

tests/data_files/scenario-dataplane.yaml
```diff
@@ -0,0 +1,21 @@
+# Helm values for data scenario tests.
+
+global:
+  plane:
+    mode: data
+
+  # This usage is deprecated, use `global.plane.mode` instead.
+  dataplane:
```
Shubham-astro: This flag is no longer present now.

tests/multi_cluster/data/test_container_no_root.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+import testinfra
+
+from tests.utils.k8s import get_pod_running_containers
+from tests.utils.k8s import kubeconfig_data as kubeconfig
+
+container_ignore_list = [
+    "kube-state",
+    "houston",
```
Shubham-astro, line 8: "houston" is not needed do data mode

tests/data_files/scenario-dataplane.yaml
```diff
@@ -0,0 +1,21 @@
+# Helm values for data scenario tests.
+
+global:
+  plane:
+    mode: data
+
```
Shubham-astro: Can we add domainSuffix as well for dataplane mode.?
This is from: https://github.com/astronomer/astronomer/blob/master/values.yaml#L15
this will be particularly useful in testing commander http and grpc ingresses.

danielhoherd: This will take some consideration. Let's talk about this at our next meeting.

tests/multi_cluster/data/test_container_no_root.py
danielhoherd: Eh, I'm going to leave it in order to avoid variety between the different test configs. Having it here does not hurt anything. It's true, houston in the dataplane can be ignored.

bin/install-ci-tools.py
```diff
@@ -1,36 +1,58 @@
...
 BIN_DIR.mkdir(parents=True, exist_ok=True)
 os.environ["PATH"] = f"{BIN_DIR}:{os.environ['PATH']}"
 
 
+def detect_os_arch():
+    # Normalize OS name
+    os_map = {
+        "linux": "linux",
+        "darwin": "darwin",
+        "windows": "windows",  # Add if you want to support Windows
```
Shubham-astro, line 36: Just a question: are there any current tests where we need windows support ?

Shubham-astro (approved):
the current changes Looks good

PR #2729, opened 2025-06-17
Fix platform deployment issues. 

Shubham-astro:
## Description

This PR addresses several deployment issues discovered during testing of the Astronomer platform with the latest plane modes.

## Changes Made

1. Bootstrap Secret Reference Fix:  Ensures consistent naming convention and resolves secret lookup failures during deployment.

2. Domain Suffix Configuration: Provides flexibility for different deployment environments and proper domain resolution.

3. Registry Wait Enabled Support: Provides consistent wait behavior for houston service so that during `control` mode houston doesn't wait for registry.

4. Registry StatefulSet Volume Fix: Resolves "volume not found" errors that prevented registry pods from starting.

## Related Issues

- astronomer/issues#7340
- astronomer/issues#7160
- astronomer/issues#7161

## Testing

- All existing tests pass
- New test cases added for registry wait functionality
- Tested the deployment creation and Airflow UI. 

## Merging

1.0

### DIFF
--- a/tests/utils/__init__.py
+++ b/tests/utils/__init__.py
@@ -4,6 +4,7 @@ import yaml
 
 from tests.utils.chart import render_chart
 
+
 def get_env_vars_dict(container_env):
     """
     Convert container environment variables list to a dictionary.
@@ -13,6 +14,7 @@ def get_env_vars_dict(container_env):
         Dictionary mapping env var names to their values or valueFrom references
     """
     return {x["name"]: x["value"] if x.get("value") else x["valueFrom"] for x in container_env}
+
 
 def get_service_ports_by_name(doc):
     """Given a single service doc, return all the ports by name."""


tests/chart_tests/test_houston_api_deployment.py
```diff
@@ -75,11 +77,17 @@ def test_houston_api_deployment_control_mode(self, kube_version):
         assert commander_wait_enabled_env is not None
         assert commander_wait_enabled_env["value"] == "false"
 
+        registry_wait_enabled_env = next(x for x in houston_env if x["name"] == "REGISTRY_WAIT_ENABLED")
```
danielhoherd: While we're working on this test case, can we switch all these assertions to use the `env_vars = {x["name"]: x["value"] if x.get("value") else x["valueFrom"] for x in c_by_name["houston"]["env"]}` form like the prior test case?

This:
```
env_vars = {x["name"]: x["value"] if x.get("value") else x["valueFrom"] for x in c_by_name["houston"]["env"]}
assert env_vars["COMMANDER_WAIT_ENABLED"] == "true"
assert env_vars["REGISTRY_WAIT_ENABLED"] == "false"
```

is a lot cleaner than

```
commander_wait_enabled_env = next(x for x in houston_env if x["name"] == "COMMANDER_WAIT_ENABLED")
assert commander_wait_enabled_env["value"] == "false"
registry_wait_enabled_init_env = next(x for x in wait_for_db_env if x["name"] == "REGISTRY_WAIT_ENABLED")
assert registry_wait_enabled_init_env["value"] == "false"
```

Also in another test case you merged recently @Shubham-astro you had put `env_vars = {x["name"]: x["value"] if x.get("value") else x["valueFrom"] for x in c_by_name["houston"]["env"]}` into a function. maybe we should store that function in the utils.py and import wherever we use it.

Shubham-astro: Sure, added `get_env_vars_dict` for the same

PR #2731, opened 2025-06-20
Remove Probes from Init Containers

Shubham-astro:
## Description

This PR removes health check probes (liveness, readiness, and startup probes) from init containers and adds comprehensive tests to ensure they remain absent from future templates.

## Related Issues

Related astronomer/issues#7375

## Motivation

Init containers are designed to run to completion before the main application containers start. Unlike regular containers, init containers don't need health check probes because:

- They have a finite lifecycle and exit when their task is complete
- Kubernetes automatically handles init container completion status
- Probes are unnecessary overhead for containers that aren't meant to run continuously

### Configuration Changes

- Removed `livenessProbe`, `readinessProbe` and `startupProbe` configurations from all init containers

###  Test Coverage

- Added new test class TestCustomProbes.test_init_containers_have_no_probes which runs against all supported Kubernetes resource types (Deployment, DaemonSet, StatefulSet) and Provides clear error messages if probes are accidentally added to init containers. 

## Testing

-  All existing tests pass
Screenshot 2025-06-20 at 12 12 21 PM
- New test validates probe removal.

## Merging

0.37.4

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -340,8 +340,6 @@ houston:
     livenessProbe: {}
 
   bootstrapper:
-    readinessProbe: {}
-    livenessProbe: {}
 
   waitForDB:
     readinessProbe: {}
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -342,8 +342,6 @@ houston:
   bootstrapper:
 
   waitForDB:
-    readinessProbe: {}
-    livenessProbe: {}
 
   taskUsageMetrics:
     readinessProbe: {}
--- a/tests/chart_tests/test_probes.py
+++ b/tests/chart_tests/test_probes.py
@@ -37,12 +37,9 @@ class TestCustomProbes:
         init_containers = template_spec.get("initContainers", [])
 
         for init_container in init_containers:
-            container_name = init_container.get("name", "unnamed")
-
+            container_name = init_container.get("name")
             assert "livenessProbe" not in init_container, f"Init container '{container_name}' should not have livenessProbe"
-
             assert "readinessProbe" not in init_container, f"Init container '{container_name}' should not have readinessProbe"
-
             assert "startupProbe" not in init_container, f"Init container '{container_name}' should not have startupProbe"
 
 
--- a/tests/chart_tests/test_probes.py
+++ b/tests/chart_tests/test_probes.py
@@ -26,6 +26,25 @@ class TestCustomProbes:
             assert container["livenessProbe"] != {}
             assert container["readinessProbe"] != {}
 
+    filtered_docs = [doc for doc in docs if doc["kind"] in include_kind_list]
+
+    @pytest.mark.parametrize("doc", filtered_docs)
+    def test_init_containers_have_no_probes(self, doc):
+        """Ensure init containers do not have liveness or readiness probes."""
+
+        spec = doc.get("spec", {})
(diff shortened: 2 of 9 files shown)


charts/astronomer/values.yaml
```diff
@@ -340,12 +340,8 @@ houston:
     livenessProbe: {}
 
   bootstrapper:
```
pgvishnuram: does this key serves any additional purpose

charts/astronomer/values.yaml
```diff
@@ -340,12 +340,8 @@ houston:
...
-    readinessProbe: {}
-    livenessProbe: {}
 
   waitForDB:
```
pgvishnuram: does this key serves any additional purpose

charts/astronomer/values.yaml
```diff
@@ -340,12 +340,8 @@ houston:
...
-    readinessProbe: {}
-    livenessProbe: {}
 
   taskUsageMetrics:
```
pgvishnuram, line 342: does this key serves any additional purpose

charts/grafana/values.yaml
```diff
@@ -85,11 +85,7 @@ extraEnvVars:
   #       key: smtp_password
 
 bootstrapper:
```
pgvishnuram: does this key serves any additional purpose

charts/grafana/values.yaml
```diff
@@ -85,11 +85,7 @@ extraEnvVars:
...
-  readinessProbe: {}
-  livenessProbe: {}
 
 waitForDB:
```
pgvishnuram: does this key serves any additional purpose

charts/astronomer/values.yaml
Shubham-astro: No, let me just get rid of them.

charts/grafana/values.yaml
Shubham-astro: removed.

charts/grafana/values.yaml
Shubham-astro: removed

tests/chart_tests/test_probes.py
```diff
@@ -26,6 +26,22 @@ def test_template_probes_with_custom_values(self, doc):
             assert container["livenessProbe"] != {}
             assert container["readinessProbe"] != {}
 
+    filtered_docs = [doc for doc in docs if doc["kind"] in include_kind_list]
```
danielhoherd, line 29: We should probably give this a different name. I'm not sure on the ordering of how pytest sets these up, but overwriting the previous variable may lead to unexpected behavior. Let's switch the previous assignment to `filtered_containers` since that's what they are, not individual yaml docs produced by helm.

tests/chart_tests/test_probes.py
```diff
@@ -26,6 +26,22 @@ def test_template_probes_with_custom_values(self, doc):
...
+
+    @pytest.mark.parametrize("doc", filtered_docs)
+    def test_init_containers_have_no_probes(self, doc):
+        """Ensure init containers do not have liveness or readiness probes."""
+
+        spec = doc.get("spec", {})
```
danielhoherd: Maybe skip things that have no spec?
```suggestion
        if not (spec := doc.get("spec", {})):
            pytest.skip("doc has no .spec")
```

danielhoherd (approved):
Looks good. I suggested a few things that are optional.

PR #2732, opened 2025-06-23
Update Houston event URL and registry events configuration

Shubham-astro:
## Description

This PR adds the correct houston auth endpoint which needs to be used by Registry. 

## Related Issues

Related astronomer/issues#7358

## Testing

- Verified that the setup is working as expected in data mode.
- Unittests are passing as expected.

## Merging

1.0

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -499,7 +499,7 @@ tlsEnabled: false
 {{- define "houston.eventsUrl" -}}
 {{- if (eq .Values.global.plane.mode "data") }}
 https://houston.{{ .Values.global.baseDomain }}/v1/authorization
-{{- else }}
+{{- else -}}
 http://{{ .Release.Name }}-houston:{{ .Values.ports.houstonHTTP }}/v1/registry/events
 {{- end }}
 {{- end -}}
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -495,3 +495,11 @@ tlsEnabled: false
 {{ toYaml .Values.securityContext | indent 16 }}
 {{- end }}
 {{- end }}
+
+{{- define "houston.eventsUrl" -}}
+{{- if (eq .Values.global.plane.mode "data") }}
+https://houston.{{ .Values.global.baseDomain }}/v1/authorization
+{{- else }}
+http://{{ .Release.Name }}-houston:{{ .Values.ports.houstonHTTP }}/v1/registry/events
+{{- end }}
+{{- end -}}
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -497,7 +497,7 @@ tlsEnabled: false
 {{- end }}
 
 {{- define "houston.eventsUrl" -}}
-{{- if (eq .Values.global.plane.mode "data") }}
+{{- if (eq .Values.global.plane.mode "data") -}}
 https://houston.{{ .Values.global.baseDomain }}/v1/authorization
 {{- else -}}
 http://{{ .Release.Name }}-houston:{{ .Values.ports.houstonHTTP }}/v1/registry/events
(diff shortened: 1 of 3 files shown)


tests/chart_tests/test_registry_configmap.py
```diff
@@ -67,3 +67,38 @@ def test_registry_configmap_with_logLevel_overrides(self, kube_version):
...
+        assert len(config["notifications"]["endpoints"]) > 0
+
+        houston_endpoint = None
+        for endpoint in config["notifications"]["endpoints"]:
+            if endpoint["name"] == "houston":
+                houston_endpoint = endpoint
+                break
+
+        assert houston_endpoint is not None
+        actual_url = houston_endpoint["url"].strip()
```
danielhoherd: This `.strip()` is obscuring a bug, which is that the value produced by helm is `'\nhttps://houston.example.com/v1/authorization'`

charts/astronomer/templates/_helpers.yaml
```diff
@@ -495,3 +495,11 @@ tlsEnabled: false
 {{ toYaml .Values.securityContext | indent 16 }}
 {{- end }}
 {{- end }}
+
+{{- define "houston.eventsUrl" -}}
+{{- if (eq .Values.global.plane.mode "data") }}
+https://houston.{{ .Values.global.baseDomain }}/v1/authorization
+{{- else }}
```
danielhoherd:
```suggestion
{{- else -}}
```

charts/astronomer/templates/_helpers.yaml
```diff
@@ -495,3 +495,11 @@ tlsEnabled: false
 {{ toYaml .Values.securityContext | indent 16 }}
 {{- end }}
 {{- end }}
+
+{{- define "houston.eventsUrl" -}}
+{{- if (eq .Values.global.plane.mode "data") }}
```
danielhoherd:
```suggestion
{{- if (eq .Values.global.plane.mode "data") -}}
```

tests/chart_tests/test_registry_configmap.py
```diff
@@ -67,3 +67,38 @@ def test_registry_configmap_with_logLevel_overrides(self, kube_version):
...
+        assert actual_url == "https://houston.example.com/v1/authorization"
```
danielhoherd:
```suggestion
        houston_endpoint = next(
            (endpoint for endpoint in config["notifications"]["endpoints"] if endpoint["name"] == "houston"),
            None,
        )
        assert houston_endpoint == "https://houston.example.com/v1/authorization"
```

danielhoherd (changes requested):
Mostly good. There was a bug. Also we I left some cleaner code suggestions.

tests/chart_tests/test_registry_configmap.py
```diff
@@ -67,3 +67,34 @@ def test_registry_configmap_with_logLevel_overrides(self, kube_version):
...
+        doc = docs[0]
+        assert doc["kind"] == "ConfigMap"
+        assert doc["apiVersion"] == "v1"
+        assert doc["metadata"]["name"] == "release-name-registry"
+
+        config = yaml.safe_load(doc["data"]["config.yml"])
+
+        assert "notifications" in config
+        assert "endpoints" in config["notifications"]
+        assert len(config["notifications"]["endpoints"]) > 0
```
danielhoherd: We can do all three of these on one line. `assert list_of_some_sort` is equivalent to `assert len(list_of_some_sort) > 0` because 0 would return boolean false and fail the assertion.

```pycon
>>> config = {}
>>> assert config.get("notifications", {}).get("endpoints", [])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
>>> assert len(config.get("notifications", {}).get("endpoints", [])) > 0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
```

```suggestion
        assert config.get("notifications", {}).get("endpoints", [])
```

danielhoherd (approved):

PR #2743, opened 2025-07-03
fix registry events url for dataplane config

pgvishnuram:
## Description

fix registry events url for dataplane config

## Related Issues

Related https://github.com/astronomer/issues/issues/7358

## Testing

QA should able to verify registry events url had public endpoint config

## Merging

1.0


pgvishnuram:
> Do we need tests for this?

we have a unit test for this change i have updated the tests to assert the new url

PR #2752, opened 2025-07-09
7394 prometheus read only root

danielhoherd:
## Description

- prometheus pod containers run with read-only root. this skips authSidecar though, which will be handled separately.

## Related Issues

- https://github.com/astronomer/issues/issues/7394
- https://github.com/astronomer/issues/issues/7407
- https://github.com/astronomer/issues/issues/7408

## Testing

Unit tests are included, and are sufficient. No QA is needed.

## Merging

This is only for 1.0

### DIFF
--- a/charts/prometheus-postgres-exporter/templates/deployment.yaml
+++ b/charts/prometheus-postgres-exporter/templates/deployment.yaml
@@ -47,7 +47,9 @@ spec:
 {{- include "prometheus-postgres-exporter.imagePullSecrets" . | indent 6 }}
       containers:
         - name: {{ .Chart.Name }}
-          securityContext: {{ toYaml .Values.securityContext| nindent 12 }}
+          securityContext:
+            readOnlyRootFilesystem: true
+            {{ toYaml .Values.securityContext | nindent 12 }}
           args:
           - "--extend.query-path=/etc/config.yaml"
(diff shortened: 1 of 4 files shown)


charts/prometheus-postgres-exporter/templates/deployment.yaml
```diff
@@ -47,7 +47,9 @@ spec:
 {{- include "prometheus-postgres-exporter.imagePullSecrets" . | indent 6 }}
       containers:
         - name: {{ .Chart.Name }}
-          securityContext: {{ toYaml .Values.securityContext| nindent 12 }}
+          securityContext:
+            readOnlyRootFilesystem: true
```
pgvishnuram, line 51: are these going to be default feature

Shubham-astro: Yes @pgvishnuram this will be by default in 1.0

Shubham-astro (approved):

charts/prometheus-postgres-exporter/templates/deployment.yaml
danielhoherd: Not only is it the default, there is no way to turn it off. read-only root will be the only configuration starting in 1.0.

PR #2764, opened 2025-07-22
Remove unused feature: natsbox

danielhoherd:
## Description

I found a feature today that we have never used: natsbox. This PR removes the chart associated with this feature.

## Related Issues

https://github.com/astronomer/issues/issues/7492

## Testing

No testing is needed. These are the only references to natsbox in the entire product. We have never enabled this, so it is safe to remove. This is all dead code.

## Merging

We can merge this everywhere.


karankhanchandani (approved):
Thanks for the PR. LGTM. I left one comment on the issue, can you please check that?

PR #2768, opened 2025-07-23
Fix registry events consumption in data mode

Shubham-astro:
## Description

Remove registry/events path restriction from nginx ingress configuration when in data mode to enable proper consumption of registry events.

## Changes
- Added conditional logic to exclude registry/events path blocking when global.plane.mode is set to "data"
- Maintains existing path restrictions for unified and control plane modes
- Ensures registry events can be properly consumed in data mode deployments

## Related Issues

https://github.com/astronomer/issues/issues/7358

## Testing

- Verified registry events are accessible in data mode

## Merging

1.0

### DIFF
--- a/charts/nats/templates/_helpers.tpl
+++ b/charts/nats/templates/_helpers.tpl
@@ -70,7 +70,7 @@ imagePullSecrets:
 
 {{- define "nats.securityContext" -}}
 {{- if or (eq ( toString ( .Values.securityContext.runAsUser )) "auto") ( .Values.global.openshiftEnabled ) }}
-{{- omit  .Values.securityContext "runAsUser" | toYaml | nindent 10 }}
+{{- omit .Values.securityContext "runAsUser" | toYaml | nindent 10 }}
 {{- else }}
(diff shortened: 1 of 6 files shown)


charts/astronomer/templates/houston/ingress.yaml
```diff
@@ -23,6 +23,13 @@ metadata:
     {{- end }}
     kubernetes.io/ingress.class: "{{ .Release.Name }}-nginx"
     nginx.ingress.kubernetes.io/custom-http-errors: "404"
+    {{- if eq .Values.global.plane.mode "data" }}
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      location ~ ^/v1/(alerts|elasticsearch|metrics) {
+        deny all;
+        return 403;
+      }
+    {{- end }}
```
pgvishnuram: this will still gets updated with below configuration snippet right - we have to do if else right

pgvishnuram (approved):

rishkarajgi:
@Shubham-astro can you link this to a GH issue

PR #2769, opened 2025-07-24
7413 registry read-only root fs

danielhoherd:
## Description

Make registry run with readOnlyRootFilesystem.

I'm also solving a bug with the alertmanager readOnlyRootFilesystem implementation because it is breaking tests here.

## Related Issues

- https://github.com/astronomer/issues/issues/7394
- https://github.com/astronomer/issues/issues/7413

## Testing

This can be tested through normal flows, no extra testing is needed.

## Merging

This is only for 1.0.

### DIFF
--- a/charts/astronomer/templates/registry/registry-statefulset.yaml
+++ b/charts/astronomer/templates/registry/registry-statefulset.yaml
@@ -68,9 +68,9 @@ spec:
 {{- end }}
 {{- include "astronomer.imagePullSecrets" . | indent 6 }}
       initContainers:
-        - name: etc_ssl_certs_copier
+        - name: etc-ssl-certs-copier
           volumeMounts:
-            - name: etc_ssl_certs
+            - name: etc-ssl-certs
               mountPath: /etc/ssl/certs_copy
           image: {{ template "registry.image" . }}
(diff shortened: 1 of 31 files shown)


karankhanchandani (approved):
Thanks for the PR. LGTM. Can you please check why circle CI unified scenario builds are failing?

pgvishnuram:
can we test this on any of the backend enabled registries while uploading before it will be locally cached as well

charts/astronomer/templates/registry/registry-statefulset.yaml
```diff
@@ -135,6 +160,8 @@ spec:
             timeoutSeconds: 5
           {{- end }}
       volumes:
+        - name: etc_ssl_certs
+          emptyDir: {}
```
pgvishnuram: you are overriding the entire system trust store

danielhoherd: Correct. I am overriding it with a writable volume that has the contents of the system trust store, which was created in an initContainer.

https://github.com/astronomer/astronomer/blob/6eb4080a60c1e6460bcf240b0c63d933216c1fdf/charts/astronomer/templates/registry/registry-statefulset.yaml#L71-L86

PR #2774, opened 2025-07-29
Vectorfix

Shubham-astro:
## Description

## Related Issues

## Testing

## Merging

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -590,7 +590,7 @@ data:
           elasticsearch:
             enabled: true
             connection:
-              host: {{ printf "%s-external-es-proxy.%s" .Release.Name .Release.Namespace }}
+              host: elasticsearch.{{ .Values.global.plane.domainSuffix }}.{{ .Values.global.baseDomain }}
               port: 9200
 
     elasticsearch:
--- a/charts/elasticsearch/templates/es-ingress.yaml
+++ b/charts/elasticsearch/templates/es-ingress.yaml
@@ -16,7 +16,7 @@ metadata:
     plane: {{ .Values.global.plane.mode }}
   annotations:
     kubernetes.io/ingress.class: "{{ .Release.Name }}-nginx"
-    
+
     nginx.ingress.kubernetes.io/backend-protocol: "HTTP"
     nginx.ingress.kubernetes.io/proxy-body-size: "100m"
     nginx.ingress.kubernetes.io/proxy-read-timeout: "300"
--- a/charts/elasticsearch/templates/es-ingress.yaml
+++ b/charts/elasticsearch/templates/es-ingress.yaml
@@ -1,12 +1,12 @@
 ########################################
 ## Elasticsearch Logging Ingress ##
 ########################################
-#{{- if .Values.global.customLogging.enabled }}
-#{{- if (eq .Values.global.plane.mode "data") }}
+{{- if .Values.global.customLogging.enabled }}
+{{- if eq .Values.global.plane.mode "data" }}
 kind: Ingress
 apiVersion: {{ template "apiVersion.Ingress" . }}
 metadata:
-  name: {{ template "elasticsearch.fullname" . }}-elasticsearch-ingress
+  name: {{ template "elasticsearch.fullname" . }}-ingress
   labels:
     tier: elasticsearch-networking
     component: elasticsearch-logging-ingress
@@ -46,6 +46,6 @@ spec:
               service:
                 name: {{ .Release.Name }}-elasticsearch
                 port:
-                  name: {{ .Values.common.ports.http }}
+                  number: {{ .Values.common.ports.http }}
 {{- end }}
 {{- end }}
--- a/charts/elasticsearch/templates/es-ingress.yaml
+++ b/charts/elasticsearch/templates/es-ingress.yaml
@@ -1,0 +1,51 @@
+########################################
+## Elasticsearch Logging Ingress ##
+########################################
+#{{- if .Values.global.customLogging.enabled }}
+#{{- if (eq .Values.global.plane.mode "data") }}
+kind: Ingress
+apiVersion: {{ template "apiVersion.Ingress" . }}
+metadata:
+  name: {{ template "elasticsearch.fullname" . }}-elasticsearch-ingress
+  labels:
+    tier: elasticsearch-networking
+    component: elasticsearch-logging-ingress
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    plane: {{ .Values.global.plane.mode }}
+  annotations:
+    kubernetes.io/ingress.class: "{{ .Release.Name }}-nginx"
+    
+    nginx.ingress.kubernetes.io/backend-protocol: "HTTP"
+    nginx.ingress.kubernetes.io/proxy-body-size: "100m"
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "300"
+    nginx.ingress.kubernetes.io/proxy-send-timeout: "300"
+    nginx.ingress.kubernetes.io/proxy-buffer-size: "16k"
+    nginx.ingress.kubernetes.io/use-regex: "true"
+
+spec:
+  {{- if or .Values.global.tlsSecret .Values.global.acme }}
+  tls:
+  {{- if .Values.global.acme }}
+    - secretName: elasticsearch-logging-tls
+  {{- end }}
+  {{- if .Values.global.tlsSecret }}
+    - secretName: {{ .Values.global.tlsSecret }}
+  {{- end }}
+      hosts:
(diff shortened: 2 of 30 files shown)


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -590,7 +590,7 @@ data:
           elasticsearch:
             enabled: true
             connection:
-              host: {{ printf "%s-external-es-proxy.%s" .Release.Name .Release.Namespace }}
+              host: elasticsearch.{{ .Values.global.plane.domainSuffix }}.{{ .Values.global.baseDomain }}
```
danielhoherd, line 593: This should use the `elasticsearch.url` template defined below, right?

charts/elasticsearch/templates/es-ingress.yaml
```diff
@@ -0,0 +1,51 @@
...
+    nginx.ingress.kubernetes.io/proxy-send-timeout: "300"
+    nginx.ingress.kubernetes.io/proxy-buffer-size: "16k"
+    nginx.ingress.kubernetes.io/use-regex: "true"
+
+spec:
+  {{- if or .Values.global.tlsSecret .Values.global.acme }}
+  tls:
+  {{- if .Values.global.acme }}
+    - secretName: elasticsearch-logging-tls
+  {{- end }}
```
danielhoherd: global.acme is being removed so we can remove this part. 

```suggestion
```

charts/astronomer/templates/houston/houston-configmap.yaml
pgvishnuram: this might no longer be valid since the metadata url will expose these informations

tests/chart_tests/test_elasticsearch.py
```diff
@@ -711,11 +711,39 @@ def test_elasticsearch_statefulset_with_scc_enabled(self, kube_version):
...
+        assert len(docs) == 1
+        doc = docs[0]
+        assert doc["kind"] == "Ingress"
+        assert doc["metadata"]["name"] == "release-name-elasticsearch-ingress"
+        assert doc["metadata"]["labels"]["component"] == "elasticsearch-logging-ingress"
+        assert doc["metadata"]["labels"]["tier"] == "elasticsearch-networking"
+        assert doc["metadata"]["labels"]["plane"] == "data"
+
+    def test_elasticsearch_ingress_disabled_when_data_mode_is_disabled(self, kube_version):
+        """Test that helm does not render Elasticsearch ingress when datamode disabled."""
```
danielhoherd:
```suggestion
        """Test that helm does not render Elasticsearch ingress in control plane mode."""
```

tests/chart_tests/test_astronomer_commander.py
```diff
@@ -63,7 +63,7 @@ def test_astronomer_commander_deployment_default(self, kube_version):
         assert env_vars["COMMANDER_REGION"] == "us-west-2"
         assert env_vars["COMMANDER_BASE_DOMAIN"] == "custom-dp-123.example.com"
         assert env_vars["COMMANDER_DATAPLANE_MODE"] == "data"
-        assert env_vars["COMMANDER_HOUSTON_JWKS_ENDPOINT"] == "https://houston.example.com/auth"
+        assert env_vars["COMMANDER_HOUSTON_JWKS_ENDPOINT"] == "https://houston.example.com"
```
danielhoherd, line 66: Is this really the endpoint? Is this not just the URL? Should we change this ENV VAR? CC @pgvishnuram

charts/astronomer/templates/houston/houston-configmap.yaml
danielhoherd: We need to resolve this before merging this PR

tests/chart_tests/test_elasticsearch.py
```diff
@@ -711,11 +711,39 @@ def test_elasticsearch_statefulset_with_scc_enabled(self, kube_version):
...
+        docs = render_chart(
+            kube_version=kube_version,
+            values={
+                "global": {
+                    "plane": {"mode": "control"},
```
danielhoherd: Let's parametrize this with "unified" and "control".

tests/chart_tests/test_elasticsearch.py
```diff
@@ -711,11 +711,39 @@ def test_elasticsearch_statefulset_with_scc_enabled(self, kube_version):
...
 
     @pytest.mark.parametrize("doc", es_component_templates)
-    def test_elasticsearch_without_controlplane_flag_disabled(self, kube_version, doc):
-        """Test that helm renders no templates when controlplane is disabled."""
+    def test_elasticsearch_without_dataplane_flag_disabled(self, kube_version, doc):
+        """Test that helm renders no templates when dataplane is disabled."""
         docs = render_chart(
             kube_version=kube_version,
-            values={"global": {"plane": {"mode": "data"}}},
+            values={"global": {"plane": {"mode": "control"}}},
```
danielhoherd: Let's parametrize this with Data and Unified.

tests/chart_tests/test_elasticsearch.py
```diff
@@ -711,11 +711,39 @@ def test_elasticsearch_statefulset_with_scc_enabled(self, kube_version):
     ]
 
     @pytest.mark.parametrize("doc", es_component_templates)
-    def test_elasticsearch_without_controlplane_flag_disabled(self, kube_version, doc):
-        """Test that helm renders no templates when controlplane is disabled."""
+    def test_elasticsearch_without_dataplane_flag_disabled(self, kube_version, doc):
```
danielhoherd: This test name is a double negative. Can we change it to a positive form?

danielhoherd (changes requested):
Tests are looking good. I'd like to see some clean up that makes it more clear what we are testing.

charts/astronomer/templates/houston/houston-configmap.yaml
pgvishnuram: this is for airflow elasticsearch so technically no public url is needed here we will need only the local url, existing format is correct in this case. since all airflow deployment will connect

tests/chart_tests/test_elasticsearch.py
Shubham-astro: it will run in unified mode, all our components work under unified mode, Let me just name the test properly to avoid this confusion.

tests/chart_tests/test_elasticsearch.py
Shubham-astro: renamed to test_elasticsearch_templates_render_in_data_and_unified_mode

tests/chart_tests/test_elasticsearch.py
Shubham-astro: This will also run in unified mode.

tests/chart_tests/test_astronomer_commander.py
Shubham-astro: @danielhoherd, In chart we only have:
```
- name: COMMANDER_HOUSTON_JWKS_ENDPOINT
  value: "https://houston.{{ .Values.global.baseDomain }}"
```

The actual URL is: https://houston.sftplatform.astro-qa.link/v1/.well-known/jwks.json, but we construct the rest of the part through commander.  That's why the verification is only for "https://houston.example.com"

danielhoherd: I don't want this to block this PR, but calling a base domain name `ENDPOINT` is inaccurate.

charts/elasticsearch/templates/es-ingress.yaml
```diff
@@ -0,0 +1,46 @@
...
+    nginx.ingress.kubernetes.io/proxy-send-timeout: "300"
+    nginx.ingress.kubernetes.io/proxy-buffer-size: "16k"
+    nginx.ingress.kubernetes.io/use-regex: "true"
+
+spec:
+  {{- if or .Values.global.tlsSecret .Values.global.acme }}
+  tls:
+  {{- if .Values.global.tlsSecret }}
+    - secretName: {{ .Values.global.tlsSecret }}
+  {{- end }}
```
danielhoherd: global.acme is being removed. We don't need this condition.
```suggestion
  {{- if .Values.global.tlsSecret }}
  tls:
    - secretName: {{ .Values.global.tlsSecret }}
```

tests/chart_tests/test_elasticsearch.py
danielhoherd: What I mean is this test should be parametrized with all of the modes that it runs in. So if it's `control` and `unified` then we should have `@pytest.mark.parametrize("mode", ["control", "unified"])` and `values={"global": {"plane": {"mode": mode}}}` and `f"Document {doc} was rendered when {mode} mode`

tests/chart_tests/test_elasticsearch.py
danielhoherd: Wait, so are you saying this test is irrelevant in unified mode because we don't need an ingress in that mode?

danielhoherd (approved):
Let's merge it. We can circle back on anything that isn't right.

PR #2775, opened 2025-07-29
Fix Kubernetes Ingress Deprecation Warnings by Migrating to IngressClass

Shubham-astro:
## Description

When installing Astronomer using bin/reset-local-dev on Kubernetes 1.29, we encounter deprecation warnings:
W0315 14:28:28.711797   88080 warnings.go:70] annotation "kubernetes.io/ingress.class" is deprecated, please use 'spec.ingressClassName' instead
The kubernetes.io/ingress.class annotation was deprecated 4 years ago in Kubernetes 1.18 ([deprecation docs](https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#deprecating-the-ingress-class-annotation)).

## Changes:
This PR migrates all Ingress resources from the deprecated annotation to the modern IngressClass approach:

### Key Changes
 1. *Created IngressClass Resource* : Added ingress-class.yaml template to create Astronomer-managed IngressClass resources.
 2. Updated All Ingress Templates: Migrated 9 ingress templates across multiple charts
 3. Preserved all existing functionality including authSidecar logic

## Related Issues

Related astronomer/issues#6220

## Testing

- Tested local rendering of on of the template:
  Screenshot 2025-07-29 at 5 59 24 PM

## Merging

1.0


charts/astronomer/templates/ingress.yaml
```diff
@@ -13,20 +13,38 @@ metadata:
...
-{{- toYaml .Values.global.extraAnnotations | nindent 4 }}
+    {{- $filteredAnnotations := dict -}}
+    {{- range $key, $value := .Values.global.extraAnnotations -}}
+      {{- if ne $key "kubernetes.io/ingress.class" -}}
+        {{- $_ := set $filteredAnnotations $key $value -}}
+      {{- end -}}
+    {{- end -}}
+    {{- if $filteredAnnotations }}
+{{- toYaml $filteredAnnotations | nindent 4 }}
+    {{- end }}
```
danielhoherd, line 38: I'd love it if we could find a way to put this long snip of code that is pasted repeatedly into the templates into something that is reusable, like a named template in _helpers.tpl. Having it here makes the template really hard to follow.

Simpcyclassy (commented):
This PR and PR #3088 (`SWAT-59.ingressClassName-in-helm-chart`) address the same problem — migrating from the deprecated `kubernetes.io/ingress.class` annotation to `spec.ingressClassName`. Both are open simultaneously and will conflict if either merges without the other being closed or updated.

Before either can move forward, the team needs to decide which approach to use and close the other. This PR adds a new `IngressClass` resource and updates 9 ingress templates; PR #3088 takes a similar approach across more templates and adds a fallback to `extraAnnotations`. They cannot both merge.

Also noting: PR #3088 has an issue with the `extraAnnotations` fallback producing an empty `ingressClassName` value when the annotation key is not present. If this PR is chosen to proceed instead, verify it does not have a similar edge case.

Left as a comment pending the duplicate resolution.

PR #2777, opened 2025-07-30
Add ENV variables needed by Kuiper reloader

Shubham-astro:
## Description

This PR intends to add env vars to prometheus chart which are needed by kuiper reloader component.

## Related Issues

Related astronomer/issues#7505

## Testing
- Added unittests for the vars.
- All unittests Passed.
image

## Merging

1.0

### DIFF
--- a/charts/prometheus/templates/prometheus-statefulset.yaml
+++ b/charts/prometheus/templates/prometheus-statefulset.yaml
@@ -142,6 +142,10 @@ spec:
               value: {{ .Release.Name }}_houston
             - name: FILESD_FILE_PATH
               value: /prometheusreloader/airflow
+            - name: ENABLE_DEPLOYMENT_SCRAPING
+              value: "False"
+            - name: ENABLE_CLUSTER_SCRAPING
+              value: "False"
 {{- if .Values.filesdReloader.extraEnv }}
 {{ toYaml .Values.filesdReloader.extraEnv | indent 12 }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


charts/prometheus/templates/prometheus-statefulset.yaml
```diff
@@ -142,6 +142,10 @@ spec:
               value: {{ .Release.Name }}_houston
             - name: FILESD_FILE_PATH
               value: /prometheusreloader/airflow
+            - name: ENABLE_DEPLOYMENT_SCRAPING
+              value: "False"
+            - name: ENABLE_CLUSTER_SCRAPING
+              value: "False"
```
pgvishnuram: this only works if not clusterRole but for  this is needed for mode control and unified those are gonna be the default, and now we should make this kuiper reloader to be running all the time

charts/prometheus/templates/prometheus-statefulset.yaml
```diff
@@ -142,13 +143,18 @@ spec:
               value: {{ .Release.Name }}_houston
             - name: FILESD_FILE_PATH
               value: /prometheusreloader/airflow
+            {{- if not $useClusterRoles }}
+            - name: ENABLE_DEPLOYMENT_SCRAPING
+              value: "False"
+            {{- end }}
+            - name: ENABLE_CLUSTER_SCRAPING
+              value: "False"
```
pgvishnuram: why is this false by default ??

pgvishnuram (approved):

.circleci/config.yml
```diff
@@ -494,6 +495,7 @@ workflows:
                 - quay.io/astronomer/ap-init:3.21.3-3
                 - quay.io/astronomer/ap-kibana:8.15.5
                 - quay.io/astronomer/ap-kube-state:2.15.0
+                - quay.io/astronomer/ap-kuiper-reloader:0.1.5
```
pgvishnuram: this should be 0.1.7 right

Shubham-astro: updated

PR #2780, opened 2025-08-04
fix kibana and fluentd config

pgvishnuram:
## Description

## Related Issues

## Testing

## Merging

### DIFF
--- a/charts/fluentd/templates/fluentd-clusterrole.yaml
+++ b/charts/fluentd/templates/fluentd-clusterrole.yaml
@@ -2,6 +2,7 @@
 ## Fluentd ClusterRole ##
 #########################
 {{- if .Values.global.rbacEnabled }}
+{{- if or (eq .Values.global.plane.mode "data") (eq .Values.global.plane.mode "unified") }}
 kind: ClusterRole
 apiVersion: {{ template "apiVersion.rbac" . }}
 metadata:
@@ -22,4 +23,5 @@ rules:
   - "get"
   - "watch"
   - "list"
+{{- end -}}
 {{- end -}}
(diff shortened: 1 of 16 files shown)


charts/fluentd/templates/fluentd-clusterrole.yaml
```diff
@@ -2,6 +2,7 @@
 ## Fluentd ClusterRole ##
 #########################
 {{- if .Values.global.rbacEnabled }}
+{{- if or (eq .Values.global.plane.mode "data") (eq .Values.global.plane.mode "unified") }}
```
pgvishnuram, line 5: shall i combine this two condition into single

charts/fluentd/templates/fluentd-clusterrolebinding.yaml
```diff
@@ -2,6 +2,7 @@
 ## Fluentd ClusterRoleBinding ##
 ################################
 {{- if .Values.global.rbacEnabled }}
+{{- if or (eq .Values.global.plane.mode "data") (eq .Values.global.plane.mode "unified") }}
```
pgvishnuram, line 5: shall i combine this two condition into single

charts/fluentd/templates/fluentd-clusterrole.yaml
danielhoherd: Lines 4 and 5 here are fine.

charts/fluentd/templates/fluentd-clusterrolebinding.yaml
danielhoherd: It's fine as it is.

danielhoherd (approved):
This is all fine. I would ask for tests on Kibana but we're going to remove it so they are not needed.

PR #2787, opened 2025-08-09
7162 start adding yamllint

danielhoherd:
## Description

This is the first part of adding yamllint to our chart validation steps.
- add an optional (disabled by default for now) yamllint step in `render_chart()`
- Fix a bunch of whitespace problems, some duplicate yaml keys, and spacing ambiguities

This is not the final PR in this effort, but this does solve some of the more important problems, most importantly duplicate keys. I decided to open a PR for this now rather than waiting, primarily so we can avoid bigger merge conflicts.

## Related Issues

astronomer/issues#7162

## Testing

No extra tests are needed here. If CI is passing we're good.

## Merging

We likely don't need to backport this, but it wouldn't hurt. Doing so would likely create a big merge conflict. Also, this is not a complete feature yet. Backporting the tests/utils/chart.py changes would allow us to at least investigate problems on older versions of the chart.

For now I think 1.0 is fine.


charts/nats/templates/statefulset.yaml
```diff
@@ -319,7 +319,6 @@ spec:
       - name: metrics
         image: {{ template "nats-exporter.image" . }}
         imagePullPolicy: {{ .Values.images.exporter.pullPolicy }}
-        securityContext: {{ template "nats.securityContext" . }}
```
danielhoherd, line 322: This was a duplicate. The other one is down on line 344.

PR #2821, opened 2025-09-03
Fix nginx-elasticsearch DNS resolution failure in unified mode

Shubham-astro:
## Description

The astronomer-elasticsearch-nginx pod is failing to start in unified mode deployments due to a DNS resolution error:
nginx: [emerg] host not found in upstream "houston.shubham-test22.astro-qa.link" in /etc/nginx/nginx.conf:42
The pod enters a CrashLoopBackOff state because nginx cannot resolve the external Houston hostname during startup validation.

## Root Cause
In unified mode, the nginx configuration attempts to use the external Houston domain `(https://houston.{{ .Values.global.baseDomain }})`  for authentication requests, but:
- The external DNS may not be immediately resolvable from within the cluster
- This creates an unnecessary external dependency when Houston runs in the same cluster

## Related Issues

## Testing

- Verified that the nginx pod started without issues after this change:
Screenshot 2025-09-03 at 11 50 03 AM

Screenshot 2025-09-03 at 11 50 34 AM

## Merging

Master

### DIFF
--- a/charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml
+++ b/charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml
@@ -57,7 +57,7 @@ data:
 
         location = /auth {
           internal;
-          {{- include "houston.proxypass" . | nindent 10 }}
+          {{- include "houston-proxy" . | nindent 10 }}
           proxy_set_header Content-Length "";
           proxy_set_header X-Original-URI $request_uri;
         }
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -68,7 +68,7 @@ imagePullSecrets:
 {{- end }}
 {{- end }}
 
-{{- define "houston.proxypass" -}}
+{{- define "houston-proxy" -}}
 {{- if eq .Values.global.plane.mode "unified" -}}
 proxy_pass http://{{ .Release.Name }}-houston.{{ .Release.Namespace }}:8871/v1/elasticsearch;
 {{- else -}}
--- a/charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
+++ b/charts/external-es-proxy/templates/external-es-proxy-configmap.yaml
@@ -54,7 +54,7 @@ data:
 
         location = /auth {
           internal;
-          {{- include "houston.proxypass" . | nindent 10 }}
+          {{- include "houston-proxy" . | nindent 10 }}
           proxy_set_header Content-Length "";
           proxy_set_header X-Original-URI $request_uri;
         }


charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml
```diff
@@ -57,7 +57,11 @@ data:
 
         location = /auth {
           internal;
+          {{- if eq .Values.global.plane.mode "unified" }}
+          proxy_pass http://{{ .Release.Name }}-houston.{{ .Release.Namespace }}:{{ .Values.common.ports.houstonHTTP }}/v1/elasticsearch;
```
pgvishnuram: use last . as well

Shubham-astro: I was seeing this issue @pgvishnuram:

```
2025/09/03 06:23:06 [emerg] 1#1: invalid port in upstream "astronomer-houston.astronomer:.8871/v1/elasticsearch" in /etc/nginx/nginx.conf:40
nginx: [emerg] invalid port in upstream "astronomer-houston.astronomer:.8871/v1/elasticsearch" in /etc/nginx/nginx.conf:40
```

templates/_helpers.tpl
```diff
@@ -67,3 +67,11 @@ imagePullSecrets:
 {{- toYaml .Values.global.podLabels }}
 {{- end }}
 {{- end }}
+
+{{- define "houston.proxypass" -}}
+{{- if eq .Values.global.plane.mode "unified" }}
+proxy_pass http://{{ .Release.Name }}-houston.{{ .Release.Namespace }}:8871/v1/elasticsearch;
+{{- else }}
```
danielhoherd: Do these need to have RHS truncation applied?
```suggestion
{{- if eq .Values.global.plane.mode "unified" -}}
proxy_pass http://{{ .Release.Name }}-houston.{{ .Release.Namespace }}:8871/v1/elasticsearch;
{{- else -}}
```

Shubham-astro: yes, it was needed

charts/elasticsearch/templates/nginx/nginx-es-configmap.yaml
```diff
@@ -57,7 +57,7 @@ data:
 
         location = /auth {
           internal;
-          proxy_pass https://houston.{{ .Values.global.baseDomain }}/v1/elasticsearch;
+          {{- include "houston.proxypass" . | nindent 10 }}
```
pgvishnuram: can it be houston-proxy

Shubham-astro: done

pgvishnuram (approved):

PR #2843, opened 2025-09-15
add elastic extra annotation support

pgvishnuram:
## Description

Currently elasticsearch ingress is hardcoded with set of nginx ingress controller annotation, which seems to be not wroking as expected. After investigation it seems es nginx dont support allowing extra ingress to be passed.

## Related Issues

- https://github.com/astronomer/issues/issues/7833

## Testing

QA should able to see custom ingress annotations  being passed ingress 

## Merging

nerge to master and release-1.0


charts/elasticsearch/templates/es-ingress.yaml
```diff
@@ -1,28 +1,36 @@
...
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
     plane: {{ .Values.global.plane.mode }}
   annotations:
+    {{- if .Values.global.authSidecar.enabled  }}
+{{- toYaml .Values.global.extraAnnotations | nindent 4 }}
+    {{- else }}
+    {{- if .Values.global.extraAnnotations }}
+{{- toYaml .Values.global.extraAnnotations | nindent 4 }}
+    {{- end }}
```
danielhoherd: Because of `{{- ... nindent` these `toYaml` can be indented however we like.
```suggestion
    {{- if .Values.global.authSidecar.enabled  }}
      {{- toYaml .Values.global.extraAnnotations | nindent 4 }}
    {{- else }}
    {{- if .Values.global.extraAnnotations }}
      {{- toYaml .Values.global.extraAnnotations | nindent 4 }}
    {{- end }}
```

tests/chart_tests/test_ingress.py
```diff
@@ -115,3 +115,39 @@ def test_prometheus_federate_ingress(self, kube_version):
...
+            assert auth_annotation not in annotations
+
+        rules = docs[0]["spec"]["rules"]
+        assert len(rules) == 1
+        paths = rules[0]["http"]["paths"]
+        assert len(paths) == 1
+        assert paths[0]["path"] == "/"
+        assert paths[0]["pathType"] == "Prefix"
+        backend = paths[0]["backend"]
+        assert backend["service"] == {"name": "release-name-elasticsearch", "port": {"number": 9200}}
```
danielhoherd: This would all be easier to read and more specific if we just did
```suggestion
        assert rules[0]["http"]["paths"] == [
            {
                "path": "/",
                "pathType": "Prefix",
                "backend": {
                    "service": {
                        "name": "release-name-elasticsearch",
                        "port": {"number": 9200},
                    }
                },
            }
        ]
```

Or maybe even `assert rules[0] == {'host': 'release-name-elasticsearch.default.svc.cluster.local.:9200', 'http': {'paths': [{'path': '/', 'pathType': 'Prefix', 'backend': {'service': {'name': 'release-name-elasticsearch', 'port': {'number': 9200}}}}]}}`

rishkarajgi:
tests failing

rishkarajgi (approved):

PR #2852, opened 2025-09-23
Remove pass as this is causing mismatch of the tokens  in CP and DP

Shubham-astro:
## Description

This PR removes the random password that we are creating in houston as this is something not useful and causes a new key to be created if houston tls secret object is not accessible. This can cause keys in DP to be different than what is present in CP, which inturn causes auth errors.

## Related Issues

https://github.com/astronomer/issues/issues/7892

## Testing

- No Testing needed, since this is optional.

## Merging

Master, 1.0


pgvishnuram:
@rishkarajgi we should have an issue this is altering existing user behaviour

PR #2853, opened 2025-09-24
Enable createJetStreamJob by default

Shubham-astro:
## Description

This PR enables createJetStreamJob by default since we have removed STAN from our charts so houston needs to NATS job need to be enabled to work without issues.

## Related Issues

Related https://github.com/astronomer/issues/issues/7818
https://github.com/astronomer/issues/issues/7351

## Testing

- Fixed unittests
- No QA Needed since anyways our QA team do this through terraform values file. 

## Merging

Master, 1.0

### DIFF
--- a/charts/nats/values.yaml
+++ b/charts/nats/values.yaml
@@ -76,7 +76,7 @@ nats:
 
   regenerateCaEachUpgrade: false
 
-  jetstream:
+  jetStream:
     enabled: true
     serviceAccount:
       create: true
--- a/charts/nats/values.yaml
+++ b/charts/nats/values.yaml
@@ -19,8 +19,6 @@ images:
 
 
 nats:
-  createJetStreamJob: true
-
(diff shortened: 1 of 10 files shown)


charts/nats/values.yaml
```diff
@@ -19,7 +19,7 @@ images:
 
 
 nats:
-  createJetStreamJob: false
+  createJetStreamJob: true
```
rishkarajgi: do we need this flag?

pgvishnuram: this will require clusterAdmin permission since we are using namespace label capabilities.

danielhoherd (approved):
This all seems good to me. I had thought about creating a pre-commit hook to disallow `jetstream` but we can't do that because the filesystem names are lowercase (which is correct) so we can't blanket reject the lower-case form. Anyhow, this looks good to me and looks good to merge.

PR #2854, opened 2025-09-25
Fix AirGapped deployment issues by adding writable Helm directories

Shubham-astro:
## Description

Commander pods fail to start on AirGapped env due to filesystem permission issues. The pod runs as a non-root user (e.g., `1000730000`) but Helm tries to write to read-only root filesystem directories (`/.cache`, `/.config`, `/.local`), causing errors like:

-  `mkdir /.cache: read-only file system`
-  `repo astronomer-ee not found`

## Related Issues

Related astronomer/issues#7900

## Testing

- Tested the local rendering of the helm: 
```helm template . --show-only charts/astronomer/templates/commander/commander-deployment.yaml --set astronomer.commander.airGapped.enabled=true```
Screenshot 2025-09-25 at 6 06 51 PM

## Merging

Master, 1.0

### DIFF
--- a/tests/chart_tests/test_astronomer_commander.py
+++ b/tests/chart_tests/test_astronomer_commander.py
@@ -431,17 +431,11 @@ class TestAstronomerCommander:
         assert doc["apiVersion"] == "apps/v1"
         assert doc["metadata"]["name"] == "release-name-commander"
         c_by_name = get_containers_by_name(doc)
-        commander_container = c_by_name["commander"]
-        env_vars = {x["name"]: get_env_value(x) for x in commander_container["env"]}
+        env_vars = {x["name"]: get_env_value(x) for x in c_by_name["commander"]["env"]}
         assert env_vars["COMMANDER_HELM_DEBUG"] == "true"
         assert env_vars["COMMANDER_MANAGE_NAMESPACE_RESOURCE"] == "false"
         assert env_vars["COMMANDER_MANUAL_NAMESPACE_NAMES"] == "true"
         assert env_vars["COMMANDER_AIRGAPPED"] == "true"
-        volume_mounts = {mount["name"]: mount["mountPath"] for mount in commander_container["volumeMounts"]}
-        assert volume_mounts["tmp-workspace"] == "/tmp"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: charts/astronomer/templates/commander/commander-deployment.yaml [hunk 1]
```diff
@@ -136,6 +136,14 @@ spec:
             {{- end }}
             {{- include "commander_metadataEnv" . | nindent 12 }}
           volumeMounts:
+            {{- if .Values.commander.airGapped.enabled }}
+            - name: helm-workspace
+              mountPath: /tmp/helm-cache
+            - name: helm-config
+              mountPath: /tmp/helm-config
+            - name: helm-data
+              mountPath: /tmp/helm-data
```
timo-astronomer: do we set somewhere helm to use `/tmp`?

in my tests, in EKS (when running as `commander`) it used `/home/commander`
and in Openshift (when running as `1000730000`) it used `/`

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -136,6 +136,14 @@ spec:
...
+            {{- end }}
```
danielhoherd: Let's just mount /tmp as emptyDir. It's something we've done for a lot of other readOnlyRootFilesystem components.

Shubham-astro: done

charts/astronomer/templates/commander/commander-deployment.yaml
Shubham-astro: @danielhoherd @pgvishnuram 
should we add env vars something like this:
```
- name: HELM_CACHE_HOME
    value: /tmp/helm-cache
```
to make helm use /tmp only.?  what do you think.?

leestro: > @danielhoherd @pgvishnuram should we add env vars something like this:
> 
> ```
> - name: HELM_CACHE_HOME
>     value: /tmp/helm-cache
> ```
> 
> to make helm use /tmp only.? what do you think.?

we should test this... when i tried setting `HELM_CACHE_HOME=/tmp/.cache` in a customer's openshift environment it seemed to get ignored.

PR #2859, opened 2025-09-26
fix ingress controller component tags

pgvishnuram:
## Description

Fix network policies and correctly use the ingress components matching the plane modes

## Related Issues

- https://github.com/astronomer/issues/issues/7824

## Testing

QA should see correct network policies when different modes are selected

## Merging

### DIFF
--- a/charts/astronomer/templates/registry/registry-networkpolicy.yaml
+++ b/charts/astronomer/templates/registry/registry-networkpolicy.yaml
@@ -99,7 +99,7 @@ spec:
     - podSelector:
         matchLabels:
           tier: nginx
-          component: ingress-controller
+          component: cp-ingress-controller
           release: {{ .Release.Name }}
     {{- end }}
     ports:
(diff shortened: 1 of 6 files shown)


rishkarajgi:
Will these new naming convention affect unified mode? We will have only one ingress in unified mode right?

charts/astronomer/templates/registry/registry-networkpolicy.yaml
```diff
@@ -99,7 +99,7 @@ spec:
     - podSelector:
         matchLabels:
           tier: nginx
-          component: ingress-controller
+          component: {{ ternary "cp-ingress-controller" "dp-ingress-controller" (eq .Values.global.plane.mode "unified") }}
```
timo-astronomer, line 102: what if mode is `control`?

danielhoherd (approved):
These changes look good. This bug does indicate a gap in our testing though, but because of the crunch for 1.0 I'm not sure it's worth addressing now.

PR #2860, opened 2025-09-26
fix all network policy configurations

pgvishnuram:
## Description

rework all network ingress policy config

## Related Issues

- https://github.com/astronomer/issues/issues/7824

## Testing

QA should able to provision astronomer chart with proper network policy in place

## Merging

### DIFF
--- a/charts/astronomer/templates/registry/registry-networkpolicy.yaml
+++ b/charts/astronomer/templates/registry/registry-networkpolicy.yaml
@@ -99,7 +99,7 @@ spec:
     - podSelector:
         matchLabels:
           tier: nginx
-          component: {{ ternary "cp-ingress-controller" "dp-ingress-controller" (or (eq .Values.global.plane.mode "unified") (eq .Values.global.plane.mode "control") )}}
+          component: {{ ternary "dp-ingress-controller" "cp-ingress-controller" (eq .Values.global.plane.mode "data")}}
           release: {{ .Release.Name }}
     {{- end }}
     ports:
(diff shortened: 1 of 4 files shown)


charts/prometheus/templates/prometheus-networkpolicy.yaml
```diff
@@ -42,7 +42,7 @@ spec:
     - podSelector:
         matchLabels:
           tier: nginx
-          component: ingress-controller
+          component: {{ ternary "cp-ingress-controller" "dp-ingress-controller" (or (eq .Values.global.plane.mode "unified") (eq .Values.global.plane.mode "control") )}}
```
timo-astronomer: i know current code is fine, but for me this is simpler :)
```suggestion
          component: {{ ternary "dp-ingress-controller" "cp-ingress-controller" (eq .Values.global.plane.mode "data") )}}
```

charts/astronomer/templates/registry/registry-networkpolicy.yaml
```diff
@@ -99,7 +99,7 @@ spec:
     - podSelector:
         matchLabels:
           tier: nginx
-          component: {{ ternary "cp-ingress-controller" "dp-ingress-controller" (eq .Values.global.plane.mode "unified") }}
+          component: {{ ternary "cp-ingress-controller" "dp-ingress-controller" (or (eq .Values.global.plane.mode "unified") (eq .Values.global.plane.mode "control") )}}
```
timo-astronomer: i know current code is fine, but for me this is simpler :)
```suggestion
          component: {{ ternary "dp-ingress-controller" "cp-ingress-controller" (eq .Values.global.plane.mode "data") )}}
```

danielhoherd: IMHO the ternary syntax is so protracted and unintuitive that either way I would not be confident that I understood it without looking up the docs and spending 5 minutes making sure I was confident in my understanding.

PR #2870, opened 2025-09-29
fix network policies on dataplane mode

pgvishnuram:
## Description

rework dataplane network policies to sync with control plane 

## Related Issues

- https://github.com/astronomer/issues/issues/7926

## Testing

QA should able to access the CP and DP ingress without any issues

## Merging

### DIFF
--- a/charts/astronomer/templates/commander/commander-networkpolicy.yaml
+++ b/charts/astronomer/templates/commander/commander-networkpolicy.yaml
@@ -48,7 +48,7 @@ spec:
           release: {{ .Release.Name }}
     {{- end }}
     {{- end }}
-    {{- if or (eq .Values.global.plane.mode "unified") (eq .Values.global.plane.mode "control") }}
+    {{- if or (eq .Values.global.plane.mode "unified") }}
     - podSelector:
         matchLabels:
           tier: astronomer
--- a/charts/astronomer/templates/commander/commander-networkpolicy.yaml
+++ b/charts/astronomer/templates/commander/commander-networkpolicy.yaml
@@ -48,7 +48,7 @@ spec:
           release: {{ .Release.Name }}
     {{- end }}
     {{- end }}
-    {{- if or (eq .Values.global.plane.mode "unified") }}
+    {{- if (eq .Values.global.plane.mode "unified") }}
     - podSelector:
         matchLabels:
           tier: astronomer
--- a/charts/astronomer/templates/commander/commander-networkpolicy.yaml
+++ b/charts/astronomer/templates/commander/commander-networkpolicy.yaml
@@ -48,6 +48,7 @@ spec:
           release: {{ .Release.Name }}
     {{- end }}
     {{- end }}
+    {{- if or (eq .Values.global.plane.mode "unified") (eq .Values.global.plane.mode "control") }}
     - podSelector:
         matchLabels:
           tier: astronomer
@@ -103,6 +104,7 @@ spec:
           tier: astronomer
           release: {{ .Release.Name }}
           component: houston-cleanup-airflow-db
+    {{- end }}
     ports:
     {{- if and .Values.global.authSidecar.enabled (eq .Values.global.plane.mode "data") }}
     - protocol: TCP
--- a/charts/elasticsearch/templates/client/es-client-networkpolicy.yaml
+++ b/charts/elasticsearch/templates/client/es-client-networkpolicy.yaml
@@ -71,11 +71,6 @@ spec:
           tier: elasticsearch
           component: es-ingress-controller
           release: {{ .Release.Name }}
-    - podSelector:
-        matchLabels:
-          tier: logging
-          component: vector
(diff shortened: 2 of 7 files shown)


charts/nginx/templates/dataplane/nginx-dp-networkpolicy.yaml
```diff
@@ -19,35 +19,19 @@ spec:
...
-    - podSelector:
-        matchLabels:
-          plane: {{ .Values.global.plane.mode }}
-          component: dp-ingress-controller
-    - podSelector:
-        matchLabels:
-          plane: {{ .Values.global.plane.mode }}
-          component: houston
+    - ipBlock:
+        cidr: 0.0.0.0/0
```
rishkarajgi, line 28: So, by this commander pod is allowed to receive traffic from any IP address. Will this affect airgapped customers?

rishkarajgi (changes requested):
please add tests

charts/nginx/templates/dataplane/nginx-dp-networkpolicy.yaml
ianbuss: It won't invalidate any other firewalls or security group rules such customers have in place. It just means the nginx ingress controller will accept incoming traffic, since the default policy is to deny all ingress traffic I believe (templates/default-deny-network-policy/networkpolicy.yaml). This isn't exposing commander directly, only via the nginx ingress.

charts/astronomer/templates/commander/commander-networkpolicy.yaml
```diff
@@ -48,6 +48,7 @@ spec:
           release: {{ .Release.Name }}
     {{- end }}
     {{- end }}
+    {{- if or (eq .Values.global.plane.mode "unified") (eq .Values.global.plane.mode "control") }}
```
ianbuss: We don't need these in `control` mode do we? Since there is no commander there.

ianbuss: In fact this whole file should be wrapped by `{{- if or (eq .Values.global.plane.mode "unified") (eq .Values.global.plane.mode "data") }}` ... `{{- end }}`.

charts/elasticsearch/templates/client/es-client-networkpolicy.yaml
```diff
@@ -26,6 +26,26 @@ spec:
...
+    {{- end }}
+    {{- else }}
+    - podSelector:
+        matchLabels:
+          tier: nginx
+          component: dp-ingress-controller
+          release: {{ .Release.Name }}
+    {{- end }}
+    {{- end }}
     - podSelector:
```
ianbuss, line 50: This one not required in `data` mode.

ianbuss: Is any of this required in `control` mode? What ES components run in the CP?

charts/prometheus/templates/prometheus-federate-networkpolicy.yaml
```diff
@@ -0,0 +1,53 @@
...
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values: {{ .Values.global.authSidecar.ingressAllowedNamespaces | toJson }}
+    {{- end }}
+    {{- else }}
+    - podSelector:
+        matchLabels:
+          tier: nginx
+          component: {{ ternary "dp-ingress-controller" "cp-ingress-controller" (eq .Values.global.plane.mode "data") }}
```
ianbuss: Since this can only be in `data` mode, do we need the switch here?

charts/prometheus/templates/prometheus-federate-networkpolicy.yaml
```diff
@@ -0,0 +1,53 @@
...
+          release: {{ .Release.Name }}
+    {{- end }}
+    # It is sometimes useful to whitelist other pods,
```
ianbuss:
```suggestion
    # It is sometimes useful to allow other pods,
```

charts/elasticsearch/templates/client/es-client-networkpolicy.yaml
pgvishnuram: on control mode we dont run any elasticsearch service , the way how we expose es differs based on the mode

ianbuss (approved):
lgtm, thanks for the changes

charts/astronomer/templates/commander/commander-networkpolicy.yaml
```diff
@@ -48,6 +48,7 @@ spec:
           release: {{ .Release.Name }}
     {{- end }}
     {{- end }}
+    {{- if or (eq .Values.global.plane.mode "unified") }}
```
danielhoherd: Shouldn't this be
```suggestion
    {{- if (eq .Values.global.plane.mode "unified") }}
```

PR #2885, opened 2025-10-03
Remove `registry.baseDomain`

Shubham-astro:
## Description

Remove registry.baseDomain exposure from control plane mode ingress configuration. The registry subdomain should only be accessible in unified plane mode.

## Related Issues

## Testing

- No QA needed registry is anyways not used in control plane.
- Added unittest.

## Merging

Master, 1.0

### DIFF
--- a/charts/astronomer/templates/ingress.yaml
+++ b/charts/astronomer/templates/ingress.yaml
@@ -34,7 +34,9 @@ spec:
       hosts:
         - {{ .Values.global.baseDomain }}
         - app.{{ .Values.global.baseDomain }}
+      {{- if eq .Values.global.plane.mode "unified" }}
         - registry.{{ .Values.global.baseDomain }}
+      {{- end }}
         - install.{{ .Values.global.baseDomain }}
   {{- end }}
   rules:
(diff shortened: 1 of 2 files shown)


charts/astronomer/templates/ingress.yaml
```diff
@@ -34,7 +34,9 @@ spec:
       hosts:
         - {{ .Values.global.baseDomain }}
         - app.{{ .Values.global.baseDomain }}
+      {{- if eq .Values.global.plane.mode "unified" }}
```
pgvishnuram: lets use variable and call in rest of the places this will reduce conditional changes in multiple places

charts/astronomer/templates/ingress.yaml
```diff
@@ -58,6 +60,7 @@ spec:
               name: {{ .Release.Name }}-astro-ui
               port:
                 name: astro-ui-http
+{{- if (eq .Values.global.plane.mode "unified") }}
```
pgvishnuram: same here

charts/astronomer/templates/ingress.yaml
Shubham-astro: done

pgvishnuram (approved):

PR #2887, opened 2025-10-03
Revert "Remove kibana (#2798)"

danielhoherd:
## Description

Revert removal of kibana.

## Related Issues

https://github.com/astronomer/issues/issues/7536

## Testing

BEFORE MERGING THIS we will need to test that kibana works in the new separated CP/DP scenario. We will likely need to update code to make kibana work with the new CP/DP separation. EG: we will probably need to update the ingress configurations.

## Merging

This is only for 1.0


Simpcyclassy (commented):
Reverting the Kibana removal for the 1.0 branch. The PR description itself says this should not merge until Kibana has been tested in the CP/DP separation scenario and ingress configurations are updated. Given that note is in the PR body, I will not add new concerns on top of it — the author already knows what needs to happen.

The revert itself (undoing PR #2798) is mechanically straightforward. Once the CP/DP testing is done and any needed ingress updates are made, this should be ready. Left as a comment rather than approving since the testing prerequisite is still open.

PR #2895, opened 2025-10-07
Add nginx-cache Volumes and VolumeMounts for External Elasticsearch Proxy

Shubham-astro:
## Description

This PR introduces a new nginx-cache volume and associated volumeMounts for the external-es-proxy Deployment template. These directories are required by OpenResty for temporary operations such as proxying, client body buffering, and fastcgi/uwsgi handling.

## Related Issues

Related astronomer/issues#7904

## Testing

- Added Unittests
- Tested on cluster 
Screenshot 2025-10-07 at 4 21 14 PM

- Confirmed that logs were stable after this:
Screenshot 2025-10-07 at 4 29 27 PM

## Merging

Master, 1.0

### DIFF
--- a/charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
+++ b/charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
@@ -87,6 +87,16 @@ spec:
               subPath: setenv.lua
             - name: tmp
               mountPath: /tmp
+            - name: nginx-cache
+              mountPath: /usr/local/openresty/nginx/client_body_temp
+            - name: nginx-cache
+              mountPath: /usr/local/openresty/nginx/proxy_temp
+            - name: nginx-cache
+              mountPath: /usr/local/openresty/nginx/fastcgi_temp
+            - name: nginx-cache
+              mountPath: /usr/local/openresty/nginx/uwsgi_temp
+            - name: nginx-cache
+              mountPath: /usr/local/openresty/nginx/scgi_temp
 
             {{- if .Values.global.customLogging.trustCaCerts -}}
             {{ $secretName := .Values.global.customLogging.trustCaCerts }}
@@ -149,6 +159,8 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       volumes:
+        - name: nginx-cache
+          emptyDir: {}
         - name: tmp
           emptyDir: {}
         - name: nginx-config-volume
--- a/tests/chart_tests/test_external_elasticsearch.py
+++ b/tests/chart_tests/test_external_elasticsearch.py
@@ -29,6 +29,13 @@ class TestExternalElasticSearch:
             ],
         )
 
+        expected_nginx_mounts = [
+        {"name": "nginx-cache", "mountPath": "/usr/local/openresty/nginx/client_body_temp"},
+        {"name": "nginx-cache", "mountPath": "/usr/local/openresty/nginx/proxy_temp"},
+        {"name": "nginx-cache", "mountPath": "/usr/local/openresty/nginx/fastcgi_temp"},
+        {"name": "nginx-cache", "mountPath": "/usr/local/openresty/nginx/uwsgi_temp"},
+        {"name": "nginx-cache", "mountPath": "/usr/local/openresty/nginx/scgi_temp"},
+        ]
         assert len(docs) == 4
         deployment, _env_configmap, _configmap, service = docs
         assert deployment["kind"] == "Deployment"
@@ -43,8 +50,11 @@ class TestExternalElasticSearch:
         assert expected_env == deployment["spec"]["template"]["spec"]["containers"][0]["env"]
         container_mounts = deployment["spec"]["template"]["spec"]["containers"][0]["volumeMounts"]
         assert {"name": "tmp", "mountPath": "/tmp"} in container_mounts
+        for mount in expected_nginx_mounts:
+            assert mount in container_mounts, f"Missing mount {mount}"
         volumes = deployment["spec"]["template"]["spec"]["volumes"]
         assert {"name": "tmp", "emptyDir": {}} in volumes
+        assert {"name": "nginx-cache", "emptyDir": {}} in volumes
 
         assert service["kind"] == "Service"
         assert service["metadata"]["name"] == "release-name-external-es-proxy"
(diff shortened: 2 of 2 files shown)


charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
```diff
@@ -87,6 +87,16 @@ spec:
               subPath: setenv.lua
             - name: tmp
               mountPath: /tmp
+            - name: nginx-cache
+              mountPath: /usr/local/openresty/nginx/client_body_temp
```
ianbuss, line 92: This means that the same volume dir will be shared by all these locations. I'm not sure that's what we want, could well result in undefined behaviour. Instead we may need to define an emptyDir volume for each f these locations.

ianbuss (approved):
Looks good from my perspective. How confident are we that these are the only locations openresty will need write access?

charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
```diff
@@ -87,6 +87,16 @@ spec:
...
+            - name: nginx-client-body-temp
+              mountPath: /usr/local/openresty/nginx/client_body_temp
+            - name: nginx-proxy-temp
+              mountPath: /usr/local/openresty/nginx/proxy_temp
+            - name: nginx-fastcgi-temp
+              mountPath: /usr/local/openresty/nginx/fastcgi_temp
+            - name: nginx-uwsgi-temp
+              mountPath: /usr/local/openresty/nginx/uwsgi_temp
+            - name: nginx-scgi-temp
+              mountPath: /usr/local/openresty/nginx/scgi_temp
```
danielhoherd: I saw @ianbuss's comment above that was resolved. I actually like the original approach of having a single volume, but agree with his concerns. The best solution for that is to use subPath, which will create a directory in the volume and mount that `volume/subPath` at the `mountPath`
```suggestion
            - name: tmp
              subPath: openresty/nginx/client_body_temp
              mountPath: /usr/local/openresty/nginx/client_body_temp
            - name: tmp
              subPath: openresty/nginx/proxy_temp
              mountPath: /usr/local/openresty/nginx/proxy_temp
            - name: tmp
              subPath: openresty/nginx/fastcgi_temp
              mountPath: /usr/local/openresty/nginx/fastcgi_temp
            - name: tmp
              subPath: openresty/nginx/uwsgi_temp
              mountPath: /usr/local/openresty/nginx/uwsgi_temp
            - name: tmp
              subPath: openresty/nginx/scgi_temp
              mountPath: /usr/local/openresty/nginx/scgi_temp
```
We do not have to do it this way, but it does create far fewer volumes, which seems cleaner.

https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath

Shubham-astro:
> How confident are we that these are the only locations openresty will need write access?

This is where i am picking these volumes from: https://github.com/astronomer/ap-vendor/blob/main/openresty/Dockerfile#L26C1-L32C55

charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
ianbuss: Agree, subPath is The Way

charts/external-es-proxy/templates/external-es-proxy-deployment.yaml
```diff
@@ -149,6 +159,16 @@ spec:
...
+        - name: nginx-client-body-temp
+          emptyDir: {}
+        - name: nginx-proxy-temp
+          emptyDir: {}
+        - name: nginx-fastcgi-temp
+          emptyDir: {}
+        - name: nginx-uwsgi-temp
+          emptyDir: {}
+        - name: nginx-scgi-temp
+          emptyDir: {}
```
danielhoherd: If we use subPath we can remove these.
```suggestion
```

tests/chart_tests/test_external_elasticsearch.py
```diff
@@ -29,6 +29,20 @@ def test_externalelasticsearch_with_secret(self, kube_version):
...
+            {"name": "nginx-uwsgi-temp", "mountPath": "/usr/local/openresty/nginx/uwsgi_temp"},
+            {"name": "nginx-scgi-temp", "mountPath": "/usr/local/openresty/nginx/scgi_temp"},
+        ]
+        expected_nginx_volumes = [
+            {"name": "nginx-client-body-temp", "emptyDir": {}},
+            {"name": "nginx-proxy-temp", "emptyDir": {}},
+            {"name": "nginx-fastcgi-temp", "emptyDir": {}},
+            {"name": "nginx-uwsgi-temp", "emptyDir": {}},
+            {"name": "nginx-scgi-temp", "emptyDir": {}},
+        ]
```
danielhoherd: This can also go away
```suggestion
```

tests/chart_tests/test_external_elasticsearch.py
```diff
@@ -43,8 +57,12 @@ def test_externalelasticsearch_with_secret(self, kube_version):
         assert expected_env == deployment["spec"]["template"]["spec"]["containers"][0]["env"]
         container_mounts = deployment["spec"]["template"]["spec"]["containers"][0]["volumeMounts"]
         assert {"name": "tmp", "mountPath": "/tmp"} in container_mounts
+        for mount in expected_nginx_mounts:
+            assert mount in container_mounts, f"Missing mount {mount}"
         volumes = deployment["spec"]["template"]["spec"]["volumes"]
         assert {"name": "tmp", "emptyDir": {}} in volumes
+        for vol in expected_nginx_volumes:
+            assert vol in volumes, f"Missing volume: {vol}"
```
danielhoherd: And this can go away
```suggestion
```

tests/chart_tests/test_external_elasticsearch.py
```diff
@@ -29,6 +29,20 @@ def test_externalelasticsearch_with_secret(self, kube_version):
             ],
         )
 
+        expected_nginx_mounts = [
+            {"name": "nginx-client-body-temp", "mountPath": "/usr/local/openresty/nginx/client_body_temp"},
+            {"name": "nginx-proxy-temp", "mountPath": "/usr/local/openresty/nginx/proxy_temp"},
+            {"name": "nginx-fastcgi-temp", "mountPath": "/usr/local/openresty/nginx/fastcgi_temp"},
+            {"name": "nginx-uwsgi-temp", "mountPath": "/usr/local/openresty/nginx/uwsgi_temp"},
+            {"name": "nginx-scgi-temp", "mountPath": "/usr/local/openresty/nginx/scgi_temp"},
+        ]
```
danielhoherd, line 43: i think this is the only block that needs to be updated for subPath test fixes

PR #2896, opened 2025-10-07
Add Image updates

Shubham-astro:
## Description

Add image updates in Master and 1.0

## Merging

Master, 1.0

### DIFF
--- a/charts/elasticsearch/values.yaml
+++ b/charts/elasticsearch/values.yaml
@@ -14,7 +14,7 @@ images:
     pullPolicy: IfNotPresent
   init:
     repository: quay.io/astronomer/ap-base # needs root permissions for sysctl changes
-    tag: 3.21.3-3
+    tag: 3.21.3-6
     pullPolicy: IfNotPresent
   curator:
     repository: quay.io/astronomer/ap-curator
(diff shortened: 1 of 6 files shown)


pgvishnuram:
Build is failing

charts/elasticsearch/values.yaml
```diff
@@ -14,7 +14,7 @@ images:
     pullPolicy: IfNotPresent
   init:
     repository: quay.io/astronomer/ap-base # needs root permissions for sysctl changes
-    tag: 3.21.3-3
+    tag: 3.21.3-6
```
pgvishnuram, line 17: if you are changing this you have to then verify our sysctl init container

values.yaml
```diff
@@ -41,7 +41,7 @@ global:
     containerdTolerations: []
     certCopier:
       repository: quay.io/astronomer/ap-base
-      tag: 3.21.3-3
+      tag: 3.21.3-6
```
pgvishnuram, line 44: cert copier run as root - it wont break right

pgvishnuram: I have some custom scripts running and this is a critical job for major customers.

danielhoherd: > I have some custom scripts running

@pgvishnuram can you elaborate on this?

Shubham-astro: @pgvishnuram if cert-copier sets root user in the image then we have no issues, it can run fine.

PR #2901, opened 2025-10-09
Support to enable federation auth token for unified mode

Shubham-astro:
## Description

Add support for `FEDERATION_AUTH_TOKEN` environment variable in Prometheus container for unified plane mode, this is needed since we also support unified mode to be used as a control plane and add DP clusters to it. 

## Related Issues

https://github.com/astronomer/issues/issues/7965

## Testing
- Added assertions in `test_prometheus_sts_defaults` to verify the `FEDERATION_AUTH_TOKEN` environment variable is present by default.

## Merging

Master, 1.0


charts/prometheus/templates/prometheus-statefulset.yaml
pgvishnuram, line 244: @Shubham-astro this one

PR #2902, opened 2025-10-09
support custom ca for prometheus

pgvishnuram:
## Description

support custom ca for prometheus

## Related Issues

- https://github.com/astronomer/issues/issues/7965

## Testing

Adds CA support for prometheus sts when global.privateCaCerts added this certs will be mounted in prometheus container as well

## Merging

### DIFF
--- a/charts/prometheus/templates/prometheus-statefulset.yaml
+++ b/charts/prometheus/templates/prometheus-statefulset.yaml
@@ -164,7 +164,7 @@ spec:
             - name: DEPLOYMENT_TABLE_NAME
               value: Deployment
             - name: DATABASE_NAME
-              value: {{ $dbName }}_houston
+              value: {{ $dbName }}
             - name: FILESD_FILE_PATH
               value: /prometheusreloader/airflow
             - name: ENABLE_DEPLOYMENT_SCRAPING


charts/prometheus/templates/prometheus-statefulset.yaml
```diff
@@ -2,6 +2,7 @@
 ## Prometheus StatefulSet ##
 ############################
 {{- $useClusterRoles := and .Values.global.rbacEnabled .Values.global.clusterRoles (not (eq .Values.rbac.role.kind "Role") ) }}
+{{- $releaseFormatConversion := .Release.Name | replace "-" "_"  }}
```
pgvishnuram: @ianbuss @Shubham-astro @rishkarajgi is this name okay

ianbuss: maybe `releaseDBName`?

rishkarajgi:
```suggestion
{{- $sanitizedReleaseName := .Release.Name | replace "-" "_"  }}
```

charts/prometheus/templates/prometheus-statefulset.yaml
```diff
@@ -2,6 +2,7 @@
 ## Prometheus StatefulSet ##
 ############################
 {{- $useClusterRoles := and .Values.global.rbacEnabled .Values.global.clusterRoles (not (eq .Values.rbac.role.kind "Role") ) }}
+{{- $sanitizedReleaseName := .Release.Name | replace "-" "_"  }}
```
danielhoherd: Since this is only used as the db name, we should change this to something like:

```suggestion
{{- $dbName := (printf "%s_houston" (.Release.Name | replace "-" "_")) }}
```

pgvishnuram: applied the change - it should be good now

pgvishnuram: seems github has some issues - my new changes are not reflecting here

PR #2908, opened 2025-10-10
readOnlyRootFilesystem for auth-proxy

danielhoherd:
## Description

Add readOnlyRootFilesystem to auth-sidecar

## Related Issues

https://github.com/astronomer/issues/issues/7443

## Testing

This needs to be tested in a cluster that has auth-sidecar, like ARO.

## Merging

Merge only to 1.0.

### DIFF
--- a/charts/nginx/templates/controlplane/nginx-cp-configmap.yaml
+++ b/charts/nginx/templates/controlplane/nginx-cp-configmap.yaml
@@ -15,6 +15,7 @@ metadata:
 data:
   allow-snippet-annotations: "{{ .Values.allowSnippetAnnotations }}"
   add-headers: {{ .Release.Namespace }}/{{ template "nginx.fullname" . }}-ingress-controller-headers
+  annotations-risk-level: Critical
   proxy-add-original-uri-header: "true"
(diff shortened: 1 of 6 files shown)


charts/alertmanager/templates/alertmanager-statefulset.yaml
```diff
@@ -142,7 +142,9 @@ spec:
         {{- if .Values.global.authSidecar.enabled  }}
         - name: auth-proxy
           image: {{ include "authSidecar.image" . }}
-          securityContext: {{ toYaml .Values.securityContext| nindent 12 }}
+          securityContext:
+            {{- $required := dict "readOnlyRootFilesystem" true }}
+            {{- toYaml (merge $required .Values.securityContext ) | nindent 12 }}
```
pgvishnuram, line 147: https://github.com/astronomer/astronomer/blob/7443-readOnlyRootFilesystem-auth-sidecar/charts/alertmanager/templates/alertmanager-auth-sidecar-configmap.yaml

-> No change required here ??

danielhoherd: We chatted about this in slack and it ended up that we didn't need to do anything.

PR #2917, opened 2025-10-13
sync secret name from astronomer control plane

pgvishnuram:
## Description

sync secret name from astronomer control plane
This PR fixes an issue where control plane secret release name is different and dataplane jwt secret name is different causing a sync issue between the secret exanchanged between the airflow2 deployments and control passed default values

## Related Issues
- https://github.com/astronomer/issues/issues/8010
## Testing

QA should able to verify the houston jwt signed secret will now have the same secret name as control plane

## Merging

### DIFF
--- a/tests/chart_tests/test_astronomer_commander_hook_job.py
+++ b/tests/chart_tests/test_astronomer_commander_hook_job.py
@@ -62,7 +62,7 @@ class TestCommanderJWKSHookJob:
 
         env_vars = get_env_vars_dict(container["env"])
         assert env_vars["CONTROL_PLANE_ENDPOINT"] == "https://houston.example.com"
-        assert env_vars["SECRET_NAME"] == "release-name-houston-jwt-signing-certificate"
+        assert env_vars["SECRET_NAME"] == "jwt-signing-certificate"
         assert env_vars["RETRY_ATTEMPTS"] == "2"
         assert env_vars["RETRY_DELAY"] == "10"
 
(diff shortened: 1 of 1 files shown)


charts/astronomer/files/commander-jwks.py
```diff
@@ -187,13 +189,18 @@ def main():
...
-    logger.info(f"  Release Name: {release_name}")
 
     try:
         jwks_data = fetch_jwks_from_endpoint(control_plane_endpoint, retry_attempts, retry_delay)
         signed_public_cert = get_base64_pem_from_jwks(jwks_data)
+        pem_cert_bytes = base64.b64decode(signed_public_cert)
+        cert = x509.load_pem_x509_certificate(pem_cert_bytes, default_backend())
+        release_name = cert.subject.get_attributes_for_oid(x509.oid.NameOID.COMMON_NAME)[0].value
+        logger.info(f"  Release Name: {release_name}")
+        secret_name = f"{release_name}-{secret_name_suffix}"
```
ianbuss, line 202: This needs to have no `.`, is that always guaranteed from the COMMON_NAME of the JWKS cert?

ianbuss: We are in control of this cert and the common name, so is safe to assume no "." in this case. See https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/houston/api/houston-jwt-certificate-secret.yaml#L6. (From an offline conversation with @pgvishnuram).

PR #2919, opened 2025-10-13
7246 put grafana back

danielhoherd:
## Description

Revert removal of grafana and add features that are required for 1.0

## Related Issues

https://github.com/astronomer/issues/issues/7246

## Testing

This definitely needs testing to ensure it works in 1.0

## Merging

This is only needed for 1.0. Grafana was not removed from prior versions.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -56,7 +56,7 @@ def dp_nginx(k8s_core_v1_client):
 
 
 @pytest.fixture(scope="function")
-def grafana(core_v1_client):
+def grafana(k8s_core_v1_client):
     """Fixture for accessing the grafana pod."""
 
     kubeconfig_file = str(KUBECONFIG_DIR / TEST_SCENARIO)


values.yaml
```diff
@@ -434,6 +439,41 @@ nginx:
...
+    # - name: GF_SMTP_HOST
+    #   value: smtp.astronomer.io
+    # - name: GF_SMTP_USER
+    #   value: grafana
+    # - name: GF_SMTP_PASSWORD
+    #   valueFrom:
+    #     secretKeyRef:
+    #       name: grafana-smtp-secret
+    #       key: smtp_password
+
```
Simpcyclassy, line 475: Why is this commented?

danielhoherd: This is a common way to show some documentation about how the feature can be customized.

charts/grafana/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+  #   value: "true"
+  # - name: GF_SMTP_HOST
+  #   value: smtp.astronomer.io
+  # - name: GF_SMTP_USER
+  #   value: grafana
+  # - name: GF_SMTP_PASSWORD
+  #   valueFrom:
+  #     secretKeyRef:
+  #       name: grafana-smtp-secret
+  #       key: smtp_password
```
Simpcyclassy, line 84: Commented out here as well.

charts/grafana/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+ports:
+  http: 3000
+
+resources: {}
+#  limits:
+#   cpu: 100m
+#   memory: 128Mi
+#  requests:
+#   cpu: 100m
+#   memory: 128Mi
```
Simpcyclassy, line 51: Here as well

PR #2928, opened 2025-10-15
fix airflow filesd regex target path

pgvishnuram:
## Description

fix airflow filesd regex target path

## Related Issues

- https://github.com/astronomer/issues/issues/8023

## Testing

QA should now see scrape targets for no cluster role

## Merging


charts/prometheus/templates/prometheus-statefulset.yaml
```diff
@@ -168,7 +170,7 @@ spec:
             - name: FILESD_FILE_PATH
               value: /prometheusreloader/airflow
             - name: ENABLE_DEPLOYMENT_SCRAPING
-              value: {{ if $useClusterRoles }}"True"{{ else }}"False"{{ end }}
```
Shubham-astro, line 171: So now deployment scraping only runs for unified mode and for rest we use cluster mode??

pgvishnuram: we wanted to enable scrapping only for unified mode and no cluster role

PR #2938, opened 2025-10-24
Respect global.clusterRoles in Vector ClusterRole and ClusterRoleBinding templates

Simpcyclassy:
## Description

Update Vector ClusterRole and ClusterRoleBinding templates to honor `global.clusterRoles` value

This change ensures that the Vector ClusterRole and ClusterRoleBinding are only rendered if `global.clusterRoles` is true, in addition to the existing checks for RBAC and plane mode. This prevents cluster-scoped resources from being created when not permitted, enabling installations in RBAC-restricted environments.

### Impact
- Prevents install failures on clusters where users do not have permission to create cluster-scoped resources.
- Addresses issue encountered in unified mode with clusterRoles: false.

## Related Issues

Related astronomer/issues#8065

## Testing

Added test coverage for the vector ClusterRole template to verify it only renders when rbacEnabled, clusterRoles, and plane mode (data/unified) conditions are all met.

## Merging

1.0.1

### DIFF
--- a/tests/chart_tests/test_nginx.py
+++ b/tests/chart_tests/test_nginx.py
@@ -495,13 +495,7 @@ class TestNginxClusterRoles:
     def test_nginx_cp_clusterrole_rbac_enabled(self, kube_version):
         """Test that helm renders nginx control plane ClusterRole when all conditions are met."""
         # Test control plane mode
-        values = {
-            "global": {
-                "rbacEnabled": True,
-                "clusterRoles": True,
-                "plane": {"mode": "control"}
-            }
-        }
+        values = {"global": {"rbacEnabled": True, "clusterRoles": True, "plane": {"mode": "control"}}}
         docs = render_chart(
             kube_version=kube_version,
             values=values,
@@ -516,7 +510,7 @@ class TestNginxClusterRoles:
         assert doc["metadata"]["labels"]["tier"] == "nginx"
         assert doc["metadata"]["labels"]["plane"] == "control"
         assert len(doc["rules"]) > 0
-        
+
         # Test unified plane mode (should also work for control plane template)
         values["global"]["plane"]["mode"] = "unified"
         docs = render_chart(
@@ -530,13 +524,7 @@ class TestNginxClusterRoles:
 
     def test_nginx_dp_clusterrole_rbac_enabled(self, kube_version):
         """Test that helm renders nginx data plane ClusterRole when all conditions are met."""
-        values = {
-            "global": {
-                "rbacEnabled": True,
-                "clusterRoles": True,
-                "plane": {"mode": "data"}
-            }
-        }
+        values = {"global": {"rbacEnabled": True, "clusterRoles": True, "plane": {"mode": "data"}}}
         docs = render_chart(
             kube_version=kube_version,
             values=values,
@@ -554,13 +542,7 @@ class TestNginxClusterRoles:
     def test_nginx_clusterrole_disabled_conditions(self, kube_version):
         """Test that nginx ClusterRoles are not rendered when required conditions are not met."""
         # Test with rbacEnabled=False (control plane)
-        values = {
-            "global": {
-                "rbacEnabled": False,
-                "clusterRoles": True,
-                "plane": {"mode": "control"}
-            }
-        }
+        values = {"global": {"rbacEnabled": False, "clusterRoles": True, "plane": {"mode": "control"}}}
         docs = render_chart(
             kube_version=kube_version,
             values=values,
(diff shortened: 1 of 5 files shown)


tests/chart_tests/test_nginx.py
```diff
@@ -484,3 +485,112 @@ def test_nginx_backend_overrides():
...
+        # Test unified plane mode (should also work for control plane template)
+        values["global"]["plane"]["mode"] = "unified"
+        docs = render_chart(
+            kube_version=kube_version,
+            values=values,
+            show_only=["charts/nginx/templates/controlplane/nginx-cp-role.yaml"],
+        )
+        assert len(docs) == 1
+        assert docs[0]["kind"] == "ClusterRole"
+        assert docs[0]["metadata"]["labels"]["plane"] == "unified"
```
danielhoherd, line 523: Lines 514-523 should be a different test case. We can parametrize the case from 495 with `plane_mode` with something like https://github.com/astronomer/astronomer/blob/9a232ebb6d/tests/chart_tests/test_grafana.py#L10-L19 to make this only one test case that has parameter values for the few things that are different between unified and control.

danielhoherd: We can probably include the test from 525 in this too.

tests/chart_tests/test_nginx.py
```diff
@@ -484,3 +485,112 @@ def test_nginx_backend_overrides():
...
+        assert len(docs) == 0
+
+        # Test data plane template with clusterRoles=False
+        values = {"global": {"rbacEnabled": True, "clusterRoles": False, "plane": {"mode": "data"}}}
+        docs = render_chart(
+            kube_version=kube_version,
+            values=values,
+            show_only=["charts/nginx/templates/dataplane/nginx-dp-role.yaml"],
+        )
+        assert len(docs) == 0
```
danielhoherd, line 596: These should be broken up into multiple test cases, some of which can probably be parametrized.

tests/chart_tests/test_vector.py
```diff
@@ -111,6 +111,72 @@ def test_vector_daemonset_private_ca_certificates(self, kube_version):
...
+
+        # Test unified plane mode
+        values["global"]["plane"]["mode"] = "unified"
+        docs = render_chart(
+            kube_version=kube_version,
+            values=values,
+            show_only=["charts/vector/templates/vector-clusterrole.yaml"],
+        )
+        assert len(docs) == 1
+        assert docs[0]["kind"] == "ClusterRole"
```
danielhoherd, line 149: This should be broken up into multiple test cases, or refactored for one `render_chart()`

tests/chart_tests/test_vector.py
```diff
@@ -111,6 +111,72 @@ def test_vector_daemonset_private_ca_certificates(self, kube_version):
...
+        assert len(docs) == 0
+
+        # Test with control plane mode (should not render)
+        values = {"global": {"rbacEnabled": True, "clusterRoles": True, "plane": {"mode": "control"}}}
+        docs = render_chart(
+            kube_version=kube_version,
+            values=values,
+            show_only=["charts/vector/templates/vector-clusterrole.yaml"],
+        )
+        assert len(docs) == 0
```
danielhoherd, line 178: This should be refactored into multiple test cases or refactored for parametrization.

tests/chart_tests/test_vector.py
```diff
@@ -111,6 +111,72 @@ def test_vector_daemonset_private_ca_certificates(self, kube_version):
...
+        docs = render_chart(
+            kube_version=kube_version,
+            values=values,
+            show_only=["charts/vector/templates/vector-clusterrole.yaml"],
+        )
+        assert len(docs) == 1
+        assert docs[0]["kind"] == "ClusterRole"
+
+    def test_vector_clusterrole_disabled_conditions(self, kube_version):
+        """Test that vector ClusterRole is not rendered when required conditions are not met."""
```
danielhoherd, line 152: This docstring is ambiguous. We need to know exactly what is being tested.

tests/chart_tests/test_vector.py
```diff
@@ -111,6 +111,72 @@ def test_vector_daemonset_private_ca_certificates(self, kube_version):
         ]
         assert pod_spec["containers"][0]["volumeMounts"] == volume_mounts
 
+    def test_vector_clusterrole_rbac_enabled(self, kube_version):
+        """Test that helm renders a good ClusterRole template for vector when all conditions are met."""
```
danielhoherd, line 115: This docstring is ambiguous. We need to know exactly what is being tested.

tests/chart_tests/test_nginx.py
```diff
@@ -484,3 +485,112 @@ def test_nginx_backend_overrides():
...
+        assert len(docs) == 1
+        doc = docs[0]
+        assert doc["kind"] == "ClusterRole"
+        assert doc["apiVersion"] == "rbac.authorization.k8s.io/v1"
+        assert doc["metadata"]["name"] == "release-name-dp-nginx"
+        assert doc["metadata"]["labels"]["tier"] == "nginx"
+        assert len(doc["rules"]) > 0
+
+    def test_nginx_clusterrole_disabled_conditions(self, kube_version):
+        """Test that nginx ClusterRoles are not rendered when required conditions are not met."""
```
danielhoherd, line 543: This docstring is ambiguous. We need to know exactly what is being tested. Same for anything else that says "when required conditions are not met." We should be specific and say "When ____ is enabled and ____ is set to ______" or something.

PR #2939, opened 2025-10-24
Add FSgroup values for registry and update blakcbox exporter image

Shubham-astro:
## Description

This PR adds FSggroup for registry component and update blakcbox exporter image.

## Related Issues

Related 
- astronomer/issues#8046
- https://github.com/astronomer/issues/issues/7705

## Testing

- Added tests
## Merging

0.37

### DIFF
--- a/charts/astronomer/values.yaml
+++ b/charts/astronomer/values.yaml
@@ -448,8 +448,6 @@ registry:
 
   podSecurityContext:
     fsGroup: 1000
-    runAsGroup: 1000
-    runAsUser: 1000
   extraEnv: []
 
   serviceAccount:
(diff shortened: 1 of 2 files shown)


charts/astronomer/values.yaml
```diff
@@ -446,7 +446,10 @@ registry:
     # periodSeconds: 10
     # timeoutSeconds: 5
 
-  podSecurityContext: {}
+  podSecurityContext:
+    fsGroup: 1000
+    runAsGroup: 1000
+    runAsUser: 1000
```
pgvishnuram: do we need these two changes ??

Shubham-astro: in 0.37 i think we don't, LEt me remove

rishkarajgi (approved):

PR #2951, opened 2025-11-03
Change CP and Unified mode nginx svc back to what it was in 0.37

danielhoherd:
## Description

Change CP and Unified mode nginx svc back to what it was before 1.0. This avoids creating a new load balancer object which switches the application to a new ip address.
- rename svc from `release-name-cp-nginx` to `release-name-nginx`
- Change labels of svc from `cp-ingress-controller` to `ingress-controller` because this field is immutable and upgrades would fail if labels changed:
    ```
    upgrade.go:487: 2025-11-04 15:37:15.27001 -0500 EST m=+119.700245668 [debug] warning: Upgrade "astronomer" failed: cannot patch "astronomer-nginx" with kind Deployment: Deployment.apps "astronomer-nginx" is invalid: spec.selector: Invalid value: {"matchLabels":{"component":"cp-ingress-controller","release":"astronomer","tier":"nginx"}}: field is immutable
    Error: UPGRADE FAILED: cannot patch "astronomer-nginx" with kind Deployment: Deployment.apps "astronomer-nginx" is invalid: spec.selector: Invalid value: {"matchLabels":{"component":"cp-ingress-controller","release":"astronomer","tier":"nginx"}}: field is immutable
    helm.go:92: 2025-11-04 15:37:15.299144 -0500 EST m=+119.729379876 [debug] cannot patch "astronomer-nginx" with kind Deployment: Deployment.apps "astronomer-nginx" is invalid: spec.selector: Invalid value: {"matchLabels":{"component":"cp-ingress-controller","release":"astronomer","tier":"nginx"}}: field is immutable
    ```

## Related Issues

<https://github.com/astronomer/issues/issues/8036>

## Testing

We should have QA do an upgrade test from 0.37 to 1.0 and verify that the `EXTERNAL-IP` address of the `{DEPLOYMENT_NAME}-nginx` svc does not change.

## Merging

This is only for 1.x

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -435,7 +435,7 @@ workflows:
                 - quay.io/astronomer/ap-postgres-exporter:0.17.1-1
                 - quay.io/astronomer/ap-postgresql:17.6.0
                 - quay.io/astronomer/ap-prometheus:3.5.0-2
-                - quay.io/astronomer/ap-redis:8.0.3
+                - quay.io/astronomer/ap-redis:8.0.3-1
                 - quay.io/astronomer/ap-registry:3.0.0-1
                 - quay.io/astronomer/ap-statsd-exporter:0.28.0-3
                 - quay.io/astronomer/ap-vector:0.47.0-5
--- a/values.yaml
+++ b/values.yaml
@@ -301,7 +301,7 @@ global:
         tag: 0.28.0-3
       redis:
         repository: quay.io/astronomer/ap-redis
-        tag: 8.0.3
+        tag: 8.0.3-1
       pgbouncer:
         repository: quay.io/astronomer/ap-pgbouncer
         tag: 1.24.1-2


charts/nginx/templates/controlplane/nginx-cp-deployment.yaml
```diff
@@ -5,10 +5,10 @@
 kind: Deployment
 apiVersion: apps/v1
 metadata:
-  name: {{ .Release.Name }}-cp-{{ template "nginx.name" . }}
```
karankhanchandani, line 8: CMMIW, but we should have a semver check here right? and check if it is unified mode..

pgvishnuram: @danielhoherd instead of changing the component name to ingress-controller can we just change the service name old one - that would save lots of changes from our end

Shubham-astro: > @danielhoherd instead of changing the component name to ingress-controller can we just change the service name old one - that would save lots of changes from our end

Yeah, with component label name change we may need to make changes in houston, airflow-chart and also run network policies again

danielhoherd: As mentioned in the PR description, we cannot `PATCH` the labels, so the component name cannot change between 0.37 and 1.0 if we keep the same ingress object.

PR #2965, opened 2025-11-12
Sync registry-auth-secret changes across Prometheus and Registry StatefulSets via checksum annotations

Shubham-astro:
## Description

This PR adds checksum annotations to the Astronomer registry StatefulSet and Prometheus StatefulSet pod templates to automatically trigger pod restarts when the registry-auth-secret.yaml template is modified.

The registry-auth-secret.yaml template was moved from astronomer/charts/astronomer/templates/registry/ to the global astronomer/templates/ directory.

## Testing

- Tested the rendering of secret locally:
Screenshot 2025-11-12 at 6 44 22 PM

## Related Issues

Related astronomer/issues#8107

## Merging

Master, 1.0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -429,9 +429,9 @@ workflows:
                 - quay.io/astronomer/ap-nginx-es:1.29.2
                 - quay.io/astronomer/ap-nginx:1.12.6
                 - quay.io/astronomer/ap-openresty:1.27.1-7
-                - quay.io/astronomer/ap-pgbouncer-exporter:0.19.0-1
-                - quay.io/astronomer/ap-pgbouncer-krb:1.17.0-20
-                - quay.io/astronomer/ap-pgbouncer:1.24.1-2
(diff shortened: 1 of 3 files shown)


templates/_helpers.tpl
```diff
@@ -81,5 +81,5 @@ proxy_pass https://houston.{{ .Values.global.baseDomain }}/v1/elasticsearch;
 {{- end }}
 
 {{ define "registry.authHeaderSecret" -}}
-{{ default (printf "%s-registry-auth-key" .Release.Name) .Values.global.authHeaderSecretName }}
```
pgvishnuram, line 84: are we changing the name

Shubham-astro: Actuallyi was testing something, to see what components currently depend on this and also to remove some extra helper definitions and i ended up removing from all the definitions, if there is some norm to add key I can add it back now.

danielhoherd: so.... are we changing the name?

danielhoherd (approved):
These changes all look good.

PR #2992, opened 2025-12-10
Remove conf-copier Init Container from PostgreSQL StatefulSet

Shubham-astro:
## Description

This PR removes the conf-copier init container from the PostgreSQL StatefulSet template. The init container is incompatible with the Chainguard-based PostgreSQL image and is no longer necessary for proper PostgreSQL initialization.

This PR also adds the `NET_BIND_SERVICE` Linux capability to the nginx ingress controller's security context, allowing it to bind to privileged ports (80/443) while running as a non-root user.

## Related Issues

Related 
- https://github.com/astronomer/issues/issues/8141
- https://github.com/astronomer/issues/issues/8131

## Testing

- PostgreSQL and nginx pod now starts successfully on first deployment.
Screenshot 2025-12-10 at 5 20 04 PM

Screenshot 2025-12-10 at 6 26 51 PM

- Database initializes properly without the conf-copier step and nginx pod also stabilised. 

- All PostgreSQL and nginx functionality remains intact, was able to create a new deployment.
Screenshot 2025-12-10 at 5 20 20 PM

## Merging

Master, 1.0

### DIFF
--- a/charts/nginx/templates/controlplane/nginx-cp-deployment.yaml
+++ b/charts/nginx/templates/controlplane/nginx-cp-deployment.yaml
@@ -72,6 +72,7 @@ spec:
       containers:
         - name: nginx
           securityContext:
+            readOnlyRootFilesystem: true
             {{ toYaml .Values.securityContext | nindent 12 }}
           image: {{ template "nginx.image" . }}
           imagePullPolicy: {{ .Values.images.nginx.pullPolicy }}
--- a/charts/nginx/templates/dataplane/nginx-dp-deployment.yaml
+++ b/charts/nginx/templates/dataplane/nginx-dp-deployment.yaml
@@ -70,6 +70,7 @@ spec:
       containers:
         - name: nginx
           securityContext:
+            readOnlyRootFilesystem: true
             {{ toYaml .Values.securityContext| nindent 12 }}
           image: {{ template "nginx.image" . }}
           imagePullPolicy: {{ .Values.images.nginx.pullPolicy }}


charts/nginx/templates/controlplane/nginx-cp-deployment.yaml
```diff
@@ -72,7 +72,6 @@ spec:
       containers:
         - name: nginx
           securityContext:
-            readOnlyRootFilesystem: true
```
Simpcyclassy, line 75: Is removing `readOnlyRootFilesystem` part of removing conf-copier Init Container from PostgreSQL StatefulSet OR we want to handle this from houston

charts/nginx/templates/dataplane/nginx-dp-deployment.yaml
```diff
@@ -70,7 +70,6 @@ spec:
       containers:
         - name: nginx
           securityContext:
-            readOnlyRootFilesystem: true
```
Simpcyclassy, line 73: Same question here

charts/nginx/templates/controlplane/nginx-cp-deployment.yaml
Shubham-astro: it is added, @Simpcyclassy , it is just that it is added: `{{ toYaml .Values.securityContext | nindent 12 }}` only to make sure the template is clean

charts/nginx/templates/dataplane/nginx-dp-deployment.yaml
Shubham-astro: it is added, @Simpcyclassy , it is just that it is added: `{{ toYaml .Values.securityContext | nindent 12 }}` only to make sure the template is clean

charts/nginx/values.yaml
```diff
@@ -17,11 +17,14 @@ images:
     pullPolicy: IfNotPresent
 
 securityContext:
+  readOnlyRootFilesystem: true
```
pgvishnuram: @Shubham-astro adding here might allows users to override - is that what we indended to do

charts/postgresql/templates/statefulset-slaves.yaml
```diff
@@ -96,8 +96,6 @@ spec:
           runAsNonRoot: {{ .Values.securityContext.runAsNonRoot }}
         {{- end }}
         env:
-        - name: BITNAMI_DEBUG
-          value: {{ ternary "true" "false" .Values.image.debug | quote }}
```
pgvishnuram, line 100: what naming convention chainguard based images have here to enable debug mode

pgvishnuram: also do master slave is supported by chainguard based postgres image

charts/nginx/values.yaml
danielhoherd: We do not want to allow users to override anything in astronomer/astronomer.

charts/nginx/templates/controlplane/nginx-cp-deployment.yaml
danielhoherd: This should be:
```
{{- $required := dict "readOnlyRootFilesystem" true }}
{{- toYaml (merge $required .Values.securityContext ) | nindent 12 }}
```
which will let users customize these values, while always enforcing RORFS.

charts/postgresql/templates/statefulset-slaves.yaml
Shubham-astro: @pgvishnuram I am not sure about this, Let me check this. with chainguard.

charts/nginx/values.yaml
Shubham-astro: ok, let me move it back to where it was.

Shubham-astro: This was moved back to it's original place

charts/nginx/templates/controlplane/nginx-cp-deployment.yaml
Shubham-astro: i have moved this rorfs back here..

PR #2426, opened 2024-11-22
Rip out cli-install

danielhoherd:
## Description

Rip out the cli-install feature.

## Related Issues

https://github.com/astronomer/issues/issues/6611

## Testing

unit-tests pass. We don't need any specific manual testing on this feature.

## Merging

This is only meant for release-0.37

PR #2996, opened 2025-12-12
fix registry ingress baseDomain generation

pgvishnuram:
## Description

This PR fixes registry URL for dataplane setup which was incorrectly generated before

## Related Issues

https://github.com/astronomer/issues/issues/8033

## Testing

QA should be able to validate below format in the ingress hosts

```
unified mode : registry.<baseDomain>.<domain-name>
data mode: registry.<dataplane-id>.<baseDomain>.<domain-name>
```

## Merging

### DIFF
--- a/tests/chart_tests/test_ingress.py
+++ b/tests/chart_tests/test_ingress.py
@@ -202,6 +202,9 @@ class TestIngress:
         for expected_host in expected_hosts:
             assert expected_host in tls_hosts, f"Expected {expected_host} in TLS hosts for {mode} mode"
 
+        # install.example.com should never be in TLS #6611
+        assert "install.example.com" not in tls_hosts
+
         # Check nginx configuration snippet (should be present in control/unified, not needed in data)
         annotations = doc["metadata"]["annotations"]
         if expected_astro_ui:


charts/astronomer/templates/ingress.yaml
```diff
@@ -2,7 +2,7 @@
 ## Astronomer Ingress ##
 ########################
 {{- if and .Values.global.baseDomain ( not .Values.global.enablePerHostIngress ) }}
-{{- $planeDataMode := or (eq .Values.global.plane.mode "data") (eq .Values.global.plane.mode "unified") }}
+{{- $planeDataMode := (eq .Values.global.plane.mode "data") }}
```
Shubham-astro: why are we removing unified mode.?

charts/astronomer/templates/ingress.yaml
```diff
@@ -65,8 +64,7 @@ spec:
               port:
                 name: astro-ui-http
 {{- end }}
-{{- if $planeDataMode }}
-  - host: registry.{{ .Values.global.baseDomain }}
+  - host: {{ if $planeDataMode }}registry.{{ .Values.global.plane.domainPrefix }}.{{ .Values.global.baseDomain }}{{ else }}registry.{{ .Values.global.baseDomain }}{{ end }}
```
Shubham-astro: ok, we are including it right here.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -563,3 +563,11 @@ https://houston.{{ .Values.global.baseDomain }}
 http://{{ .Release.Name }}-houston.{{ .Release.Namespace}}:{{ .Values.ports.houstonHTTP }}
 {{- end }}
 {{- end -}}
+
+{{- define "registry.ingressURL" -}}
+{{- if (eq .Values.global.plane.mode "data") -}}
+registry.{{ .Values.global.plane.domainPrefix }}.{{ .Values.global.baseDomain }}
+{{- else -}}
+registry.{{ .Values.global.baseDomain }}
```
Shubham-astro, line 571: this will get rendered for control mode as well right.?? Do we want that.?

Shubham-astro:
@pgvishnuram, since we've had issues with component rendering issues, Let's also add a test to confirm that registry deployment is not rendered in the plane it shouldn't render (control plane). what do you think.??

tests/chart_tests/test_ingress.py
```diff
@@ -204,9 +202,6 @@ def test_ingress_per_plane_mode(
         for expected_host in expected_hosts:
             assert expected_host in tls_hosts, f"Expected {expected_host} in TLS hosts for {mode} mode"
 
-        # install.example.com should always be in TLS
-        assert "install.example.com" in tls_hosts
-
```
danielhoherd: Wow. Is this test even running? How was this passing?

danielhoherd: It looks like some code was missed in https://github.com/astronomer/astronomer/pull/2426

https://github.com/astronomer/astronomer/blob/cf814dee51/charts/astronomer/templates/ingress.yaml#L43

danielhoherd: I'll fix this in a separate PR.

pgvishnuram: its not there daniel - i rechecked the code

pgvishnuram: this PR removes them from the master branch after it gets merged

danielhoherd: This is the bigger cleanup I was referring to https://github.com/astronomer/astronomer/pull/3001

PR #3000, opened 2025-12-12
PINF-972 - extend global labels support for astronomer components

pgvishnuram:
## Description

extend global labels support for astronomer components

## Related Issues

- https://github.com/astronomer/issues/issues/8213

## Testing

setting  global.labels now should be applied to all astronomer platform objects

## Merging


danielhoherd (changes requested):
We need a test that finds all pods and ensures that these labels have been added to them.

danielhoherd (changes requested):
The current state of this pr applies `podLabels` to items that are not pods. This is unintuitive, and makes an assumption that those things should always align. I don't think we can make that assumption. I think we should introduce `ingressLabels` for this.

The parent ticket has no description, and no labels saying what release this is for. We need to know what is driving this work, what its priority is, when we are expected to release it, what QA needs to know about it, what the docs team needs to know about it, etc..

rishkarajgi (changes requested):
*(superseded)*

tests/chart_tests/test_ingress.py
```diff
@@ -34,14 +36,26 @@ def test_basic_ingress(self, kube_version):
...
 
+    def test_basic_ingress_with_labels(self, kube_version):
+        docs = render_chart(
+            kube_version=kube_version,
+            show_only=["charts/astronomer/templates/ingress.yaml"],
+            values={"global": {"podLabels": self.labels}},
+        )
+
+        assert len(docs) == 1
+
```
rishkarajgi, line 47: This is passing `podLabels` instead of `labels` — so this test isn't actually exercising the new `global.labels` feature at all. The template uses `.Values.global.labels` but the test renders with `.Values.global.podLabels`, which is a completely separate existing field. The feature has no real test coverage as written.

tests/chart_tests/test_ingress.py
```diff
@@ -55,12 +69,14 @@ def test_astro_ui_per_host_ingress(self, mode, expected, kube_version):
```
rishkarajgi, line 54: Same issue here — `podLabels` should be `labels` to match the new value being tested.

tests/chart_tests/test_ingress.py
```diff
@@ -83,14 +99,15 @@ def test_registry_per_host_ingress(self, kube_version):
```
rishkarajgi, line 82: And here too. All three modified tests need `labels` instead of `podLabels`.

templates/_helpers.tpl
```diff
@@ -72,6 +72,12 @@ imagePullSecrets:
 {{- end }}
 {{- end }}
 
+{{- define "global.labels" -}}
+{{- if .Values.global.labels }}
+{{- toYaml .Values.global.labels }}
+{{- end }}
+{{- end }}
+
 {{- define "houston-proxy" -}}
```
rishkarajgi, line 173: If a user sets a key in `global.labels` that already exists in the template (e.g. `release`, `heritage`, `chart`), the YAML will have duplicate keys — Kubernetes will reject the manifest with a confusing error. Worth calling this out in the `values.yaml` comment, or documenting that `global.labels` keys must not conflict with the standard Helm labels already applied.

values.yaml
```diff
@@ -18,6 +18,9 @@ global:
   # Labels to add to every pod
   podLabels: {}
 
+  # Labels to add on all components
+  labels: {}
+
   # Name of secret containing TLS certificate
   tlsSecret: astronomer-tls
```
rishkarajgi, line 77: The comment `# Labels to add on all components` is a bit broad — this only applies to ingress resources right now, not deployments, services, configmaps, etc. Either scope the comment down (`# Labels to add to all ingress resources`) or expand the implementation to cover other resource types.

rishkarajgi (changes requested):
Adds a `global.labels` Helm value and helper that propagates custom labels onto all Astronomer ingress resources. The concept is straightforward but the tests have a critical bug — see inline comments.

**Verdict:** ⚠️ Needs changes — the tests all pass `global.podLabels` instead of `global.labels`, so the new feature is effectively untested.

Simpcyclassy (commented):
Extends global label support to all Astronomer platform objects. The approach of using `global.labels` is consistent with how other global values (annotations, node selectors, tolerations) are handled in this chart.

A couple of things to verify before merge:

Are labels being applied to resources that should not receive arbitrary labels? Specifically, check that labels are not applied to PVC templates or StatefulSet selectors, where adding labels after initial deploy would break an in-place upgrade. Kubernetes does not allow modifying selector labels on existing StatefulSets — if any selector labels derive from `global.labels`, that is an upgrade-blocking change.

The PR description says to merge to master and release-1.0 but does not address whether 0.x branches are affected. If there are customers on 0.37.x who would benefit from this, add cherry-pick targets.

The testing section says "setting global.labels now should be applied" but does not describe how this was verified. A chart test in `tests/chart_tests/` covering that labels propagate to the key resource types would make this reviewable with more confidence.

PR #3003, opened 2025-12-12
8200 namespace labels

danielhoherd:
## Description

Add namespaceLabels support for commander and namespacePools.

- Add `namespaceLabels` data structure to `commander-metadata.yaml`.
- Add `labels` to namespaces created by `namespacePools.enabled: true` && `namespacePools.namespaces.create: true`.
- Enable `COMMANDER_MANAGE_NAMESPACE_RESOURCE` based off of global.rbacEnabled.
- Rename tests to be more accurate and specific about what they are testing.
- Improve some existing test code syntaxes and structure.
- Update commander to a version that will enforce the labels.

## Related Issues

- <https://github.com/astronomer/issues/issues/7455>

## Testing

This will need to be tested by QA. Specifically, tests that add `pod-security.kubernetes.io/enforce: baseline` should be run.

## Merging

This is only for 1.0.

Merging this now should be 100% safe. However, only namespacePools features will work until commander has been changed to apply the namespaceLabels in its metadata.yaml.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -408,7 +408,7 @@ workflows:
                 - quay.io/astronomer/ap-auth-sidecar:1.29.2
                 - quay.io/astronomer/ap-awsesproxy:1.5.0-20
                 - quay.io/astronomer/ap-base:3.21.3-6
-                - quay.io/astronomer/ap-commander:1.0.23
+                - quay.io/astronomer/ap-commander:1.0.24
                 - quay.io/astronomer/ap-configmap-reloader:0.14.0
                 - quay.io/astronomer/ap-curator:8.0.21-7
                 - quay.io/astronomer/ap-dag-deploy:0.7.2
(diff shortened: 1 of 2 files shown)


charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -158,7 +160,7 @@ spec:
               value: "true"
             {{- end }}
             - name: COMMANDER_MANAGE_NAMESPACE_RESOURCE
-              value: "false"
+              value: {{ if .Values.global.rbacEnabled }}"true"{{ else }}"false"{{ end }}
```
danielhoherd, line 162: @pgvishnuram this was set to `false` in 1.0. Will enabling this cause more code paths to turn on? Will we need additional testing because of this change?

pgvishnuram: houston api is not impacted with this change - no code path being affected on this change

PR #3005, opened 2025-12-16
Use kerberos pgbouncer image if kerberos is enabled

karankhanchandani:
## Description

This PR adds necessary flags in houston to make the kerberos work. It also updates the pgbouncer image based on if kerberos is enabled or not.

## Related Issues

Related astronomer/issues#8204

## Testing

- Unit tests
- Dev tested the changes by manually doing changes in deployment

## Merging

0.37

### DIFF
--- a/charts/pgbouncer/templates/_helpers.yaml
+++ b/charts/pgbouncer/templates/_helpers.yaml
@@ -50,29 +50,17 @@
 
 {{- define "pgbouncer_environment" }}
 - name: DB_USER
-  valueFrom:
-    secretKeyRef:
-      name: {{ default "astronomer-db" .Values.global.pgbouncer.credentialsSecret.name }}
-      key: {{ default "username" .Values.global.pgbouncer.credentialsSecret.usernameKey }}
-      optional: false
+  value: postgres
 - name: DB_PASSWORD
-  valueFrom:
-    secretKeyRef:
-      name: {{ default "astronomer-db" .Values.global.pgbouncer.credentialsSecret.name }}
-      key: {{ default "password" .Values.global.pgbouncer.credentialsSecret.passwordKey }}
-      optional: false
+  value: postgres
 - name: DB_HOST
-  value: {{ default "astronomer-postgresql.astronomer.svc.cluster.local" .Values.global.pgbouncer.host | quote }}
+  value: "astronomer-postgresql.astronomer.svc.cluster.local"
 - name: DB_PORT
-  value: {{ default "5432" (toString .Values.global.pgbouncer.servicePort) | quote }}
+  value: "5432"
 - name: POOL_MODE
   value: session
 - name: ADMIN_USERS
-  valueFrom:
-    secretKeyRef:
-      name: {{ default "astronomer-db" .Values.global.pgbouncer.credentialsSecret.name }}
-      key: {{ default "username" .Values.global.pgbouncer.credentialsSecret.usernameKey }}
-      optional: false
+  value: postgres
 - name: AUTH_TYPE
   value: plain
 - name: LOG_DISCONNECTIONS
--- a/charts/pgbouncer/templates/_helpers.yaml
+++ b/charts/pgbouncer/templates/_helpers.yaml
@@ -8,12 +8,20 @@
 
 
 {{ define "pgbouncer.image" -}}
+{{- $kerbRepo := .Values.kerberosImage.repository -}}
+{{- $kerbTag := .Values.kerberosImage.tag -}}
+{{- $repo := .Values.image.repository -}}
+{{- $tag := .Values.image.tag -}}
 {{- if .Values.global.privateRegistry.enabled -}}
-{{ .Values.global.privateRegistry.repository }}/ap-pgbouncer-krb:{{ .Values.image.tag }}
+{{- $kerbRepo = printf "%s/ap-pgbouncer-krb" .Values.global.privateRegistry.repository -}}
+{{- $repo = printf "%s/ap-pgbouncer" .Values.global.privateRegistry.repository -}}
+{{- end -}}
+{{- if .Values.global.kerberos.enabled -}}
+{{- printf "%s:%s" $kerbRepo $kerbTag -}}
 {{- else -}}
-{{ printf "%s:%s" .Values.image.repository .Values.image.tag }}
-{{- end }}
-{{- end }}
+{{- printf "%s:%s" $repo $tag -}}
+{{- end -}}
+{{- end -}}
 
 {{ define "pgbouncer_host" -}}
 {{ printf "%s-pgbouncer" .Release.Name }}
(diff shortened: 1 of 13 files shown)


values.yaml
```diff
@@ -131,6 +131,11 @@ global:
 
   disableManageClusterScopedResources: false
 
+  manualConnectionStrings:
+    enabled: false
+
+  upsertExtraIniAllowed: false
```
pgvishnuram, line 137: do we really need to do this - can't we update via docs on enabling this feature

pgvishnuram: the case with dagOnlyDeploy and other flags were different it touches various layers of the components. In this case it just updates flag inside astronomer.houston.config.deployments

charts/pgbouncer/templates/pgbouncer-deployment.yaml
```diff
@@ -55,11 +55,10 @@ spec:
       serviceAccountName: {{ template "pgbouncer.serviceAccountName" . }}
       containers:
         - name: pgbouncer
-          image: {{ template "pgbouncer.image" . }}
+          image: {{ include "pgbouncer.image" . | trim }}
           securityContext: {{ toYaml .Values.securityContext| nindent 12 }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           env:
-            {{- include "pgbouncer_environment" . | indent 12 }}
```
pgvishnuram, line 62: we have to also remove 
https://github.com/astronomer/astronomer/blob/68f7c8c03a1487af0d7de02c914c4b1f5e3cdcaf/charts/pgbouncer/templates/_helpers.yaml#L43

charts/pgbouncer/templates/_helpers.yaml
```diff
@@ -8,12 +8,20 @@
...
+{{- end -}}
+{{- if .Values.global.kerberos.enabled -}}
+{{- printf "%s:%s" $kerbRepo $kerbTag -}}
 {{- else -}}
-{{ printf "%s:%s" .Values.image.repository .Values.image.tag }}
-{{- end }}
-{{- end }}
+{{- printf "%s:%s" $repo $tag -}}
+{{- end -}}
+{{- end -}}
```
pgvishnuram, line 24: @Shubham-astro do you think we should do something like this cant we simply them

charts/pgbouncer/templates/pgbouncer-deployment.yaml
pgvishnuram, line 69: @Shubham-astro can we also simplyfy this similar to what we have done for other subcharts

charts/pgbouncer/values.yaml
```diff
@@ -6,6 +6,10 @@ image:
   tag: 1.17.0-22
   pullPolicy: IfNotPresent
 
+kerberosImage:
+  repository: quay.io/astronomer/ap-pgbouncer-krb
+  tag: 1.25.0-1
```
pgvishnuram, line 11: both images are kerbrized one - we can use the above section and make them default and this additional image policy is not required - since we dont have any real team using this image

values.yaml
karankhanchandani: So they have to manually modify the configmap? What happens if they upgrade? I thought of adding it here so that these are managed by helm.

charts/pgbouncer/templates/pgbouncer-deployment.yaml
karankhanchandani: Should I remove the entire define block?

charts/pgbouncer/values.yaml
karankhanchandani: They both should not be kerberized. What if we want to use a non-kerberized image? Hence i kept separate sections.

Should I merge into a single section?

karankhanchandani:
To be evaluated further, I might not need these changes.

charts/pgbouncer/templates/pgbouncer-deployment.yaml
```diff
@@ -55,11 +55,10 @@ spec:
       serviceAccountName: {{ template "pgbouncer.serviceAccountName" . }}
       containers:
         - name: pgbouncer
-          image: {{ template "pgbouncer.image" . }}
+          image: {{ include "pgbouncer.image" . | trim }}
```
danielhoherd, line 58: Why is `trim` needed here? We don't use this syntax in any other `include "whatever.image"` lines.

charts/pgbouncer/templates/pgbouncer-deployment.yaml
```diff
@@ -91,4 +90,31 @@ spec:
...
+                path: pgbouncer.ini
+              - key: users.txt
+                path: users.txt
+        {{- if .Values.global.kerberos.enabled }}
+        - name: krb
+          emptyDir: {}
+        {{- end }}
+        {{- with .Values.extraVolumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
```
danielhoherd, line 119: When there is an empty `extraVolumes` list in the pgbouncer/values.yaml (which there should be, as an indicator that the value exists) this is functionally equivalent to
```
        {{- .Values.extraVolumes | toYaml | nindent 8 }}
```

charts/pgbouncer/templates/pgbouncer-deployment.yaml
```diff
@@ -91,4 +90,31 @@ spec:
...
+            - name: pgbouncer-config
+              mountPath: /etc/pgbouncer
+              readOnly: true
+            {{- if .Values.global.kerberos.enabled }}
+            - name: krb
+              mountPath: /krb
+            {{- end }}
+            {{- with .Values.extraVolumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
```
danielhoherd, line 103: Same thing as the `extraVolumes` comment below.

charts/pgbouncer/values.yaml
```diff
@@ -6,6 +6,10 @@ image:
...
+
```
danielhoherd, line 12: Regarding the above comments about switching to a single `toYaml` line:
```suggestion
# extraVolumes allows you to specify extra volumes that will be included in the pgbouncer deployment pod.
extraVolumes: []

# extraVolumesMounts allows you to specify extra volumes that will be included in the pgbouncer deployment pod containers. (TODO: Make sure this is correct. Maybe this doesn't go into all containers.)
extraVolumeMounts: []
```

charts/pgbouncer/values.yaml
danielhoherd: Would it be helpful to write a test to validate that the image has the expected features? If so, we do have a place in the unit tests where docker images are inspected only if the docker daemon is running, and are skipped if it isn't. https://github.com/astronomer/astronomer/blob/68f7c8c03a1487af0d7de02c914c4b1f5e3cdcaf/tests/chart_tests/test_docker_images.py#L33

PR #3008, opened 2025-12-18
add metadata checksum commander service

pgvishnuram:
## Description

Adds metadata checksum for commander deployment hash restarts 

example scenario when there is a no-op update that touch the metadata this will make sure commander deployment gets restarted due to hash changes that would allow reflection of data in the meta api

## Related Issues

- related https://github.com/astronomer/issues/issues/7455

## Testing

NA

## Merging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -405,33 +405,33 @@ workflows:
                 - quay.io/astronomer/airflow-operator-controller:1.5.2
                 - quay.io/astronomer/ap-alertmanager:0.28.1-5
                 - quay.io/astronomer/ap-astro-ui:1.0.28
-                - quay.io/astronomer/ap-auth-sidecar:1.29.2
+                - quay.io/astronomer/ap-auth-sidecar:1.29.3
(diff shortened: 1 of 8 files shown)


charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -14,9 +14,6 @@ metadata:
     heritage: {{ .Release.Service }}
     plane: {{ .Values.global.plane.mode }}
 spec:
-  {{- if .Values.houston.hostAliases }}
```
ianbuss, line 17: Related?

pgvishnuram: this is accidentally pushed from my local which is a totally different PR - so  im reverting this change

pgvishnuram: this is the PR which is being worked upon https://github.com/astronomer/astronomer/pull/2997

PR #3035, opened 2026-01-21
Script for checking TLS certificates.

htech7x:
## Description

This script validates TLS certificates, checks SANs, expiration, and verifies CA trust on macOS and Linux systems

### DIFF
--- a/scripts/certificate_check/cert_check.sh
+++ b/scripts/certificate_check/cert_check.sh
@@ -1,0 +1,215 @@
+#!/usr/bin/env bash
+#set -euo pipefail
+
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+VIOLET='\033[0;35m'
+NC='\033[0m'
+
+CN_VALID=false
+EXPIRATION_VALID=true
+CERT_TRUSTED=false
+SAN_VALID=false
+
+CERT="$1"
+VALUES_YAML="$2"
+ROOT_CA="${3:-}"  # optional root CA
+WARN_DAYS=90
+WARN_SECONDS=$((WARN_DAYS * 86400))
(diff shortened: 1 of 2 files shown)


scripts/certificate_check/cert_check.sh
```diff
@@ -0,0 +1,215 @@
...
+echo
+
+# Common Name (CN)
+CN=$(openssl x509 -in "$CERT" -noout -subject \
+     | sed -n 's/.*CN *= *\([^,/]*\).*/\1/p')
+
+echo -e "${GREEN}Common Name (CN):${NC} ${CN:-<none>}"
+echo
+# Compare CN with BASE_DOMAIN from values.yaml
+if [[ "$CN" == "$BASE_DOMAIN" ]]; then
```
timo-astronomer, line 75: do we want to support case, where `CN=app.$BASE_DOMAIN`?

if so, we probably need to support case, where SANs do not include `app.$BASE_DOMAIN`

htech7x: We can cover this if necessary.

Simpcyclassy (commented):
The TLS validation script itself is reasonable — checks certificate validity, SANs, expiration, and CA trust on macOS and Linux. A few questions before this merges:

Where does this script live in the repo's mental model? The `bin/` directory is for operational scripts used in CI or by the platform team. If this is intended for customer support or field use, it should live somewhere with that context. A brief comment at the top of the script explaining the intended audience and usage would help.

The script has no tests and the PR description has no testing section. For a bash utility, even a simple smoke test or the equivalent of running it against a known-valid cert would be useful.

The PR description also does not list target branches for merging.

Not blocking — this is clearly useful tooling — but the context and merge target need to be filled in.

PR #3054, opened 2026-01-27
dag-deploy 0.9.0

danielhoherd:
## Description

- ap-dag-deploy v0.9.0

## Related Issues

- <https://github.com/astronomer/issues/issues/8275>
- <https://linear.app/astronomer/issue/PINF-189/dag-downloader-emitting-excessive-logs-before-first-deploy>
- <https://github.com/astronomer/issues/issues/8272>
- <https://linear.app/astronomer/issue/APC-1088/support-kubernetes-135>

## Testing

Since this is a decent refactor, and since we've had a few bugs with dag-deploy, it would be a good idea to run all of our tests. The overall behavior on the application has not changed, just the implementation details, so all existing tests should still be valid.

## Merging

This should go into both 1.1.0 and 0.37.7


Simpcyclassy (approved):
LGTM

PR #3068, opened 2026-01-30
PINF-216 - fix network policies in es proxy

pgvishnuram:
## Description

external es proxy ingress is broken due to network policies returning gateway timedout.
This PR adds correct network policy, so that controlplane  can reach the dataplane elastic ingress without issues

## Related Issues

- PINF-216 https://linear.app/astronomer/issue/PINF-216/es-proxy-logs-are-not-showing-up-because-of-the-network-policy-issue

## Testing

QA should not see gateway timeout and should be able to stream logs

## Merging


danielhoherd (commented):
We need tests for each of these `if` and `else` code paths.

PR #3085, opened 2026-02-19
Remove unit test stages from stage cluster build

sudarshanrampuria:
## Description

Removing `run_pre_commit`, `test-unit`, and `check-commander-airflow-version` jobs from stage cluster workflow because this already run in all PR merge and master push. It is not required to be run in stage cluster workflow.

## Testing

NA

## Merging

NA

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -590,15 +590,8 @@ workflows:
         - << pipeline.parameters.qa-release >>
         - not: << pipeline.parameters.scan-docker-images >>
     jobs:
-      - run_pre_commit
-      - test-unit
-      - check-commander-airflow-version
       - build-artifact:
           qa_release: true
-          requires:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -590,15 +590,8 @@ workflows:
         - << pipeline.parameters.qa-release >>
         - not: << pipeline.parameters.scan-docker-images >>
     jobs:
-      - run_pre_commit
-      - test-unit
-      - check-commander-airflow-version
```
pgvishnuram, line 595: why are we removing check-commander-airflow-version and run_pre_commit

sudarshanrampuria: We are already running those checks in every PR merge/master build. 

I thought we don't need to run in scheduled pipeline/ QA builds because we have already ran it when we merged the branch.

PR #3086, opened 2026-02-19
PINF-176 kuiper-reloader 0.1.17 with k8s 1.35 support

danielhoherd:
## Description

kuiper-reloader 0.1.17 with k8s 1.35 support, which is the last component that needed to be upgraded to 1.35.

## Related Issues

https://linear.app/astronomer/issue/PINF-176/support-kubernetes-135

## Testing

Tests have passed, but I have not deployed this into a k8s cluster.

## Merging

This should be merged to all supported branches.


pgvishnuram:
approved the image push pipeline and the CI is passing now

PR #3088, opened 2026-02-20
SWAT-59 / PINF-72 - helm chart support for ingressClassName

stefan-raducan:
## Description

Our helm chart creates ingress configs with [deprecated annotations](https://stackoverflow.com/questions/73915526/kubernetes-ingress-kubernetes-io-ingress-class-vs-ingressclassname) (there's no option to add spec.ingressClassName [here](https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/ingress.yaml#L31).

If global.ingressClassName is set → uses that value
Else if global.extraAnnotations contains kubernetes.io/ingress.class → uses that value
Else → defaults to "{{ .Release.Name }}-nginx"

## Related Issues

- [SWAT-59](https://astronomer.atlassian.net/browse/SWAT-59)
- <https://github.com/astronomer/issues/issues/6220>
- <https://linear.app/astronomer/issue/PINF-72>

## Testing

Pytests have been added for the configuration updates.
TODO - test install chart on a cluster and verify ingress with standard and bespoke ingress controller.

## Merging

Master, 1.1


charts/astronomer/templates/ingress.yaml
```diff
@@ -29,6 +29,17 @@ metadata:
...
+  {{- if .Values.global.ingressClassName }}
+  ingressClassName: {{ .Values.global.ingressClassName }}
+  {{- else if .Values.global.extraAnnotations }}
+    {{- range $key, $value := .Values.global.extraAnnotations }}
+      {{- if eq $key "kubernetes.io/ingress.class" }}
+  ingressClassName: {{ $value }}
+      {{- end }}
+    {{- end }}
+  {{- else }}
+  ingressClassName: "{{ .Release.Name }}-nginx"
```
danielhoherd:
```suggestion
  ingressClassName: {{ index .Values.global.extraAnnotations "kubernetes.io/ingress.class" | default (printf "%s-nginx" .Release.Name) }}
```

danielhoherd (approved):
LGTM. I left one optimization that reduces LOC while keeping the same functionality.

charts/astronomer/templates/ingress.yaml
danielhoherd: This same type of change can also be applied above.

stefan-raducan: updated on all templates

Simpcyclassy (commented):
This PR migrates all ingress templates from the deprecated `kubernetes.io/ingress.class` annotation to `spec.ingressClassName`. The motivation is valid (the annotation has been deprecated since K8s 1.18) and the approach is consistent across all 14 ingress templates.

Two things to resolve before merge:

**Potential conflict with PR #2775**

PR #2775 (`6220` branch) also touches ingress templates for the same migration. Before either merges, the two PRs need to be reconciled to avoid conflicts or duplicated work. Check with the author of #2775 about which should take precedence.

**The `extraAnnotations` fallback path will fail on most configurations**

The fallback logic for when `global.ingressClassName` is not set but `global.extraAnnotations` is:

```yaml
ingressClassName: {{ index .Values.global.extraAnnotations "kubernetes.io/ingress.class" | default (printf "%s-nginx" .Release.Name) }}
```

This will produce an incorrect YAML value if `global.extraAnnotations` doesn't contain `kubernetes.io/ingress.class`. The `index` function returns an empty string for a missing key, and empty string is truthy in Helm templates, so `default` won't kick in. The rendered output would be `ingressClassName: ` (empty), which will cause an invalid Ingress spec. The condition should be tightened:

```yaml
{{- if .Values.global.ingressClassName }}
ingressClassName: {{ .Values.global.ingressClassName }}
{{- else }}
ingressClassName: "{{ .Release.Name }}-nginx"
{{- end }}
```

and drop the `extraAnnotations` read-through entirely (or handle it with `hasKey`).

The test in `test_ingress.py` should cover the case where neither `global.ingressClassName` nor the annotation is set, to verify the default is rendered correctly.

charts/astronomer/templates/ingress.yaml
pgvishnuram: do we need this much of changes can we straight set the ingressClassName for the nginx ingress

stefan-raducan: I've not touched this PR in a while. We do need to set ingressClassName as part of all ingress templates. Chioma's comment below summarizes a potential fix that was not applied. I can implement it if we decide we go with this PR in favor of #2775 
IMO the templating logic in #2775 are harder to read but I might be biased. I'm ok with any of the PRs moving further

charts/astronomer/templates/houston/ingress.yaml
```diff
@@ -35,6 +35,13 @@ metadata:
     {{- toYaml .Values.houston.ingress.annotation  | nindent 4  }}
     {{- end }}
 spec:
+  {{- if .Values.global.ingressClassName }}
+  ingressClassName: {{ .Values.global.ingressClassName }}
```
rishkarajgi: When `global.ingressClassName` is set this renders the custom class, but the deprecated `kubernetes.io/ingress.class: "{{ .Release.Name }}-nginx"` annotation just above (line 21) is left untouched — so the ingress carries two disagreeing class signals, and the deprecated annotation this PR is meant to move off of is still emitted. Suppressing or aligning that annotation when a class is set would actually close out the original goal. (Same pattern in all ~12 ingress templates.)

charts/nginx/templates/controlplane/nginx-cp-ingressclass.yaml
```diff
@@ -0,0 +1,21 @@
...
+  name: {{ .Release.Name }}-nginx
+  labels:
+    tier: {{ template "nginx.name" . }}
+    component: cp-ingress-controller
+    chart: {{ template "nginx.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    plane: {{ .Values.global.plane.mode }}
+spec:
+  controller: k8s.io/ingress-nginx
```
rishkarajgi: This uses the generic `k8s.io/ingress-nginx` controller string, and the bundled controllers run with the default `--ingress-class-by-name=false` and no `--controller-class` override — so `ingressClassName` matches by controller class, which is broader than the old name-scoped annotation. In a cluster with two Astronomer nginx controllers each could claim the other's ingresses. Worth confirming the topology or setting a per-install `--controller-class`.

tests/chart_tests/test_ingress.py
```diff
@@ -91,6 +91,7 @@ def test_single_ingress_per_host(self, kube_version):
         assert all(doc["apiVersion"] == "networking.k8s.io/v1" for doc in ingresses)
         assert all(doc["kind"] == "Ingress" for doc in ingresses)
         assert all(doc["metadata"]["annotations"]["kubernetes.io/ingress.class"] == "release-name-nginx" for doc in ingresses)
+        assert all(doc["spec"]["ingressClassName"] == "release-name-nginx" for doc in ingresses)
```
rishkarajgi, line 94: Good that the new field is asserted — but there's no test for the deprecated-annotation side when a custom `global.ingressClassName` is set, so the annotation/spec contradiction (class set to a custom value while the annotation stays `<release>-nginx`) isn't caught. A case asserting the annotation is dropped/aligned in that scenario would cover it.

rishkarajgi (commented):
### Verdict: PASS, no blockers, 3 should-fix:
Traced the upgrade path carefully since this touches every platform ingress. The good news is existing installs are safe; the gap is that the feature's own goal — stop using the deprecated annotation — isn't actually met on the path that sets it. Nothing blocking, but I'd want the annotation handling sorted before merge.

**Existing-install upgrade is a clean, continuity-preserving migration (verified):** with `global.ingressClassName` unset (the default, `~`), the deprecated `kubernetes.io/ingress.class: <release>-nginx` annotation is preserved AND `spec.ingressClassName` resolves to the same `<release>-nginx` via the else-branch, so both signals agree. The new IngressClass is created with a matching controller, and it does NOT set `ingressclass.kubernetes.io/is-default-class`, so it can't hijack other tenants' class-less ingresses. No window where a platform ingress is claimed by nothing. That part is solid.

**Should-fix — the deprecated annotation is never dropped/aligned when `global.ingressClassName` is set, so that ingress ends up self-contradictory.** Every template still hardcodes `kubernetes.io/ingress.class: "<release>-nginx"` (e.g. houston ingress line 21) in the default authSidecar-off branch, and this PR only *adds* `spec.ingressClassName`. So `global.ingressClassName: foo` renders the annotation as `<release>-nginx` and the spec as `foo` — two disagreeing class signals on the same ingress, and the stated goal (get off the deprecated annotation) isn't reached on any path. Depending on how the shipped ap-nginx resolves a disagreement (drop vs ignore), this could bite adopters harder than Should-fix — I couldn't confirm that precedence from the chart, so flagging it there. The fix is to suppress or align the annotation when a class is set.

**Should-fix — setting `global.ingressClassName` doesn't reconfigure/gate the bundled nginx controller.** The bundled controller watches `<release>-nginx` (a separate value), so with the annotation retained it can still claim the ingress while an operator's own `foo` controller also matches — a double-claim rather than a clean hand-off.

**Should-fix — test gap:** the chart tests assert `spec.ingressClassName` is set but never assert the annotation is dropped/aligned when a custom class is set, so the contradiction above goes uncaught.

**Open question for platform/SRE:** both new IngressClasses use the generic `controller: k8s.io/ingress-nginx`, and the controllers run with the default `--ingress-class-by-name=false` / no `--controller-class` override, so matching is by controller-class rather than the old name-scoped annotation. Two co-located Astronomer nginx controllers could then cross-claim each other's ingresses. Worth confirming the topology (or setting a per-install `--controller-class`) before this goes out — this is the piece I'm least able to verify from the chart alone.

Nits: the class-resolution block is copied inline across ~12 ingress templates rather than a shared `_helpers.tpl` include (drift risk); the DP IngressClass omits the `plane:` label the CP one has; and values.yaml drops the inline `#kubernetes.io/ingress.class` example comment.

Fleet-wide ingress change, so worth an SRE eyeball on the multi-controller question in addition to the usual approval.

tests/chart_tests/test_ingress.py
stefan-raducan: added test_ingress_class_annotation_absent_with_ingressClassName

charts/astronomer/templates/houston/ingress.yaml
stefan-raducan: added `global.ingressClassName` in templates/_helpers.tpl

charts/nginx/templates/controlplane/nginx-cp-ingressclass.yaml
stefan-raducan: I will update the startup and other logic using the [official docs](https://kubernetes.github.io/ingress-nginx/user-guide/multiple-ingress/#using-ingressclasses)
I initially thought `--ingress-class` would be enough but you're right that `--controller-class` is also needed to correctly pair the ingressClass with the running nginx deployment. I've used the release name as the unique identifier and tested propagation in test_default_ingressclass_rendered `spec.controller` check

stefan-raducan:
Updates in the last 2 commits since the review:
1. Bundled nginx controller deployed in all scenarios
The bundled nginx ingress controller and default backend are rendered even when using a custom global.ingressClassName, so both the bundled controller and an external one would run simultaneously.
  - Fix: Introduced a nginx.deployBundledController helper that gates all bundled controller resources. Behavior:
    - No ingressClassName set: bundled controller + default backend rendered
    - ingressClassName set to a non-default value: bundled controller + default backend skipped
    - ingressClassName matching default (<release>-nginx): bundled controller rendered
1. Double-claim on Ingress resources within a single install
Both the kubernetes.io/ingress.class annotation (targeting the bundled controller) and spec.ingressClassName
(potentially targeting an external controller) were present on the same Ingress resource, causing two controllers to reconcile it (in this specific scenario).
  - Fix: The kubernetes.io/ingress.class annotation is suppressed when global.ingressClassName is set, leaving only spec.ingressClassName for controller matching.
1. Cross-claim between multiple installs in the same cluster
The IngressClass resources all used the generic `spec.controller: k8s.io/ingress-nginx` and controllers ran with the same default `--controller-class=k8s.io/ingress-nginx`. Two installs in the same cluster would both claim each other's IngressClass.
  - Fix:
      - spec.controller and --controller-class are now release-scoped: `k8s.io/ingress-nginx-<release>`
    - IngressClass is rendered only when the bundled controller is deployed and no `kubernetes.io/ingress.class` annotation exists in extraAnnotations
    - If installing multiple charts, users must set global.ingressClassName to a non-default value for at least one of them to avoid IngressClass name collision.
      - TBD: does it make sense to also make the IngressClass default name include the release name? if yes, multiple default helm chart installs can be done, each with their own IngressClass and controller. IMO this isn't needed because in all scenarios I've seen of multiple installs on the same cluster a single ingressClass and controller was used

charts/nginx/templates/controlplane/nginx-cp-ingressclass.yaml
```diff
@@ -0,0 +1,22 @@
+##########################
+## NGINX IngressClass ##
+##########################
+{{- if include "nginx.deployBundledController" . }}
+{{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
+{{- if not (index (.Values.global.extraAnnotations | default dict) "kubernetes.io/ingress.class") }}
```
Simpcyclassy, line 6: IngressClass is cluster-scoped, and this chart gates every other cluster-scoped object on `global.clusterRoles` (see `charts/nats/templates/jetstream-job.yaml`, commander's role/rolebinding, prometheus). On a `global.clusterRoles: false` install this will either be rejected outright or, worse, get skipped while every Ingress still renders `ingressClassName: <release>-nginx`, at which point nginx resolves the class, finds nothing, and stops serving the platform ingresses. Please add the gate, and decide explicitly what `spec.ingressClassName` should be when we can't create the class. Good that the plane-mode gating on the line above is here.

charts/nginx/templates/_helpers.tpl
```diff
@@ -48,10 +48,25 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{ define "nginx.ingress.class" -}}
-{{- if .Values.ingressClass -}}
-{{- .Values.ingressClass -}}
+{{- if .Values.global.ingressClassName -}}
```
Simpcyclassy, line 51: This drops `.Values.ingressClass` without removing it, so `charts/nginx/values.yaml:47` is now a value that silently does nothing for anyone already setting it. It also replaces the `nginx.fullname` fallback with a hardcoded `<release>-nginx`, so installs using `nameOverride`/`fullnameOverride` get a different class than before. Either keep honouring both or delete the dead value and call the break out in the body.

charts/nginx/templates/_helpers.tpl
```diff
@@ -48,10 +48,25 @@ Create chart name and version as used by the chart label.
...
+{{- printf "%s-nginx" .Release.Name -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Whether the bundled nginx controller should be deployed. Skipped when
+global.ingressClassName is set to a non-default value (i.e. the operator is
+bringing their own controller). The default is <release>-nginx.
+*/}}
+{{- define "nginx.deployBundledController" -}}
```
Simpcyclassy, line 63: This ties two independent decisions to one value: what class our Ingress objects reference, and whether we ship a controller at all. An operator who wants our nginx but a differently named class has no way to say so, and anyone setting this on an existing install has their nginx Deployment and Service deleted on the next upgrade, which on a cloud provider means losing the LoadBalancer and its address. I'd split the controller toggle into its own value and leave `global.ingressClassName` as naming only.

charts/nginx/templates/_helpers.tpl
```diff
@@ -48,10 +48,25 @@ Create chart name and version as used by the chart label.
...
+{{- if .Values.global.ingressClassName -}}
+  {{- if eq .Values.global.ingressClassName (printf "%s-nginx" .Release.Name) -}}
+    true
+  {{- end -}}
```
Simpcyclassy, line 67: The negative case here falls off the end and returns an empty string, so this helper is "true" or "" and only works with the bare `if` at the call sites. `templates/_helpers.tpl` documents the opposite convention (`Returns the string "true" or "false" - compare with eq`). If someone later makes this consistent by adding `{{- else -}}false`, every nginx resource silently starts rendering again. Return "true"/"false" and compare with `eq` at the call sites.

charts/nginx/templates/controlplane/nginx-cp-deployment.yaml
```diff
@@ -79,6 +80,7 @@ spec:
             - --default-backend-service={{ .Release.Namespace }}/{{ template "nginx.fullname" . }}-default-backend
             - --configmap={{ .Release.Namespace }}/{{ .Release.Name }}-cp-{{ template "nginx.name" . }}-ingress-controller
             - --ingress-class={{ template "nginx.ingress.class" . }}
+            - --controller-class=k8s.io/ingress-nginx-{{ .Release.Name }}
```
Simpcyclassy, line 83: This closes out the controller-class scoping question, but it needs an upgrade note. During the rolling update the old-generation pods still run the default `k8s.io/ingress-nginx` controller class, so once the IngressClass exists with `k8s.io/ingress-nginx-<release>` those pods no longer match it and drop the platform ingresses from their config until they're replaced. Worth confirming on a real upgrade and saying so in the body if there's a window.

values.yaml
```diff
@@ -99,6 +99,9 @@ global:
     containerdCertCopier:
       startupProbe: {}
     priorityClassName: ~
+  # If you do NOT want to use the default configured nginx ingress controller, add the ingressClassName you want to replace it with
+  # If no value is provided, the default will be set to <release-name>-nginx and the nginx ingress controller will be created with that name
+  ingressClassName: ~
```
Simpcyclassy, line 104: Good that the default is `~`. The comment should say what actually happens though: setting this doesn't just rename the class, it removes the bundled nginx Deployment, Service, ConfigMaps, RBAC and default backend, so on an existing install it deletes the LoadBalancer. That's not something to discover from a diff.

templates/_helpers.tpl
```diff
@@ -187,3 +187,18 @@ Returns the string "true" or "false" — compare with eq.
...
+
+{{/*
+Resolve the ingressClassName for Ingress resources. Uses global.ingressClassName if set,
+otherwise falls back to the value from the kubernetes.io/ingress.class annotation in
+global.extraAnnotations, or the default <release>-nginx.
+*/}}
+{{- define "global.ingressClassName" -}}
+{{- if .Values.global.ingressClassName -}}
+{{- .Values.global.ingressClassName -}}
+{{- else if .Values.global.extraAnnotations -}}
```
Simpcyclassy, line 199: This branch and the `else` below it produce the same value whenever the annotation key is absent, so the two can collapse into one `{{- else -}}{{- index (.Values.global.extraAnnotations | default dict) "kubernetes.io/ingress.class" | default (printf "%s-nginx" .Release.Name) -}}`. That also matches the `| default dict` shape used in the new IngressClass templates.

tests/chart_tests/test_nginx.py
```diff
@@ -493,3 +493,166 @@ def test_nginx_backend_overrides():
...
+            and not d["metadata"].get("labels", {}).get("chart", "").startswith("elasticsearch")
+        ]
+
+    def test_default_renders_bundled_controller(self, plane_mode):
+        """Without ingressClassName, the bundled nginx controller resources are rendered."""
+        docs = render_chart(values={"global": {"plane": {"mode": plane_mode}}})
+        nginx_resources = self._nginx_resources(docs)
+        assert len(nginx_resources) > 0, f"Expected bundled nginx resources in {plane_mode} mode"
+
+    def test_custom_ingressclassname_skips_bundled_controller(self, plane_mode):
```
Simpcyclassy, line 626: This is the right shape for this kind of gate: a whole-chart render asserting the resources are actually gone rather than checking one template in isolation. The error message listing the leaked names will save someone real time.

Simpcyclassy (changes requested):
The new IngressClass is cluster-scoped but not gated on `global.clusterRoles`, so namespace-scoped installs end up with platform ingresses pointing at a class that was never created and no controller claiming them. Separately, `global.ingressClassName` now doubles as an off switch for the entire nginx stack including its LoadBalancer Service, which should be its own value rather than a side effect of naming a class. The `extraAnnotations` fallback I flagged in May is resolved and I got the mechanism wrong there, so the clusterRoles gate is what's blocking.

charts/astronomer/templates/ingress.yaml
Simpcyclassy: Agree with the original suggestion here, and I owe a correction on my May review: I claimed `index` returns an empty string for a missing key so `default` wouldn't fire. That's wrong. `index` on a `map[string]interface{}` returns a nil interface, sprig's `default` treats nil as empty, and the fallback resolves to `<release>-nginx` as intended. So this shape is sound and the added `else` branch covers the empty-map case.

One thing to still add: `spec.ingressClassName` is typed `["string", "null"]` in our k8s schemas, so a regression that renders it empty would pass object validation silently. None of the new parametrize tables cover `global.extraAnnotations` set *without* `kubernetes.io/ingress.class` (the OpenShift `route.openshift.io/termination` case), which is the branch in question. One row there closes it.

Also, is #2775 closed out now? My other May comment was about reconciling the two, and I'd rather not leave that hanging at merge time.

charts/astronomer/templates/houston/ingress.yaml
Simpcyclassy: Agree, and this is fixed for the chart's own annotation: the `{{- if not .Values.global.ingressClassName }}` guard now suppresses the hardcoded `<release>-nginx` on all the templates, and `test_ingress_class_annotation_absent_with_ingressClassName` locks it in.

One case is still open though. `global.extraAnnotations` is dumped verbatim right after that guard, and in the authSidecar branch it's dumped with no guard at all. So `global.ingressClassName: foo` plus an `extraAnnotations` entry carrying `kubernetes.io/ingress.class: bar` still produces the two disagreeing signals, just user-supplied rather than chart-supplied. Stripping that key from `extraAnnotations` when `global.ingressClassName` is set would finish the job.

charts/nginx/templates/controlplane/nginx-cp-ingressclass.yaml
Simpcyclassy: Agree, and the fix looks right: `spec.controller` and the deployment's new `--controller-class` both derive from the release name, so two Astronomer nginx controllers in one cluster can no longer claim each other's ingresses. `test_default_ingressclass_rendered` asserting the exact controller string is the right guard.

Two follow-ons I've left on the files themselves: the IngressClass isn't gated on `global.clusterRoles` even though it's cluster-scoped, and adding `--controller-class` looks like it creates a window during the nginx rollout where old-generation pods no longer match the new class.

charts/nginx/templates/controlplane/nginx-cp-config-role.yaml
```diff
@@ -1,6 +1,7 @@
 #####################################
 ## NGINX Control Plane Config Role ##
 #####################################
+{{- if include "nginx.deployBundledController" . }}
```
pgvishnuram, line 4: why are we adding this condition here -> this role is for leader election right

charts/nginx/templates/controlplane/nginx-cp-configmap.yaml
```diff
@@ -1,6 +1,7 @@
 ###################################
 ## NGINX Control Plane ConfigMap ##
 ###################################
+{{- if include "nginx.deployBundledController" . }}
```
pgvishnuram, line 4: gating here is not necessary for rolebinding too

charts/nginx/templates/controlplane/nginx-cp-config-rolebinding.yaml
```diff
@@ -1,6 +1,7 @@
 ###############################
 ## NGINX Config Role Binding ##
 ###############################
+{{- if include "nginx.deployBundledController" . }}
```
pgvishnuram, line 4: gating here is not necessary for rolebinding too

rishkarajgi:
@stefan-raducan is this still WIP?

PR #3096, opened 2026-02-28
Make index.yaml upload transactional

nickstenning:
To avoid accidentally overwriting index.yaml updates from other repositories and workflows, we need to make the updates to index.yaml transactional.

To do this, we first fetch the current file generation, and then upload only if the generation being overwritten matches (i.e. hasn't been updated in the meantime).

As this appears to be a thing that does happen regularly in practice, I've also added some retry logic so that we're not constantly battling failing uploads.

### DIFF
--- a/bin/release-helm-chart
+++ b/bin/release-helm-chart
@@ -79,6 +79,25 @@ pre-flight-setup() {
   export BOTO_CONFIG=/dev/null
 }
 
+update_index() {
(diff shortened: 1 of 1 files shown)


nickstenning:
Closing to clean up my personal WIP. Feel free to reopen if we do actually want to integrate something like this.

PR #3110, opened 2026-03-18
danielhoherd/pinf-333 dp-link helm chart features

danielhoherd:
## Description

Add dp-link helm chart features.

## Related Issues

https://linear.app/astronomer/issue/PINF-333/houston-dp-link-deployment-helm-chart-setup

## Testing

Unit tests are included. Manual testing is still ongoing. Will update when it's complete.

## Merging

This is only meant for feature/apc-dr and 2.0

### DIFF
--- a/tests/chart_tests/test_dp_link_deployment.py
+++ b/tests/chart_tests/test_dp_link_deployment.py
@@ -44,7 +44,8 @@ class TestDpLinkDeployment:
 
         assert dp_link_container["image"].startswith("quay.io/astronomer/ap-houston-api:")
         assert dp_link_container["securityContext"]["readOnlyRootFilesystem"]
-        assert dp_link_container["command"] == ["yarn", "run", "dplink"]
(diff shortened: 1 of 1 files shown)


configs/disable-logging.yaml
```diff
@@ -0,0 +1,2 @@
+tags:
+  logging: true
```
pgvishnuram, line 2:
```suggestion
tags:
  logging: false
```

configs/disable-monitoring.yaml
```diff
@@ -0,0 +1,2 @@
+tags:
+  monitoring: true
```
pgvishnuram, line 2:
```suggestion
tags:
  monitoring: false
```

danielhoherd: That config is specifically to disable monitoring.

configs/disable-logging.yaml
danielhoherd: That config is specifically to disable logging.

PR #3115, opened 2026-03-21
Pinf 217 authz dp side grpc chart changes

adhipjos:
## Description

1. Added Helm chart support for shared secret authentication between pilot and commander gRPC services.
2. No env vars needed. Both commander and pilot already default `INTERNAL_SERVICE_TOKEN_FILE` to `/etc/astronomer/internal-service-auth/token`

Key changes:

- `charts/astronomer/templates/secrets/internal-service-auth-secret.yaml`: Kubernetes Secret that holds the shared token, rendered for data and unified plane modes

- `charts/astronomer/templates/_helpers.yaml`: Added astronomer.internalServiceToken helper: uses a user-provided value if set, preserves the existing secret across upgrades via lookup, or auto-generates a token with randBytes 32

- `values.yaml`: Added global.internalServiceAuth.token (empty by default, auto-generated on install)

- `charts/astronomer/templates/commander/commander-deployment.yaml`: Mounts the secret at `/etc/astronomer/internal-service-auth` (read-only)

- `charts/astronomer/templates/pilot/pilot-deployment.yaml`: Same volume mount so pilot can read the shared token

## Related Issues

[PINF-217: Implement Authz mechanism for DP local unary gRPC calls](https://linear.app/astronomer/issue/PINF-217/implement-authz-mechanism-for-dp-local-unary-grpc-calls)

## Testing

## Merging

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -20,10 +20,6 @@ global:
   internalServiceAuth:
     token: ""
 
-  # Enable or disable the flightDeck feature
-  flightDeck:
-    enabled: false
-
   # Labels to add to every pod
   podLabels: {}
 


charts/astronomer/templates/secrets/cluster-local-service-auth-secret.yaml
```diff
@@ -0,0 +1,19 @@
+############################################
+## Internal Service Auth Shared Secret    ##
+############################################
+{{- if or (eq .Values.global.plane.mode "data") (eq .Values.global.plane.mode "unified") }}
```
pgvishnuram, line 4: This is not going under feature flag this shared will be mounted even without DR mode is that fine

danielhoherd: We don't yet have a single feature flag to toggle failover features on and off.

adhipjos: That is fine. We can add feature flag gating in a future PR once we add a global DP failover feature flag. This secret is to grant local cluster scoped access to RPCs to our internal services like Pilot, it shouldn't affect any existing features

charts/astronomer/templates/secrets/internal-service-auth-secret.yaml
```diff
@@ -0,0 +1,19 @@
+############################################
+## Internal Service Auth Shared Secret    ##
```
danielhoherd: I feel like we should call this something other than "internal". It's plane-local. We have `charts/astronomer/templates/cluster-local-data.yaml` so maybe this should be `cluster-local-service-auth-secret` ?

adhipjos: Updating

tests/chart_tests/test_internal_service_auth.py
```diff
@@ -0,0 +1,175 @@
...
+            kube_version=kube_version,
+            values={"global": {"plane": {"mode": plane_mode}}},
+            show_only=self.show_only,
+        )
+        assert len(docs) == 1
+        labels = docs[0]["metadata"]["labels"]
+        assert labels["component"] == "internal-service-auth"
+        assert labels["tier"] == "astronomer"
+        assert labels["release"] == "release-name"
+        assert labels["plane"] == plane_mode
```
danielhoherd: These two test cases can be merged. We try to do this because `helm template` operations are expensive.

Alternatively, you could pre-render the chart with the three different plane modes and then have test cases that assert against that rendered output, which would avoid the expensive `helm template` call. Usually though we have a test case for default chart properties for a given config.

tests/chart_tests/test_internal_service_auth.py
```diff
@@ -0,0 +1,175 @@
...
+        """When no token is provided, a non-empty token should be auto-generated."""
+        docs = render_chart(
+            kube_version=kube_version,
+            values={"global": {"plane": {"mode": "data"}}},
+            show_only=self.show_only,
+        )
+        assert len(docs) == 1
+        token_b64 = docs[0]["data"]["token"]
+        raw = base64.b64decode(token_b64).decode()
+        assert len(raw) > 0
```
danielhoherd: This can be merged into the test case above.

tests/chart_tests/test_internal_service_auth.py
```diff
@@ -0,0 +1,175 @@
...
+            kube_version=kube_version,
+            values={"global": {"plane": {"mode": plane_mode}}},
+            show_only=self.show_only,
+        )
+        assert len(docs) == 1
+        c_by_name = get_containers_by_name(docs[0])
+        vm_by_name = {m["name"]: m for m in c_by_name["commander"]["volumeMounts"]}
+        assert "internal-service-auth" in vm_by_name
+        assert vm_by_name["internal-service-auth"]["mountPath"] == "/etc/astronomer/internal-service-auth"
+        assert vm_by_name["internal-service-auth"]["readOnly"] is True
```
danielhoherd, line 104: These can be merged for the same reasons listed above.

tests/chart_tests/test_internal_service_auth.py
```diff
@@ -0,0 +1,175 @@
...
+                "astronomer": {"pilot": {"enabled": True}},
+            },
+            show_only=self.show_only,
+        )
+        assert len(docs) == 1
+        c_by_name = get_containers_by_name(docs[0])
+        vm_by_name = {m["name"]: m for m in c_by_name["pilot"]["volumeMounts"]}
+        assert "internal-service-auth" in vm_by_name
+        assert vm_by_name["internal-service-auth"]["mountPath"] == "/etc/astronomer/internal-service-auth"
+        assert vm_by_name["internal-service-auth"]["readOnly"] is True
```
danielhoherd: These can be merged for the same reasons listed above.

tests/chart_tests/test_internal_service_auth.py
```diff
@@ -0,0 +1,175 @@
...
+        """Pilot deployment should not render at all when disabled."""
+        docs = render_chart(
+            kube_version=kube_version,
+            values={
+                "global": {"plane": {"mode": "data"}},
+                "astronomer": {"pilot": {"enabled": False}},
+            },
+            show_only=self.show_only,
+        )
+        assert len(docs) == 0
```
danielhoherd: This test isn't needed. We already assert that pilot templates are not rendered when it is disabled in https://github.com/astronomer/astronomer/blob/c9c01d97d7/tests/chart_tests/test_astronomer_pilot.py#L35-L42

tests/chart_tests/test_internal_service_auth.py
adhipjos: updated

tests/chart_tests/test_internal_service_auth.py
adhipjos: updated

tests/chart_tests/test_internal_service_auth.py
adhipjos: updated

tests/chart_tests/test_internal_service_auth.py
adhipjos: updated

tests/chart_tests/test_internal_service_auth.py
adhipjos: removed

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -260,6 +260,9 @@ spec:
             - mountPath: /app/metadata.yaml
               name: metadata
               subPath: metadata.yaml
+            - name: internal-service-auth
+              mountPath: /etc/astronomer/internal-service-auth
```
danielhoherd, line 264: We renamed the secret, but we didn't rename these. Now we have an inconsistency. Do we want to fix that now? I feel like everything in this PR should be renamed to match the template naming change. I should have spelled that out more explicitly in my earlier comment.

adhipjos: That's fine. Easy to make these changes right now. Let me update these for consistency

adhipjos: Actually let's do this later, since changing the mount path would also require changes to the commander and pilot code which have this mount path as the default. I'll create a Linear issue for this change for later

PR #3119, opened 2026-03-24
PINF-379 removetrivy 

sudarshanrampuria:
## Description
* Remove trivy
* Changed setup-kind.py script to get images list from `show-docker-images.py`. It was using circle ci config and the config was removed when trivy got removed. 

## Related Issues

https://linear.app/astronomer/issue/PINF-379/remove-trivy-from-astro-private-cloud-repos

## Merging

master, 0.x, 1.x, 2.x

### DIFF
--- a/bin/setup-cp-dp-k3d.py
+++ b/bin/setup-cp-dp-k3d.py
@@ -27,7 +27,7 @@ import time
 from dataclasses import dataclass
 from pathlib import Path
 
-GIT_ROOT_DIR = next(iter([x for x in Path(__file__).resolve().parents if (x / ".git").is_dir()]), None)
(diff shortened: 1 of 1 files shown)


Shubham-astro (approved):

bin/setup-kind.py
```diff
@@ -74,10 +82,7 @@ def kind_load_docker_images(cluster: str) -> None:
     Args:
         cluster: Name of the KIND cluster to load images into.
     """
-    circleci_config = yaml.safe_load((GIT_ROOT_DIR / ".circleci" / "config.yml").read_text())
-    image_list = circleci_config["workflows"]["scan-docker-images"]["jobs"][0]["trivy-scan-docker"]["matrix"]["parameters"][
```
danielhoherd, line 78: Hmm.... I wonder if we're relying on this anywhere. We'll have to be attentive to things breaking.

PR #3122, opened 2026-03-25
PINF-306: add gitRyncRelay.metrics option in helm

Shubham-astro:
## Description

This PR makes the gitSyncRelay.metrics.enabled setting configurable via the platform's global Helm values, instead of hardcoding it to false in the Houston configmap.

## Related Issues

https://github.com/astronomer/issues/issues/7928

## Merging

Master, 1.2

### DIFF
--- a/.github/workflows/pr_checks.yml
+++ b/.github/workflows/pr_checks.yml
@@ -22,6 +22,7 @@ jobs:
         env:
           PR_BODY: ${{ github.event.pull_request.body }}
           KTLO: ${{ contains(github.event.pull_request.labels.*.name, 'ktlo') }}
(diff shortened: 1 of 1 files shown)


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -251,6 +251,8 @@ data:
       {{- end }}
 
         gitSyncRelay:
+          metrics:
+            enabled: {{ .Values.global.gitSyncRelay.metrics.enabled }}
```
pgvishnuram, line 255: should we hold this config till 2.0 or lets proceed with this

PR #3124, opened 2026-03-25
Dho dr fixes

danielhoherd:
## Description

This includes fixes from Adhip and myself that we came up with while manually testing the end-to-end backend failover logic.


pgvishnuram:
@danielhoherd all these are going through one feature flag right like glonal.failoverEnabled something like this

charts/astronomer/templates/navigator/navigator-deployment.yaml
pgvishnuram, line 159: @danielhoherd you can move tall this to helper templates

danielhoherd: We can circle back on this. Right now it's working and I don't want that to block merging this.

PR #3125, opened 2026-03-26
PINF-340: global feature flag for dp failover

adhipjos:
## Description

Added a single global feature flag **global.features.dataPlaneFailover.enabled** that acts as a master switch. The same flag is set on both CP and DP chart installs. The templates use global.plane.mode to determine which components to enable:

  - mode=data: enables pilot, flightDeck, COMMANDER_DATAPLANE_FAILOVER_ENABLED
  - mode=control: enables navigator, dp-link, houston dispatcher
  - mode=unified: no effect. Unified installs must enable components individually          

### Standalone vs Global Flag Enablement Matrix

| Component  | Standalone (`.enabled`) works in | Global flag auto-enables in | Never renders in |
|------------|----------------------------------|-----------------------------|------------------|
| Pilot      | data, unified                    | data only                   | control          |
| FlightDeck | data, unified                    | data only                   | control          |
| Navigator  | control, unified                 | control only                | data             |
| DP-Link    | control, unified                 | control only                | data             |
| Dispatcher | control, unified                 | control only                | data             |                                                                                                                    

## Related Issues

[PINF-340: DP registration updates and Failover feature flag](https://linear.app/astronomer/issue/PINF-340/dp-registration-updates-and-failover-feature-flag)

## Testing

## Merging

### DIFF
--- a/charts/astronomer/templates/commander/commander-deployment.yaml
+++ b/charts/astronomer/templates/commander/commander-deployment.yaml
@@ -81,7 +81,7 @@ spec:
             {{- $required := dict "readOnlyRootFilesystem" true }}
             {{- merge $required .Values.securityContext | toYaml | nindent 12 }}
 
-      {{- if or .Values.flightDeck.enabled (and .Values.global.features.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data")) }}
+      {{- if or .Values.flightDeck.enabled (and .Values.global.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data")) }}
         - name: flightdeck-bootstrapper
           image: {{ template "dbBootstrapper.image" . }}
           imagePullPolicy: {{ .Values.images.dbBootstrapper.pullPolicy }}
@@ -204,7 +204,7 @@ spec:
             - name: {{ $config.name }}
               value: {{ $config.value | quote }}
             {{- end }}
-            {{- if or .Values.flightDeck.enabled (and .Values.global.features.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data")) }}
+            {{- if or .Values.flightDeck.enabled (and .Values.global.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data")) }}
             - name: COMMANDER_FLIGHTDECK_DSN
               valueFrom:
                 secretKeyRef:
--- a/charts/astronomer/templates/dp-link/dp-link-serviceaccount.yaml
+++ b/charts/astronomer/templates/dp-link/dp-link-serviceaccount.yaml
@@ -2,7 +2,7 @@
 ## DP-Link Service Account ##
 #############################
 {{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
-{{- if and (or .Values.dpLink.enabled (and .Values.global.features.dataPlaneFailover.enabled (eq .Values.global.plane.mode "control"))) .Values.dpLink.serviceAccount.create .Values.global.rbacEnabled }}
+{{- if and (or .Values.dpLink.enabled (and .Values.global.dataPlaneFailover.enabled (eq .Values.global.plane.mode "control"))) .Values.dpLink.serviceAccount.create .Values.global.rbacEnabled }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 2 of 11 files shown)


charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -81,7 +81,7 @@ spec:
             {{- $required := dict "readOnlyRootFilesystem" true }}
             {{- merge $required .Values.securityContext | toYaml | nindent 12 }}
 
-      {{- if .Values.flightDeck.enabled }}
+      {{- if and (or .Values.flightDeck.enabled (and .Values.global.features.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data"))) (or (eq .Values.global.plane.mode "data") (eq .Values.global.plane.mode "unified")) }}
```
pgvishnuram: why do we need a "or" logic here all these can be grpuped under .Values.global.features.dataPlaneFailover.enabled

adhipjos: This is to allow individual components to be reusable for other features. Flightdeck is essentially a DP scoped data store and can be used to store state if required by other features

danielhoherd: We have a check for the plane on line 4, so this can be vastly simplified which makes it way easier to understand:

```suggestion
      {{- if or (.Values.flightDeck.enabled (and .Values.global.features.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data"))) }}
```

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -204,7 +204,7 @@ spec:
             - name: {{ $config.name }}
               value: {{ $config.value | quote }}
             {{- end }}
-            {{- if .Values.flightDeck.enabled }}
+            {{- if and (or .Values.flightDeck.enabled (and .Values.global.features.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data"))) (or (eq .Values.global.plane.mode "data") (eq .Values.global.plane.mode "unified")) }}
```
danielhoherd: Same as above
```suggestion
            {{- if or (.Values.flightDeck.enabled (and .Values.global.features.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data"))) }}
```

values.yaml
```diff
@@ -70,6 +70,13 @@ global:
         create: false
         names: []
 
+    # dataPlaneFailover: master switch for DR failover
+    # On data plane (mode=data): enables pilot, flightDeck, and commander StartFlight RPC
+    # On control plane (mode=control): enables navigator, dp-link, and houston dispatcher
+    # On unified (mode=unified): no effect — enable components individually
+    dataPlaneFailover:
+      enabled: false
```
danielhoherd: IMHO we don't need `global.features`, it should just be `global`. Technically, everything in the app is a feature. `namespacePools` is the only feature that lives under `global.features` and IMHO we should work to back that out to `global`.

charts/astronomer/templates/commander/commander-deployment.yaml
adhipjos: ah that makes sense! way simpler. updating

values.yaml
adhipjos: Valid point. Question is, is this change better left to do for both namespacepool as well as failover together at a later time or we just do failover now and have that inconsistency

adhipjos: Updated to global.dataPlaneFailover

charts/astronomer/templates/dp-link/dp-link-serviceaccount.yaml
```diff
@@ -2,7 +2,7 @@
 ## DP-Link Service Account ##
 #############################
 {{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
-{{- if and .Values.dpLink.enabled .Values.dpLink.serviceAccount.create .Values.global.rbacEnabled }}
+{{- if and (or .Values.dpLink.enabled (and .Values.global.dataPlaneFailover.enabled (eq .Values.global.plane.mode "control"))) .Values.dpLink.serviceAccount.create .Values.global.rbacEnabled }}
```
karankhanchandani, line 5: We will need to rework this PR a bit in the end when we merge both feature flag config hierarchy and this, because we have now

```
global:
   rbac:
      enabled: true
```

instead of 

```
global
    rbacEnabled : true
```

PR #3141, opened 2026-03-31
PINF-407: Add kubelet and cAdvisor scrape jobs for pod-level metrics

Shubham-astro:
## Description

This PR brings back kubelet and cAdvisor Prometheus scrape targets and associated recording rules enabling collection of pod-level CPU and memory metrics via the Kubernetes API server proxy.

## Related Issues

https://linear.app/astronomer/issue/PINF-407/add-helm-opt-in-flags-for-node-exporter-and-pod-metrics

## Testing

- Added the tests back

## Merging

Master, 2.0

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -202,6 +202,8 @@ global:
   postgresqlEnabled: false
 
   prometheusPostgresExporterEnabled: false
+  nodeExporter:
+    enabled: false
 
   # Configure vector to gather logs from all available namespaces
   manualNamespaceNamesEnabled: false
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -33,6 +33,10 @@ dependencies:
     condition: global.kubeStateEnabled
     tags:
       - monitoring
+  - name: prometheus-node-exporter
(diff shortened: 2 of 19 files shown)


charts/prometheus/values.yaml
```diff
@@ -7,6 +7,9 @@ affinity: {}
 tolerations: []
 podLabels: {}
 
+# Enables kubelet and cAdvisor scrape jobs for pod-level metrics
+cadvisorEnabled: false
```
pgvishnuram: can this be wrapped under prometheus config section ??

pgvishnuram: we already have them

values.yaml
```diff
@@ -494,6 +494,10 @@ grafana:
 #################################
 prometheus:
   podLabels: {}
+
+  # Enables kubelet and cAdvisor scrape jobs for pod-level metrics
+  cadvisorEnabled: false
```
rishkarajgi:
```suggestion
  cadvisor
      enabled: false
```

PR #3145, opened 2026-04-01
Added create_linear_milestone job

sreenusuuda:
---
  Automate Linear milestone creation on Helm chart release

  Summary

  - Adds a new create_linear_milestone CircleCI job that creates a milestone in the Linear [APC RC tracker](https://linear.app/astronomer/project/apc-rc-tracker-5234a43e361d/overview) project whenever a new beta Helm chart is released
  - The job runs as a downstream step in the build-and-release-helm-chart workflow, triggered after sign-release-image on release-*.* branches
  - The chart version is derived automatically from the workspace artifact (astronomer-<version>.tgz) — no manual input required
  - Milestone is named <version> - <UTC datetime> with description "New Beta chart is released"
  
  Issue:
  
  https://linear.app/astronomer/issue/APC-1213/create-a-script-to-create-milestone-when-ever-new-rc-is-released


.circleci/config.yml
```diff
@@ -339,7 +339,61 @@ jobs:
...
+              --arg projectId "$LINEAR_PROJECT_ID" \
+              --arg name "$MILESTONE_NAME" \
+              --arg description "New Beta chart is released" \
+              '{"query": "mutation CreateMilestone($projectId: String!, $name: String!, $description: String!) { projectMilestoneCreate(input: { projectId: $projectId, name: $name, description: $description }) { success projectMilestone { id name } } }", "variables": { "projectId": $projectId, "name": $name, "description": $description }}')
+
+            echo "Creating Linear milestone: $MILESTONE_NAME"
+            curl -sS -X POST "$GRAPHQL_URL" \
+              -H "Content-Type: application/json" \
+              -H "Authorization: ${LINEAR_API_KEY}" \
+              --data "$CREATE_MILESTONE_PAYLOAD" | jq .
```
danielhoherd: This should be a shell script in `bin`

Also you could write it in python which would probably be better all around.

danielhoherd (changes requested):
All these changes need to go into `.circleci/config.yml.j2` or they will be destroyed when we run pre-commit..

.circleci/config.yml
```diff
@@ -339,7 +339,61 @@ jobs:
...
+      - image: cimg/base:stable
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - run:
+          name: Install jq
+          command: |
+            sudo apt-get update
+            sudo apt-get install -y jq
+
```
danielhoherd: If you use ci-pre-commit you won't have to install `jq`

```suggestion
    docker:
      - image: quay.io/astronomer/ci-pre-commit:{{ ci_runner_version }}
    steps:
      - attach_workspace:
          at: /tmp/workspace

```

.circleci/config.yml
```diff
@@ -339,7 +339,61 @@ jobs:
...
+      - run:
+          name: Create Linear milestone
+          command: |
+            set -eo pipefail
+
+            : "${LINEAR_API_KEY:?LINEAR_API_KEY is required}"
```
danielhoherd: This is an obscure syntax and I seem to recall it being problematic in some cases. We should just check for a populated string `[[ -n "${LINEAR_API_KEY" ]] || { echo "LINEAR_API_KEY is required" ; exit 1 ; }`

.circleci/config.yml
sreenusuuda: This code isn’t related to the Astronomer repository—it’s only used to update the Linear project whenever a beta release happens. Given that, do you still think it makes sense to keep it in the Astronomer repo’s bin directory?

sreenusuuda: Removed the above code and created python script in bin directory bin/create-linear-beta-milestone.py

.circleci/config.yml
sreenusuuda: addressed

.circleci/config.yml
sreenusuuda: NA as we have removed this code and added python script.

Simpcyclassy (commented):
The Linear milestone creation script and CircleCI job look correct. The script reads the chart version from the workspace artifact, queries Linear for the project ID, and creates a milestone — no hardcoded IDs, uses the CircleCI `linear` context for the API token.

One thing to verify: the `linear` CircleCI context needs to expose a `LINEAR_API_KEY` environment variable that the script reads. The script uses `os.environ.get('LINEAR_API_KEY')` — make sure the context variable name matches exactly what the script expects. If it does, this is fine.

Also, both `config.yml` and `config.yml.j2` are updated consistently, which is correct.

Left as a comment rather than a block — the question about the context variable name is worth confirming before merge but is not a code defect.

.circleci/config.yml.j2
```diff
@@ -253,6 +253,26 @@ jobs:
...
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - checkout
+      - run:
+          name: Install uv
+          command: |
+            set -eo pipefail
+            export UV_VERSION="{{ uv_version }}"
+            curl -fsSL https://astral.sh/uv/install.sh | sh
```
danielhoherd: ci-pre-commit already has `uv` installed
```suggestion
```

bin/create-linear-beta-milestone.py
```diff
@@ -0,0 +1,242 @@
...
+_FIND_PROJECT_QUERY = """
+query FindProject($name: String!) {
+  projects(filter: { name: { eq: $name } }) {
+    nodes {
+      id
+      name
+    }
+  }
+}
+"""
```
danielhoherd, line 34: Is this necessary? Won't this always return the same project, or is this going to be different depending on if we are working on release-1.x vs release-2.x ?

danielhoherd: Found answer on a meeting. This will be different per-release.

danielhoherd (approved):
Looks good. I reviewed these changes with @sreenusuuda on a call. These changes do not affect the software that is shipped to customers, only dev processes, so these changes do not need any QA.

PR #3150, opened 2026-04-06
PINF-384: Add a precommit hook to check consistency between chart versions

Shubham-astro:
## Description

There was no automated check to ensure version and appVersion are consistent across Chart.yaml files in the repo. This allowed mismatches to slip through undetected, which previously caused deployment failures when Commander tried to install a chart version that didn't match what was indexed.

## Related Issues

https://linear.app/astronomer/issue/PINF-384/incorrect-version-tooltip-after-helm-upgrade

## Testing

- Tested that if versions matched pre-commits passed:
Screenshot 2026-04-06 at 3 33 00 PM

- If versions were different precommit failed: 
Screenshot 2026-04-06 at 3 34 56 PM

## Merging

Master, 1.0,1.1,1.2,2.0,0.37

### DIFF
--- a/bin/check-chart-version-consistency.sh
+++ b/bin/check-chart-version-consistency.sh
@@ -5,7 +5,7 @@ unique_versions=$(awk '/^(version|appVersion)/ {print $2}' Chart.yaml charts/ast
 
 if [[ "$unique_versions" -ne 1 ]]; then
   echo "ERROR: version/appVersion mismatch between Chart.yaml and charts/astronomer/Chart.yaml"
-  awk '/^(version|appVersion)/ {print FILENAME": "$0}' Chart.yaml charts/astronomer/Chart.yaml
+  grep -E '^(version|appVersion)' Chart.yaml charts/astronomer/Chart.yaml
   exit 1
 fi
 


bin/check-chart-version-consistency.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -e
+
+unique_versions=$(awk '/^(version|appVersion)/ {print $2}' Chart.yaml charts/astronomer/Chart.yaml | sort -u | wc -l)
+
+if [[ "$unique_versions" -ne 1 ]]; then
+  echo "ERROR: version/appVersion mismatch between Chart.yaml and charts/astronomer/Chart.yaml"
+  awk '/^(version|appVersion)/ {print FILENAME": "$0}' Chart.yaml charts/astronomer/Chart.yaml
```
danielhoherd: I can't believe I never knew about the FILENAME var in awk! However this can be easier done with
```suggestion
  grep -E '^(version|appVersion)' Chart.yaml charts/astronomer/Chart.yaml
```

PR #3156, opened 2026-04-08
PINF-432: Containerd 2.x support for private CA cert daemonset

rishkarajgi:
## Summary

GKE 1.33+ ships **containerd 2.x**, which uses the CRI plugin namespace `io.containerd.cri.v1.images` instead of `io.containerd.grpc.v1.cri`. The previous private-CA DaemonSet logic could target the wrong namespace or miss `config_path`, so registry TLS trust did not apply reliably on newer nodes.

This change **replaces the inlined shell script** with a **Python 3 script** (`files/update-containerd-certs.py`) that runs **on the host** via `nsenter` and the node's `python3`. The script **detects containerd major version at runtime** (`containerd --version`) and applies the correct plugin namespace, `certs.d` / `hosts.toml` updates, and one-time `config_path` injection plus restart when required.

The Helm chart **embeds** the script in the ConfigMap (`.Files.Get`) and passes **environment variables** on the cert-copier container: `REGISTRY_HOST` (respects data-plane `domainPrefix` when `global.plane.mode` is `data`), `CONTAINERD_HOST_PATH`, `CERT_CONFIG_PATH`, and `PRIVATE_CA_CERTS_DIR`.

The **`containerdVersion` chart value is removed**; behaviour is driven by live detection instead of operator-configured version strings.

## What the Python script does (high level)

- **Version detection:** `nsenter` + `containerd --version` → major version 1 or 2 (defaults to 2 if detection fails).
- **Backups:** timestamped `config.toml` copies plus a stable `config.toml.bak` for idempotent edits.
- **containerd 2.x without `config_path`:** injects `config_path` under `io.containerd.cri.v1.images` and restarts containerd once.
- **When `config_path` is present (typical 2.x and some 1.x):** copies CA PEMs into `certs.d/<registry>/`, writes `hosts.toml`, uses **SHA-256** checksums to avoid unnecessary work; **no restart** on cert rotation.
- **Legacy path (containerd 1.x without `config_path`):** inline `config.toml` updates and restart on change (same class of behaviour as before).

## Linear issue

https://linear.app/astronomer/issue/PINF-432/containerd-cert-patch-not-supported-with-new-gke-versions

## CI tooling

`bin/install-ci-tools.py`: `tarfile.extract()` no longer uses `filter=` so Helm installs work on Python before 3.12.

## Testing

- Chart rendering: `tests/chart_tests/test_containerd_privateca.py`
- Script unit tests and GKE `config.toml` fixtures: `tests/chart_tests/test_update_containerd_certs.py`, `tests/fixtures/gke_1_32_containerd_1x_config.toml`, `tests/fixtures/gke_1_33_containerd_2x_config.toml`

## Release note

**Upgrade note:** Private-CA containerd integration now uses a **host Python** script and **runtime containerd version detection**. Operators **no longer set** `global.privateCaCertsAddToHost.containerdVersion`; remove it from values if present.

**GKE 1.33+ / containerd 2.x:** Registry trust is applied under the correct plugin namespace with `certs.d` / `hosts.toml` where applicable.

**Older nodes (containerd 1.x):** The script still supports the legacy inline `config.toml` path when `config_path` is not set.


templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
```diff
@@ -15,76 +15,124 @@ metadata:
...
         echo "No /hostcontainerd/config.toml found, exiting. Is this a containerd node?";
         exit 1;
     fi
 
-    # backup the original ca certificates it is not already backed up
+    # backup the original config.toml if not already backed up
     if [ ! -f /hostcontainerd/config.toml.bak ]; then
-        # backup /hostcontainerd/config.toml
         cp /hostcontainerd/config.toml /hostcontainerd/config.toml.bak
     fi
```
danielhoherd: IMHO this backup should have had epoch seconds appended to it, so successive runs would not overwrite the previous backup.

templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
```diff
@@ -15,76 +15,124 @@ metadata:
 data:
   update-containerd-certs.sh: |
     #!/usr/bin/env sh
+    set -e
+
+    CONTAINERD_VERSION="{{ .Values.global.privateCaCertsAddToHost.containerdVersion }}"
+    REGISTRY_HOST="registry.{{ .Values.global.baseDomain }}"
```
pgvishnuram: this is incorrect and will fail with DP mode

templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
```diff
@@ -15,76 +15,124 @@ metadata:
...
-          else
-            echo "no change in config sleep for 1 second";
-            sleep 1;
-          fi
+        if [ "${CONFIG_PATH_SET}" = "true" ]; then
+            # --- hosts.toml path (containerd 2.x, or 1.x with config_path set) ---
+            {{- if .Values.global.privateCaCertsAddToHost.containerdConfigToml }}
+            # Use custom hosts.toml content from containerdConfigToml if provided
+            cat << CUSTOM_EOF > "${CERTS_DIR}/hosts.toml.new"
+{{ include "containerd.configToml" . | indent 8 }}
```
danielhoherd:
```suggestion
            {{- include "containerd.configToml" . | nindent 8 }}
```

danielhoherd (changes requested):
I have very low confidence that this is going to succeed without breaking things. The script already had some issues, and I feel like this exacerbates those issues.

## Here's what I think we should do
1. rewrite this in python. the python script should include a docstring that explains exactly what happens for each supported version of containerd. (See the "Supplemental data" section below.)
2. rewrite it in a way that does not modify the contents of the script depending helm values. the script should always have the same content, and flow control should happen within the script itself, not by modifying the contents of the script based off of the helm values. any inputs to the script should be given through env vars, not through script modification. these env vars should be injected through the `templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml` container env.
3. Add tests that run the script on actual known containerd toml files, and then parse the resulting toml file and ensure it has contents that we expect. (See the "Supplemental data" section below.)
4. We should probably determine the containerd version from the environment, not through a configuration value in helm. If a cluster administrator upgrades the nodes, we do not want the application administrators to need to modify their helm values and do another `helm upgrade`. (See the "Supplemental data" section below.)
5. The script should be a standalone file that can be linted like any other file in our repo, and should be inserted into the configmap with `.Files.Get`

## Supplemental data

### containerd 1.x running in gke k8s 1.32
```
gke-containerd-test-1-default-pool-5b36e8fc-rc9m / # cat /etc/containerd/config.toml
version = 2
required_plugins = ["io.containerd.grpc.v1.cri"]
# Kubernetes doesn't use containerd restart manager.
disabled_plugins = ["io.containerd.internal.v1.restart"]
oom_score = -999

[debug]
  level = "info"

[grpc]
  gid = 412

[plugins."io.containerd.grpc.v1.cri"]
  enable_cdi = true
  stream_server_address = "127.0.0.1"
  max_container_log_line_size = 262144
  sandbox_image = "us-central1-artifactregistry.gcr.io/gke-release/gke-release/pause:3.8@sha256:880e63f94b145e46f1b1082bb71b85e21f16b99b180b9996407d61240ceb9830"
  image_pull_progress_timeout = "5m"
[plugins."io.containerd.grpc.v1.cri".cni]
  bin_dir = "/home/kubernetes/bin"
  conf_dir = "/etc/cni/net.d"
  conf_template = "/home/containerd/cni.template"
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
  SystemdCgroup = true
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
  endpoint = ["https://mirror.gcr.io","https://registry-1.docker.io"]
[metrics]
 address = "127.0.0.1:1338"
[plugins."io.containerd.grpc.v1.cri".containerd]
  default_runtime_name = "runc"
  discard_unpacked_layers = true
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
  runtime_type = "io.containerd.runc.v2"
gke-containerd-test-1-default-pool-5b36e8fc-rc9m / # containerd --version
containerd github.com/containerd/containerd 1.7.29 442cb34bda9a6a0fed82a2ca7cade05c5c749582
gke-containerd-test-1-default-pool-5b36e8fc-rc9m / # python --version
Python 3.8.19
```

### containerd 2.x running in gke k8s 1.33

```
gke-dev-cluster-terraform-20240611092-5bc44158-ipo1 / # cat /etc/containerd/config.toml
version = 2
required_plugins = ["io.containerd.grpc.v1.cri"]
# Kubernetes doesn't use containerd restart manager.
disabled_plugins = ["io.containerd.internal.v1.restart"]
oom_score = -999

[debug]
  level = "info"

[grpc]
  gid = 412

[plugins."io.containerd.grpc.v1.cri"]
  enable_cdi = true
  stream_server_address = "127.0.0.1"
  max_container_log_line_size = 262144
  sandbox_image = "gke.gcr.io/pause:3.8@sha256:880e63f94b145e46f1b1082bb71b85e21f16b99b180b9996407d61240ceb9830"
  image_pull_progress_timeout = "5m"
[plugins."io.containerd.grpc.v1.cri".cni]
  bin_dir = "/home/kubernetes/bin"
  conf_dir = "/etc/cni/net.d"
  conf_template = ""
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
  SystemdCgroup = true
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
  endpoint = ["https://mirror.gcr.io","https://registry-1.docker.io"]
[metrics]
 address = "127.0.0.1:1338"
[plugins."io.containerd.grpc.v1.cri".containerd]
  default_runtime_name = "runc"
  discard_unpacked_layers = true
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
  runtime_type = "io.containerd.runc.v2"
gke-dev-cluster-terraform-20240611092-5bc44158-ipo1 / # containerd --version
containerd github.com/containerd/containerd/v2 2.0.7 4ac6c20c7bbf8177f29e46bbdc658fec02ffb8ad
gke-dev-cluster-terraform-20240611092-5bc44158-ipo1 / # python --version
Python 3.11.11
```

templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
rishkarajgi: done

templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
rishkarajgi: added a check for mode "data"

templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
rishkarajgi: updated

configs/use-docker-io.yaml
```diff
@@ -57,7 +57,7 @@ global:
     repository: docker.io/astronomerinc/ap-vector
   privateCaCertsAddToHost:
     certCopier:
-      repository: docker.io/astronomerinc/ap-base
+      repository: docker.io/astronomerinc/ap-db-bootstrapper
```
rishkarajgi, line 60: using db-bootstrapper as it contains python binaries

tests/fixtures/gke_1_33_containerd_2x_config.toml
```diff
@@ -0,0 +1,33 @@
+version = 2
```
danielhoherd, line 1: This should go in `tests/data_files`

tests/fixtures/gke_1_32_containerd_1x_config.toml
```diff
@@ -0,0 +1,33 @@
+version = 2
```
danielhoherd, line 1: This should go in `tests/data_files`

tests/chart_tests/test_containerd_privateca.py
```diff
@@ -242,5 +234,72 @@ def test_containerd_privateca_daemonset_enabled_with_affinity_and_toleration(sel
...
+
+        # Verify the daemonset env var uses base registry host (no prefix)
+        container = docs[0]["spec"]["template"]["spec"]["containers"][0]
+        env = {e["name"]: e["value"] for e in container["env"]}
+        assert env["REGISTRY_HOST"] == "registry.example.com"
+
+        # Verify the daemonset volume path
+        volumes = docs[0]["spec"]["template"]["spec"]["volumes"]
+        hostcerts_vol = next(v for v in volumes if v["name"] == "hostcerts")
+        assert hostcerts_vol["hostPath"]["path"] == "/etc/containerd/certs.d/registry.example.com/"
```
danielhoherd, line 305: These two tests should be one test that is parametrized on `plane`. There are several other test cases like this that can be used as a reference.

tests/chart_tests/test_containerd_privateca.py
```diff
@@ -123,6 +122,55 @@ def test_containerd_privateca_daemonset_enabled(self, kube_version):
...
+            },
+        )
+
+        configmap = docs[1]
+        assert configmap["kind"] == "ConfigMap"
+        script = configmap["data"]["update-containerd-certs.py"]
+        # Verify it's the Python script with key features
+        assert "detect_containerd_version" in script
+        assert "generate_hosts_toml" in script
+        assert "REGISTRY_HOST" in script
```
danielhoherd, line 172: These two tests could be merged into one test case named `test_containerd_privateca_defaults`. We often do this because `render_chart()` is expensive, so we try to reduce the number of times we call it.

tests/chart_tests/test_update_containerd_certs.py
```diff
@@ -0,0 +1,252 @@
...
+    @patch("subprocess.run")
+    def test_detects_containerd_2x(self, mock_run):
+        mock_run.return_value = MagicMock(
+            stdout="containerd github.com/containerd/containerd/v2 2.0.7 4ac6c20c7bbf8177f29e46bbdc658fec02ffb8ad",
+        )
+        assert script.detect_containerd_version() == 2
+
+    @patch("subprocess.run")
+    def test_defaults_to_2_on_failure(self, mock_run):
+        mock_run.side_effect = FileNotFoundError("nsenter not available")
```
danielhoherd, line 64: It's great that we have a test on this specific failure scenario, except that if we do not have `nsenter` then the whole thing will fail, so this test case is actually invalid. We should rework the script to die with useful information if `nsenter` is not available, and test that behavior here.

tests/chart_tests/test_update_containerd_certs.py
```diff
@@ -0,0 +1,252 @@
...
+        assert script.detect_containerd_version() == 2
+
+    @patch("subprocess.run")
+    def test_defaults_to_2_on_empty_output(self, mock_run):
+        mock_run.return_value = MagicMock(stdout="")
+        assert script.detect_containerd_version() == 2
```
danielhoherd, line 70: I don't think this is a behavior we want. If we do not get an expected output, that is an indicator that we are running in an unknown environment, and IMHO we should die with useful information in that case.

tests/chart_tests/test_update_containerd_certs.py
```diff
@@ -0,0 +1,252 @@
...
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+FIXTURES_DIR = Path(__file__).parent.parent / "fixtures"
+SCRIPT_PATH = Path(__file__).parent.parent.parent / "files" / "update-containerd-certs.py"
+
+# Load the script module dynamically
+spec = importlib.util.spec_from_file_location("update_containerd_certs", SCRIPT_PATH)
+script = importlib.util.module_from_spec(spec)
```
danielhoherd, line 21: Because we do things like `script.CONFIG_TOML = config_toml` we should turn `script` into a function-scoped fixture.

files/update-containerd-certs.py
```diff
@@ -0,0 +1,260 @@
...
+
+    This triggers a one-time containerd restart.
+    """
+    plugin_ns = PLUGIN_NS[containerd_version]
+    stanza = f'\n[{plugin_ns}]\n  config_path = "{CERT_CONFIG_PATH}"\n'
+
+    content = CONFIG_TOML_BAK.read_text()
+    content += stanza
+
+    CONFIG_TOML.write_text(content)
```
danielhoherd, line 151: This does string editing of a serialized data file. IMHO we should instead load the file into python, change the variables, and then write the file back out to disk. Doing it that way will be less brittle.

files/update-containerd-certs.py
```diff
@@ -0,0 +1,260 @@
...
+    Returns 1 or 2. Defaults to 2 if detection fails.
+    """
+    try:
+        result = subprocess.run(
+            ["nsenter", "--target", "1", "--mount", "--uts", "--ipc", "--net", "--pid", "containerd", "--version"],
+            capture_output=True,
+            text=True,
+            timeout=10,
+        )
+        version_output = result.stdout.strip()
```
Shubham-astro, line 84: If nsenter exits non-zero (wrong permissions, not on a systemd node, etc.), result.stdout is empty and parsing silently falls through to the `log.warning("Defaulting to containerd version 2")`. On a containerd 1.x cluster this would inject the wrong plugin namespace (`io.containerd.cri.v1.images` instead of `io.containerd.grpc.v1.cri`) and silently misconfigure the node. This can be something like:
```
if result.returncode != 0:
    log.warning("containerd --version exited %d: %s", result.returncode, result.stderr.strip())
```

files/update-containerd-certs.py
```diff
@@ -0,0 +1,260 @@
...
+
+            current_checksum = checksum_of_files(ca_crt_path)
+            content_checksum = hashlib.sha256(hosts_content.encode()).hexdigest()
+            combined = current_checksum + content_checksum
+
+            if combined != last_checksum:
+                hosts_toml_path.write_text(hosts_content)
+                log.info("Updated %s", hosts_toml_path)
+                last_checksum = combined
+                log.info("No containerd restart required for hosts.toml updates.")
```
Shubham-astro, line 237: I think we can add a sleep here as well as after writing `hosts.toml`, the loop immediately calls `copy_ca_certs()` and recomputes checksums again. It works (since last_checksum is updated), but it has a tight cycle. A continue or restructuring the loop to always sleep at the bottom would be cleaner

files/update-containerd-certs.py
```diff
@@ -0,0 +1,260 @@
...
+        # Example outputs:
+        #   containerd github.com/containerd/containerd 1.7.29 ...
+        #   containerd github.com/containerd/containerd/v2 2.0.7 ...
+        for token in version_output.split():
+            if token and token[0].isdigit():
+                major = int(token.split(".")[0])
+                if major in (1, 2):
+                    log.info("Detected containerd major version: %d", major)
+                    return major
+    except (OSError, subprocess.SubprocessError, ValueError, IndexError) as exc:
```
Shubham-astro, line 96: `IndexError` can't be raised by subprocess.run(), it can only raise exceptions related to starting the process (e.g., `FileNotFoundError`) or `subprocess.CalledProcessError`, hence it is not needed here.

files/update-containerd-certs.py
```diff
@@ -0,0 +1,260 @@
...
+        for pem_file in secret_dir.glob("*.pem"):
+            shutil.copy2(pem_file, ca_dest)
+            log.info("Copied %s -> %s", pem_file, ca_dest)
+
+
+# ---------------------------------------------------------------------------
+# Main loop
+# ---------------------------------------------------------------------------
+def main() -> None:
+    if not REGISTRY_HOST:
```
Shubham-astro, line 188: I think we need to add more checks here like malformed hostname (spaces, slashes) would produce an invalid hosts.toml and broken TOML/directory paths. At a minimum, log a warning for obviously bad values.

rishkarajgi:
> I have very low confidence that this is going to succeed without breaking things. The script already had some issues, and I feel like this exacerbates those issues.
> 
> ## Here's what I think we should do
> 1. rewrite this in python. the python script should include a docstring that explains exactly what happens for each supported version of containerd. (See the "Supplemental data" section below.)
> 2. rewrite it in a way that does not modify the contents of the script depending helm values. the script should always have the same content, and flow control should happen within the script itself, not by modifying the contents of the script based off of the helm values. any inputs to the script should be given through env vars, not through script modification. these env vars should be injected through the `templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml` container env.
> 3. Add tests that run the script on actual known containerd toml files, and then parse the resulting toml file and ensure it has contents that we expect. (See the "Supplemental data" section below.)
> 4. We should probably determine the containerd version from the environment, not through a configuration value in helm. If a cluster administrator upgrades the nodes, we do not want the application administrators to need to modify their helm values and do another `helm upgrade`. (See the "Supplemental data" section below.)
> 5. The script should be a standalone file that can be linted like any other file in our repo, and should be inserted into the configmap with `.Files.Get`
> 
> ## Supplemental data
> ### containerd 1.x running in gke k8s 1.32
> ```
> gke-containerd-test-1-default-pool-5b36e8fc-rc9m / # cat /etc/containerd/config.toml
> version = 2
> required_plugins = ["io.containerd.grpc.v1.cri"]
> # Kubernetes doesn't use containerd restart manager.
> disabled_plugins = ["io.containerd.internal.v1.restart"]
> oom_score = -999
> 
> [debug]
>   level = "info"
> 
> [grpc]
>   gid = 412
> 
> [plugins."io.containerd.grpc.v1.cri"]
>   enable_cdi = true
>   stream_server_address = "127.0.0.1"
>   max_container_log_line_size = 262144
>   sandbox_image = "us-central1-artifactregistry.gcr.io/gke-release/gke-release/pause:3.8@sha256:880e63f94b145e46f1b1082bb71b85e21f16b99b180b9996407d61240ceb9830"
>   image_pull_progress_timeout = "5m"
> [plugins."io.containerd.grpc.v1.cri".cni]
>   bin_dir = "/home/kubernetes/bin"
>   conf_dir = "/etc/cni/net.d"
>   conf_template = "/home/containerd/cni.template"
> [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
>   SystemdCgroup = true
> [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
>   endpoint = ["https://mirror.gcr.io","https://registry-1.docker.io"]
> [metrics]
>  address = "127.0.0.1:1338"
> [plugins."io.containerd.grpc.v1.cri".containerd]
>   default_runtime_name = "runc"
>   discard_unpacked_layers = true
> [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
>   runtime_type = "io.containerd.runc.v2"
> gke-containerd-test-1-default-pool-5b36e8fc-rc9m / # containerd --version
> containerd github.com/containerd/containerd 1.7.29 442cb34bda9a6a0fed82a2ca7cade05c5c749582
> gke-containerd-test-1-default-pool-5b36e8fc-rc9m / # python --version
> Python 3.8.19
> ```
> 
> ### containerd 2.x running in gke k8s 1.33
> ```
> gke-dev-cluster-terraform-20240611092-5bc44158-ipo1 / # cat /etc/containerd/config.toml
> version = 2
> required_plugins = ["io.containerd.grpc.v1.cri"]
> # Kubernetes doesn't use containerd restart manager.
> disabled_plugins = ["io.containerd.internal.v1.restart"]
> oom_score = -999
> 
> [debug]
>   level = "info"
> 
> [grpc]
>   gid = 412
> 
> [plugins."io.containerd.grpc.v1.cri"]
>   enable_cdi = true
>   stream_server_address = "127.0.0.1"
>   max_container_log_line_size = 262144
>   sandbox_image = "gke.gcr.io/pause:3.8@sha256:880e63f94b145e46f1b1082bb71b85e21f16b99b180b9996407d61240ceb9830"
>   image_pull_progress_timeout = "5m"
> [plugins."io.containerd.grpc.v1.cri".cni]
>   bin_dir = "/home/kubernetes/bin"
>   conf_dir = "/etc/cni/net.d"
>   conf_template = ""
> [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
>   SystemdCgroup = true
> [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
>   endpoint = ["https://mirror.gcr.io","https://registry-1.docker.io"]
> [metrics]
>  address = "127.0.0.1:1338"
> [plugins."io.containerd.grpc.v1.cri".containerd]
>   default_runtime_name = "runc"
>   discard_unpacked_layers = true
> [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
>   runtime_type = "io.containerd.runc.v2"
> gke-dev-cluster-terraform-20240611092-5bc44158-ipo1 / # containerd --version
> containerd github.com/containerd/containerd/v2 2.0.7 4ac6c20c7bbf8177f29e46bbdc658fec02ffb8ad
> gke-dev-cluster-terraform-20240611092-5bc44158-ipo1 / # python --version
> Python 3.11.11
> ```

refactored in-line script to python and added relevant tests

tests/data_files/gke_1_33_containerd_2x_config.toml
```diff
@@ -0,0 +1,33 @@
+version = 2
```
rishkarajgi: done

tests/data_files/gke_1_32_containerd_1x_config.toml
```diff
@@ -0,0 +1,33 @@
+version = 2
```
rishkarajgi: done

tests/chart_tests/test_containerd_privateca.py
rishkarajgi: done

tests/chart_tests/test_containerd_privateca.py
rishkarajgi: done

rishkarajgi:
@copilot resolve the merge conflicts in this pull request

tests/chart_tests/test_update_containerd_certs.py
rishkarajgi: updated to throw a runtime error on no version detected

files/update-containerd-certs.py
rishkarajgi: updated to throw a runtime error on no version detected

files/update-containerd-certs.py
rishkarajgi: updated

files/update-containerd-certs.py
rishkarajgi: added regex validation on registry hostname

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
```diff
@@ -56,8 +56,26 @@ spec:
...
+            nsenter --target 1 --mount --uts --ipc --net --pid \
+              python3 {{ .Values.global.privateCaCertsAddToHost.containerdHostPath }}/update-containerd-certs.py
         imagePullPolicy: {{ .Values.global.privateCaCertsAddToHost.certCopier.pullPolicy }}
+        env:
+        - name: REGISTRY_HOST
+          {{- if eq .Values.global.plane.mode "data" }}
+          value: "registry.{{ .Values.global.plane.domainPrefix }}.{{ .Values.global.baseDomain }}"
+          {{- else }}
+          value: "registry.{{ .Values.global.baseDomain }}"
+          {{- end }}
```
pgvishnuram: we have a helper template inside charts/astronomer/templates we can move to common templates and use the same reference everywhere for registry hosts 

example this one {{- define "registry.ingressURL" -}}

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
```diff
@@ -93,7 +111,11 @@ spec:
         name: hostcontainerd
       - name: hostcerts
         hostPath:
+          {{- if eq .Values.global.plane.mode "data" }}
+          path: {{ .Values.global.privateCaCertsAddToHost.containerdCertConfigPath }}/registry.{{ .Values.global.plane.domainPrefix }}.{{ .Values.global.baseDomain }}/
+          {{- else }}
           path: {{ .Values.global.privateCaCertsAddToHost.containerdCertConfigPath }}/registry.{{ .Values.global.baseDomain }}/
+          {{- end }}
```
pgvishnuram: we can reuse the same helper templates to reduce the code duplications

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
```diff
@@ -56,8 +56,26 @@ spec:
           - "sh"
           - "-c"
         args:
-          - sh /cert-copy-and-toml-update.sh
+          - |
+            # Copy the Python script to the host filesystem so nsenter can access it,
+            # then execute it using the host's Python interpreter.
+            cp /scripts/update-containerd-certs.py /hostcontainerd/update-containerd-certs.py
```
pgvishnuram: why do we have to copy this script inside hostcontainerd

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
```diff
@@ -56,8 +56,26 @@ spec:
...
+            nsenter --target 1 --mount --uts --ipc --net --pid \
+              python3 {{ .Values.global.privateCaCertsAddToHost.containerdHostPath }}/update-containerd-certs.py
```
pgvishnuram: @sudarshanrampuria can you see if the containerd version can be detected w/o nsenter

tests/chart_tests/test_update_containerd_certs.py
sudarshanrampuria: Added `ensure_nsenter_available` and test case for it.

tests/chart_tests/test_update_containerd_certs.py
sudarshanrampuria: Updated

files/update-containerd-certs.py
sudarshanrampuria: Used `tomllib` to parse and update toml files.

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
sudarshanrampuria: Updated

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
sudarshanrampuria: Updated

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
sudarshanrampuria: removed copy and nsenter from root trigger anyways we are using nsenter to check containerd version and to restart

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
sudarshanrampuria: Removed nsenter from here, we are anyways using it in script subprocess call.

sudarshanrampuria:
## Testing

Deployed the chart in an existing 1.32 cluster, then updated that cluster to 1.33 to check that the script works fine for both containerd version and gke version.

templates/trust-private-ca-on-all-nodes/containerd-daemonset.yaml
```diff
@@ -56,8 +56,23 @@ spec:
...
+        - name: CONTAINERD_HOST_PATH
+          value: "/hostcontainerd"
+        - name: CERT_CONFIG_PATH
+          value: "{{ .Values.global.privateCaCertsAddToHost.containerdCertConfigPath }}"
+        - name: PRIVATE_CA_CERTS_DIR
+          value: "/private-ca-certs"
+        {{- with .Values.global.privateCaCertsAddToHost.containerdConfigToml }}
+        - name: CONTAINERD_CONFIG_TOML
+          value: |-
+{{ include "containerd.configToml" $ | indent 12 }}
```
pgvishnuram: instead of this approach should we mount them as as a configmap

tests/chart_tests/test_containerd_privateca.py
```diff
@@ -66,9 +67,8 @@ def test_containerd_privateca_daemonset_enabled(self, kube_version):
 
         assert len(docs) == 2
         self.common_tests_daemonset(docs[0])
-        assert len(docs[0]["spec"]["template"]["spec"]["containers"]) == 1
         cert_copier = docs[0]["spec"]["template"]["spec"]["containers"][0]
-        cert_copier["image"].startswith("alpine:3")
+        assert cert_copier["image"] == "quay.io/astronomer/ap-db-bootstrapper:2.0.1"
```
danielhoherd: This tag will surely change. Testing just the registry repo and image portions is sufficient.
```suggestion
        assert cert_copier["image"] == "quay.io/astronomer/ap-db-bootstrapper:"
```

tests/chart_tests/test_containerd_privateca.py
```diff
@@ -123,81 +123,100 @@ def test_containerd_privateca_daemonset_enabled(self, kube_version):
...
             },
         )
 
         assert len(docs) == 2
-        self.common_tests_daemonset(docs[0])
-        assert len(docs[0]["spec"]["template"]["spec"]["containers"]) == 1
-        cert_copier = docs[0]["spec"]["template"]["spec"]["containers"][0]
-        cert_copier["image"].startswith("alpine:3")
+        container = docs[0]["spec"]["template"]["spec"]["containers"][0]
+        env = {e["name"]: e["value"] for e in container["env"]}
```
danielhoherd: We have a function for this that handles edge cases https://github.com/astronomer/astronomer/blob/20d3701194/tests/utils/__init__.py#L8-L16

tests/chart_tests/test_containerd_privateca.py
```diff
@@ -123,81 +123,100 @@ def test_containerd_privateca_daemonset_enabled(self, kube_version):
...
-                "name": "private-ca-cert-foo",
-                "mountPath": "/private-ca-certs/private-ca-cert-foo/private-ca-cert-foo.pem",
-                "subPath": "cert.pem",
-            },
-            {
-                "name": "private-ca-cert-bar",
-                "mountPath": "/private-ca-certs/private-ca-cert-bar/private-ca-cert-bar.pem",
-                "subPath": "cert.pem",
+        container = docs[0]["spec"]["template"]["spec"]["containers"][0]
+        env = {e["name"]: e["value"] for e in container["env"]}
```
danielhoherd: We have a function for this that handles edge cases https://github.com/astronomer/astronomer/blob/20d3701194/tests/utils/__init__.py#L8-L16

tests/chart_tests/test_containerd_privateca.py
```diff
@@ -242,5 +258,47 @@ def test_containerd_privateca_daemonset_enabled_with_affinity_and_toleration(sel
...
+                        "enabled": True,
+                        "addToContainerd": True,
+                    },
+                }
+            },
+        )
+
+        assert len(docs) == 2
+        container = docs[0]["spec"]["template"]["spec"]["containers"][0]
+        env = {e["name"]: e["value"] for e in container["env"]}
```
danielhoherd: We have a function for this that handles edge cases https://github.com/astronomer/astronomer/blob/20d3701194/tests/utils/__init__.py#L8-L16

files/update-containerd-certs.py
```diff
@@ -0,0 +1,597 @@
...
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+def validate_registry_host(host: str) -> None:
+    """Ensure REGISTRY_HOST is non-empty and safe for paths and TOML.
+
+    Raises:
+        ValueError: If the value is empty, has unsafe characters, or fails validation.
+    """
```
danielhoherd, line 110: It would be good to have examples of good and bad values.

tests/chart_tests/test_update_containerd_certs.py
```diff
@@ -0,0 +1,869 @@
...
+def _completed_process(stdout: str, *, returncode: int = 0, stderr: str = "") -> MagicMock:
+    """Build a CompletedProcess-like mock for subprocess.run."""
+    return MagicMock(stdout=stdout, returncode=returncode, stderr=stderr)
+
+
+class TestValidateRegistryHost:
+    """Tests for REGISTRY_HOST validation."""
+
+    def test_accepts_typical_registry_hostname(self, script) -> None:
+        script.validate_registry_host("registry.example.com")
```
danielhoherd: We could parametrize this with more examples of valid hosts.

Are ports supported? `user:pass@host` style credentials? If not we should add them to the bad_host list below, otherwise we should parametrize them in this test.

tests/chart_tests/test_update_containerd_certs.py
```diff
@@ -0,0 +1,869 @@
...
+    def _prepare(self, script, containerd_env, fixture: Path) -> tuple[Path, Path]:
+        config_toml = containerd_env["containerd_dir"] / "config.toml"
+        config_toml_bak = containerd_env["containerd_dir"] / "config.toml.bak"
+        shutil.copy2(fixture, config_toml)
+        shutil.copy2(fixture, config_toml_bak)
+
+        script.CONFIG_TOML = config_toml
+        script.CONFIG_TOML_BAK = config_toml_bak
+        script.CERT_CONFIG_PATH = "/etc/containerd/certs.d"
+        return config_toml, config_toml_bak
```
danielhoherd, line 321: Ideally this should be a fixture.

danielhoherd (changes requested):
I left a few suggestions about functions that should be replaced, and a container comparison that will break next time we update the image which needs to be fixed. This is a lot of code, so I didn't go through it line by line. I did look at most of it though and it looks mostly good. Once those are fixed I will approve this.

templates/trust-private-ca-on-all-nodes/containerd-ca-update-script.yaml
```diff
@@ -5,86 +5,18 @@
...
-              echo "no change in config sleep for 1 second";
-              sleep 1;
-          fi
-        fi
-    done
+  update-containerd-certs.py: |-
+{{ .Files.Get "files/update-containerd-certs.py" | indent 4 }}
+  {{- with .Values.global.privateCaCertsAddToHost.containerdConfigToml }}
+  containerd-config-toml: |-
+{{ include "containerd.configToml" $ | indent 4 }}
```
danielhoherd: Style nitpick:
```suggestion
    {{- include "containerd.configToml" $ | nindent 4 }}
```

danielhoherd (approved):
There is too much code to go through line by line, but having looked over most of this it looks good to me.

tests/chart_tests/test_update_containerd_certs.py
danielhoherd: Due to the inputs, this actually doesn't work as a fixture, or at least not in a straightforward way.

PR #3161, opened 2026-04-09
PLX-300: use cspPolicy.cdn.enabled for CDN CSP toggle

Simpcyclassy:
## Description
Rename the nginx CSP “CDN sources” toggle from a flat flag to nested `cdn.enabled`, matching the unified config style.

**Rename:**

| Before | After |
|--------|--------|
| `nginx.cspPolicy.cdnEnabled` | `nginx.cspPolicy.cdn.enabled` |

**What changed**

1. **Chart defaults** — Default in `charts/nginx/values.yaml`; both ingress header ConfigMaps read the new key. If `cdn.enabled` is omitted, behavior matches the old default (CDN-related CSP entries on). Old overrides that still set `cdnEnabled` keep working until updated.

2. **Migration** — The 1.x→2.x and 0.37.x→2.x values migration scripts rewrite the old key when present (`helm_chart_values_migration_shared.py` + the two `bin/migrate-helm-chart-values-*.py` entrypoints).

3. **Docs** — Upgrade guides note this mapping and that most other `nginx` keys are unchanged.

4. **Tests** — `TestNginxCspPolicyMigration` in `tests/migration/test_migrate_helm_chart_values.py`.

## Related Issues
- https://linear.app/astronomer/issue/PLX-300/cdnenabled-feature-flag-is-not-as-per-unified-schema

## Testing

- tested migrations work
- ran pytests

## Merging

- Master
- 2.0.0

### DIFF
--- a/bin/helm_chart_values_migration_shared.py
+++ b/bin/helm_chart_values_migration_shared.py
@@ -466,6 +466,28 @@ def apply_houston_deployment_migrations(root: CommentedMap) -> list[MigrationCha
     return all_changes
 
 
+def apply_nginx_csp_policy_migrations(root: CommentedMap) -> list[MigrationChange]:
+    """Migrate ``nginx.cspPolicy.cdnEnabled`` to ``nginx.cspPolicy.cdn.enabled``.
+
+    Aligns customer Helm overrides with chart 2.x / PLX-300 unified nested
+    ``.enabled`` shape under ``nginx.cspPolicy``.
+
+    Parameters:
+        root: The parsed YAML document root.
+
+    Returns:
+        Migration changes applied under ``nginx.cspPolicy``, if any.
+    """
+    nginx = _get_nested_mapping(root, ["nginx"])
+    if nginx is None:
+        return []
+    csp_policy = _get_nested_mapping(nginx, ["cspPolicy"])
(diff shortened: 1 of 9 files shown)


bin/helm_chart_values_migration_shared.py
```diff
@@ -466,6 +466,28 @@ def apply_houston_deployment_migrations(root: CommentedMap) -> list[MigrationCha
     return all_changes
 
 
+def apply_nginx_csp_policy_migrations(root: CommentedMap) -> list[MigrationChange]:
+    """Migrate ``nginx.cspPolicy.cdnEnabled`` to ``nginx.cspPolicy.cdn.enabled``.
+
+    Aligns customer Helm overrides with chart 2.x / PLX-300 unified nested
+    ``.enabled`` shape under ``nginx.cspPolicy``.
```
danielhoherd, line 473: Nit: Why are we using double-ticks here? I've never seen this with python code before. I see this is used other places in the values migration code, but nowhere outside of it. Seems nonidiomatic.

danielhoherd (approved):
I guess this looks good. There are some syntaxes in here I'm not too happy with, but they are from prior code that was introduced as part of the migration, not this PR itself.

PR #3163, opened 2026-04-10
PINF-366 use least_conn load balancing for commander

danielhoherd:
## Description

- Use `least_conn` load balancing for commander.
    > Why this fixes the problem: GRPC uses long-lived HTTP/2 connections. With the default round_robin algorithm, nginx distributes connections in sequence — but because GRPC RPCs can run for varying durations, one replica can end up with far more active concurrent calls than others. least_conn routes each new connection to the replica with the fewest active connections, naturally rebalancing as calls complete and ensuring fair distribution across commander replicas.
- Add functional test agent skill

## Related Issues

https://linear.app/astronomer/issue/PINF-366/update-nginx-config-to-enable-fairer-load-balancing-across-commander

## Testing

Unit tests are included. I have not done performance testing under high load.

## Merging

This could theoretically be merged everywhere.


charts/astronomer/templates/commander/commander-grpc-ingress.yaml
```diff
@@ -14,20 +14,21 @@ metadata:
...
     {{- if .Values.global.extraAnnotations }}
-{{- toYaml .Values.global.extraAnnotations | nindent 4 }}
+      {{- toYaml .Values.global.extraAnnotations | nindent 4 }}
     {{- end }}
     nginx.ingress.kubernetes.io/backend-protocol: "GRPC"
     nginx.ingress.kubernetes.io/enable-http2: "true"
     kubernetes.io/ingress.class: "{{ .Release.Name }}-nginx"
     nginx.ingress.kubernetes.io/custom-http-errors: "404"
     nginx.ingress.kubernetes.io/proxy-buffer-size: "16k"
+    nginx.ingress.kubernetes.io/load-balance: "least_conn"
```
pgvishnuram, line 28: we will scope out authsidecar change in 2.0 right

danielhoherd: This PR is not meant to address authSidecar, which does not have a known commander load balancing solution that is configurable through our app.

pgvishnuram (approved):

PR #3166, opened 2026-04-14
PLX-167: Fix migration script for nodeExporter and daemonsetLogging

karankhanchandani:
## Description

Fix migration script for nodeExporter and daemonsetLogging

## Related Issues

https://linear.app/astronomer/issue/PLX-167/cpdp-chart-migration-tooling-and-backwards-compatibility-for-the

## Testing

- Unit tests
- Tested locally by running the scripts

## Merging

2.0, main

### DIFF
--- a/bin/helm_chart_values_migration_shared.py
+++ b/bin/helm_chart_values_migration_shared.py
@@ -308,6 +308,7 @@ GLOBAL_FEATURE_FLAG_RULES: list[BoolToNested | InvertedBoolToNested | SubtreeMov
     BoolToNested("podDisruptionBudgetsEnabled", ["podDisruptionBudgets", "enabled"]),
     BoolToNested("postgresqlEnabled", ["postgresql", "enabled"]),
     BoolToNested("prometheusPostgresExporterEnabled", ["prometheusPostgresExporter", "enabled"]),
+    BoolToNested("nodeExporterEnabled", ["nodeExporter", "enabled"]),
     BoolToNested("manualNamespaceNamesEnabled", ["namespaceManagement", "manualNamespaceNames", "enabled"]),
     BoolToNested("enablePerHostIngress", ["perHostIngress", "enabled"]),
     BoolToNested("enableArgoCDAnnotation", ["argoCD", "annotation", "enabled"]),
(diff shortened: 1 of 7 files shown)


tests/migration/test_migrate_helm_chart_values_037x.py
```diff
@@ -1264,6 +1274,38 @@ def test_set_value_already_at_new_value(self):
         assert result["global"]["pgbouncer"]["servicePort"] == "6543"
         assert len(changes) == 0
 
+    def test_vector_enabled_conflict_keeps_new(self):
```
Shubham-astro, line 1277: Why are we testing this.? I am confused about the condition:
```
global:
              vectorEnabled: true
              daemonsetLogging:
                enabled: false
```

bin/helm_chart_values_migration_shared.py
```diff
@@ -320,6 +321,7 @@ def apply(self, global_mapping: CommentedMap) -> list[MigrationChange]:
     BoolToNested("prometheusEnabled", ["prometheus", "enabled"]),
     BoolToNested("elasticsearchEnabled", ["elasticsearch", "enabled"]),
     BoolToNested("vectorEnabled", ["daemonsetLogging", "enabled"]),
+    BoolToNested("fluentdEnabled", ["daemonsetLogging", "enabled"]),
```
pgvishnuram, line 324: why do we need fluend its removed in 1.x itself having or not having doesnt gonna impact us

pgvishnuram: as per shubham
if either its fluend or vector we will set the daemonsetLogging enabled as true

rishkarajgi: @pgvishnuram `fluentdEnabled` still exists in 0.x versions, so this line addresses for those versions

tests/migration/test_migrate_helm_chart_values_037x.py
karankhanchandani: This is to remove the key.
Since we already have `daemonsetLogging` set, we should discard the old value.

Shubham-astro (approved):

PR #3170, opened 2026-04-15
PLX-335: Fix nested global values migration

karankhanchandani:
## Description

This PR fixes Helm values migration gaps in the `1.x -> 2.x` and `0.37.x -> 2.x` migration scripts.

It updates the shared migration helpers to recurse through nested `*.global` blocks, so full customer values dumps from `helm get values --all` are migrated correctly instead of only the top-level `global` section.

This PR also:
- adds regression coverage for nested subchart `global` values

## Related Issues

https://linear.app/astronomer/issue/PLX-335/migration-is-not-applying-to-all-the-feature-flags-in-all-the-modules

## Testing

Tested with:

- Tested locally
- Unit tests

## Merging
2.0, main


Shubham-astro (approved):
simple script changes, lgtm.

PR #3172, opened 2026-04-15
PINF-495: add emptyDir mount for Prisma client path in db-migration job

Shubham-astro:
## Description

Upgrading from `0.37.x` to `2.0.x` fails during the houston-db-migrations job with:
`Error: EROFS: read-only file system`, unlink '/houston/node_modules/.prisma/client/index.js'
Houston 2.x enforces `readOnlyRootFilesystem: true`, but Prisma needs to write/unlink files under node_modules/.prisma/client/ at migration time.

## Related Issues

https://linear.app/astronomer/issue/PINF-495/upgrade-from-0374-to-200-beta2-is-failed

## Testing

- Added unittest

## Merging

Master, 2.0

### DIFF
--- a/charts/astronomer/templates/houston/helm-hooks/houston-db-migration-job.yaml
+++ b/charts/astronomer/templates/houston/helm-hooks/houston-db-migration-job.yaml
@@ -160,6 +160,8 @@ spec:
               mountPath: /tmp
             - name: tmp
               mountPath: /houston/node_modules/.cache
+            - name: tmp
+              mountPath: /houston/node_modules/.prisma
           {{- include "houston_volume_mounts" . | indent 12 }}
           {{- include "custom_ca_volume_mounts" . | indent 12 }}
       volumes:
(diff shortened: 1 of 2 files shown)


charts/astronomer/templates/houston/helm-hooks/houston-db-migration-job.yaml
```diff
@@ -160,6 +160,8 @@ spec:
               mountPath: /tmp
             - name: tmp
               mountPath: /houston/node_modules/.cache
+            - name: tmp
+              mountPath: /houston/node_modules/.prisma
```
pgvishnuram, line 164: volumeMount name should be unique right

danielhoherd: no, volumeMount[].name is a reference to the volume of the same name. this series of `tmp` volumeMounts mounts the same volume in multiple places.

That being said, they should probably have subPath entries so they are not clobbering each others' data. 


```suggestion
- name: tmp
mountPath: /houston/node_modules/.cache
            - name: tmp
              subPath: .prisma
              mountPath: /houston/node_modules/.prisma
```

The next one up could be `subPath: .cache`

PR #3173, opened 2026-04-15
SWAT-96 Correctly handle ES ingress spec when no TLS secret is specified

stefan-raducan:
## Description

Fix the Elasticsearch ingress template so that hosts is correctly nested under the tls block and will only render when global.tlsSecret is set. Previously, the {{- end }} closing the TLS conditional was placed before hosts making hosts render at a wrong indent when global.tlsSecret is not set, resulting in an invalid Ingress spec.

## Related Issues

Related [SWAT-96](https://astronomer.atlassian.net/browse/SWAT-96)

## Testing

Added two new parametrized tests covering both data and unified plane modes:

test_elasticsearch_ingress_without_tls — verifies that when tlsSecret is empty, the tls block is absent and rules is the first key in spec.
test_elasticsearch_ingress_with_tls — verifies that when tlsSecret is set, the tls block is present with the correct secretName and hosts, and rules is still present.
Both tests set domainPrefix appropriately ("cluster1" for data plane, "" for unified) and assert the expected ingress hostname.

## Merging

master, release-2.0


stefan-raducan:
Closing this as changes were already added via [this PR](https://github.com/astronomer/astronomer/commit/0b47989920472f11274f9e4619e076916bb05335).

PR #3178, opened 2026-04-17
PLX-342 Fix GCP audit logging sink resource wiring for Houston sidecars

piyush-shrivastava:
## Description

This fixes the Houston Vector sidecar wiring for the GCP Cloud Logging sink.

The chart was rendering invalid `k8s_container` monitored resource metadata for the `gcp_stackdriver_logs` sink:
- `location` was incorrectly derived from the pod namespace
- `cluster_name` could silently fall back to `"unknown"`

In practice, this caused Vector to send requests that GCP rejected with `400 Bad Request`, so Houston audit logs were dropped instead of reaching Cloud Logging.

This change:
- wires `location` from `houston.logging.loggingSidecar.gcpCloudLogging.resource.location`
- wires `cluster_name` from `houston.logging.loggingSidecar.gcpCloudLogging.resource.clusterName`
- adds validation so `projectId`, `resource.location`, and `resource.clusterName` are required when `gcpCloudLogging.enabled=true`
- adds the new GCP resource fields to both chart values files

This keeps the configuration under the `houston.logging.loggingSidecar.gcpCloudLogging.*` schema and does not change CloudWatch or Elasticsearch sink behavior.

## Related Issues

https://linear.app/astronomer/issue/PLX-342/vector-in-gcp-sink-is-giving-400-error-due-to-bad-config-map

## Testing

- Debugged the issue on a live GKE cluster where Houston audit logs were not reaching Cloud Logging
- Confirmed Vector sidecars were attempting writes and receiving `400 Bad Request`
- Patched the live Houston Vector configmaps with the same `project_id`, `location`, and `cluster_name` wiring implemented in this PR
- Restarted Houston and Houston Worker
- Triggered fresh audit events through Houston API
- Verified the events landed successfully in GCP Cloud Logging with:
  - `resource.type = k8s_container`
  - correct `project_id`
  - correct `location`
  - correct `cluster_name`
- Verified the fix path is isolated to the GCP sink and does not change CloudWatch or Elasticsearch templates

## Merging

Merge into master

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -83,6 +83,11 @@
 {{- if not (or $hasBuiltInSink $hasExtraSinks) -}}
 {{- fail "houston.logging.loggingSidecar.enabled requires at least one sink: enable cloudwatch, gcpCloudLogging, elasticsearch, or set extraSinks" -}}
 {{- end -}}
+{{- if .Values.houston.logging.loggingSidecar.gcpCloudLogging.enabled -}}
+{{- required "houston.logging.loggingSidecar.gcpCloudLogging.projectId must be set when gcpCloudLogging.enabled is true" .Values.houston.logging.loggingSidecar.gcpCloudLogging.projectId -}}
+{{- required "houston.logging.loggingSidecar.gcpCloudLogging.resource.location must be set when gcpCloudLogging.enabled is true" .Values.houston.logging.loggingSidecar.gcpCloudLogging.resource.location -}}
+{{- required "houston.logging.loggingSidecar.gcpCloudLogging.resource.clusterName must be set when gcpCloudLogging.enabled is true" .Values.houston.logging.loggingSidecar.gcpCloudLogging.resource.clusterName -}}
+{{- end -}}
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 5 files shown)


charts/astronomer/templates/houston/api/houston-vector-configmap.yaml
```diff
@@ -133,11 +133,11 @@ data:
         resource:
           type: {{ .Values.houston.logging.loggingSidecar.gcpCloudLogging.resource.type | quote }}
           project_id: {{ .Values.houston.logging.loggingSidecar.gcpCloudLogging.projectId | quote }}
-          location: "{{ "{{" }} kubernetes.pod_namespace {{ "}}" }}"
+          location: {{ .Values.houston.logging.loggingSidecar.gcpCloudLogging.resource.location | quote }}
```
Shubham-astro: since a lot of this block is duplicated in houston-worker, consider using helper template for this. @piyush-shrivastava  This helps in reducing duplication of the same code across templates.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -83,6 +83,11 @@
 {{- if not (or $hasBuiltInSink $hasExtraSinks) -}}
 {{- fail "houston.logging.loggingSidecar.enabled requires at least one sink: enable cloudwatch, gcpCloudLogging, elasticsearch, or set extraSinks" -}}
 {{- end -}}
+{{- if .Values.houston.logging.loggingSidecar.gcpCloudLogging.enabled -}}
```
Shubham-astro, line 86: Nit comment:
since these are required value there is an edge case for example if someone passes:
`location: "   "` them helm considers it a non-empty string, so to avoid this you can use `trim` here.

Shubham-astro: something like: `{{- required "location must be set" (trim .Values.houston.logging.loggingSidecar.gcpCloudLogging.resource.location) -}}`

Shubham-astro (approved):
looks good..

PR #3199, opened 2026-04-23
PINF-514 fix broken test container user is not root test

danielhoherd:
## Description

- Fix a logic error that was exposed in <https://linear.app/astronomer/issue/PLX-364/fix-failing-apc-ui-test-capabilities-test> and caused confusion.
- Exclude all chainguard distroless containers from the root user test, since it requires a shell and there is no shell available in the container.

## Related Issues

https://linear.app/astronomer/issue/PINF-514/fix-broken-test-container-user-is-not-root-test-in

## Testing

CI tests passing is fine. These are only test changes.

## Merging

master, all supported >= 1.x branches

### DIFF
--- a/tests/functional/unified/test_container_user_is_not_root.py
+++ b/tests/functional/unified/test_container_user_is_not_root.py
@@ -5,20 +5,19 @@ from tests.utils.k8s import KUBECONFIG_UNIFIED, get_pod_running_containers
 
 container_ignore_list = ["kube-state", "houston", "astro-ui"]
 
+_all_containers = get_pod_running_containers(kubeconfig=KUBECONFIG_UNIFIED, namespace="astronomer")
+_test_containers = {k: v for k, v in _all_containers.items() if v["_name"] not in container_ignore_list}
 
-def test_container_user_is_not_root():
-    container_list = get_pod_running_containers(kubeconfig=KUBECONFIG_UNIFIED, namespace="astronomer")
-    for container in container_list.values():
-        if container["_name"] in container_ignore_list:
-            pytest.skip("Info: Unsupported container: " + container["_name"])
 
-        pod_client = testinfra.get_host(
-            f"kubectl://{container['pod_name']}?container={container['_name']}&namespace={container['namespace']}",
-            kubeconfig=KUBECONFIG_UNIFIED,
-        )
+@pytest.mark.parametrize("container", _test_containers.values(), ids=_test_containers.keys())
+def test_container_user_is_not_root(container):
+    pod_client = testinfra.get_host(
+        f"kubectl://{container['pod_name']}?container={container['_name']}&namespace={container['namespace']}",
+        kubeconfig=KUBECONFIG_UNIFIED,
+    )
 
-        user_info = pod_client.user()
-        assert user_info.name != "root"
-        assert user_info.group != "root"
-        assert user_info.gid != 0
-        assert user_info.uid != 0
+    user_info = pod_client.user()
+    assert user_info.name != "root"
+    assert user_info.group != "root"
+    assert user_info.gid != 0
+    assert user_info.uid != 0
--- a/tests/functional/unified/test_container_user_is_not_root.py
+++ b/tests/functional/unified/test_container_user_is_not_root.py
@@ -3,7 +3,16 @@ import testinfra
 
 from tests.utils.k8s import KUBECONFIG_UNIFIED, get_pod_running_containers
 
-container_ignore_list = ["kube-state", "houston", "astro-ui"]
+container_ignore_list = [
+    "astro-ui",
+    "configmap-reloader",
+    "default-backend",
+    "houston",
+    "kube-state",
+    "nginx",
+    "postgresql",
+    "vector",
+]
 
 _all_containers = get_pod_running_containers(kubeconfig=KUBECONFIG_UNIFIED, namespace="astronomer")
 _test_containers = {k: v for k, v in _all_containers.items() if v["_name"] not in container_ignore_list}
--- a/tests/functional/unified/test_container_user_is_not_root.py
+++ b/tests/functional/unified/test_container_user_is_not_root.py
@@ -5,12 +5,12 @@ from tests.utils.k8s import KUBECONFIG_UNIFIED, get_pod_running_containers
 
 container_ignore_list = [
     "astro-ui",
+    "astronomer-postgresql",
     "configmap-reloader",
     "default-backend",
     "houston",
     "kube-state",
     "nginx",
-    "postgresql",
     "vector",
 ]
 


rishkarajgi (commented):
*(superseded)*

tests/functional/unified/test_container_user_is_not_root.py
```diff
@@ -1,24 +1,130 @@
...
 
 from tests.utils.k8s import KUBECONFIG_UNIFIED, get_pod_running_containers
 
-container_ignore_list = ["kube-state", "houston", "astro-ui"]
-
+_all_containers = get_pod_running_containers(kubeconfig=KUBECONFIG_UNIFIED, namespace="astronomer")
 
-def test_container_user_is_not_root():
-    container_list = get_pod_running_containers(kubeconfig=KUBECONFIG_UNIFIED, namespace="astronomer")
-    for container in container_list.values():
```
rishkarajgi, line 11: This runs at module import time, so `get_pod_running_containers` is called during pytest collection — not just when tests execute. If the kubeconfig isn't available or the namespace is empty at collection time, parametrize silently produces zero tests with no warning. Worth wrapping in a session-scoped fixture so it only runs when the tests actually execute?

tests/functional/unified/test_container_user_is_not_root.py
```diff
@@ -1,24 +1,130 @@
...
-            pytest.skip("Info: Unsupported container: " + container["_name"])
 
-        pod_client = testinfra.get_host(
-            f"kubectl://{container['pod_name']}?container={container['_name']}&namespace={container['namespace']}",
-            kubeconfig=KUBECONFIG_UNIFIED,
+def _wait_for_ephemeral_container_terminated(pod_name, debug_container_name, namespace, kubeconfig, timeout=60):
+    """Poll until the ephemeral container has terminated (or raise on timeout)."""
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        result = subprocess.run(
```
rishkarajgi: Nit: `timeout=60` is hardcoded here and in two other `subprocess.run` calls below. Would be cleaner as a module-level constant like `_KUBECTL_TIMEOUT = 60` so it's easy to change in one place.

tests/functional/unified/test_container_user_is_not_root.py
```diff
@@ -1,24 +1,130 @@
...
+            text=True,
+            timeout=10,
         )
+        if result.returncode == 0:
+            pod = json.loads(result.stdout)
+            for cs in pod.get("status", {}).get("ephemeralContainerStatuses", []):
+                if cs["name"] == debug_container_name:
+                    if "terminated" in cs.get("state", {}):
+                        return
+        time.sleep(1)
```
rishkarajgi: If `debug_container_name` is never found in `ephemeralContainerStatuses` (e.g. because the name got truncated or the debug container failed silently), this loop just spins until the `TimeoutError` with no indication of why. A `break`-with-warning or a check after the loop would speed up debugging a lot.

tests/functional/unified/test_container_user_is_not_root.py
```diff
@@ -1,24 +1,130 @@
...
+        return exec_result.stdout
+
+    # Distroless container — no cat binary. Use an ephemeral busybox container
+    # with --attach=false to avoid the race condition between container exit and
+    # kubectl attach, then wait for it to terminate before reading logs.
+    debug_container_name = f"proc-status-{container_name}"
+    subprocess.run(
+        [
+            "kubectl",
+            f"--kubeconfig={kubeconfig}",
```
rishkarajgi: Kubernetes caps container names at 63 chars — `f"proc-status-{container_name}"` could exceed that if the container name is long. Might be worth truncating: `f"proc-status-{container_name}"[:63]`.

rishkarajgi (commented):
Rewrites the root-user test to use direct `kubectl` subprocess calls instead of `testinfra`, and adds a fallback via ephemeral busybox containers for distroless images. The approach is solid — a few things worth looking at inline before merging.

One thing I couldn't pin to a specific line: the original code had `container_ignore_list = ["kube-state", "houston", "astro-ui"]` which skipped those containers. The PR description only mentions excluding chainguard/distroless images — worth clarifying whether those three are now intentionally included (and passing) or if the ignore list was accidentally dropped.

**Verdict:** A few questions — nothing blocking if the ignore list removal is intentional, but the collection-time side effect is worth addressing.

tests/functional/unified/test_container_user_is_not_root.py
```diff
@@ -1,24 +1,130 @@
...
+            "/proc/1/status",
+        ],
+        capture_output=True,
+        text=True,
+        timeout=60,
+        check=True,
+    )
+    _wait_for_ephemeral_container_terminated(pod_name, debug_container_name, namespace, kubeconfig)
+    logs_result = subprocess.run(
+        [
```
rishkarajgi: `check=True` will raise `CalledProcessError` if the debug container fails to spawn, but stderr won't surface in the test output unless explicitly captured. Something like:
```python
try:
    subprocess.run([...], check=True, ...)
except subprocess.CalledProcessError as e:
    raise RuntimeError(f"kubectl debug failed for {pod_name}: {e.stderr}") from e
```
...would make failures much easier to diagnose.

tests/functional/unified/test_container_user_is_not_root.py
```diff
@@ -3,22 +3,30 @@
...
+    "kube-state",
+    "nginx",
+    "vector",
+]
 
+_all_containers = get_pod_running_containers(kubeconfig=KUBECONFIG_UNIFIED, namespace="astronomer")
+_test_containers = {k: v for k, v in _all_containers.items() if v["_name"] not in container_ignore_list}
 
-def test_container_user_is_not_root():
-    container_list = get_pod_running_containers(kubeconfig=KUBECONFIG_UNIFIED, namespace="astronomer")
```
rishkarajgi, line 10: This runs at collection time rather than test execution — if the kubeconfig isn't available when pytest collects, `get_pod_running_containers` will either raise or return empty, and `parametrize` will silently produce zero test cases instead of a visible failure. Worth wrapping in a session-scoped fixture or gating with `pytest.importorskip`-style logic so you get an explicit skip/error rather than a quiet pass?

tests/functional/unified/test_container_user_is_not_root.py
```diff
@@ -3,22 +3,30 @@
...
-    for container in container_list.values():
```
rishkarajgi, line 11: Nit: if a new container shows up that isn't in the ignore list but also happens to be distroless/shell-less, the test will fail with a confusing testinfra error rather than a clear skip. Might be worth a try/except around `pod_client.user()` in the test body that converts connection failures to `pytest.skip` with a message, so the ignore list only needs to cover intentionally-root containers.

tests/functional/unified/test_container_user_is_not_root.py
```diff
@@ -3,22 +3,30 @@
...
 
-        pod_client = testinfra.get_host(
-            f"kubectl://{container['pod_name']}?container={container['_name']}&namespace={container['namespace']}",
-            kubeconfig=KUBECONFIG_UNIFIED,
-        )
+@pytest.mark.parametrize("container", _test_containers.values(), ids=_test_containers.keys())
+def test_container_user_is_not_root(container):
+    pod_client = testinfra.get_host(
+        f"kubectl://{container['pod_name']}?container={container['_name']}&namespace={container['namespace']}",
+        kubeconfig=KUBECONFIG_UNIFIED,
```
rishkarajgi, line 25: Good call switching to `parametrize` — the old loop with `pytest.skip()` would bail out of the entire test on the first ignored container, silently skipping every container after it. This gives proper per-container pass/fail visibility.

rishkarajgi (approved):
Fixes the broken root-user test by switching from an internal loop (where `pytest.skip` aborted the whole test on first ignored container) to `@pytest.mark.parametrize`, and expands the ignore list to exclude chainguard distroless containers that lack a shell.

**Verdict:** ✅ Looks good — the parametrize approach is the right fix. One thing to consider around collection-time execution noted inline.

PR #3204, opened 2026-04-24
PLX-350 - remove unused logging configs

pgvishnuram:
## Description

This PR removes unused configs from the houston config map 

## Related Issues

https://linear.app/astronomer/issue/PLX-350/bug-airflowelasticsearch-config-takes-priority-over

## Testing

NA

## Merging

merge to master and release-2.0

### DIFF
--- a/bin/show-docker-images.py
+++ b/bin/show-docker-images.py
@@ -109,7 +109,7 @@ def get_images_from_houston_configmap(doc, args):
     auth_sidecar_tag = deployments["authSideCar"]["tag"]
     if auth_sidecar_repository and auth_sidecar_tag:
         images.append(f"{auth_sidecar_repository}:{auth_sidecar_tag}")
-    if logging_sidecar_image := deployments.get("logging", {}).get("loggingSidecar", {}).get("image"):
+    if logging_sidecar_image := (deployments.get("logging") or {}).get("loggingSidecar") or {}.get("image"):
         images.append(logging_sidecar_image)
     dag_server_repository = deployments.get("dagDeploy", {}).get("images", {}).get("dagServer", {}).get("repository")
     dag_server_tag = deployments.get("dagDeploy", {}).get("images", {}).get("dagServer", {}).get("tag")
--- a/bin/show-docker-images.py
+++ b/bin/show-docker-images.py
@@ -109,7 +109,7 @@ def get_images_from_houston_configmap(doc, args):
     auth_sidecar_tag = deployments["authSideCar"]["tag"]
     if auth_sidecar_repository and auth_sidecar_tag:
         images.append(f"{auth_sidecar_repository}:{auth_sidecar_tag}")
-    if logging_sidecar_image := (deployments.get("logging") or {}).get("loggingSidecar") or {}.get("image"):
+    if logging_sidecar_image := (deployments.get("logging") or {}).get("loggingSidecar", {}).get("image"):
         images.append(logging_sidecar_image)
     dag_server_repository = deployments.get("dagDeploy", {}).get("images", {}).get("dagServer", {}).get("repository")
     dag_server_tag = deployments.get("dagDeploy", {}).get("images", {}).get("dagServer", {}).get("tag")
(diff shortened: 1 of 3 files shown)


rishkarajgi (dismissed):
Removes the elasticsearch connection config (`logging.enabled`, `logging.provider`, `logging.elasticsearch.*`) from the houston configmap — clean removal, the configmap change itself looks good. The companion Python change in `show-docker-images.py` has an operator precedence bug though that needs a fix before merge.

**Verdict:** ⚠️ Needs changes — the `bin/show-docker-images.py` change drops the `.get("image")` call due to operator precedence.

---

**`bin/show-docker-images.py` line ~112** — `or {}.get("image")` doesn't do what you'd expect here. Python evaluates `{}.get("image")` first (giving `None`) before the `or`, so the full expression collapses to:

```python
(deployments.get("logging") or {}).get("loggingSidecar") or None
```

The `.get("image")` is never actually called on the `loggingSidecar` dict. When `loggingSidecar` is truthy (e.g. `{"image": "repo/img:tag"}`), `logging_sidecar_image` gets the whole dict rather than the image string, and `images.append(logging_sidecar_image)` silently appends a dict instead of a string.

The fix is to add an extra set of parens so the fallback `{}` wraps the right thing:

```python
if logging_sidecar_image := ((deployments.get("logging") or {}).get("loggingSidecar") or {}).get("image"):
```

Or equivalently, since you only needed `or {}` to handle an explicit `null` on `logging`:

```python
if logging_sidecar_image := (deployments.get("logging") or {}).get("loggingSidecar", {}).get("image"):
```

bin/show-docker-images.py
```diff
@@ -109,7 +109,7 @@ def get_images_from_houston_configmap(doc, args):
     auth_sidecar_tag = deployments["authSideCar"]["tag"]
     if auth_sidecar_repository and auth_sidecar_tag:
         images.append(f"{auth_sidecar_repository}:{auth_sidecar_tag}")
-    if logging_sidecar_image := deployments.get("logging", {}).get("loggingSidecar", {}).get("image"):
+    if logging_sidecar_image := (deployments.get("logging") or {}).get("loggingSidecar", {}).get("image"):
         images.append(logging_sidecar_image)
```
rishkarajgi, line 113: Good catch — after removing the `logging.enabled` / `logging.elasticsearch` block from the configmap, the `logging:` key can render as `None` (YAML key with no children when `loggingSidecar` is disabled). `.get("logging", {})` would return `None` in that case since the key *exists* but has a `None` value, so `or {}` is the right fix here.

tests/chart_tests/test_houston_configmap.py
```diff
@@ -992,7 +992,7 @@ def test_houston_configmap_features_elasticsearch_defaults():
         show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
     )
     prod = yaml.safe_load(docs[0]["data"]["production.yaml"])
-    es = prod["deployments"]["logging"]["elasticsearch"]
+    es = prod["deployments"]["helm"]["airflow"]["elasticsearch"]
     assert es["enabled"] is True
```
rishkarajgi, line 990: These tests now read ES config from `deployments["helm"]["airflow"]["elasticsearch"]` — can you confirm that path is actually rendered in the configmap? The removed block was under `deployments.logging.elasticsearch`, and I don't see a corresponding `helm.airflow.elasticsearch` block being added in the configmap template in this PR. If it already exists elsewhere in the template that's fine, just want to make sure these assertions will still pass.

tests/chart_tests/test_houston_configmap.py
```diff
@@ -1005,7 +1005,7 @@ def test_houston_configmap_features_elasticsearch_custom_logging():
...
```
rishkarajgi, line 1003: Same question as above — this test checks the custom-logging variant (`customLogging.enabled: true`) and asserts the host ends with `-external-es-proxy`. If the ES connection config was only rendered by the block that's now deleted, these assertions would fail. Worth a quick `helm template` sanity check if CI isn't covering this path.

rishkarajgi (commented):
Removes the `logging.enabled`, `logging.provider`, and `logging.elasticsearch` config block from the houston configmap — the ES connection config now lives under `helm.airflow.elasticsearch` instead. Updates the `show-docker-images` script to handle `logging` being `None`, and points the ES tests at the new path.

**Verdict:** 💬 A few questions — the configmap deletion looks correct, but I want to verify the test path updates will actually pass since the `helm.airflow.elasticsearch` block isn't added in this PR. If it's already rendered elsewhere in the template, this is good to go.

danielhoherd (approved):
LGTM

bin/show-docker-images.py
danielhoherd: Yeah, I'm not sure what's going on with this new code. Why is the old code not the right code?

```
>>> deployments = {'foo': 'FOO'}
>>> deployments.get("logging") or {}
{}
>>> deployments.get("logging", {})
{}
```

PR #3205, opened 2026-04-24
PLX-370: wire dagTarballVersionValidation.enabled and migrate old key

Simpcyclassy:
## Description

This is the chart-side follow-up to `enabledagtarballversionvalidation` and the coordinated **houston-api** change: the DAG tarball presence check for deploy rollback is controlled by **`deployments.deploymentLifecycle.deployRollback.dagTarballVersionValidation.enabled`**, not a flat `enableDagTarballVersionValidation` on `deployRollback`.

This PR:

- **Values:** `global.deploymentLifecycle.deployRollback.dagTarballVersionValidation.enabled` in `values.yaml` (default aligned with Houston).
- **Houston ConfigMap:** render that path under `astronomer.houston.config.deployments.deploymentLifecycle.deployRollback` so new installs and upgrades get the right shape in Houston’s config.
- **Helm values migration:** `HOUSTON_DEPLOYMENT_NESTED_PATH_MIGRATIONS` in `bin/helm_chart_values_migration_shared.py` mirrors Houston’s `DEPLOYMENTS_NESTED_PATH_MIGRATIONS` — renames `deploymentLifecycle.deployRollback.enableDagTarballVersionValidation` → `...dagTarballVersionValidation.enabled` when present in `astronomer.houston.config.deployments`.

## Related Issues

- https://linear.app/astronomer/issue/PLX-370/update-missed-enabledagtarballversionvalidation-flag-in-houston-and

## Testing

- **Helm template sanity:** e.g. `helm template` (or your chart CI) to confirm `global.deploymentLifecycle.deployRollback.dagTarballVersionValidation.enabled` flows into the Houston configmap under `deployments.deploymentLifecycle.deployRollback` without template errors.
- **Migration dry-run:** sample `astronomer.houston.config.deployments` YAML that still has `enableDagTarballVersionValidation` under `deploymentLifecycle.deployRollback` should produce `dagTarballVersionValidation.enabled` after `migrate-helm-chart-values` (1.x/0.37.x → 2.x) per your script entrypoint.

## Merging

- main
- 2.0.0

### DIFF
--- a/bin/helm_chart_values_migration_shared.py
+++ b/bin/helm_chart_values_migration_shared.py
@@ -406,7 +406,8 @@ GLOBAL_FEATURE_FLAG_RULES: list[BoolToNested | InvertedBoolToNested | SubtreeMov
 
 HOUSTON_DEPLOYMENTS_PREFIX = "astronomer.houston.config.deployments"
 
-# Renders houston-api ``DEPLOYMENTS_PATH_MIGRATIONS`` for Helm
+# Renders houston-api ``DEPLOYMENTS_PATH_MIGRATIONS`` and
+# ``HOUSTON_DEPLOYMENT_NESTED_PATH_MIGRATIONS`` for Helm
 # (``src/lib/deployments-config-path-migration/index.js``). Order is significant.
 HOUSTON_DEPLOYMENT_PATH_MIGRATIONS: list[tuple[str, str | None, str]] = [
     ("performanceOptimizationModeEnabled", "performanceOptimization.enabled", "boolean-to-enabled"),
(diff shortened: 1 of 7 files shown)


bin/helm_chart_values_migration_shared.py
```diff
@@ -729,6 +767,30 @@ def _apply_houston_deployments_path_migrations(
...
+            )
+            continue
+        old_value = _get_dotted_value(deployments, old_dotted)
+        new_value = _transform_houston_deployment_value(old_value, transform)
+        _set_dotted_path_in_mapping(deployments, new_dotted, new_value)
+        _unset_dotted_path(deployments, old_dotted)
+        p_old = f"{HOUSTON_DEPLOYMENTS_PREFIX}.{old_dotted}"
+        p_new = f"{HOUSTON_DEPLOYMENTS_PREFIX}.{new_dotted}"
+        changes.append(MigrationChange(p_old, p_new, f"Migrated {p_old} -> {p_new}"))
     return changes
```
rishkarajgi, line 796: Nit: this nested-migration loop is structurally identical to the flat one above — same transform → set → unset → append pattern. Not blocking, but if a third nested migration shows up it'd be worth extracting a shared helper to avoid the duplication.

values.yaml
```diff
@@ -101,6 +101,8 @@ global:
   deploymentLifecycle:
     deployRollback:
       enabled: false
+      dagTarballVersionValidation:
+        enabled: true
 
   # Deploy mechanisms
   deployMechanisms:
```
rishkarajgi, line 108: The default for `deployRollback.enabled` is `false` but `dagTarballVersionValidation.enabled` defaults to `true` — so the validation is armed even when rollback itself is off. That's fine because the check is only reached when rollback runs, but worth a quick comment in the values file so future readers don't wonder about the mismatch.

rishkarajgi (approved):
Wires `dagTarballVersionValidation.enabled` into the Helm configmap and adds a nested-path migration to rename the old flat `enableDagTarballVersionValidation` key. Companion to houston-api#2460.

**Verdict:** ✅ Looks good — clean migration with solid test coverage. A couple of minor nits below.

danielhoherd (approved):
Seems fine to me.

PR #3209, opened 2026-04-28
PINF-515 fix helm value key

danielhoherd:
## Description

Fix helm value key. Ideally we should fix this before releasing 2.0.0, because otherwise we won't be able to fix it until 3.0.

This does not close PINF-515. It only fixes the customer-facing parts of this problem, the least amount of work we can do that will let us fix the full bug within the lifecycle of 2.0 instead of having to wait until 3.0.

## Related Issues

https://linear.app/astronomer/issue/PINF-515/correct-typo-in-external-secret-manager-failover-key

## Testing

I have tested this in a local cluster where dataPlaneFailover was enabled and I was able to create a deployment that came up healthy.

## Merging

This is for master, release-2.0

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -326,7 +326,7 @@ gcp_cloud_logging:
   value: {{ ternary "true" "false" .Values.global.dataPlaneFailover.enabled | quote }}
 {{- if .Values.global.dataPlaneFailover.enabled }}
 - name: COMMANDER_EXTERNAL_SECRET_MANAGER_SECRET_NAME
-  value: {{ .Values.global.dataPlaneFailover.externalSecretManagerSecretName | quote }}
+  value: {{ .Values.global.dataPlaneFailover.externalSecretManagerName | quote }}
 {{- end }}
 {{- end }}
 {{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -91,6 +91,8 @@ global:
   # On unified (mode=unified): no effect — enable components individually
   dataPlaneFailover:
     enabled: false
+    # The name of a manually created ClusterSecretStore
+    # externalSecretManagerName: your-css-name
 
   # Metrics and reporting features
   metricsReporting:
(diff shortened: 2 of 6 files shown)


values.yaml
```diff
@@ -91,6 +91,8 @@ global:
   # On unified (mode=unified): no effect — enable components individually
   dataPlaneFailover:
     enabled: false
+    # The name of a manually created ClusterSecretStore
+    # externalSecretManagerName: your-css-name
 
   # Metrics and reporting features
   metricsReporting:
```
rishkarajgi, line 98: Nit: nice addition — the comment makes it clear this expects a ClusterSecretStore name, not a Secret name. Helps avoid the exact confusion the old key caused.

rishkarajgi (approved):
Renames the helm value key from `externalSecretManagerSecretName` to `externalSecretManagerName` across templates, configs, tests, and values.yaml. The old name was misleading — it referenced a ClusterSecretStore, not a Secret. Good catch fixing this before 2.0.0 ships so it doesn't become a breaking change later.

**Verdict:** ✅ Looks good — clean rename, tests updated, configs consistent, and the new values.yaml comments clarify intended usage.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -326,7 +326,7 @@ gcp_cloud_logging:
   value: {{ ternary "true" "false" .Values.global.dataPlaneFailover.enabled | quote }}
 {{- if .Values.global.dataPlaneFailover.enabled }}
 - name: COMMANDER_EXTERNAL_SECRET_MANAGER_SECRET_NAME
-  value: {{ .Values.global.dataPlaneFailover.externalSecretManagerSecretName | quote }}
+  value: {{ .Values.global.dataPlaneFailover.externalSecretManagerName | quote }}
```
pgvishnuram, line 329: does this name really make sense externalSecretManagerName doesn't sync with clustersecretstore or secret store

danielhoherd: It is deliberately ambiguous. We do not want to prescribe the type of secret manager here. That lets us change the type of secret manager without having to change this key.

PR #3210, opened 2026-04-28
PINF-533: Make federated-dataplanes scrape_interval configurable and add scrape_timeout

Shubham-astro:
## Summary
- Make the `federated-dataplanes` scrape job's `scrape_interval` configurable via `prometheus.config.scrape_configs.federated_dataplanes.scrape_interval`, defaulting to `15s` (previously hardcoded to `5s`)
- Add `scrape_timeout: 15s` to the `federated-dataplanes` scrape job

## Related Issues
https://linear.app/astronomer/issue/PINF-533

## Testing

- We are just increasing the scrape_interval of prometheus, no logic changes.

### DIFF
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -72,7 +72,7 @@ data:
         metrics_path: '/federate'
         scheme: https
         scrape_interval: {{ .Values.config.scrape_configs.federated_dataplanes.scrape_interval }}
-        scrape_timeout: 15s
+        scrape_timeout: {{ .Values.config.scrape_configs.federated_dataplanes.scrape_timeout }}
         params:
           'match[]':
             - '{job=~".*-kube-state$|airflow|kubernetes-nodes-cadvisor|node-exporter"}'
--- a/charts/prometheus/templates/prometheus-config-configmap.yaml
+++ b/charts/prometheus/templates/prometheus-config-configmap.yaml
@@ -71,7 +71,8 @@ data:
         honor_labels: true
         metrics_path: '/federate'
         scheme: https
-        scrape_interval: 5s
+        scrape_interval: {{ .Values.config.scrape_configs.federated_dataplanes.scrape_interval }}
+        scrape_timeout: 15s
         params:
           'match[]':
             - '{job=~".*-kube-state$|airflow|kubernetes-nodes-cadvisor|node-exporter"}'
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -209,6 +209,7 @@ config:
         insecure_skip_verify: false
     federated_dataplanes:
       scrape_interval: 15s
+      scrape_timeout: 15s
 
 rbac:
   role:
--- a/charts/prometheus/values.yaml
+++ b/charts/prometheus/values.yaml
@@ -209,7 +209,7 @@ config:
         insecure_skip_verify: false
     federated_dataplanes:
       scrape_interval: 15s
-      scrape_timeout: 14s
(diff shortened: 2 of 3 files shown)


charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -71,7 +71,8 @@ data:
         honor_labels: true
         metrics_path: '/federate'
         scheme: https
-        scrape_interval: 5s
+        scrape_interval: {{ .Values.config.scrape_configs.federated_dataplanes.scrape_interval }}
+        scrape_timeout: 15s
```
pgvishnuram: make this also configurable

Shubham-astro: done @pgvishnuram

rishkarajgi (commented):
Makes the `federated-dataplanes` scrape job configurable by pulling `scrape_interval` and `scrape_timeout` out of the template and into `values.yaml`, with the default interval bumped from `5s` to `15s`.

**Verdict:** 💬 A few questions — the interval default triples from `5s` to `15s`, which is a meaningful resolution change worth confirming is intentional; and setting `scrape_timeout` equal to `scrape_interval` leaves no buffer for slow scrapes.

charts/prometheus/values.yaml
```diff
@@ -207,6 +207,9 @@ config:
     kubernetes_apiservers:
       tls_config:
         insecure_skip_verify: false
+    federated_dataplanes:
+      scrape_interval: 15s
+      scrape_timeout: 15s
 
 rbac:
   role:
```
rishkarajgi, line 215: The previous hardcoded value was `5s`, so this default triples the scrape interval. Just want to confirm that's intentional — `15s` means noticeably coarser time resolution for federated dataplane metrics.

rishkarajgi, line 215: Also, setting `scrape_timeout` equal to `scrape_interval` at `15s` leaves zero slack — if a federated scrape runs even slightly long, Prometheus will cancel it right at the deadline. Common practice is to set the timeout a second or two under (e.g., `14s`) to give some breathing room.

Shubham-astro: This is what we were discussing in the standup @rishkarajgi ..
So currently what is hapenning is with 5s interval in CP/DP mode with more nummber of. deployments prometheus is getting overloaded, there are a few improvements that i am planning to do for this, but for now this frees up the prometheus without hurting the UI status indicator..

Shubham-astro: yeahh, makes sense, added it to 14s

charts/prometheus/values.yaml
```diff
@@ -207,6 +207,9 @@ config:
     kubernetes_apiservers:
       tls_config:
         insecure_skip_verify: false
+    federated_dataplanes:
+      scrape_interval: 15s
+      scrape_timeout: 14s
```
pgvishnuram: it should be higher than scrape_interval

charts/prometheus/values.yaml
rishkarajgi: ok, thanks for the clarification

charts/prometheus/values.yaml
```diff
@@ -207,6 +207,9 @@ config:
     kubernetes_apiservers:
       tls_config:
         insecure_skip_verify: false
+    federated_dataplanes:
+      scrape_interval: 15s
+      scrape_timeout: 10s
```
pgvishnuram, line 212: lets add a test case as well to validate overrides vs defaults

Shubham-astro: done

rishkarajgi (approved):

PR #3212, opened 2026-04-29
PINF-539 fix incorrect version for registry in meta endpoint

pgvishnuram:
## Description

The dataplane metadata endpoint returns an empty string for `registry.version`. This PR fixes it by populating the field from the Helm value `images.registry.tag` in the commander metadata ConfigMap template.

### Changes

| File | What changed |
|------|-------------|
| `charts/astronomer/templates/commander/commander-metadata.yaml` | Adds `registry.version: {{ .Values.images.registry.tag }}` to the metadata template so the registry image tag is surfaced in the metadata endpoint |

### Improvements

- `registry.version` in the metadata endpoint now reflects the actual registry image tag instead of an empty string
- Consumers of the metadata API (Houston-API, commander) can now identify which registry version is running on the dataplane

## Related Issues

https://linear.app/astronomer/issue/PINF-539/fix-registry-version-missing-from-dataplane-metadata-endpoint

## Testing

1. Deploy with the patched chart
2. Call the dataplane metadata endpoint and verify `registry.version` matches the value of `images.registry.tag` in the Helm release values
3. Confirm no other metadata fields are affected

## Merging

Merge to master.

### DIFF
--- a/tests/chart_tests/test_astronomer_commander.py
+++ b/tests/chart_tests/test_astronomer_commander.py
@@ -32,7 +32,7 @@ class TestAstronomerCommander:
                 "namespaceLabels": namespace_labels,
             },
             "astronomer": {
-                "images": {"registry": {"tag": "3.0.1-9"}},
+                "images": {"registry": {"tag": "99.88.77"}},
             },
         }
         docs = render_chart(
@@ -50,13 +50,13 @@ class TestAstronomerCommander:
         if namespace_labels and rbac_enabled:
             assert metadata_file_contents == {
                 "namespaceLabels": namespace_labels,
-                "registry": {"version": "3.0.1-9"},
+                "registry": {"version": "99.88.77"},
                 "customLogging": {"enabled": False},
             }
         else:
             assert metadata_file_contents == {
                 "namespaceLabels": {},
-                "registry": {"version": "3.0.1-9"},
+                "registry": {"version": "99.88.77"},
                 "customLogging": {"enabled": False},
             }
 
@@ -68,7 +68,7 @@ class TestAstronomerCommander:
                 "customLogging": {"enabled": enabled},
             },
             "astronomer": {
-                "images": {"registry": {"tag": "3.0.1-10"}},
+                "images": {"registry": {"tag": "99.88.77"}},
             },
         }
         docs = render_chart(
@@ -86,7 +86,7 @@ class TestAstronomerCommander:
         assert metadata_file_contents == {
             "namespaceLabels": {},
             "customLogging": {"enabled": enabled},
-            "registry": {"version": "3.0.1-10"},
+            "registry": {"version": "99.88.77"},
         }
 
     def test_commander_deployment_default(self, kube_version):
--- a/tests/chart_tests/test_astronomer_commander.py
+++ b/tests/chart_tests/test_astronomer_commander.py
@@ -1,3 +1,5 @@
+from ensurepip import version
+
 import jmespath
 import pytest
 import yaml
@@ -30,7 +32,10 @@ class TestAstronomerCommander:
             "global": {
                 "rbac": {"enabled": rbac_enabled},
                 "namespaceLabels": namespace_labels,
-            }
+            },
+            "astronomer": {
+                "images": {"registry": {"tag": "3.0.1-9"}},
+            },
         }
         docs = render_chart(
             kube_version=kube_version,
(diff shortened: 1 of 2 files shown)


rishkarajgi (approved):
Fixes the empty `registry.version` field in the dataplane metadata endpoint by pulling `images.registry.tag` from Helm values into the commander metadata ConfigMap.

**Verdict:** ✅ Looks good — minimal, focused fix with a clear reproduction path.

**Additional notes** (could not anchor inline):

**charts/astronomer/templates/commander/commander-metadata.yaml** (line ~19): Nit: worth quoting the tag value with `| quote` — semver-style tags like `1.0` will be parsed as YAML floats rather than strings, which could trip up anything downstream that expects `registry.version` to be a string. The existing `plane.mode` field has the same pattern so no obligation to fix here, just flagging it.

tests/chart_tests/test_astronomer_commander.py
```diff
@@ -30,7 +30,10 @@ def test_commander_metadata_yaml(self, kube_version, rbac_enabled, namespace_lab
             "global": {
                 "rbac": {"enabled": rbac_enabled},
                 "namespaceLabels": namespace_labels,
-            }
+            },
+            "astronomer": {
+                "images": {"registry": {"tag": "3.0.1-9"}},
```
danielhoherd: Test values should use obviously wrong data that is easy to distinguish from something that might be a real value. We should change this to something like:
```suggestion
                "images": {"registry": {"tag": "99.88.77"}},
```

tests/chart_tests/test_astronomer_commander.py
```diff
@@ -47,18 +50,26 @@ def test_commander_metadata_yaml(self, kube_version, rbac_enabled, namespace_lab
         if namespace_labels and rbac_enabled:
             assert metadata_file_contents == {
                 "namespaceLabels": namespace_labels,
+                "registry": {"version": "3.0.1-9"},
```
danielhoherd:
```suggestion
                "registry": {"version": "99.88.77"},
```

tests/chart_tests/test_astronomer_commander.py
```diff
@@ -47,18 +50,26 @@ def test_commander_metadata_yaml(self, kube_version, rbac_enabled, namespace_lab
...
                 "customLogging": {"enabled": False},
             }
         else:
-            assert metadata_file_contents == {"namespaceLabels": {}, "customLogging": {"enabled": False}}
+            assert metadata_file_contents == {
+                "namespaceLabels": {},
+                "registry": {"version": "3.0.1-9"},
```
danielhoherd:
```suggestion
                "registry": {"version": "99.88.77"},
```

tests/chart_tests/test_astronomer_commander.py
```diff
@@ -47,18 +50,26 @@ def test_commander_metadata_yaml(self, kube_version, rbac_enabled, namespace_lab
...
     @pytest.mark.parametrize("enabled", [True, False], ids=["custom_logging_enabled", "custom_logging_disabled"])
     def test_commander_metadata_custom_logging(self, kube_version, enabled):
         """Test that helm renders custom logging in metadata.yaml template for astronomer/commander."""
         values = {
             "global": {
                 "customLogging": {"enabled": enabled},
-            }
+            },
+            "astronomer": {
+                "images": {"registry": {"tag": "3.0.1-10"}},
```
danielhoherd:
```suggestion
                "images": {"registry": {"tag": "99.88.77"}},
```

tests/chart_tests/test_astronomer_commander.py
```diff
@@ -72,7 +83,11 @@ def test_commander_metadata_custom_logging(self, kube_version, enabled):
         assert doc["apiVersion"] == "v1"
 
         metadata_file_contents = yaml.safe_load(doc["data"]["metadata.yaml"])
-        assert metadata_file_contents == {"namespaceLabels": {}, "customLogging": {"enabled": enabled}}
+        assert metadata_file_contents == {
+            "namespaceLabels": {},
+            "customLogging": {"enabled": enabled},
+            "registry": {"version": "3.0.1-10"},
```
danielhoherd:
```suggestion
            "registry": {"version": "99.88.77"},
```

danielhoherd (approved):
Small fix looks good and safe to merge. I left some suggestions to improve the tests, but those should not block merging this PR since they are minor QOL improvements, not fixing bugs or anything major.

PR #3217, opened 2026-04-30
docs(cp-dp-k3d-setup-guide): make local-setup work end-to-end

leestro:
## Summary

Doc-only updates to `docs/cp-dp-k3d-setup-guide.md` that reflect what was actually needed to get a fresh OrbStack-based local install working through the UI's "Open Airflow" flow.

- **Step 11 led with host-mapped ports** (Option A) which makes Houston's port-less "Open Airflow" link unreachable in a browser. Promoting LB-IP routing (Option B) to the default works because OrbStack auto-routes its docker subnet to the host, so each cluster's nginx serves on its own `:443` natively.
- **Step 8 / Step 10's manual `kubectl edit configmap coredns` flow** leaves the install fragile across OrbStack restarts. The repo already ships `bin/reconcile-k3d-orbstack-network.py` for exactly this; it should be the recommended path. Manual edit is preserved as a `<details>` fallback.
- **CP cp-values had `mode: unified`** even though the architecture diagram and rest of the guide describe a CP/DP split.
- **`publicSignups` / `emailConfirmation` nested form** trips a Houston 1.1.x resolver bug (`AuthConfig.publicSignup` declared Boolean, resolver returns wrapper object) that breaks the login page. Flat Boolean form works. 
- **Hardcoded `/Users/karankhanchandani/...` paths** in a public repo should be parameterized.
- **Dead config** (DP `elasticsearch:` block; redundant `astroUI.env` override; broken `pip install -r tests/requirements.in`) was confusing and broke first-run.

### DIFF
--- a/docs/cp-dp-k3d-setup-guide.md
+++ b/docs/cp-dp-k3d-setup-guide.md
@@ -41,14 +41,9 @@ This guide walks you through setting up Astronomer with a Control Plane (CP) and
 ## Step 1: Install Prerequisites
 
 ```bash
-# Navigate to the astronomer directory
-cd /Users/karankhanchandani/codebase/python/astronomer
-
-# Activate the existing virtual environment
-source venv/bin/activate
-
-# Install required Python packages
-pip install -r tests/requirements.in
+# Set this to wherever you cloned astronomer/astronomer; the rest of this guide uses it.
+export ASTRONOMER_REPO="${ASTRONOMER_REPO:-$HOME/astronomer/astronomer}"
+cd "$ASTRONOMER_REPO"
 
 # Install k3d
 curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
@@ -70,7 +65,7 @@ kubectl version --client
 ## Step 2: Generate TLS Certificates
 
 ```bash
-cd /Users/karankhanchandani/codebase/python/astronomer
+cd "$ASTRONOMER_REPO"
 
 # Install mkcert if not already installed
 # macOS: brew install mkcert
@@ -278,11 +273,12 @@ kubectl --context k3d-data -n astronomer get secrets
 ```bash
 cat > /tmp/cp-values.yaml << 'EOF'
 # Control Plane Helm Values for k3d
-# Control Plane Helm Values for k3d
 global:
   baseDomain: localtest.me
   plane:
-    mode: unified
+    # mode: control matches the CP/DP architecture in this guide. Use mode: unified
+    # only if you intentionally want CP + DP components in one cluster.
+    mode: control
     domainPrefix: ""
   tlsSecret: astronomer-tls
   postgresqlEnabled: true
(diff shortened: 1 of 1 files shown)


docs/cp-dp-k3d-setup-guide.md
```diff
@@ -337,10 +326,12 @@ astronomer:
...
-      emailConfirmation:
-        enabled: false
-      publicSignups:
-        enabled: false
+      # Use flat Boolean form for these two fields — the nested form
+      # (publicSignups: { enabled: false }) trips a resolver bug in Houston
+      # 1.1.x where AuthConfig.publicSignup is declared Boolean but the
+      # resolver returns the wrapper object, breaking the login page.
+      emailConfirmation: false
+      publicSignups: false
```
rishkarajgi, line 334: The flat-Boolean workaround for the AuthConfig.publicSignup resolver bug is great context to capture here, but is there an open Houston issue tracking the actual fix? Worth linking so we know when the nested form becomes safe again and the doc note can be removed.

rishkarajgi (approved):
Doc-only refactor of the CP/DP k3d setup guide that swaps in `$ASTRONOMER_REPO`, promotes the `bin/reconcile-k3d-orbstack-network.py` helper as the primary path (with the manual CoreDNS edit kept as a `<details>` fallback), defaults to LB-IP routing for port-less Houston links, and prunes dead config (DP `elasticsearch:` block, redundant `astroUI.env`, broken `pip install`).

**Verdict:** ✅ Looks good — opinionated cleanup of a guide that nobody could follow end-to-end before; the rationales in the PR body line up with the actual diff and the manual flow is preserved as a fallback.

karankhanchandani (approved):
Thanks @leestro for updating the doc. 

Can you run `pre-commit` once and push?
Also, if the PR description does not have a ticket or does not makes sense to have a ticket, you can add `ktlo` label to the PR.

Approved the PR.

docs/cp-dp-k3d-setup-guide.md
leestro: No open issue — this is already fixed in release-2.0. The publicSignup resolver was updated from config.get("publicSignups") to config.get("publicSignups.enabled") in houston-api src/resolvers/types/auth-config/index.js, so the nested form works correctly there. The doc note can be dropped once this guide targets 2.0+.

leestro:
Done — ran pre-commit (codespell caught a re-use → reuse fix in the stuck-release section), pushed the fix, and added the ktlo label. Thanks for the review!

Simpcyclassy (approved):
Docs-only update to the CP/DP k3d local setup guide. The changes are well-motivated: promoting LB-IP routing (Option B) to the default, pointing to the existing `bin/reconcile-k3d-orbstack-network.py` script instead of the manual CoreDNS edit, fixing the CP values `mode: unified` typo, and removing dead/broken config. No template or runtime changes. Good to merge.

karankhanchandani:
This is an outdated guide and code has changed after that. Closing this PR.

PR #3222, opened 2026-05-05
PINF-559: Add securitycontext null

Shubham-astro:
## Description

Adds securitycontext as null in houston configmap as we are doing merge in houston which merges current securitycontext 'empty' values to default values in airflow-chart which is causing issues in openshift cluster.

## Related Issues
https://linear.app/astronomer/issue/PINF-559/git-sync-merging-security-context-with-default-values

## Testing

- Added SecurityContext as Null in houston:
Screenshot 2026-05-05 at 6 04 21 PM

- Confirmed pods are up and deployment healthy: 
Screenshot 2026-05-05 at 6 06 03 PM

Screenshot 2026-05-05 at 6 06 33 PM

## Merging

Master, 2.0

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -266,7 +266,7 @@ data:
             gitSync:
               repository: {{ .Values.global.gitSyncRelay.images.gitSync.repository }}
               tag: {{ .Values.global.gitSyncRelay.images.gitSync.tag }}
-          {{ if .Values.global.openshift.enabled }}
+          {{- if .Values.global.openshift.enabled }}
       # We have to set this to null to override the default security context that is not compatible with OpenShift. The individual containers have their own security contexts that are compatible with OpenShift.
       securityContext: null
           gitDaemon:
@@ -277,7 +277,7 @@ data:
             securityContext:
                 readOnlyRootFilesystem: true
                 runAsNonRoot: true
-          {{ end }}
+          {{- end }}
 
 
         airflow:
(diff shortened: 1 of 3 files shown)


rishkarajgi (approved):
Switches `securityContext: {}` to `securityContext: null` in the houston gitSyncRelay configmap on OpenShift, so houston's merge logic doesn't end up applying the airflow-chart defaults that aren't OpenShift-compatible. Test updated to assert the new `None` value.

**Verdict:** ✅ Looks good — small, well-scoped fix for an OpenShift compatibility issue. Reasoning is captured nicely in the inline YAML comment.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -267,7 +267,7 @@ data:
               repository: {{ .Values.global.gitSyncRelay.images.gitSync.repository }}
               tag: {{ .Values.global.gitSyncRelay.images.gitSync.tag }}
           {{ if .Values.global.openshift.enabled }}
-          securityContext: {}
+          securityContext: null # We have to set this to null to override the default security context that is not compatible with OpenShift. The individual containers have their own security contexts that are compatible with OpenShift.
```
pgvishnuram: can we add comment above this 
```suggestion
      # We have to set this to null to override the default security context that is not compatible with OpenShift. The individual containers have their own security contexts that are compatible with OpenShift.
      securityContext: null
```

pgvishnuram: This also can introduce a risk of customer adding runAsUser or fsGroup here and that will make their deployments to break again can we also add the omit config in the airflowChart

pgvishnuram (approved):

PR #3223, opened 2026-05-05
PINF-560 Add update-ca-certificates in houston wrapper configMap

sudarshanrampuria:
## Description

- Houston API on the logging-sidecar path bypassed the image's entrypoint (bin/entrypoint), which is where `update-ca-certificates` runs. As a result, certs from `global.privateCaCerts` were mounted at `/usr/local/share/ca-certificates/` but never added to the trust bundle, so Node TLS calls (e.g. Houston → data-plane Commander) failed.
- Fix: invoke `update-ca-certificates` inside the shared log-wrapper script. Same ConfigMap is mounted by houston-api and houston-worker, so both stay covered uniformly.

## Related Issues

https://linear.app/astronomer/issue/PINF-560/houston-does-not-trust-private-ca-during-deployment-creation

## Testing

https://astronomer.slack.com/archives/C0A8A36G7HT/p1777458575571209

## Merging

This needs to go in `2.0.1`

### DIFF
--- a/charts/astronomer/templates/houston/api/houston-log-wrapper-configmap.yaml
+++ b/charts/astronomer/templates/houston/api/houston-log-wrapper-configmap.yaml
@@ -21,6 +21,7 @@ data:
     #!/bin/sh
     # Wrapper script to redirect Houston logs to files for Vector sidecar
     # stdout/stderr are tee'd to log files so kubectl logs still works
+    echo "update ca certs on startup..."
     update-ca-certificates
     exec "$@" \
       1> >(tee -a /var/log/sidecar-log-consumer/out.log) \
--- a/charts/astronomer/templates/houston/api/houston-log-wrapper-configmap.yaml
+++ b/charts/astronomer/templates/houston/api/houston-log-wrapper-configmap.yaml
@@ -21,6 +21,7 @@ data:
     #!/bin/sh
     # Wrapper script to redirect Houston logs to files for Vector sidecar
     # stdout/stderr are tee'd to log files so kubectl logs still works
+    update-ca-certificates
     exec "$@" \
       1> >(tee -a /var/log/sidecar-log-consumer/out.log) \
       2> >(tee -a /var/log/sidecar-log-consumer/err.log)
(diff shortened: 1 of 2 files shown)


charts/astronomer/templates/houston/api/houston-log-wrapper-configmap.yaml
```diff
@@ -21,6 +21,7 @@ data:
     #!/bin/sh
     # Wrapper script to redirect Houston logs to files for Vector sidecar
     # stdout/stderr are tee'd to log files so kubectl logs still works
+    update-ca-certificates
```
pgvishnuram, line 25: echo "update ca certs on startup..." # atleast to make sure it started 
update-ca-certificates

sudarshanrampuria: Added

charts/astronomer/templates/houston/api/houston-log-wrapper-configmap.yaml
```diff
@@ -21,6 +21,8 @@ data:
     #!/bin/sh
     # Wrapper script to redirect Houston logs to files for Vector sidecar
     # stdout/stderr are tee'd to log files so kubectl logs still works
+    echo "update ca certs on startup..."
+    update-ca-certificates
     exec "$@" \
```
rishkarajgi, line 26: Quick check: `update-ca-certificates` typically needs to write to `/etc/ssl/certs/`, which usually requires root. Does the houston-api container in the logging-sidecar path actually run as root? If it doesn't, this will fail silently (the script has no `set -e` and we discard the exit code), and the original cert-loading bug stays in place. Worth adding a quick log line on failure or a runtime check so we don't ship a no-op fix.

rishkarajgi (commented):
Adds an `update-ca-certificates` invocation at the top of the houston log-wrapper script so private CAs from `global.privateCaCerts` get loaded on the logging-sidecar code path.

**Verdict:** 💬 A few questions — directionally correct, but want to make sure the command actually runs successfully at the privilege level the wrapper executes at.

charts/astronomer/templates/houston/api/houston-log-wrapper-configmap.yaml
sudarshanrampuria: Root is NOT required here. `/etc/ssl/certs/` is drwxrwxrwx because the chart's etc-ssl-certs-copier init container mounts an emptyDir there and the houston Dockerfile does `chmod -R ug+rwx /etc/ssl/certs`

rishkarajgi (approved):

pgvishnuram:
can we add tests pls

PR #3224, opened 2026-05-05
Disable kernel_hung collector on node-exporter to reduce log noise

Shubham-astro:
## Description

The node-exporter DaemonSet on EKS clusters running Amazon Linux 2023 (`6.12.73-95.123.amzn2023.x86_64`) logs repeated collector failed errors for the `kernel_hung` collector on every scrape interval across all nodes.

The kernel on these nodes is compiled without `CONFIG_DETECT_HUNG_TASK`, so `/host/proc/sys/kernel/hung_task_detect_count` is absent. The collector fails every time it tries to read this file, generating significant log noise that makes it harder to spot real issues.

This does not cause metric loss — but the volume of error logs is operationally disruptive.

## Related Issues

https://linear.app/astronomer/issue/PINF-528/disable-kernel-hung-collector-on-amazon-linux-2023-nodes ﻿

## Testing

- Just addition of an args, the lognoise should be reduced with change.
- Unittests added
 
## Merging

Master, 2.0

### DIFF
--- a/charts/prometheus-node-exporter/templates/daemonset.yaml
+++ b/charts/prometheus-node-exporter/templates/daemonset.yaml
@@ -48,6 +48,7 @@ spec:
             - --path.procfs=/host/proc
             - --path.sysfs=/host/sys
             - --web.listen-address=$(HOST_IP):{{ .Values.service.port }}
+            - --no-collector.kernel_hung
 {{- if .Values.extraArgs }}
 {{ toYaml .Values.extraArgs | indent 12 }}
 {{- end }}
--- a/tests/chart_tests/test_prometheus_node_exporter.py
+++ b/tests/chart_tests/test_prometheus_node_exporter.py
@@ -161,3 +161,50 @@ class TestPrometheusNodeExporterDaemonset:
         assert len(spec["tolerations"]) > 0
         assert spec["nodeSelector"] == values["prometheus-node-exporter"]["nodeSelector"]
         assert spec["tolerations"] == values["prometheus-node-exporter"]["tolerations"]
+
+    def test_prometheus_node_exporter_daemonset_extra_args(self, kube_version):
+        """Test to validate node exporter with extraArgs configured."""
+        docs = render_chart(
+            kube_version=kube_version,
+            values={
+                **enable_node_exporter,
+                "prometheus-node-exporter": {
+                    "extraArgs": [
+                        "--something-to-test",
+                    ],
+                },
+            },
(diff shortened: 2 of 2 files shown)


charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -48,6 +48,7 @@ spec:
             - --path.procfs=/host/proc
             - --path.sysfs=/host/sys
             - --web.listen-address=$(HOST_IP):{{ .Values.service.port }}
+            - --no-collector.kernel_hung
 {{- if .Values.extraArgs }}
```
rishkarajgi, line 52: Nit: this is hardcoded rather than gated on a value. If someone ever needs to re-enable the kernel_hung collector for debugging they can do it via `extraArgs` (which renders after this line and so wins), but worth a one-liner comment here noting the precedence so the next person doesn't think they have to fork the chart.

tests/chart_tests/test_prometheus_node_exporter.py
```diff
@@ -161,3 +161,50 @@ def test_node_exporter_nodepool_config_overrides(self, kube_version, global_plat
...
+        )
+
+        assert len(docs) == 1
+        doc = docs[0]
+        self.node_exporter_common_tests(doc)
+
+        c_by_name = get_containers_by_name(doc)
+        args = c_by_name["node-exporter"]["args"]
+        assert len(args) == 4
+        assert "--path.procfs=/host/proc" in args
```
rishkarajgi, line 207: Nit: `assert len(args) == 4` will break the next time someone adds another default flag. The `in` checks below already prove the four flags you care about are present — feels safe to drop the length check, or at least leave a comment explaining why the count is load-bearing.

rishkarajgi (commented):
Disables the `kernel_hung` collector on the prometheus-node-exporter daemonset to cut log noise, with a test pair that exercises both the with-extraArgs and no-extraArgs branches.

**Verdict:** ✅ Looks good — small, well-tested change. Couple of nit-level suggestions inline but nothing blocking.

charts/prometheus-node-exporter/templates/daemonset.yaml
pgvishnuram: can we use similar to what we had for kube-state-metrics collector

pgvishnuram: collector:[]
and the lets join them

pgvishnuram: can we also remove the collector which we dont even use for our product

PR #3225, opened 2026-05-06
PINF-528: Add disabledCollectors support to node-exporter and disable unused collectors

Shubham-astro:
## Description

The node-exporter DaemonSet logs repeated collector failed errors for the kernel_hung collector on EKS clusters running Amazon Linux 2023 (kernel compiled without CONFIG_DETECT_HUNG_TASK). Beyond this immediate issue, node-exporter runs all default collectors, many of which produce metrics the Astronomer platform never queries — adding unnecessary scrape load, storage cost, and log noise.

## Approach

Instead of hardcoding individual `--no-collector.*` flags via `extraArgs`, this PR introduces a dedicated `disabledCollectors` list in values (similar to how kube-state-metrics uses collectors: [] with --resources={{ join "," }}). The template iterates over the list and renders --no-collector.<name> args.

## Related Issues

## Testing

Testing is pending on this PR

## Merging


charts/prometheus-node-exporter/templates/daemonset.yaml
```diff
@@ -48,6 +48,10 @@ spec:
             - --path.procfs=/host/proc
             - --path.sysfs=/host/sys
             - --web.listen-address=$(HOST_IP):{{ .Values.service.port }}
+            - --collector.disable-defaults
+{{- range .Values.collectors }}
+            - --collector.{{ . }}
+{{- end }}
```
pgvishnuram, line 50: {{ .Values.collectors | join "," }}  -> can you also see this

Shubham-astro: @pgvishnuram I think there is no `--collectors=cpu,diskstats,filesystem` equivalent in node-exporter..
that's why i used range method here.

Simpcyclassy (changes requested):
This PR aims to fix the `kernel_hung` collector error on AL2023 nodes, which is a real problem worth solving. The implementation is broken in a way that will impact all existing deployments.

**Breaking change for all installations**

The template unconditionally adds `--collector.disable-defaults` to every node-exporter DaemonSet, then re-enables only the 10 collectors listed in `values.yaml`. The effect is that any installation upgrading to a chart version containing this PR will immediately lose metrics from all default collectors not on the whitelist — `conntrack`, `hwmon`, `ipvs`, `mdadm`, `buddyinfo`, `drbd`, `rapl`, `schedstat`, `sockstat`, `tapestats`, `thermal_zone`, and others. Any Prometheus rules or Grafana dashboards that rely on those metrics will stop working silently.

The right approach here is to make `--collector.disable-defaults` opt-in. Add a boolean value like `disableDefaultCollectors: false` and wrap the entire block:

```yaml
{{- if .Values.disableDefaultCollectors }}
- --collector.disable-defaults
{{- range .Values.collectors }}
- --collector.{{ . }}
{{- end }}
{{- end }}
```

Or, follow the PR description's original intent and implement a `disabledCollectors` list that appends `--no-collector.<name>` flags instead, which is additive and non-breaking.

**No tests**

The PR description says "Testing is pending." The chart test suite in `tests/chart_tests/` should have a test that verifies the rendered DaemonSet args for the base case and for the disabled-collectors case. The test must follow the existing parametrization pattern for K8s 1.31-1.35.

**Merging section not filled in**

No release branches listed.

The kernel_hung fix is worth having, but not at the cost of silently breaking metric collection for every existing installation. Needs a redesign before merge.

PR #3226, opened 2026-05-06
Refactor k3d cluster names

danielhoherd:
## Description

- Refactor k3d setup to not map `data-N` cluster to `dp0N`, and to use `cp0N` instead of `control` for the cp cluster.
- Fix devcontainer ruff complaint
- Reduce prometheus resources in local dev config

## Related Issues

This was prerequisite work for https://linear.app/astronomer/issue/PLA-211/dr-test-dp-link-crash-recovery

## Testing

I've tested this locally and it works as expected. It does change the command line arguments though, which is a breaking change, so devs will need to account for that in their scripts.

None of these changes are shipped to the customer, they are all docs and dev related changes, so no QA work is needed.

## Merging

master, release-2.0

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -11,7 +11,10 @@
         "ms-python.vscode-pylance",
         "ms-vscode-remote.remote-containers",
         "ms-vscode.makefile-tools"
-      ]
+      ],
+      "settings": {
+        "ruff.path": ["/root/.local/bin/ruff"]
+      }
     }
   }
 }
--- a/bin/reconcile-k3d-orbstack-network.py
+++ b/bin/reconcile-k3d-orbstack-network.py
@@ -188,12 +188,12 @@ class Settings:
 
 def _settings_from_env() -> Settings:
     return Settings(
-        cp_context=os.environ.get("CP_CONTEXT", "k3d-control"),
-        dp_context=os.environ.get("DP_CONTEXT", "k3d-data"),
+        cp_context=os.environ.get("CP_CONTEXT", "k3d-cp01"),
+        dp_context=os.environ.get("DP_CONTEXT", "k3d-dp01"),
         platform_namespace=os.environ.get("PLATFORM_NAMESPACE", "astronomer"),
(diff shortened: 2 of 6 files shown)


rishkarajgi (commented):
<ignore>

rishkarajgi (commented):
Renames k3d clusters from `control`/`data-N` to `cp01`/`dp01`, and refactors the single `cp_cluster_name` + `Ports` pair into a `ControlPlane` dataclass that mirrors `DataPlane`. The auto-derived DP port defaults (`cp_base_https_port + cp_count`) are a nice touch — no extra args needed to avoid port collisions.

**Verdict:** 💬 A few questions — mostly around how multi-CP is intended to work once `--cp-count > 1` is used in practice.

---

**`bin/setup-cp-dp-k3d.py`, DNS reconcile always pins to `control_planes[0]` (~lines 1490, 1514, 1094)**

When `--cp-count > 1`, `_run_dns_reconcile` and `_ensure_dp_node_houston_hosts_pin` always drive every DP's CoreDNS + `/etc/hosts` pinning off `control_planes[0]` (cp01 only). If multiple CPs are meant to be HA peers behind a shared VIP this is correct — but if the intent is independent CPs each paired with its own DP, you'd want each DP reconciled against its own CP instead. Worth a comment or a brief doc note clarifying the expected topology?

**`bin/setup-cp-dp-k3d.py`, `_print_host_etc_hosts_instructions` loops over all CPs, same hostnames (~line 1124)**

Nit: with `--cp-count > 1`, the loop emits a `/etc/hosts` line for `{base}`, `app.{base}`, `houston.{base}`, etc. for *each* CP's distinct LB IP. On macOS/Linux the last matching entry wins, so for 2+ CPs only the last one's IP would resolve those hostnames. Probably fine given multi-CP is a future case, but a `# TODO` comment here would help whoever tackles it.

danielhoherd:
I've pinned cp-count to 1 for now since we do not yet handle multiple cp's. I think this addresses all of the concerns left by @rishkarajgi.

rishkarajgi (approved):

karankhanchandani (approved):

PR #3227, opened 2026-05-06
pla-182: make stage cluster automation part of internal chart releases

sudarshanrampuria:
## Description

The PR adds stage-cluster build alongside the existing release path. Previously, this only ran when the qa-release pipeline parameter was manually set, in a separate workflow (That option still exists; I am not removing the manual workflow):

1. **approve-stage-cluster-test** - Manual approval gate.
2. **build-artifact (stage-cluster instance)** - Runs with `qa_release: true` and the new `artifact_subdir: qa` parameter. Default artifact_subdir: "." preserves existing behavior (chart .tgz at workspace root); when set, the chart is written to $(pwd)/<artifact_subdir>/. This keeps prod and stage-cluster artifacts separate in the shared workspace so downstream jobs don't pick up the wrong .tgz via glob.
3. **release-to-internal (stage-cluster instance)** - Also takes artifact_subdir, so it only publishes the QA build.
4. **trigger-stage-cluster-build** - kicks off the downstream stage-cluster pipeline and waits for it to complete (replaces the previous fire-and-forget behavior).

### Other changes

Removed `upgrade-test` jobs from the release pipeline. I don't think we are using that. Also, downstream doesn't support that as of now.

## Related Issues

https://linear.app/astronomer/issue/PLA-182/make-stage-cluster-automation-block-internal-chart-releases

## Testing

Did a test by removing the release branch filter and removing the release script and just printing artefacts present in `release-to-internal` stage.
```
echo "Releasing the Astronomer chart to internal with the following artifact: /tmp/workspace/<< parameters.artifact_subdir >>/astronomer-*.tgz"
ls -l /tmp/workspace/<< parameters.artifact_subdir >>/astronomer-*.tgz
```

Before:
image

After:
Screenshot 2026-05-06 at 9 54 31 PM


sudarshanrampuria:
**NOTE:** Stage-cluster trigger is intentionally parallel to, not a prerequisite of, approve-internal-release. In 2.x RC cycles we saw that the stage-cluster pipeline frequently surfaced unrelated failures that would have blocked internal releases if it were a hard gate.

bin/trigger_stage_cluster_update.py
```diff
@@ -50,6 +50,25 @@ def get_job_state(circleci_token: str, pipeline_id: str):
...
 
 
+def wait_for_pipeline_completion(circleci_token: str, pipeline_id: str, wait_time_min: int = 120, check_interval_sec: int = 30):
+    elapsed_time = 0
+    while elapsed_time < wait_time_min * 60:
+        job_state_resp = get_job_state(circleci_token=circleci_token, pipeline_id=pipeline_id)
+        pipeline_state = json.loads(job_state_resp)["items"][0]["status"]
+
+        if pipeline_state == "success":
+            print("INFO: Pipeline completed successfully.")
```
rishkarajgi, line 60: This now blocks the CI job for up to 120 minutes by default. The old approach only waited ~60 seconds before proceeding regardless. Was this intentional? If the stage cluster update is expected to take that long and a timeout is a hard failure, that's fine — just want to make sure this won't be holding up release pipelines unexpectedly.

bin/trigger_stage_cluster_update.py
```diff
@@ -88,23 +107,8 @@ def main(circleci_token: str, astro_path: str, branch: str):
     # Printing Info
```
rishkarajgi, line 107: Nit: `TimeoutError` is a built-in but raising it here will cause the CircleCI step to exit non-zero, which is correct. Just worth confirming this is the desired behavior — if the stage cluster update is still in-flight after 2 hours, we want the pipeline to fail loudly rather than silently succeed.

rishkarajgi (commented):
Adds a stage-cluster build path to the release workflow, running in parallel with the existing prod internal release gate. The `artifact_subdir` parameter cleanly separates stage and prod chart artifacts in the shared workspace, and renaming `trigger_feature_stack_update.py` → `trigger_stage_cluster_update.py` makes the intent clearer.

**Verdict:** 💬 A few questions — mostly around the blocking behavior of the new wait loop and the removal of upgrade test triggering.

**On the upgrade test removal:** `trigger-upgrade-test` and `approve-upgrade-test` are removed from the automated workflow, and `bin/trigger_rc_tests.py` is deleted. The PR description mentions "the manual workflow still exists" — is there a separate CircleCI config or pipeline that covers this path? Just want to make sure upgrade test coverage isn't accidentally dropped from the release process.

bin/trigger_stage_cluster_update.py
sudarshanrampuria: The old approach didn't wait for downstream stage cluster pipeline to complete(success or fail). It just waited till the pipeline goes from pending state -> running state. 

Generally, the stage cluster pipeline take 1hr including upgrading helm release and running all test cases. I have just given some buffer because the time increases if there is any infra changes to the stage cluster.

bin/trigger_stage_cluster_update.py
sudarshanrampuria: Yes, correct.  This is intentional. We want the pipeline to fail.

rishkarajgi (approved):

PR #3235, opened 2026-05-11
APC-1298: Update Platform images

Shubham-astro:
## Description

This PR updates gitsync image and revert vector for 1.1.3 and also updates to other platform images

## Merging

1.1

### DIFF
--- a/charts/vector/values.yaml
+++ b/charts/vector/values.yaml
@@ -16,7 +16,7 @@ vector:
     runAsUser: 0
   image:
     repository: quay.io/astronomer/ap-vector
-    tag: 0.55.0
+    tag: 0.54.0
     pullPolicy: IfNotPresent
   resources: ~
   livenessProbe: ~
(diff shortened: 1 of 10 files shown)


sudarshanrampuria (approved):

rishkarajgi (commented):
<ignore>

charts/vector/values.yaml
```diff
@@ -16,7 +16,7 @@ vector:
     runAsUser: 0
   image:
     repository: quay.io/astronomer/ap-vector
-    tag: 0.55.0
+    tag: 0.54.0-1
     pullPolicy: IfNotPresent
```
rishkarajgi, line 20: Nit: just confirming the vector downgrade from 0.55.0 → 0.54.0-1 is intentional for the 1.1.3 branch? PR description calls it out, but worth double-checking there's no regression with rolling back a minor version.

rishkarajgi (commented):
<ignore>

PR #3237, opened 2026-05-13
PLX-426: Fix COMMANDER_HOUSTON_JWKS_ENDPOINT being overridden on helm upgrades

karankhanchandani:
## Summary

Fixes a bug where `COMMANDER_HOUSTON_JWKS_ENDPOINT` was reset on every Helm upgrade, 
ignoring any override set via `commander.env`. Reported by Tesco, who had to manually 
patch the value after each upgrade to add `.svc.cluster.local` to avoid connection timeouts.

## Root Cause

Two issues combined to cause this:

1. **Env ordering**: `commander.env` (user overrides) was rendered *before* 
   `commander_metadataEnv` in the deployment template. Since Kubernetes uses 
   last-wins for duplicate env var names, the hardcoded default from 
   `commander_metadataEnv` always silently won.

2. **Missing `.svc.cluster.local`**: `houston.authServiceURL` in unified plane 
   mode was generating `http://<release>-houston.<namespace>:<port>` without 
   `.svc.cluster.local`, causing DNS resolution timeouts in some cluster 
   configurations.

## Changes

- **`commander-deployment.yaml`**: Moved `commander.env` block to after 
  `commander_metadataEnv` so user-provided values always take precedence.

- **`_helpers.yaml` (`commander_metadataEnv`)**: Added a conditional check that 
  skips rendering `COMMANDER_HOUSTON_JWKS_ENDPOINT` when it is already present 
  in `commander.env`. This prevents duplicate env var names, which Kubernetes 
  rejects via strategic merge patch on upgrade.

- **`_helpers.yaml` (`houston.authServiceURL`)**: Unified plane URL now includes 
  `.svc.cluster.local` for reliable in-cluster DNS resolution.

## Related Issues
https://linear.app/astronomer/issue/PLX-426/commander-jwks-endpoint-override-ignored-on-helm-upgrades

## Testing

- Unit tests
- generated the template locally, rendering fine

## Merging

1.1

### DIFF
--- a/charts/astronomer/templates/commander/commander-deployment.yaml
+++ b/charts/astronomer/templates/commander/commander-deployment.yaml
@@ -126,10 +126,6 @@ spec:
             periodSeconds: 10
           {{- end }}
           env:
-            {{- range $i, $config := .Values.commander.env }}
-            - name: {{ $config.name }}
-              value: {{ $config.value | quote }}
-            {{- end }}
             - name: HELM_CACHE_HOME
               value: /tmp/helm-cache
             - name: HELM_CONFIG_HOME
@@ -163,6 +159,10 @@ spec:
               value: "true"
             {{- end }}
             {{- include "commander_metadataEnv" . | nindent 12 }}
+            {{- range $i, $config := .Values.commander.env }}
+            - name: {{ $config.name }}
+              value: {{ $config.value | quote }}
+            {{- end }}
           volumeMounts:
             - name: etc-ssl-certs
               mountPath: /etc/ssl/certs
--- a/charts/astronomer/templates/commander/commander-deployment.yaml
+++ b/charts/astronomer/templates/commander/commander-deployment.yaml
@@ -126,6 +126,10 @@ spec:
             periodSeconds: 10
           {{- end }}
           env:
+            {{- range $i, $config := .Values.commander.env }}
+            - name: {{ $config.name }}
+              value: {{ $config.value | quote }}
+            {{- end }}
             - name: HELM_CACHE_HOME
               value: /tmp/helm-cache
             - name: HELM_CONFIG_HOME
@@ -159,10 +163,6 @@ spec:
               value: "true"
             {{- end }}
             {{- include "commander_metadataEnv" . | nindent 12 }}
-            {{- range $i, $config := .Values.commander.env }}
-            - name: {{ $config.name }}
-              value: {{ $config.value | quote }}
-            {{- end }}
           volumeMounts:
             - name: etc-ssl-certs
               mountPath: /etc/ssl/certs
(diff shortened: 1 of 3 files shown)


Shubham-astro:
@karankhanchandani  does this fix exist in master.? if not let's create a PR to master branch then cherry-pick to 1.1

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -163,6 +159,10 @@ spec:
               value: "true"
             {{- end }}
             {{- include "commander_metadataEnv" . | nindent 12 }}
+            {{- range $i, $config := .Values.commander.env }}
+            - name: {{ $config.name }}
+              value: {{ $config.value | quote }}
+            {{- end }}
```
pgvishnuram: this change will make all the env overidable even from the defaults

charts/astronomer/templates/commander/commander-deployment.yaml
```diff
@@ -163,6 +159,10 @@ spec:
               value: "true"
             {{- end }}
             {{- include "commander_metadataEnv" . | nindent 12 }}
+            {{- range $i, $config := .Values.commander.env }}
+            - name: {{ $config.name }}
+              value: {{ $config.value | quote }}
```
rishkarajgi: This is the key ordering fix — user-supplied env vars now come after `commander_metadataEnv`, so Kubernetes last-wins semantics ensure overrides take effect. Looks right.

rishkarajgi (approved):
Fixes PLX-426 with a two-pronged approach: moves `commander.env` rendering to after `commander_metadataEnv` so user overrides win via Kubernetes last-wins semantics, and adds a `$jwksOverridden` guard in the metadata template to suppress the default entry entirely when the user has already provided the value.

**Verdict:** ✅ Looks good — both root causes are addressed, tests cover the override regression explicitly, and the `len(jwks_entries) == 1` assertion will catch any future re-introduction of the duplicate. The `$` vs `.` context fix in the `include` call is also correct for a named template invoked via `include`.

**Additional notes** (could not anchor inline):

**charts/astronomer/templates/_helpers.yaml** (line ~218): Nit: the `$jwksOverridden` scan is a linear walk over `$.Values.commander.env` at template render time. For the typical case (small env list) this is totally fine. Just worth noting that it relies on `commander.env` entries having a `name` field — if someone passes a map instead of a list-of-objects, Helm would error here rather than silently skipping the guard. The existing `range` loop in the deployment template has the same assumption, so this is consistent.

Shubham-astro:
@karankhanchandani  do fix the pre-commit before merging, LGTM.

Shubham-astro (approved):

charts/astronomer/templates/commander/commander-deployment.yaml
pgvishnuram: @rishkarajgi on the templates it looks right but if you see the underlying commander_metadataEnv helper tpl we generate default values that will always takes precedence even if customer pass those values as a part of env - we have to decide first whether do he have to allow overrides over our defaults to give full contol of the variables and a internal discussion is already going on it

PR #3238, opened 2026-05-13
PLX-426: Fix COMMANDER_HOUSTON_JWKS_ENDPOINT being overridden on helm upgrades

karankhanchandani:
## Summary

Fixes a bug where `COMMANDER_HOUSTON_JWKS_ENDPOINT` was reset on every Helm upgrade, 
ignoring any override set via `commander.env`. Reported by Tesco, who had to manually 
patch the value after each upgrade to add `.svc.cluster.local` to avoid connection timeouts.

## Root Cause

Two issues combined to cause this:

1. **Env ordering**: `commander.env` (user overrides) was rendered *before* 
   `commander_metadataEnv` in the deployment template. Since Kubernetes uses 
   last-wins for duplicate env var names, the hardcoded default from 
   `commander_metadataEnv` always silently won.

2. **Missing `.svc.cluster.local`**: `houston.authServiceURL` in unified plane 
   mode was generating `http://<release>-houston.<namespace>:<port>` without 
   `.svc.cluster.local`, causing DNS resolution timeouts in some cluster 
   configurations.

## Changes

- **`commander-deployment.yaml`**: Moved `commander.env` block to after 
  `commander_metadataEnv` so user-provided values always take precedence.

- **`_helpers.yaml` (`commander_metadataEnv`)**: Added a conditional check that 
  skips rendering `COMMANDER_HOUSTON_JWKS_ENDPOINT` when it is already present 
  in `commander.env`. This prevents duplicate env var names, which Kubernetes 
  rejects via strategic merge patch on upgrade.

- **`_helpers.yaml` (`houston.authServiceURL`)**: Unified plane URL now includes 
  `.svc.cluster.local` for reliable in-cluster DNS resolution.

## Related Issues
https://linear.app/astronomer/issue/PLX-426/commander-jwks-endpoint-override-ignored-on-helm-upgrades

## Testing

- Unit tests
- generated the template locally, rendering fine

## Merging

master

### DIFF
--- a/tests/chart_tests/test_astronomer_commander.py
+++ b/tests/chart_tests/test_astronomer_commander.py
@@ -203,7 +203,7 @@ class TestAstronomerCommander:
         assert env_vars["COMMANDER_ELASTICSEARCH_ENABLED"] == "true"
         assert env_vars["COMMANDER_ELASTICSEARCH_LOG_LEVEL"] == "info"
         assert env_vars["COMMANDER_ELASTICSEARCH_NODE"] == "http://release-name-elasticsearch.default.svc.cluster.local.:9200"
-        assert env_vars["COMMANDER_HOUSTON_JWKS_ENDPOINT"] == "http://release-name-houston.default:8871"
+        assert env_vars["COMMANDER_HOUSTON_JWKS_ENDPOINT"] == "http://release-name-houston.default.svc.cluster.local:8871"
         assert env_vars["COMMANDER_MANAGE_NAMESPACE_RESOURCE"] == "true"
         assert env_vars["COMMANDER_REGION"] == "us-west-2"
         assert env_vars["COMMANDER_UPGRADE_TIMEOUT"] == "600"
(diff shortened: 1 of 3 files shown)


tests/chart_tests/test_astronomer_commander.py
```diff
@@ -914,3 +914,29 @@ def test_commander_user_provided_env_vars(self, kube_version):
...
+        """
+        docs = render_chart(
+            kube_version=kube_version,
+            values={"global": {"plane": {"mode": plane_mode}}},
+            show_only=["charts/astronomer/templates/commander/commander-deployment.yaml"],
+        )
+        assert len(docs) == 1
+        commander_env = get_containers_by_name(docs[0])["commander"]["env"]
+        jwks_entries = [e for e in commander_env if e["name"] == "COMMANDER_HOUSTON_JWKS_ENDPOINT"]
+        assert len(jwks_entries) == 1, "duplicate COMMANDER_HOUSTON_JWKS_ENDPOINT entries would break helm upgrade"
```
rishkarajgi, line 941: The `len(jwks_entries) == 1` assertion here will still fail on a cluster where someone has `commander.env` with `COMMANDER_HOUSTON_JWKS_ENDPOINT` set, because the ordering issue in `commander-deployment.yaml` means the default entry still appears after the user's override, producing duplicates. This test only covers the no-override case.

rishkarajgi (commented):
Applies the `.svc.cluster.local` suffix fix to `COMMANDER_HOUSTON_JWKS_ENDPOINT` and adds a parameterized test covering both plane modes.

**Verdict:** ⚠️ Needs changes — this addresses the URL root cause but doesn't fix the env ordering issue that allows the default to override a user-supplied value in `commander.env`. Without the accompanying ordering fix (moving the `{{- range .Values.commander.env }}` block to after `commander_metadataEnv` in `commander-deployment.yaml`), a user who sets `COMMANDER_HOUSTON_JWKS_ENDPOINT` via `commander.env` will end up with two entries for the same env var, which breaks Kubernetes strategic merge patch on helm upgrade. PR #3237 covers both root causes — is this PR intentionally scoped down, or should the deployment template ordering fix be included here too?

tests/chart_tests/test_astronomer_commander.py
karankhanchandani: I removed this based on the slack thread discussion: https://astronomer.slack.com/archives/C08CL6S53J6/p1778651261689699

Shubham-astro (approved):
LGTM.

PR #3239, opened 2026-05-13
PINF-621 fix external es-proxy internal url

pgvishnuram:
## Description

In `unified` plane mode, the ES proxy URL template in `_helpers.yaml` was pointing to `<release>-es-proxy` (the in-cluster ES proxy) instead of `<release>-external-es-proxy`. This caused commander to route external-ES traffic to the wrong service, while the equivalent `data` plane mode was correctly using the external proxy.

This PR fixes the service name and updates the affected unit test expectation.

## Improvements

- Unified plane deployments now route external ES traffic to the correct `external-es-proxy` service, matching the data plane mode behaviour
- Eliminates a divergence between unified and data plane mode that could cause silent routing failures or 5xx errors when commander tried to reach the wrong service

## Related Issues

https://linear.app/astronomer/issue/PINF-621

## Testing

1. Render the chart with `global.plane.mode=unified` and confirm the ES proxy URL helper resolves to `<release>-external-es-proxy.<namespace>.svc.cluster.local:9201`
2. Render with `global.plane.mode=data` and confirm the existing `es-proxy.<domainPrefix>.<baseDomain>` URL is unchanged
3. Run `tests/chart_tests/test_astronomer_commander.py::test_commander_data_plane_failover_permissions_disabled` — should pass with the updated expected URL

## Merging

Merge to master.

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -756,7 +756,7 @@ http://{{ .Release.Name }}-houston:{{ .Values.ports.houstonHTTP }}/v1/registry/e
 {{- if eq .Values.global.plane.mode "data" -}}
 es-proxy.{{ .Values.global.plane.domainPrefix }}.{{ .Values.global.baseDomain }}
 {{- else if eq .Values.global.plane.mode "unified" -}}
-{{ .Release.Name }}-es-proxy.{{ .Release.Namespace }}.svc.cluster.local:9201
+{{ .Release.Name }}-external-es-proxy.{{ .Release.Namespace }}.svc.cluster.local:9201
 {{- end -}}
 {{- end -}}
 
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -753,11 +753,7 @@ http://{{ .Release.Name }}-houston:{{ .Values.ports.houstonHTTP }}/v1/registry/e
 {{- end -}}
 
 {{- define "esproxy.url" -}}
-{{- if eq .Values.global.plane.mode "data" -}}
-es-proxy.{{ .Values.global.plane.domainPrefix }}.{{ .Values.global.baseDomain }}
-{{- else if eq .Values.global.plane.mode "unified" -}}
 {{ .Release.Name }}-external-es-proxy.{{ .Release.Namespace }}.svc.cluster.local:9201
-{{- end -}}
 {{- end -}}
 
 {{- define "elasticsearch.url" -}}
--- a/tests/chart_tests/test_astronomer_commander.py
+++ b/tests/chart_tests/test_astronomer_commander.py
@@ -605,7 +605,7 @@ class TestAstronomerCommander:
             (
                 "unified",
                 True,
-                "http://release-name-es-proxy.default.svc.cluster.local:9201",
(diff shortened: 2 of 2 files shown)


Shubham-astro (approved):
Small fix LGTM.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -756,7 +756,7 @@ http://{{ .Release.Name }}-houston:{{ .Values.ports.houstonHTTP }}/v1/registry/e
 {{- if eq .Values.global.plane.mode "data" -}}
 es-proxy.{{ .Values.global.plane.domainPrefix }}.{{ .Values.global.baseDomain }}
 {{- else if eq .Values.global.plane.mode "unified" -}}
-{{ .Release.Name }}-es-proxy.{{ .Release.Namespace }}.svc.cluster.local:9201
+{{ .Release.Name }}-external-es-proxy.{{ .Release.Namespace }}.svc.cluster.local:9201
 {{- end -}}
```
rishkarajgi, line 757: Quick question: is `external-es-proxy` always deployed in unified mode, or only when external ES is configured? If it's conditional, hitting this template with a cluster that doesn't have the service would fail silently at routing rather than at render time. If there's a values guard somewhere upstream that ensures this branch is only reachable when external ES is set, this is fine — just wanted to confirm.

rishkarajgi (approved):
Fixes a clear service name mismatch in `unified` plane mode — the helper was pointing to `{release}-es-proxy` (the in-cluster proxy) while the equivalent `data` plane path correctly uses the external service. The test expectation is updated to match.

**Verdict:** ✅ Looks good — focused one-line fix with a corresponding test update.

charts/astronomer/templates/_helpers.yaml
pgvishnuram: {{- define "esproxy.url" -}} -> this function called inside commander deployment template with a guard checking the customLogging flag and renders them according - im refactoring this change to make it cleaner as well

Vasu-Madaan: esproxy.url should reference external-es-proxy.fullname directly instead of hardcoding the string:

```
  {{- define "esproxy.url" -}}
  {{- if eq .Values.global.plane.mode "data" -}}
  es-proxy.{{ .Values.global.plane.domainPrefix }}.{{ .Values.global.baseDomain }}
  {{- else if eq .Values.global.plane.mode "unified" -}}
  {{ include "external-es-proxy.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local:9201
  {{- end -}}
  {{- end -}}
```

  This handles all three cases automatically:
  - Default: fordastro-external-es-proxy ✓
  - nameOverride: foo: fordastro-foo ✓
  - fullnameOverride: bar: bar ✓

  This PR only handles the first case. The other two would still be broken after the fix.

Vasu-Madaan: Referencing https://github.com/astronomer/astronomer/blob/df64d0411d3fd05212b70290865ca6537706e49f/charts/external-es-proxy/templates/_helpers.tpl#L13C1-L24C11

pgvishnuram: @Vasu-Madaan direct referencing wont work here since both are two different charts.

PR #3242, opened 2026-05-13
Update images and Airflow Chart with 1.1.3-beta2 

Shubham-astro:
## Description

This PR Updates images and Airflow Chart, and also update 1.1.3-beta2

## Merging

1.1


currentvalues.yaml
```diff
@@ -0,0 +1,106 @@
+USER-SUPPLIED VALUES:
+alertmanager:
```
rishkarajgi: This looks like the output of `helm get values` from a live environment — the `USER-SUPPLIED VALUES:` header and the environment-specific content (`release-1.astronomer-azuredev.com`, `core.harbor.astro-qa.link`, `ocptestting`) are a giveaway. This file shouldn't be committed to the repo; it captures deploy-time state from a specific cluster rather than chart defaults. Please remove it before merging.

airflowvalues.yaml
```diff
@@ -0,0 +1,1300 @@
+USER-SUPPLIED VALUES:
+airflow:
```
rishkarajgi: Same issue here — this is a `helm get values` dump from a live Airflow release (1300 lines). It contains environment-specific configuration that doesn't belong in the chart repo. Please remove it.

rishkarajgi (changes requested):
Bumps several component image tags (`commander` 1.1.10→1.1.12, `houston-api` 1.1.5→1.1.9, `astroUI` 1.1.1→1.1.3, `dbBootstrapper` 2.0.2→2.0.3 across astronomer/grafana/nats charts) and advances `airflowChartVersion` to 1.17.31. The image tag updates look correct — but two new files (`currentvalues.yaml` and `airflowvalues.yaml`) appear to be `helm get values` dumps from a live QA environment and shouldn't be committed here.

**Verdict:** ⚠️ Needs changes — please drop the two stray values files before merging.

airflowvalues.yaml
Shubham-astro: removed

currentvalues.yaml
Shubham-astro: removed

sudarshanrampuria (approved):

charts/astronomer/values.yaml
```diff
@@ -5,7 +5,7 @@
 # This version number controls the default Airflow chart version that will be installed
 # when creating a new deployment in the system. This is also used to ensure all
 # child airflow deployments are kept up to date and on the latest version.
-airflowChartVersion: 1.17.30
+airflowChartVersion: 1.17.31
```
pgvishnuram, line 8: i dont see commander image ??

pgvishnuram (approved):

PR #3250, opened 2026-05-19
PINF-642 - disable default es ingress tpl render when on customLogging

pgvishnuram:
## Description

Stops the in-cluster Elasticsearch ingress from rendering when `global.customLogging.enabled` is `true`. Without this gate, both the customLogging ingress and the in-cluster ES ingress render together, producing duplicate Ingress objects for the same host. customLogging is intended to take precedence, and this PR enforces that at the template level.

### Changes

| File | What changed |
|------|-------------|
| `charts/elasticsearch/templates/es-ingress.yaml` | Wraps the entire manifest in `{{- if not .Values.global.customLogging.enabled }} ... {{- end }}` so the in-cluster ES ingress is skipped when customLogging owns logging. |
| `tests/chart_tests/test_external_elasticsearch.py` | Adds `charts/elasticsearch/templates/es-ingress.yaml` to `show_only` in `test_external_elasticsearch_nginx_deployment_defaults` to assert gating behaviour under the external-ES path. |

### Improvements

- Eliminates duplicate ingress objects when customLogging and in-cluster ES are both configured
- customLogging cleanly takes precedence — no manual override or post-render patching needed
- Covered by the existing external-ES test path

## Related Issues

https://linear.app/astronomer/issue/PINF-642/fix-duplicate-elasticsearch-ingress-when-customlogging-is-enabled

## Testing

- Already tested on the cluster.
- `pytest tests/chart_tests/test_external_elasticsearch.py::test_external_elasticsearch_nginx_deployment_defaults` passes with the new `show_only` entry.
- Rendering with `global.customLogging.enabled=true` produces no in-cluster ES Ingress; rendering with it `false` (default) is unchanged.

## Merging

Master, 1.1, 2.0


charts/elasticsearch/templates/es-ingress.yaml
```diff
@@ -1,6 +1,7 @@
 ########################################
 ## Elasticsearch Logging Ingress ##
 ########################################
+{{- if not .Values.global.customLogging.enabled }}
```
Simpcyclassy, line 4: Correct outer guard placement — `{{- if not .Values.global.customLogging.enabled }}` short-circuits before the plane-mode check, so the ingress is suppressed for all modes when a custom ES backend is active.

AI-assisted review via Claude

tests/chart_tests/test_external_elasticsearch.py
```diff
@@ -777,6 +777,7 @@ def test_external_elasticsearch_nginx_deployment_defaults(self, kube_version):
             },
             show_only=[
                 "charts/external-es-proxy/templates/external-es-proxy-deployment.yaml",
+                "charts/elasticsearch/templates/es-ingress.yaml",
```
Simpcyclassy, line 780: Good addition — including `es-ingress.yaml` in `show_only` with `customLogging.enabled: True` and asserting `len(docs) == 1` is a real negative-case test for this guard; consider adding a corresponding positive-case test in `test_elasticsearch.py` that verifies the ingress renders when `customLogging` is disabled.

AI-assisted review via Claude

Simpcyclassy (dismissed):
The template change is correct — the outer `customLogging.enabled` guard at line 4 prevents the internal ES ingress from rendering when a custom logging backend is in use, and the plane-mode gate is preserved inside it. The test addition is meaningful: including `es-ingress.yaml` in `show_only` while asserting `len(docs) == 1` with `customLogging.enabled: True` confirms the ingress produces no output in that configuration. The PR description is entirely unmodified boilerplate — all three sections (Description, Related Issues, Testing) still contain the "Do not merge this PR until..." placeholder text from the template. The title has a typo ("custonLogging" should be "customLogging") and the related issue link is still the `XXXX` placeholder. Replace the description with the actual problem statement and link the Linear issue before this merges.

AI-assisted review via Claude

Shubham-astro (approved):
Already tested this on a cluster, LGTM.

tests/chart_tests/test_external_elasticsearch.py
```diff
@@ -777,6 +777,7 @@ def test_external_elasticsearch_nginx_deployment_defaults(self, kube_version):
...
             ],
```
rishkarajgi, line 781: Nit: adding `es-ingress.yaml` to `show_only` here means the gating is verified only indirectly via `assert len(docs) == 1` further down — i.e. the es-ingress template gets requested but doesn't render. That works, but if someone later changes the assertion or adds another rendered manifest, the regression check on the new conditional disappears silently. Worth either (a) adding a short comment near the `show_only` block noting that es-ingress.yaml is included to assert it does NOT render when `customLogging.enabled=True`, or (b) splitting in a dedicated tiny test that renders only `es-ingress.yaml` with `customLogging.enabled=True` and asserts the result is empty, plus a positive case with `customLogging.enabled=False` to lock in the default behaviour.

charts/elasticsearch/templates/es-ingress.yaml
```diff
@@ -1,6 +1,7 @@
...
 {{- if or (eq .Values.global.plane.mode "data") (eq .Values.global.plane.mode "unified") }}
```
rishkarajgi, line 5: Quick sanity check — when `global.customLogging` is omitted entirely from the values (not just `enabled: false`), this evaluates `not nil` → true, so the ingress still renders by default. That matches the intended pre-PR behaviour. Worth confirming `global.customLogging.enabled` has a defined default in `values.yaml` (or the parent chart) so future template changes don't accidentally depend on it being unset.

rishkarajgi (commented):
Wraps the in-cluster Elasticsearch ingress template in `{{- if not .Values.global.customLogging.enabled }}` so it no longer renders alongside the customLogging-owned ingress, eliminating the duplicate-Ingress-for-same-host issue. Test coverage is added by including `es-ingress.yaml` in the `show_only` list of `test_external_elasticsearch_nginx_deployment_defaults`, which exercises the `customLogging.enabled=True` path.

**Verdict:** 💬 A few questions — change looks correct and minimal, just a couple of nits around test explicitness and default-value safety.

charts/elasticsearch/templates/es-ingress.yaml
pgvishnuram: for control mode we don't provision elasticsearch any way and only for data and unified we provision elasticsearch, the whole point is to give the customLogging the highest order to make the validation and not generate ingress object for the incluster ingress.so whether we place the guard before or after the plane.mode  condition its  still going to work.

if you see any issues on this conditional placement can you share on what scenario it can fail

PR #3265, opened 2026-05-27
PINF-682 - rework openshift enabled to support podsecurity context

pgvishnuram:
## Description

Replaces all hardcoded toYaml .Values.podSecurityContext calls with named Helm helpers that omit fsGroup, runAsGroup, runAsUser when global.openshift.enabled=true — OpenShift manages UID/GID assignment at the platform level.

## Related Issues

 https://linear.app/astronomer/issue/PINF-682/allow-podsecuritycontext-independent-of-openshift-flag

## Testing

NA

## Merging

merge to master


Shubham-astro (approved):
LGTM

PR #3271, opened 2026-05-29
PLX-514: Upgrade the postgres image for operator

karankhanchandani:
## Description

Upgrade the postgres image for operator. Also refactors the setup scripts for local cluster.
 
## Related Issues

https://linear.app/astronomer/issue/PLX-514/restore-crud-parity-for-operator-mode-deployments-from-0x-to-2x

## Testing

Tested by deploying locally

## Merging

master


charts/airflow-operator/templates/crds/airflow.yaml
```diff
@@ -1500,7 +1500,7 @@ spec:
                     type: object
                     x-kubernetes-preserve-unknown-fields: true
                   image:
-                    default: docker.io/bitnami/postgresql:11.10.0-debian-10-r60
+                    default: astrocrpublic.azurecr.io/astronomer/ap-postgresql:17.5.0
```
Simpcyclassy, line 1503: This is a major version jump. Did we confirm if we need a migration path for the db?

bin/setup-037x-k3d.py
```diff
@@ -822,6 +860,12 @@ def _values_yaml(settings: Settings) -> str:
 postgresql:
   postgresqlUsername: postgres
   postgresqlPassword: postgres
+
+airflow-operator:
+  crd:
+    create: true
```
Simpcyclassy, line 866: What happens on reinstall after crd had been created the first time? Will this be fine or will there be an error on recreating?

bin/setup-cp-dp-k3d.py
```diff
@@ -627,6 +702,28 @@ def _write_values_file(path: Path, content: str) -> None:
...
+    # The airflow-operator subchart is conditionally installed by
+    # global.airflowOperator.enabled. We set that flag on the CP so houston-configmap
+    # learns operator mode is on, but the actual operator controller belongs on the DP.
+    # Neutralize the subchart on the CP: zero replicas, no CRDs (DP owns them), no
+    # webhook configs (no apiserver -> empty-endpoint calls), no cert-manager dance.
+    operator_subchart_block = (
+        """\
+airflow-operator:
+  crd:
+    create: false
```
Simpcyclassy, line 715: Same here as well


What happens on reinstall after crd had been created the first time? Will this be fine or will there be an error on recreating?

charts/airflow-operator/templates/crds/airflow.yaml
karankhanchandani: APC Operator based deployments are not being used anywhere AFAIK.

bin/setup-037x-k3d.py
karankhanchandani: This will be fine. It will not throw an error. It will keep on updating if there are new changes.

bin/setup-cp-dp-k3d.py
karankhanchandani: This will be fine. It will not throw an error. It will keep on updating if there are new changes.

sudarshanrampuria (approved):
LGTM. Just need to fix pre-commit failure before merging

bin/setup-037x-k3d.py
```diff
@@ -180,6 +185,33 @@ def _get_registry_config_path(docker_network: str) -> Path:
 HELM_CHART = f"{HELM_REPO_NAME}/astronomer"
 HELM_REPO_URL = "https://internal-helm.astronomer.io"
 
+CERT_MANAGER_VERSION = "v1.5.4"
```
Shubham-astro: Since we are bumping and writing tests anyway, let's also try to use latest CERT_MANAGER_VERSION now. It wouldn't affect this as this is just test setup, but it would be useful for our customers, if we are aware of something breaking in operator with latest cert manager version. 
```suggestion
CERT_MANAGER_VERSION = "v1.19.4"
```

Shubham-astro (approved):
One small change, otherwise LGTM

bin/setup-037x-k3d.py
ianbuss: We should probably mirror the version in the operator repo and ensure that is updated as well.

bin/setup-037x-k3d.py
```diff
@@ -893,6 +937,12 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument("--helm-timeout", default=os.environ.get("HELM_TIMEOUT", "60m"))
     parser.add_argument("--helm-debug", action="store_true")
     parser.add_argument("--recreate-cluster", action="store_true", help="Delete and recreate k3d cluster if it exists")
+    parser.add_argument(
+        "--agents",
```
ianbuss: nit: I think avoid the use of `agents` here if possible (we already have multiple meanings for that in Astronomer) and its not a general term for nodes in Kubernetes. Maybe prefer `--num-compute-nodes`.

bin/setup-cp-dp-k3d.py
```diff
@@ -167,6 +169,9 @@ def _get_registry_config_path(docker_network: str) -> Path:
   "registry.k8s.io":
     endpoint:
       - "http://{proxy_k8s}:5000"
+  "astrocrpublic.azurecr.io":
```
ianbuss, line 172: Why is this required?

bin/setup-cp-dp-k3d.py
```diff
@@ -729,11 +830,49 @@ def _cp_values_yaml(settings: Settings) -> str:
...
+      repository: quay.io/astronomer/airflow-operator-controller
+      tag: "1.5.2"
+  manager:
+    replicas: 1
+    # Pin the operator controller (which also serves the admission webhook) to
+    # the control-plane node.  The kube-apiserver runs on that same node and
+    # reaches webhook pods via its own pod-network (10.42.0.x).  If the
+    # controller lands on an agent node (10.42.1.x), the apiserver has to
+    # traverse the Flannel VXLAN overlay between Docker containers, which
+    # fails with "502 Bad Gateway" in k3d.
```
ianbuss, line 861: Interesting. I wonder if there's a way around this. Not a blocker but it is interesting. Does the same thing not happen for certmanager webhooks?

bin/setup-cp-dp-k3d.py
karankhanchandani: We have certain images that we pull from this registry. Those images don't change often, so I added a cache for this registry. This way we can spin up clusters locally (in your laptop) faster.

bin/setup-cp-dp-k3d.py
karankhanchandani: Yes this fails for every webhook calls. The webhook should be able to reach  k8s API server. For each node, k3d spins up a new docker container. We have to do some extra networking stuff for making a connection between two docker containers. I attempted to do that, but I was not able to succeed. Our local CP/DP planes thankfully require only one nodes each so they work. Another thing is we don't have enough resources in laptop to spin up more than one airflow deployment. So I decided to not proceed with the networking fix.

bin/setup-037x-k3d.py
karankhanchandani: @ianbuss Operator is older version of cert manager. Should I update 1.19.4 in operator as well?

bin/setup-037x-k3d.py
karankhanchandani: Fixed.

ianbuss: When I wrote this comment originally, I didn't realise `--agents` was what `k3d` itself uses. So no worries if you didn't change it.

PR #3273, opened 2026-06-01
pinf-724 update-container-ca-script-to-preserve-mirror-registry

sudarshanrampuria:
### Issue

GKE's default `/etc/containerd/config.toml` ships an inline mirror for Docker Hub:
```
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
  endpoint = ["https://mirror.gcr.io/","https://registry-1.docker.io/"]
```

`mirror.gcr.io` is GKE's pull-through cache — Docker Hub images are served from Google's regional cache first, with `registry-1.docker.io` as the fallback.
Our private-CA script needs to enable containerd's `config_path` (the hosts.d / containerd v2 mechanism) so it can write `certs.d/<our-registry-url>/hosts.toml` with the CA trust:
```
[plugins."io.containerd.cri.v1.images".registry]
  config_path = "/etc/containerd/certs.d"
```

`mirrors.*` and `config_path` cannot coexist — this is a containerd 2.x constraint. When both are set under the registry plugin, containerd 2.x silently disables `config_path` and falls back to the inline schema. So we cannot leave the `docker.io` block in place and still get private-CA trust on a different registry. Previously the script removed the `registry.mirrors` block entirely, which restored private-CA trust but disabled the GKE cache.

Side note: GKE's defaults still use the legacy containerd 1.x inline schema even on 2.x runtimes. If GKE shipped `mirror.gcr.io` as a hosts.d file out of the box, there would be no conflict to resolve.

More info on this is in slack thread: https://astronomer.slack.com/archives/C08QKH6LX3R/p1776966109295129

### Solution

Instead of just stripping the `mirrors.*` block, the script now reads it first and translates each entry into a per-registry `hosts.toml` under `/etc/containerd/certs.d/<host>/`. For GKE's Docker Hub mirror, that produces `/etc/containerd/certs.d/docker.io/hosts.toml`:

```
server = "https://docker.io"

[host."https://mirror.gcr.io"]
  capabilities = ["pull", "resolve"]
[host."https://registry-1.docker.io"]
  capabilities = ["pull", "resolve"]
```
Same endpoints in the same order, expressed in the schema containerd 2.x reads via config_path. The translation is cloud-agnostic — no registry URLs are hard-coded, so any inline mirror the cloud provider ships (GKE, EKS, AKS, on-prem) gets carried over verbatim.

## Related Issues

https://linear.app/astronomer/issue/PINF-724/update-script-for-walmart-containerd-ca-patch

## Testing

Tested end-to-end on a fresh GKE 1.33 / containerd 2.0.7 cluster.
After the new script ran:
```
/etc/containerd/certs.d/
├── docker.io/
 │   └── hosts.toml          ← NEW: GKE mirror behavior preserved
└── registry.sud.astrodev.io/
    ├── hosts.toml          ← private-registry trust (unchanged)
    └── private-ca-root-cert.pem
```
certs.d/docker.io/hosts.toml contents matched the translated GKE mirror block (mirror.gcr.io listed first, registry-1.docker.io second).

Verified the cache is actually being used: ran crictl pull docker.io/library/alpine:3.20 on the node and sampled containerd's open TCP connections during the pull. Both established connections went to 173.194.45.82:443 (Google's 173.194.0.0/16 netblock, the same range mirror.gcr.io resolves into). Docker Hub's registry-1.docker.io resolves to AWS IPs (2600:1f18:...), which were never contacted. Confirms the pull went through the GKE cache.

## Merging

release-2.0

### DIFF
--- a/files/update-containerd-certs.py
+++ b/files/update-containerd-certs.py
@@ -276,6 +276,102 @@ _LEGACY_MIRRORS_HEADER_RE = re.compile(
 )
 
 
+def _extract_legacy_mirrors(config: dict) -> dict[str, list[str]]:
+    """Pull inline `registry.mirrors.<host>` blocks out of a parsed config.toml.
+
+    Walks both the 1.x (`io.containerd.grpc.v1.cri`) and 2.x
+    (`io.containerd.cri.v1.images`) plugin namespaces — current cloud-provider
+    images (GKE 1.33 et al) still declare their default mirrors under the 1.x
+    namespace even on a containerd 2.x runtime, but a future image could move
+    them.
+
+    Returns {registry_host: [endpoint_url, ...]}. Endpoint order is preserved —
+    containerd tries them in the order listed, so the first endpoint is the
+    primary (e.g. a pull-through cache) and the rest are fallbacks.
+    """
+    out: dict[str, list[str]] = {}
+    for version in (1, 2):
+        subtree = _registry_subtree(config, version)
+        if not subtree:
+            continue
+        mirrors = subtree.get("mirrors")
+        if not isinstance(mirrors, dict):
+            continue
+        for host, block in mirrors.items():
+            if not isinstance(block, dict):
+                continue
+            endpoints = block.get("endpoint")
+            if not isinstance(endpoints, list) or not endpoints:
+                continue
+            # Last namespace wins if both 1.x and 2.x have an entry for the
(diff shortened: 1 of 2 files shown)


pgvishnuram:
is this tested with upgrade path and also with other cloud providers

pgvishnuram (approved):

PR #3274, opened 2026-06-02
PINF-688 - Allow extraAnnotation metadata support commander

pgvishnuram:
## Description

Surfaces `global.extraAnnotations` in the commander metadata ConfigMap so commander can read global annotation defaults at runtime.

```mermaid
sequenceDiagram
    actor User
    participant Helm as Helm
    participant K8s as Kubernetes API
    participant CM as ConfigMap<br/>(metadata.yaml)
    participant Deploy as Commander<br/>Deployment / Pod
    participant LC as LoadConfig (http.go)
    participant EP as /metadata endpoint

    Note over User,EP: 1. Configure & upgrade
    User->>Helm: update global.extraAnnotations in values.yaml
    User->>Helm: helm upgrade
    Helm->>Helm: render metadata.yaml (extraAnnotations: {...})
    Helm->>K8s: apply manifests
    K8s->>CM: update ConfigMap (metadata.yaml)

    Note over K8s,Deploy: 2. Roll out
    K8s->>Deploy: restart Commander deployment
    CM-->>Deploy: metadata.yaml mounted into new pod
    Deploy->>Deploy: pod becomes Ready

    Note over Helm,User: helm upgrade completes ✓

    Note over User,EP: 3. Read updated values
    User->>EP: GET http://localhost:8880/metadata
    EP->>LC: LoadConfig(metadata.yaml)
    LC->>Deploy: os.ReadFile(metadata.yaml)
    Deploy-->>LC: raw YAML
    LC->>LC: yaml.Unmarshal → MetadataConfig
    LC->>LC: env.Parse overrides + default nil maps → {}
    LC-->>EP: *MetadataConfig
    EP-->>User: 200 JSON { extraAnnotations: {...} }
```
### Changes

| File | What changed |
|------|-------------|
| `charts/astronomer/templates/_helpers.yaml` | New `commander.metadata.extraAnnotations` helper that emits `global.extraAnnotations` as YAML, falling back to `{}` when empty |
| `charts/astronomer/templates/commander/commander-metadata.yaml` | Adds `extraAnnotations:` to the ConfigMap, rendered via the new helper |
| `tests/chart_tests/test_astronomer_commander.py` | Updates existing tests to expect the new `extraAnnotations` field; adds `test_commander_metadata_extra_annotations` covering ingress-class and OpenShift route annotations |

### Improvements

- Commander can now read `global.extraAnnotations` from its metadata file and propagate them to managed resources (e.g. ingress class, OpenShift route termination)
- Empty annotation maps render as `{}` rather than producing malformed YAML
- Behaviour is unchanged when `global.extraAnnotations` is unset — new key just renders as `{}`

## Related Issues

https://linear.app/astronomer/issue/PINF-688/auto-sync-extraannotations-to-airflow-deployments-via-dataplane

## Testing

1. `pytest tests/chart_tests/test_astronomer_commander.py::test_commander_metadata_extra_annotations` passes
2. Existing `test_commander_metadata_yaml` and `test_commander_metadata_custom_logging` pass with the updated `extraAnnotations: {}` expectation
3. Render the chart with `global.extraAnnotations` set to a sample map — confirm the value appears verbatim in the commander metadata ConfigMap
4. Render with `global.extraAnnotations` unset — confirm `extraAnnotations: {}` in the ConfigMap

## Merging

Merge to master.

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -24,6 +24,14 @@
 {{- end }}
 {{- end }}
 
+{{ define "commander.metadata.extraAnnotations" }}
+{{- if not (empty .Values.global.extraAnnotations) -}}
+{{ .Values.global.extraAnnotations | toYaml | nindent 6 -}}
+{{- else -}}
+{}
+{{- end -}}
+{{- end }}
+
 {{ define "commander.image" -}}
 {{- if .Values.global.privateRegistry.enabled -}}
 {{ .Values.global.privateRegistry.repository }}/ap-commander:{{ .Values.images.commander.tag }}
--- a/tests/chart_tests/test_astronomer_commander.py
+++ b/tests/chart_tests/test_astronomer_commander.py
@@ -50,12 +50,14 @@ class TestAstronomerCommander:
         if namespace_labels and rbac_enabled:
             assert metadata_file_contents == {
                 "namespaceLabels": namespace_labels,
+                "extraAnnotations": {},
                 "registry": {"version": "99.88.77"},
                 "customLogging": {"enabled": False},
             }
         else:
             assert metadata_file_contents == {
                 "namespaceLabels": {},
+                "extraAnnotations": {},
                 "registry": {"version": "99.88.77"},
                 "customLogging": {"enabled": False},
             }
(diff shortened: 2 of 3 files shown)


Shubham-astro (approved):
LGTM.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -24,6 +24,14 @@
 {{- end }}
 {{- end }}
 
+{{ define "commander.metadata.extraAnnotations" }}
+{{- if not (empty .Values.global.extraAnnotations) -}}
+{{ .Values.global.extraAnnotations | toYaml | nindent 6 -}}
```
rishkarajgi, line 29: Hardcoded nindent 6 is fragile. The indent depth is tied to this helper's call site inside data.metadata.yaml: |. If commander-metadata.yaml ever restructures (e.g., key moves up/down a level), this silently lands values at the wrong indent — the YAML still parses, just incorrectly. Please add a comment here noting the expected call-site indent, or compute it from context.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -24,6 +24,14 @@
...
+{{- else -}}
```
rishkarajgi, line 30: Inconsistent rendering style between branches. The if branch emits a block-style YAML on a new line (via nindent 6); the else branch emits inline {}. Both parse, but mixed shapes can surprise downstream consumers and produce noisy diffs. Consider rendering both branches consistently — either always block or always inline.

tests/chart_tests/test_astronomer_commander.py
```diff
@@ -85,10 +87,35 @@ def test_commander_metadata_custom_logging(self, kube_version, enabled):
...
+        }
+        docs = render_chart(
+            kube_version=kube_version,
+            values=values,
+            show_only=["charts/astronomer/templates/commander/commander-metadata.yaml"],
+        )
+
+        assert len(docs) == 1
+        metadata_file_contents = yaml.safe_load(docs[0]["data"]["metadata.yaml"])
+        assert metadata_file_contents["extraAnnotations"] == extra_annotations
```
rishkarajgi, line 117: Test only asserts the new field. This assertion verifies extraAnnotations round-trips, but doesn't check that namespaceLabels, registry, and customLogging remain intact when annotations are set. Cheap to extend — assert the full metadata_file_contents dict so we catch regressions if the new helper ever interferes with sibling keys.

charts/astronomer/templates/_helpers.yaml
pgvishnuram: I can move them extraAnnotations: {{ include "commander.metadata.extraAnnotations" . }} in here

rishkarajgi (approved):
Approved with a few comments

Also 
Missing default in values.yaml. The helper relies on Helm returning nil for missing .Values.global.extraAnnotations plus empty nil == true. It works, but adding global.extraAnnotations: {} to the default values.yaml documents the contract, prevents typos from silently no-op'ing, and makes the new key discoverable. Could you add it (and a one-line comment describing how it flows into commander)?

PR #3278, opened 2026-06-03
PINF-687 - allow extra ingress overrides precendence ordering

pgvishnuram:
## Description

Reorders `global.extraAnnotations` so they render **after** each chart's built-in nginx ingress annotations. Because Kubernetes keeps the last value when an annotation key is repeated, user-supplied `extraAnnotations` now take precedence over the chart defaults instead of being silently overridden by them.

### Improvements

- Values set in `global.extraAnnotations` now override the chart's default ingress annotations (e.g. `kubernetes.io/ingress.class`, `nginx.ingress.kubernetes.io/*`) instead of being overridden by them.
- Consistent annotation ordering across all 12 ingress templates, so override behavior is predictable regardless of which component's ingress is being configured.
- No behavioral change when `extraAnnotations` does not collide with a default key — only the precedence on duplicate keys changes.

## Related Issues

https://linear.app/astronomer/issue/PINF-687/allow-extraannotations-to-take-precedence-over-injected-ingress

## Testing

Render the affected charts with an `extraAnnotations` value that collides with a default and confirm the override wins in the output:

```bash
helm template . 
  --set global.extraAnnotations.'kubernetes\.io/ingress\.class'=custom-class \\
  --show-only charts/astronomer/templates/ingress.yaml
```

Expected: the rendered ingress `metadata.annotations` shows `kubernetes.io/ingress.class: custom-class` rather than the default `{{ .Release.Name }}-nginx`. Repeat for any of the other ingress templates and for the `global.authSidecar.enabled` path (unchanged) to confirm no regression.

## Merging

Merge to \`master\`. Cherry-pick to active release branches if the override-precedence fix is needed in supported releases.

### DIFF
--- a/charts/alertmanager/templates/ingress.yaml
+++ b/charts/alertmanager/templates/ingress.yaml
@@ -18,13 +18,13 @@ metadata:
     {{- if .Values.global.authSidecar.enabled  }}
 {{- toYaml .Values.global.extraAnnotations | nindent 4 }}
     {{- else }}
-    {{- if .Values.global.extraAnnotations }}
-{{- toYaml .Values.global.extraAnnotations | nindent 4 }}
-    {{- end }}
     kubernetes.io/ingress.class: "{{ .Release.Name }}-nginx"
     {{ include "houston.internalauthurl" . | indent 4 }}
     nginx.ingress.kubernetes.io/auth-signin: https://app.{{ .Values.global.baseDomain }}/login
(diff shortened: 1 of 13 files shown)


ianbuss (approved):
Approved. But are `global.extraAnnotations` just for ingresses? It would be a bit weird for users to apply an ingress.class to other resources.

PR #3280, opened 2026-06-04
pinf-689 Airflow operator 1.6rc2

sudarshanrampuria:
## Description

* Syncs upstream airflow-operator chart
* Gated airflow-operator chart to only be deployed when `.Values.global.airflowOperator.enabled` and plane_mode is set to `unified` or `data`

crd folder is just exactly same/synced from upstream so we can skip it? 

## Related Issues

https://linear.app/astronomer/issue/PINF-689/re-sync-and-update-airflow-operator-subchart-from-upstream-to-version

## Testing

* Updated local setup script and tested by running locally and creating AF2 deployment.
* AF3 is not working. We would need to make some changes in operator/houston

## Merging

2.1


charts/airflow-operator/values.yaml
```diff
@@ -1,7 +1,8 @@
 images:
   manager:
-    repository: quay.io/astronomer/airflow-operator-controller
-    tag: 1.5.2
+    # TODO: Change it to public image once 1.6.0 is released
+    repository: astrocr.azurecr.io/astronomer/airflow-operator-dev
```
pgvishnuram: we have to use quay before releasing else customer has to mirror 2 different image registries

sudarshanrampuria: Updated.

karankhanchandani (approved):
Changes looks good.

PR #3288, opened 2026-06-10
PINF-765 Fix security context diverged config

pgvishnuram:
## Description

Fixes init/sidecar containers that hardcoded `securityContext.readOnlyRootFilesystem: true` and ignored user-configured `securityContext` values. They now merge the configured security context while still enforcing `readOnlyRootFilesystem`, bringing init containers in line with the main containers.

### Changes

| File | What changed |
|------|-------------|
| `charts/elasticsearch/templates/_helpers.tpl` | Hoist `$required` dict so it is defined on both branches |
| `charts/elasticsearch/templates/{client,data,master}/*.yaml` | Init container security context now merges the per-role `securityContext` (falls back to the shared helper) |
| `charts/prometheus/templates/_helpers.yaml` | Add `filesdreloader.securityContext` helper; honor configured `securityContext` in the else branch |
| `charts/prometheus/templates/prometheus-statefulset.yaml` | filesd-reloader sidecar merges configured security context with enforced `readOnlyRootFilesystem` |
| `charts/prometheus/values.yaml` | Add `filesdReloader.securityContext` default (`{}`) |

## Related Issues

PINF-765 - Container security context not respecting passed values
https://linear.app/astronomer/issue/PINF-765/container-security-context-not-respecting-passed-values

## Testing

`helm template` the elasticsearch and prometheus charts with and without a custom `securityContext` and confirm init/sidecar containers render the configured fields while `readOnlyRootFilesystem: true` is always present.

## Merging

Merge to \`master\`; cherry-pick to active release branches as needed.

### DIFF
--- a/charts/elasticsearch/templates/client/es-client-deployment.yaml
+++ b/charts/elasticsearch/templates/client/es-client-deployment.yaml
@@ -86,7 +86,12 @@ spec:
         imagePullPolicy: {{ .Values.images.init.pullPolicy }}
         resources: {{- toYaml .Values.client.initResources | nindent 10 }}
         securityContext:
-          readOnlyRootFilesystem: true
+        {{- if .Values.client.securityContext }}
+          {{- $required := dict "readOnlyRootFilesystem" true }}
+          {{- merge $required .Values.client.securityContext | toYaml | nindent 10}}
+        {{- else }}
+          {{- include "elasticsearch.securityContext" . | nindent 10 }}
+        {{- end }}
         volumeMounts:
           - name: es-config-dir
             mountPath: /usr/share/elasticsearch/config_copy
--- a/charts/elasticsearch/templates/data/es-data-statefulset.yaml
+++ b/charts/elasticsearch/templates/data/es-data-statefulset.yaml
@@ -90,7 +90,12 @@ spec:
         imagePullPolicy: {{ .Values.images.init.pullPolicy }}
         resources: {{- toYaml .Values.data.initResources | nindent 10 }}
         securityContext:
-          readOnlyRootFilesystem: true
+          {{- if .Values.data.securityContext }}
+          {{- $required := dict "readOnlyRootFilesystem" true }}
+          {{- merge $required .Values.data.securityContext | toYaml | nindent 10 }}
+          {{- else }}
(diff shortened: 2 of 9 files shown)


charts/elasticsearch/templates/client/es-client-deployment.yaml
```diff
@@ -86,7 +86,12 @@ spec:
         imagePullPolicy: {{ .Values.images.init.pullPolicy }}
         resources: {{- toYaml .Values.client.initResources | nindent 10 }}
         securityContext:
-          readOnlyRootFilesystem: true
+        {{- if .Values.client.securityContext }}
```
Shubham-astro, line 89: is there a condition where we would need separate securitycontext for es client, can we not use es securitycontext only @pgvishnuram .??

charts/elasticsearch/templates/data/es-data-statefulset.yaml
```diff
@@ -90,7 +90,12 @@ spec:
         imagePullPolicy: {{ .Values.images.init.pullPolicy }}
         resources: {{- toYaml .Values.data.initResources | nindent 10 }}
         securityContext:
-          readOnlyRootFilesystem: true
+          {{- if .Values.data.securityContext }}
```
Shubham-astro, line 93: same question here, is there a condition where we would need separate securitycontext for data, can we not use es securitycontext only ??

Shubham-astro:
Can we add tests as well for this.? if we don't have it already @pgvishnuram .?

charts/elasticsearch/templates/client/es-client-deployment.yaml
pgvishnuram: it supports both ways by default all elasticsearch security contexts are applied to all elasticsearch components - if they want to overrdide specific component they can use this option

charts/elasticsearch/templates/data/es-data-statefulset.yaml
pgvishnuram: it supports both ways by default all elasticsearch security contexts are applied to all elasticsearch components - if they want to overrdide specific component they can use this option

Shubham-astro (approved):
LGTM..

PR #3290, opened 2026-06-10
add support for Istio ingress

htech7x:
## Description
Adding support for Istio ingress

## Related Issues
https://linear.app/astronomer/issue/APC-1423/add-optional-istio-ingress-support-to-apc-helm-chart

## Testing
Testing was conducted for APC 1.1.5 in unified mode.


values.yaml
```diff
@@ -86,10 +86,28 @@ global:
...
-  # Enables necessary components for compatibility with Istio Service Mesh
+  # Enable/disable nginx ingress controller
+  nginxEnabled: true
+
+  # Enable Istio VirtualService for ingress (requires Istio Gateway to be pre-configured)
+  # When enabled, creates VirtualService resources to route traffic through existing Istio Gateway
+  istioEnabled: false
+
+  # Istio configuration
   istio:
```
Simpcyclassy, line 97: This removes global.istio.enabled and global.istio.rootNamespace, which were previously available config keys, with no backward-compatible alias or deprecation path; any customer already setting these loses Istio sidecar behavior silently on upgrade.

values.yaml
```diff
@@ -86,10 +86,28 @@ global:
   # Use cluster roles
   clusterRoles: true
 
-  # Enables necessary components for compatibility with Istio Service Mesh
+  # Enable/disable nginx ingress controller
+  nginxEnabled: true
```
Simpcyclassy, line 90: nginxEnabled is added here and referenced in the new configs/istio-gateway.yaml example, but no ingress template in this diff actually checks it, so setting it to false has no effect and the nginx controller keeps running.

charts/astronomer/templates/istio/sidecars.yaml
```diff
@@ -1,57 +1,47 @@
...
-    - "istio-system/*"
-    # allow egress to pods in the nginx-elasticsearch service
-    - "{{ .Release.Namespace }}/{{ .Release.Name }}-elasticsearch-nginx.{{ .Release.Namespace }}.svc.cluster.local"
-    # allow egress to sql proxy
-    - "{{ .Release.Namespace }}/pg-sqlproxy-gcloud-sqlproxy.{{ .Release.Namespace }}.svc.cluster.local"
----
-# This is used to overwrite the default configuration
-# for the platform namespace. This is the defaults
-# for sidecars in the platform namespace.
+{{- if .Values.global.istioEnabled }}
```
Simpcyclassy, line 1: This switches the gate from the old global.istio.enabled to global.istioEnabled and also drops the previous root-namespace default-sidecar-config Sidecar resource entirely, both are undocumented breaking changes for anyone with the old key already set.

charts/astronomer/templates/istio/astro-ui-virtualservice.yaml
```diff
@@ -0,0 +1,50 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    component: istio-virtualservice
+    tier: astronomer
+    release: {{ .Release.Name }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    heritage: {{ .Release.Service }}
+spec:
+  hosts:
+  - "{{ .Values.global.baseDomain }}"
```
Simpcyclassy, line 17: Unlike the equivalent Ingress template, this VirtualService is not gated on global.baseDomain being set, so enabling istioEnabled without a baseDomain renders hosts with an empty string.

charts/astronomer/templates/houston/ingress.yaml
```diff
@@ -1,7 +1,7 @@
 #####################
 ## Houston Ingress ##
 #####################
-{{- if .Values.global.baseDomain }}
+{{- if and .Values.global.baseDomain (not .Values.global.istioEnabled) }}
```
Simpcyclassy, line 4: Good, this istioEnabled suppression is applied consistently across every existing ingress template in the chart, nice thoroughness.

tests/istio-enabled.yaml
```diff
@@ -0,0 +1,39 @@
+# Test values for Istio Gateway configuration
```
Simpcyclassy, line 1: This is a values fixture only, there is no tests/chart_tests/*.py assertion anywhere in this PR that actually renders and checks the new VirtualService templates or the ingress-suppression logic, please add pytest coverage parametrized like the rest of tests/chart_tests.

Simpcyclassy (changes requested):
Three problems need fixing before this merges. First, upgrade safety: values.yaml silently renames global.istio.enabled and global.istio.rootNamespace to global.istioEnabled and a restructured global.istio block with no backward-compatible alias, so any customer already setting the old keys loses Istio sidecar config on upgrade with no warning in the PR description, and sidecars.yaml also quietly drops the old root-namespace-scoped default-sidecar-config resource entirely. Second, global.nginxEnabled is introduced in values.yaml and used in the new configs/istio-gateway.yaml example config as if it disables the nginx ingress controller, but no template in this diff (or the ingress files it touches) actually reads nginxEnabled, so following the shipped example leaves nginx running alongside Istio, wasted resources and a misleading config surface. Third, there are no tests/chart_tests/*.py additions for any of the ten new VirtualService templates or the istioEnabled gating added to nine existing ingress templates; tests/istio-enabled.yaml is only a values fixture, not an assertion, so a broken VirtualService render would not be caught by CI. The ingress-suppression gating itself (not .Values.global.istioEnabled) is applied consistently and correctly across every existing ingress template, that part is solid. Please add backward compatibility or an explicit migration note for the renamed istio keys, wire or remove nginxEnabled, and add render-and-assert pytest coverage before this merges.

rishkarajgi (commented):
Independently checked all three things Simpcyclassy's review already raised, since they're still live in the current diff -- all three confirmed real:

- `global.istio.enabled`/`global.istio.rootNamespace` are gone with no fallback anywhere in this diff (grepped for an `or`-style alias, there isn't one) -- any install with those set today loses Istio sidecar config silently on upgrade.
- `nginxEnabled` is set in three places (the default, and both new fixtures) but grepped the whole chart and it's read nowhere -- setting it to false does not stop nginx from running.
- Still no chart-test coverage for any of the ten new VirtualService templates or the istioEnabled gating on the nine existing ingress templates.

One more thing worth flagging given this is explicitly the companion to airflow-chart#735 (same Linear ticket, APC-1423): that PR's toggle is `airflow.istio.enabled` (nested, per-airflow-deployment), this one is `global.istioEnabled` (flat, platform-level). Worth confirming that split is intentional -- different scopes -- rather than the two PRs just not having coordinated on naming, since a customer configuring "turn on Istio" across the platform will need to set two differently-shaped flags in two different charts.

This is a mesh/auth-adjacent networking change (mTLS routing, gateway config) on top of the values-rename risk above, so it'll want a security review in addition to the CODEOWNERS approval.

PR #3302, opened 2026-06-15
PINF-768 Astronomer chart values for Control Plane HA

adhipjos:
## Description

- Adds Helm chart support for **Control Plane HA** (multiple active CPs serving traffic concurrently within a region), gated behind a new default-off `global.controlPlaneHA.enabled` flag. 
- Single-CP installs render identically (no behavior change). 
- Companion houston-api code changes (`ui()`/`houston()` URL helpers, `getCookieName()`, `setJWTCookie`, CORS) and ingress/TLS host templating are tracked separately.

When `global.controlPlaneHA.enabled: true`:

- **Skips chart generation of the Houston JWT signing keypair.** New `jwks.generation.enabled` value (default `true`) gates `houston-jwt-certificate-secret.yaml` which is auto-skipped in HA mode so every CP consumes the same externally-provisioned / ESO-synced signing material. 
- **Generates a stable `cp-identity` Secret** using the lookup-or-generate pattern (cf. `astronomer.clusterLocalServiceToken`), so the UUID is created once and preserved across `helm upgrade`.
- **Mounts `CP_ID`** (from `cp-identity`) on every CP pod that runs a state-mutating / reconciliation loop (Houston API + workers, dp-link, and navigator) via a shared `houston.cpIdEnv` helper so it can be used to gate state-mutating work to the active region.
- **Templates the global domain / cookie values** into the houston config: `helm.globalBaseDomain` (required when HA enabled), plus optional `helm.cookieDomain` / `helm.cookieName` pass-through (emitted only when explicitly overridden. Houston computes the `.<domain>` / snakeCase defaults).

### New values

| Value | Default | Notes |
|---|---|---|
| `global.controlPlaneHA.enabled` | `false` | Master flag |
| `global.controlPlaneHA.globalBaseDomain` | `~` | Required when enabled |
| `global.controlPlaneHA.cookieDomain` | `~` | Optional override (advanced single-cookie mode) |
| `global.controlPlaneHA.cookieName` | `~` | Optional override (migration back-compat) |
| `jwks.generation.enabled` (astronomer subchart) | `true` | Auto-skipped in HA mode |

## Related Issues

[PINF-768](https://linear.app/astronomer/issue/PINF-768)

## Testing

- New `tests/chart_tests/test_control_plane_ha.py` (85 cases): cp-identity present/absent + required-`globalBaseDomain` guard; JWT keypair gating (HA + `jwks.generation.enabled`). `CP_ID` present in HA / absent otherwise across Houston API, worker, dp-link, navigator; configmap `globalBaseDomain` / cookie pass-through.
- Verified via `helm template` (HA on/off) that the HA-off render is unchanged, the HA-on render is correct, and output is whitespace-clean at all injection sites.
- Full chart unit suite green: **4597 passed, 72 skipped**.

## Merging

This PR targets the **`feature/cp-ha-dr`** feature branch. It is part of the Control Plane HA/DR workstream and will reach `master` / the **2.1.0** release when the feature branch is integrated.

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -388,6 +388,21 @@ s3:
   regionendpoint: {{ .Values.registry.s3.regionendpoint }}
 {{- end }}
 
+{{- /*
+  CP_ID env var, sourced from the cp-identity Secret. Single source of truth for the
+  secret ref, shared by houston, dp-link and navigator. Guard call sites with
+  `if .Values.global.controlPlaneHA.enabled` — in HA mode it is mounted on every
+  control-plane pod that runs a state-mutating/reconciliation loop so it can resolve
+  its region's active state (Region-gating).
+*/}}
+{{- define "houston.cpIdEnv" -}}
+- name: CP_ID
+  valueFrom:
+    secretKeyRef:
+      name: cp-identity
+      key: cp_id
+{{- end -}}
+
 {{- define "dp_link_environment" }}
 - name: NODE_ENV
   value: "production"
@@ -448,6 +463,9 @@ s3:
   value: {{ template "houston.nats.servers" . }}
 - name: NATS__CLUSTER_ID
   value: {{ .Release.Name }}-stan
+{{- if .Values.global.controlPlaneHA.enabled }}
+{{ include "houston.cpIdEnv" . }}
+{{- end }}
 {{- end }}
 
 
@@ -467,6 +485,9 @@ s3:
 {{- end }}
 - name: NODE_ENV
   value: "production"
+{{- if .Values.global.controlPlaneHA.enabled }}
+{{ include "houston.cpIdEnv" . }}
+{{- end }}
 - name: DATABASE__CONNECTION
   valueFrom:
     secretKeyRef:
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -464,7 +464,7 @@ s3:
 - name: NATS__CLUSTER_ID
   value: {{ .Release.Name }}-stan
 {{- if .Values.global.controlPlaneHA.enabled }}
-{{ include "houston.cpIdEnv" . }}
+{{- include "houston.cpIdEnv" . | nindent 0 }}
 {{- end }}
 {{- end }}
 
@@ -486,7 +486,7 @@ s3:
 - name: NODE_ENV
   value: "production"
 {{- if .Values.global.controlPlaneHA.enabled }}
-{{ include "houston.cpIdEnv" . }}
+{{- include "houston.cpIdEnv" . | nindent 0 }}
 {{- end }}
 - name: DATABASE__CONNECTION
   valueFrom:
--- a/charts/astronomer/templates/houston/api/cp-identity-secret.yaml
+++ b/charts/astronomer/templates/houston/api/cp-identity-secret.yaml
@@ -7,17 +7,23 @@
 {{- end }}
 {{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
 {{- /*
-  Stable per-CP UUID, generated once at install and preserved across `helm upgrade`
-  via the lookup-or-generate pattern (cf. astronomer.clusterLocalServiceToken).
-  Mounted as the CP_ID env var on every CP pod that runs a reconciliation loop so
-  they can resolve their region's active state (Region-gating).
+  Stable per-CP UUID, mounted as the CP_ID env var on every CP pod that runs a
+  reconciliation loop so they can resolve their region's active state (Region-gating).
+  Precedence (cf. astronomer.clusterLocalServiceToken):
+    1. explicit global.controlPlaneHA.cpId  - deterministic for GitOps / offline render
+    2. existing cp-identity Secret           - preserved across `helm upgrade`
+    3. freshly generated UUID                - first install
 */}}
-{{- $existing := lookup "v1" "Secret" .Release.Namespace "cp-identity" }}
 {{- $cpId := "" }}
+{{- if .Values.global.controlPlaneHA.cpId }}
+{{- $cpId = .Values.global.controlPlaneHA.cpId | toString }}
+{{- else }}
+{{- $existing := lookup "v1" "Secret" .Release.Namespace "cp-identity" }}
 {{- if and $existing $existing.data $existing.data.cp_id }}
 {{- $cpId = index $existing.data "cp_id" | b64dec }}
 {{- else }}
 {{- $cpId = uuidv4 }}
+{{- end }}
 {{- end }}
 apiVersion: v1
 kind: Secret
--- a/charts/astronomer/templates/houston/api/cp-identity-secret.yaml
+++ b/charts/astronomer/templates/houston/api/cp-identity-secret.yaml
@@ -1,0 +1,41 @@
+######################################################
+## Control Plane identity Secret (Control Plane HA)  ##
+######################################################
+{{- if .Values.global.controlPlaneHA.enabled }}
+{{- if not .Values.global.controlPlaneHA.globalBaseDomain }}
+{{- fail "global.controlPlaneHA.globalBaseDomain is required when global.controlPlaneHA.enabled is true" }}
+{{- end }}
+{{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
+{{- /*
+  Stable per-CP UUID, generated once at install and preserved across `helm upgrade`
+  via the lookup-or-generate pattern (cf. astronomer.clusterLocalServiceToken).
+  Mounted as the CP_ID env var on every CP pod that runs a reconciliation loop so
+  they can resolve their region's active state (Region-gating).
+*/}}
+{{- $existing := lookup "v1" "Secret" .Release.Namespace "cp-identity" }}
+{{- $cpId := "" }}
+{{- if and $existing $existing.data $existing.data.cp_id }}
+{{- $cpId = index $existing.data "cp_id" | b64dec }}
+{{- else }}
+{{- $cpId = uuidv4 }}
+{{- end }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cp-identity
+  labels:
+    component: cp-identity
+    tier: astronomer
+    release: {{ .Release.Name }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    heritage: {{ .Release.Service }}
+    plane: {{ .Values.global.plane.mode }}
+  {{ if .Values.global.argoCD.annotation.enabled }}
+  annotations:
+    argocd.argoproj.io/sync-wave: "-1"
+  {{- end }}
+type: Opaque
+data:
+  cp_id: {{ $cpId | b64enc | quote }}
+{{- end }}
+{{- end }}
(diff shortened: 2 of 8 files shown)


values.yaml
```diff
@@ -20,6 +20,24 @@ global:
   clusterLocalServiceAuth:
     token: ""
 
+  # Control Plane High Availability: run multiple active CPs serving traffic concurrently within a region.
+  # Default false keeps single-CP installs unchanged (no behavior change).
+  controlPlaneHA:
+    enabled: false
+    # Parent domain of the global customer-facing hostname (e.g. astro.<region>.<...>.link
+    # when the global URL is app.astro.<region>...). REQUIRED when enabled.
```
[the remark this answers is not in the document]
adhipjos: Fixed

charts/astronomer/templates/houston/api/cp-identity-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+{{- if and $existing $existing.data $existing.data.cp_id }}
+{{- $cpId = index $existing.data "cp_id" | b64dec }}
+{{- else }}
+{{- $cpId = uuidv4 }}
+{{- end }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cp-identity
+  labels:
```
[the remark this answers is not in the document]
adhipjos: This is by design. We do not want to generate a new id for a new release but reuse the same

charts/astronomer/templates/houston/api/cp-identity-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+  Mounted as the CP_ID env var on every CP pod that runs a reconciliation loop so
+  they can resolve their region's active state (Region-gating).
+*/}}
+{{- $existing := lookup "v1" "Secret" .Release.Namespace "cp-identity" }}
+{{- $cpId := "" }}
+{{- if and $existing $existing.data $existing.data.cp_id }}
+{{- $cpId = index $existing.data "cp_id" | b64dec }}
+{{- else }}
+{{- $cpId = uuidv4 }}
+{{- end }}
```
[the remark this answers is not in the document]
adhipjos: Valid for offline workflows. Added a override cp_id value to the chart that can be specified when a lookup is not possible

charts/astronomer/templates/houston/api/cp-identity-secret.yaml
```diff
@@ -0,0 +1,47 @@
+######################################################
+## Control Plane identity Secret (Control Plane HA)  ##
+######################################################
+{{- if .Values.global.controlPlaneHA.enabled }}
+{{- if not .Values.global.controlPlaneHA.globalBaseDomain }}
+{{- fail "global.controlPlaneHA.globalBaseDomain is required when global.controlPlaneHA.enabled is true" }}
+{{- end }}
+{{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
```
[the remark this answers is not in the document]
adhipjos: This makes sense. No need to render this in data mode so the mode should be checked before checking existence of globalbasedomain

charts/astronomer/templates/_helpers.yaml
```diff
@@ -448,6 +463,9 @@ s3:
   value: {{ template "houston.nats.servers" . }}
 - name: NATS__CLUSTER_ID
   value: {{ .Release.Name }}-stan
+{{- if .Values.global.controlPlaneHA.enabled }}
+{{ include "houston.cpIdEnv" . }}
+{{- end }}
 {{- end }}
```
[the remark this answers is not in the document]
adhipjos: updated to trim-safe form

charts/astronomer/templates/_helpers.yaml
```diff
@@ -467,6 +485,9 @@ s3:
...
 - name: NODE_ENV
   value: "production"
+{{- if .Values.global.controlPlaneHA.enabled }}
+{{ include "houston.cpIdEnv" . }}
+{{- end }}
```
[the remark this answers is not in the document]
adhipjos: updated to trim-safe form

charts/astronomer/templates/houston/api/houston-jwt-certificate-secret.yaml
```diff
@@ -1,7 +1,7 @@
 ############################
 ## Houston JWT Key Secret ##
 ############################
-{{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
+{{- if and (or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified")) .Values.jwks.generation.enabled (not .Values.global.controlPlaneHA.enabled) }}
```
pgvishnuram: jwks.generation.enabled -> can be under houston.* 
since all of the jwks references are under houston values spec 
ex: houston.regenerateCaEachUpgrade

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -67,6 +67,15 @@ data:
...
       baseDomain: {{ .Values.global.baseDomain }}
+      {{- if .Values.global.controlPlaneHA.enabled }}
+      globalBaseDomain: {{ .Values.global.controlPlaneHA.globalBaseDomain }}
+      {{- with .Values.global.controlPlaneHA.cookieDomain }}
+      cookieDomain: {{ . }}
+      {{- end }}
+      {{- with .Values.global.controlPlaneHA.cookieName }}
+      cookieName: {{ . }}
+      {{- end }}
+      {{- end }}
```
pgvishnuram, line 78: this can be loaded from heplers.tpl and sourced here

charts/astronomer/templates/_helpers.yaml
```diff
@@ -448,6 +463,9 @@ s3:
   value: {{ template "houston.nats.servers" . }}
 - name: NATS__CLUSTER_ID
   value: {{ .Release.Name }}-stan
+{{- if .Values.global.controlPlaneHA.enabled }}
+{{- include "houston.cpIdEnv" . | nindent 0 }}
+{{- end }}
```
pgvishnuram, line 468: just a thought we have 2 references  - one common helper can be added pulled as include right

charts/astronomer/values.yaml
```diff
@@ -46,6 +46,13 @@ securityContext:
 # ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
 podSecurityContext: {}
 
+# Houston JWT (JWKS) signing keypair generation. Default true generates a
+# self-signed keypair at install. Set false (or enable global.controlPlaneHA)
+# to consume an externally-provisioned / ESO-synced shared keypair instead.
+jwks:
+  generation:
+    enabled: true
+
```
pgvishnuram: since the jwks is a part of houston and we already have a config inside houston - we should either move all flags inside houston and keep everything above houston

charts/astronomer/templates/houston/api/cp-identity-secret.yaml
```diff
@@ -0,0 +1,49 @@
...
+######################################################
+{{- if .Values.global.controlPlaneHA.enabled }}
+{{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
+{{- /* globalBaseDomain is only consumed on control/unified planes, so only enforce it here.
+   This keeps data-plane renders working when HA is enabled via a shared global values file. */}}
+{{- if not .Values.global.controlPlaneHA.globalBaseDomain }}
+{{- fail "global.controlPlaneHA.globalBaseDomain is required when global.controlPlaneHA.enabled is true" }}
+{{- end }}
+{{- /*
+  Stable per-CP UUID, mounted as the CP_ID env var on every CP pod that runs a
```
danielhoherd, line 12: Don't we already have a per-CP ID? https://github.com/astronomer/astronomer/blob/07b3ca914c/charts/astronomer/templates/cluster-local-data.yaml#L24

charts/astronomer/templates/houston/api/houston-jwt-certificate-secret.yaml
danielhoherd: I agree. As it is now, the value path is `astronomer.jwks.generation.enabled` but that is not specific enough for potential future growth.

tests/chart_tests/test_control_plane_ha.py
```diff
@@ -0,0 +1,204 @@
...
+
+def _ha_values(**overrides):
+    """global.controlPlaneHA enabled with a valid globalBaseDomain, control plane."""
+    cpha = {"enabled": True, "globalBaseDomain": "astro.example.com"}
+    cpha.update(overrides)
+    return {"global": {"plane": {"mode": "control"}, "controlPlaneHA": cpha}}
+
+
+def _any_container_has_cp_id(doc):
+    """True if any (init) container in the pod has the CP_ID env wired to cp-identity."""
```
danielhoherd:
```suggestion
    """True if any container or initContainer in the pod has the CP_ID env wired to cp-identity."""
```

tests/chart_tests/test_control_plane_ha.py
```diff
@@ -0,0 +1,204 @@
...
+        values = self._merge({"global": {"plane": {"mode": "control"}}}, extra_values)
+        docs = render_chart(kube_version=kube_version, show_only=[component_file], values=values)
+        assert len(docs) == 1
+        assert _no_container_has_cp_id(docs[0])
+
+
+@pytest.mark.parametrize("kube_version", supported_k8s_versions)
+class TestHoustonConfigHelmValues:
+    @staticmethod
+    def _helm_block(docs):
```
danielhoherd: Nit: We should probably change this from `_helm_block` to something like `production_yaml_prod_helm` because `_helm_block` is really ambiguous at a glance.

danielhoherd (changes requested):
This looks mostly good. I think we should:
- Figure out can we merge the existing `cluster-local-data` id with the new `cpId`. Having these two things seems confusing and suboptimal.
- Move the jwks values under `astronomer.houston`

charts/astronomer/templates/houston/houston-configmap.yaml
adhipjos: This exact combination is only used in this one place so I don't see much benefit in it and it just an indirection. We can revisit if we find this being reused in the future

charts/astronomer/templates/_helpers.yaml
adhipjos: the secret ref is already centralized adding a helper would re-introduce the whitespace/indent problem that was fixed in the comment below. Not worth it

charts/astronomer/values.yaml
adhipjos: Makes sense. Moved under houston

tests/chart_tests/test_control_plane_ha.py
adhipjos: Fixed

tests/chart_tests/test_control_plane_ha.py
adhipjos: Fixed

charts/astronomer/templates/houston/api/houston-jwt-certificate-secret.yaml
adhipjos: Agreed. Moved under houston

charts/astronomer/templates/houston/api/cp-identity-secret.yaml
adhipjos: I looked at the cluster-local-data generated but as we discussed it's not a uuid so we can't use as it. I thought about changing local-cluster-data to a uuid but that maybe an issue for flightdeck as it might break the db name as uuid could have hyphens plus currently this is used only on the commander side. Will keep this separate as is and keep cp_id restricted to only CPs

danielhoherd (approved):

PR #3303, opened 2026-06-16
PINF-784: support dots in dag_id and task_id for Vector log path parsing 

Shubham-astro:
## Description

Task logs from TaskGroup tasks were silently dropped by Vector when using the Kubernetes executor with DS logging. This happened as the regex in the `parse_airflowV3_path` transform used `[a-zA-Z0-9_-]` for `dag_id `and `task_id` path segments. Airflow natively permits dots in these identifiers, and TaskGroup task IDs use dot notation by default `(prefix_group_id=True)`, e.g. `my_group.my_task`. Vector failed to match these paths and dropped the logs entirely.

## Related Issues

https://linear.app/astronomer/issue/PINF-784/task-logs-from-taskgroup-tasks-with-prefix-group-idtrue-the-default

## Merging
Master

### DIFF
--- a/charts/vector/templates/vector-configmap.yaml
+++ b/charts/vector/templates/vector-configmap.yaml
@@ -155,7 +155,7 @@ data:
           file_str = to_string!(.file)
 
           # Try task execution logs: /dag_id=xxx/run_id=yyy/task_id=zzz/attempt=n.log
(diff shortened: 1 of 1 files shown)


karankhanchandani (approved):
LGTM. It was already verified by field team that this fix works.

PR #3309, opened 2026-06-17
PINF-829 - fix values mangling in commander metadata config

pgvishnuram:
## Description

metadata config values are mangled inside the configmap and makes the reading harder while editing this changes fixes the whitespace problems

before:

image

after:

image

## Related Issues

https://linear.app/astronomer/issue/PINF-829/fix-commander-metadata-configmap-mangled-into-quoted-string-by

## Testing

when editing the commander metadata config they should not see mangled output

## Merging

merge to master and release-2.x

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -16,21 +16,21 @@
 {{- end }}
 {{- end }}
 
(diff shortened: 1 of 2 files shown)


Shubham-astro (approved):
Small lint changes, LGTM

PR #3313, opened 2026-06-18
PINF-842 - fix registry and elasticsearch naming overrides

pgvishnuram:
## Description

This PR allows fullname override for registry and elasticsearch services to reuse the same name on control plane and dataplane split change , also gives customer more flexibility on name selection

## Related Issues

https://linear.app/astronomer/issue/PINF-842/allow-fullname-overrides-for-more-platform-components

## Testing

QA should able to pass custom name by passing 
```
elasticsearch:
   fullnameOverride: customsearch-elasticsearch
astronomer:
   registry:
       fullnameOverride: custom-registry   
```

## Merging

merge to master and release-2.0


charts/astronomer/templates/_helpers.yaml
```diff
@@ -885,3 +885,29 @@ Return registry podSecurityContext, omitting fsGroup,runAsGroup and runAsUser fi
 {{- toYaml .Values.registry.podSecurityContext }}
 {{- end -}}
 {{- end }}
+
+
+{{/*
+Expand the name of the chart.
```
danielhoherd: Incorrect comment.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -885,3 +885,29 @@ Return registry podSecurityContext, omitting fsGroup,runAsGroup and runAsUser fi
...
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "registry.fullname" -}}
+{{- if .Values.registry.fullnameOverride }}
+{{- .Values.registry.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default "registry" .Values.registry.nameOverride }}
```
danielhoherd, line 906: Isn't this redundant to the prior `if` condition?

tests/chart_tests/test_houston_api_deployment.py
```diff
@@ -356,3 +356,21 @@ def test_houston_user_provided_env_vars_and_secret_vars(self, kube_version):
         assert env_vars["ANOTHER_VAR"] == "another-value"
         assert env_vars["MY_SECRET_VAR"] == {"secretKeyRef": {"name": "my-secret", "key": "my-key"}}
         assert env_vars["ANOTHER_SECRET"] == {"secretKeyRef": {"name": "other-secret", "key": "value"}}
+
+    def test_houston_registry_host_overrides(self, kube_version):
```
danielhoherd, line 360: `registry.fullname` has 4 different code paths, but we are only testing one of them. We should test the other code paths too.

charts/astronomer/templates/_helpers.yaml
```diff
@@ -885,3 +885,29 @@ Return registry podSecurityContext, omitting fsGroup,runAsGroup and runAsUser fi
...
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
```
danielhoherd, line 909: I think this has a bug. What if `.Release.Name` is `foo` and somebody set `.registry.fullnameOverride` to `foo-registry-dev`. Wouldn't this overwrite that with simply `foo` ?

Or what if `.registry.fullnameOverride` is `dev-registry-foo`, wouldn't this change it back to just `foo`?

charts/astronomer/templates/_helpers.yaml
danielhoherd:
```suggestion
Expand the name of the registry component.
```

danielhoherd (approved):
LGTM. I left one non-code related comment that should be merged.

PR #3321, opened 2026-06-18
PINF-844 fix missing scc privileges

pgvishnuram:
## Description

Adds missing nginx cp and dp ingress controller service account to scc privileges

## Related Issues

https://linear.app/astronomer/issue/PINF-844/sccenabled-prevents-nginx-pod-creation-due-to-incorrect-service

## Testing

QA should able to provision nginx ingress controller with sccEnabled flag

## Merging

merge to master and release-2.x and 1.x

### DIFF
--- a/templates/scc/astronomer-scc-anyuid.yaml
+++ b/templates/scc/astronomer-scc-anyuid.yaml
@@ -23,11 +23,10 @@ users:
 - system:serviceaccount:{{ .Release.Namespace }}:default
 - system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-commander
 - system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-houston-bootstrapper
-{{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
-- system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-nginx-cp
-{{- end }}
-{{- if eq .Values.global.plane.mode "data" }}
-- system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-nginx-dp
+{{- $mode := .Values.global.plane.mode }}
+{{- if has $mode (list "control" "unified" "data") }}
+- system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-nginx-{{ ternary "dp" "cp" (eq $mode "data") }}
+- system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-nginx-default-backend
 {{- end }}
 - system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-prometheus
 volumes:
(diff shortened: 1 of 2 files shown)


tests/chart_tests/test_scc.py
```diff
@@ -60,6 +61,24 @@ def test_scc_enabled(self, kube_version):
         )
         houston_values = yaml.safe_load(docs[1]["data"]["production.yaml"])
 
-        assert len(docs) == 2
+        assert len(docs) == 3
         assert commander_expected_result in docs[0]["rules"]
         assert houston_values["deployments"]["helm"]["sccEnabled"] is True
+
+    def test_scc_privilges(self, kube_version):
```
rishkarajgi:
```suggestion
    def test_scc_privileges(self, kube_version):
```

templates/scc/astronomer-scc-anyuid.yaml
```diff
@@ -23,7 +23,11 @@ users:
 - system:serviceaccount:{{ .Release.Namespace }}:default
 - system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-commander
 - system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-houston-bootstrapper
-- system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-nginx
+{{- $mode := .Values.global.plane.mode }}
+{{- if has $mode (list "control" "unified" "data") }}
```
danielhoherd: These are all of the possible options. Do we need this conditional check?

pgvishnuram: for unified and control its nginx-cp
for data its  nginx-dp
i wasn't able to find a proper way

tests/chart_tests/test_scc.py
```diff
@@ -60,6 +61,24 @@ def test_scc_enabled(self, kube_version):
         )
         houston_values = yaml.safe_load(docs[1]["data"]["production.yaml"])
 
-        assert len(docs) == 2
+        assert len(docs) == 3
         assert commander_expected_result in docs[0]["rules"]
         assert houston_values["deployments"]["helm"]["sccEnabled"] is True
+
+    def test_scc_privileges(self, kube_version):
```
Shubham-astro, line 68: can we parametrize this test for all the modes.? control,data and unified? 

Something like:
```
@pytest.mark.parametrize(
    "plane_mode,expected_nginx_sa",
    [
        ("control", "nginx-cp"),
        ("unified", "nginx-cp"),
        ("data", "nginx-dp"),
    ],
)
```

templates/scc/astronomer-scc-anyuid.yaml
Shubham-astro: @pgvishnuram I do not think if is needed, we do not expect $mode to be null right.? it would always be one of the three modes, i think we can drop the if condition and only have what you have below.

pgvishnuram: removed all the conditions kept them cleaner

Shubham-astro (approved):
LGTM

PR #3329, opened 2026-06-23
PINF-872: Exclude filesd-reloader sidecar from data plane Prometheus deployments

Shubham-astro:
## Description

In data plane clusters, the astronomer-prometheus StatefulSet was always injecting the filesd-reloader (ap-kuiper-reloader) sidecar regardless of plane mode. On startup, the sidecar unconditionally attempts to connect to astronomer_houston — the Houston database — which does not exist in the data plane. This caused the container to crash-loop immediately with:
```
FATAL: database "astronomer_houston" does not exist
```
## Related Issues

https://linear.app/astronomer/issue/PINF-872/astronomer-prometheus-0-pod-is-in-crashloopbackoff-after-upgrading-to

## Testing

- Removed it from a DP cluster and confirmed that this did not affect the working of prometheus.
Screenshot 2026-06-23 at 4 49 56 PM

- Added unittests.
## Merging

Master, 2.0, 1.1

### DIFF
--- a/charts/prometheus/templates/prometheus-statefulset.yaml
+++ b/charts/prometheus/templates/prometheus-statefulset.yaml
@@ -150,6 +150,7 @@ spec:
               mountPath: /etc/prometheus/alerts.d
             - name: prometheus-config-volume
               mountPath: /etc/prometheus/config
+        {{- if $planeCPMode }}
         - name: filesd-reloader
           image: {{ include "filesdReloader.image" . }}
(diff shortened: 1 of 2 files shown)


charts/prometheus/templates/prometheus-statefulset.yaml
pgvishnuram, line 239: this should also be under condition

pgvishnuram, line 301: this should also be under condition

pgvishnuram (approved):

PR #3335, opened 2026-06-25
PINF-888 - add ingress config support cp -- dp split

pgvishnuram:
## Description

This PR extends ingress configuration to support CP–DP split deployments within the same cluster.
The core challenge was around how domain resolution behaves when global.plane.mode is set to data but no domainPrefix is defined. In this case, we now treat the installation as a same-cluster CP–DP split, where the data plane components communicate internally rather than over external ingress endpoints.
To support this, we restructured the domain helper templates across several charts so that domainPrefix is required for external ingress to be active in data plane mode. When it is absent, certain ingress resources are intentionally disabled — since in a same-cluster split, those services are reachable via in-cluster DNS and don't need to be exposed externally.
Affected ingress resources: Commander (gRPC + metadata), Registry, Elasticsearch, Prometheus, and the external ES proxy.

## Related Issues

https://linear.app/astronomer/issue/PINF-888/add-ingress-config-support-for-cp-dp-split-same-cluster-mode

## Testing

Existing tests were updated to reflect the new domainPrefix-gated rendering behavior, and new parametrized tests were added to cover both the same-cluster (no domainPrefix) and split-with-external-ingress (domainPrefix set) cases.

## Merging

### DIFF
--- a/templates/NOTES.txt
+++ b/templates/NOTES.txt
@@ -32,6 +32,9 @@ The control plane is responsible for platform management, orchestration, and use
 You can access the data plane at:
 - Astronomer dashboard:   {{ if and $astronomerEnabled $platformEnabled }}https://{{ $domainPrefix }}.{{ .Values.global.baseDomain }}{{ else }}Disabled{{ end }}
 - Grafana dashboard:      {{ if and $grafanaEnabled $monitoringEnabled }}https://grafana.{{ .Values.global.baseDomain }}{{ else }}Disabled{{ end }}
+{{ else }}
+- Commander API:   {{ if and $astronomerEnabled $platformEnabled }}commander.{{ .Values.global.baseDomain }}:443{{ else }}Disabled{{ end }}
+- Commander Metadata: {{ if and $astronomerEnabled $platformEnabled }}http://{{ .Release.Name}}-commander.{{ .Release.Namespace }}.svc.cluster.local.:8880{{ else }}Disabled{{ end }}
 {{- end }}
 
 The data plane is responsible for executing and monitoring Airflow workloads.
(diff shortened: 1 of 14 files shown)


templates/NOTES.txt
```diff
@@ -32,6 +32,9 @@ The control plane is responsible for platform management, orchestration, and use
 You can access the data plane at:
 - Astronomer dashboard:   {{ if and $astronomerEnabled $platformEnabled }}https://{{ $domainPrefix }}.{{ .Values.global.baseDomain }}{{ else }}Disabled{{ end }}
 - Grafana dashboard:      {{ if and $grafanaEnabled $monitoringEnabled }}https://grafana.{{ .Values.global.baseDomain }}{{ else }}Disabled{{ end }}
+{{ else }}
+- Commander API:   {{ if and $astronomerEnabled $platformEnabled }}commander.{{ .Values.global.baseDomain }}:443{{ else }}Disabled{{ end }}
+- Commander Metadata: {{ if and $astronomerEnabled $platformEnabled }}http://{{ .Release.Name}}-commander.{{ .Release.Namespace }}.svc.cluster.local.:8880{{ else }}Disabled{{ end }}
```
danielhoherd: Align the URLs:
```suggestion
- Commander API:      {{ if and $astronomerEnabled $platformEnabled }}commander.{{ .Values.global.baseDomain }}:443{{ else }}Disabled{{ end }}
- Commander Metadata: {{ if and $astronomerEnabled $platformEnabled }}http://{{ .Release.Name}}-commander.{{ .Release.Namespace }}.svc.cluster.local.:8880{{ else }}Disabled{{ end }}
```

danielhoherd (approved):
Seems good to me

templates/NOTES.txt
pgvishnuram: let me apply the suggestions directly

PR #3342, opened 2026-06-29
PLX-575: Adds logic in migration scripts to remove hard delete deployment flag

karankhanchandani:
## Description

Adds logic in migration scripts to remove hard delete deployment flag. 

## Related Issues

Resolves [PLX-575](https://linear.app/astronomer/issue/PLX-575/remove-soft-delete-and-default-to-hard-delete-in-apc)

## Testing

- Unit tests

## Merging

2.1, master


Simpcyclassy (approved):
LGTM on the premise we truly want to remove soft delete entirely instead of simply making hard delete deployment the default

PR #3356, opened 2026-07-01
PINF-894/PINF-895 Auth-sidecar support under control-plane HA

adhipjos:
## Summary

Closes two auth-sidecar / customer-BYO-ingress chart gaps that block auth-sidecar from working in control-plane HA (multi-CP active-active) mode out of the box. Both were surfaced by [PINF-745](https://linear.app/astronomer/issue/PINF-745) and live on the same auth-sidecar/BYO chart surface, so they ship together.

### PINF-894: auth-sidecar configmap base-domain precedence
The three monitoring auth-sidecar nginx configmaps (prometheus, alertmanager, grafana) templated the auth subrequest `Host` header, the auth `proxy_pass`, and the login `302` redirect from `global.baseDomain`. Under HA the customer enters via the **global** host and the session cookie is scoped to `controlPlaneHA.globalBaseDomain`, so a user hitting the global host was authenticated/redirected against the per-CP host.

These now reuse the shared `global.authBaseDomain` helper (added by [PINF-893](https://linear.app/astronomer/issue/PINF-893)): HA-on + `globalBaseDomain` set → global host, else `baseDomain`. Single-CP / HA-off rendering is byte-identical to before.

### PINF-895: commander gRPC ingress protocol annotations
`charts/astronomer/templates/commander/commander-grpc-ingress.yaml` emitted `nginx.ingress.kubernetes.io/backend-protocol: "GRPC"` and `enable-http2: "true"` **only** in the auth-sidecar-off branch, so auth-sidecar installs lost the CP→DP gRPC/HTTP-2 backend protocol unless manually patched with `commander.ingress.annotation`. Both annotations are now hoisted above the branch and rendered unconditionally.

While fixing this, guarded `extraAnnotations` in the auth-sidecar branch (matching the else branch) to avoid a bare `{}` YAML render when `extraAnnotations` is empty which was previously masked, now exposed once real keys precede it.

Non-NGINX BYO controllers (ALB/Istio/etc.) ignore the `nginx.ingress.*` annotations harmlessly and must enable gRPC backend + HTTP/2 via their own mechanism. Documented as an inline template comment.

## Tests
New `tests/chart_tests/test_auth_sidecar_global_base_domain.py` (40 cases across supported k8s versions):
- **PINF-894**: all three configmaps: HA-on resolves `Host` / `proxy_pass` / login-302 to the global host; HA-off resolves to `baseDomain` with no broken empty-host URLs.
- **PINF-895**: commander gRPC ingress renders both annotations in auth-sidecar mode (without a manual override) **and** in default mode (regression guard).

Verified: new suite + existing `test_auth_flow_global_base_domain.py`, `test_dual_host_ingress.py`, and `test_astronomer_commander_ingress.py` all pass. Manual `helm template` spot-checks confirm correct rendering across auth-sidecar-on/off and HA-on/off.

## Merging

Target: `feature/cp-ha-dr`

### DIFF
--- a/tests/chart_tests/test_auth_sidecar_global_base_domain.py
+++ b/tests/chart_tests/test_auth_sidecar_global_base_domain.py
@@ -1,0 +1,120 @@
+"""Tests for the auth-sidecar / commander-gRPC chart gaps in control-plane HA (PINF-894, PINF-895).
+
+Sibling to test_auth_flow_global_base_domain.py (PINF-893). Two auth-sidecar / BYO-ingress
+surfaces still assumed the per-CP host under control-plane HA:
+
+PINF-894 — the three monitoring auth-sidecar nginx configmaps (prometheus / alertmanager /
+grafana) templated the auth subrequest `Host`, the auth `proxy_pass`, and the login `302`
+redirect from `global.baseDomain`. Under HA the customer enters via the global host and the
+session cookie is scoped to `controlPlaneHA.globalBaseDomain`, so these must resolve to the
+global host. They now reuse the shared `global.authBaseDomain` helper: HA-on +
+globalBaseDomain -> global host, else baseDomain (single-CP / HA-off rendering unchanged).
+
+PINF-895 — the commander gRPC ingress emitted `backend-protocol: GRPC` and
+`enable-http2: true` only in the auth-sidecar-off branch, so auth-sidecar installs lost the
+CP->DP gRPC/HTTP-2 backend protocol unless manually patched. Both annotations are now
+emitted unconditionally.
+"""
+
+import pytest
+
+from tests import supported_k8s_versions
+from tests.utils.chart import render_chart
+
+PROMETHEUS_CONFIGMAP = "charts/prometheus/templates/prometheus-auth-sidecar-configmap.yaml"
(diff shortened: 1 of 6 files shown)


tests/chart_tests/test_auth_sidecar_global_base_domain.py
```diff
@@ -0,0 +1,119 @@
...
+        """Auth-sidecar mode must render gRPC/HTTP-2 without a manual commander.ingress.annotation."""
+        annotations = self._annotations(kube_version, auth_sidecar=True)
+        assert annotations[BACKEND_PROTOCOL] == "GRPC"
+        assert annotations[ENABLE_HTTP2] == "true"
+
+    def test_auth_sidecar_off_keeps_grpc_annotations(self, kube_version):
+        """Regression guard: default (non-auth-sidecar) mode still has both annotations."""
+        annotations = self._annotations(kube_version, auth_sidecar=False)
+        assert annotations[BACKEND_PROTOCOL] == "GRPC"
+        assert annotations[ENABLE_HTTP2] == "true"
```
danielhoherd, line 119: Nit: these two could be a single parametrized test

danielhoherd (approved):
Looks good

PR #3369, opened 2026-07-06
Refactor local setup scripts and remove etc host dependency

karankhanchandani:
## Description

Refactor local setup scripts and remove etc host dependency. It now has interactive setup where you can choose what type of setup you want and which release. Now you can also select main from the list of releases and it will install main branch of houston, commander, UI, etc.

## Related Issues
NA

## Testing

- Tested by creating clusters locally

## Merging

master


karankhanchandani:
@danielhoherd I have tested this setup couple of times locally and @Simpcyclassy also tested it as well. Is it possible to merge this PR? Why I'm asking to merge this, is it has a couple of changes like removal of etc/host dependency which survives mac/docker restarts. Currently when the docker restarts, the LB IP of both CP and DP changes, and it breaks the local dev.

danielhoherd (approved):
I added an `--interactive` flag, which fixes this change for my workflow (and I think make the default behavior what most people would expect)

PR #3370, opened 2026-07-06
PLX-263: Adds operator.adoption feature flag

karankhanchandani:
## Description

Adds operatorAdoption feature flag

## Related Issues

Resolves [PLX-263](https://linear.app/astronomer/issue/PLX-263)

## Testing

- unit tests
- tested locally

## Merging

master, 2.0

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -81,10 +81,11 @@ dependencies:
 
   # Airflow Operator
   - name: airflow-operator
-    # First-found path wins. Leave `airflow-operator.enabled` unset (the default) and the global
-    # flag governs as before. Set `airflow-operator.enabled: false` to SKIP installing the operator
-    # subchart — e.g. on a data plane whose cluster already runs a standalone Astro Runtime Operator —
-    # while keeping `global.airflowOperator.enabled: true` so Houston/Commander stay operator-aware.
+    # First-found path wins: `airflow-operator.enabled`, if set, decides whether this subchart
+    # installs; if unset, `global.airflowOperator.enabled` decides instead. Set
+    # `airflow-operator.enabled: false` to skip installing the operator subchart on a data plane
+    # whose cluster already runs a standalone Astro Runtime Operator, while keeping
+    # `global.airflowOperator.enabled: true` so Houston/Commander stay operator-aware.
     condition: airflow-operator.enabled,global.airflowOperator.enabled
     tags:
       - airflow-operator
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -81,7 +81,11 @@ dependencies:
 
   # Airflow Operator
   - name: airflow-operator
-    condition: global.airflowOperator.enabled
+    # First-found path wins. Leave `airflow-operator.enabled` unset (the default) and the global
+    # flag governs as before. Set `airflow-operator.enabled: false` to SKIP installing the operator
+    # subchart — e.g. on a data plane whose cluster already runs a standalone Astro Runtime Operator —
+    # while keeping `global.airflowOperator.enabled: true` so Houston/Commander stay operator-aware.
+    condition: airflow-operator.enabled,global.airflowOperator.enabled
     tags:
       - airflow-operator
 
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -673,8 +673,9 @@ data:
     strictSchemaCheck:
       enabled: {{ .Values.houston.strictSchemaCheck.enabled }}
 
-    operatorAdoption:
-      enabled: {{ .Values.global.operatorAdoption.enabled }}
+    operator:
+      adoption:
+        enabled: {{ .Values.global.operator.adoption.enabled }}
 
   # These are any user-specified config overrides.
   local-production.yaml: |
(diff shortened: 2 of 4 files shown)


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -673,6 +673,9 @@ data:
     strictSchemaCheck:
       enabled: {{ .Values.houston.strictSchemaCheck.enabled }}
 
+    operatorAdoption:
+      enabled: {{ .Values.global.operatorAdoption.enabled }}
```
pgvishnuram: do you want to make them auto detected in metadata

pgvishnuram: adding here may be on initial install it will work - later we have to update in clusterConfig

karankhanchandani: @pgvishnuram I mean I don't see this changing with every release, it is not a dynamic property. We can target it in next release if required. We need to make changes in commander right?

karankhanchandani: @pgvishnuram Does it make sense to make it a control plane property, i.e, remove it from deployments.* block?
Adopting a deployment anyhow takes clusterId as param, so if customers don't want to adopt a deployment, they don't select the cluster.

karankhanchandani: Ok so it is not the property of deployments.*,  so it is a control plane property. The question is should we make it per dataplane, but it does not make sense since we pass clusterId in adopt deployments flow,

pgvishnuram: operatorAdoption:
      enabled: {{ .Values.global.operatorAdoption.enabled }}

this is a subdepency of operator right

pgvishnuram: also one more clarification if ir-respective of operator feature this can be enabled - then it can be control plane properly - but every data plane hold operator right - if we are still validating based on operator then it cant be a control plane property

pgvishnuram (approved):

charts/astronomer/templates/houston/houston-configmap.yaml
danielhoherd: I think this should be `operator.adoption.enabled` so that `operator` can be the config key for everything operator related.

karankhanchandani: > if ir-respective of operator feature this can be enabled - then it can be control plane properly 

No, operator can exist independently for the adoption flow. So theoritically, this feature can exist without enabling the operator via helm in dataplane, but it sounds weird if we are installing the dataplane as well (requirement for adoption).

Chart.yaml
```diff
@@ -81,7 +81,11 @@ dependencies:
 
   # Airflow Operator
   - name: airflow-operator
-    condition: global.airflowOperator.enabled
+    # First-found path wins. Leave `airflow-operator.enabled` unset (the default) and the global
+    # flag governs as before. Set `airflow-operator.enabled: false` to SKIP installing the operator
+    # subchart — e.g. on a data plane whose cluster already runs a standalone Astro Runtime Operator —
+    # while keeping `global.airflowOperator.enabled: true` so Houston/Commander stay operator-aware.
```
danielhoherd: This comment needs to be improved. It references conversational context that the reader will not have, like "as before"

values.yaml
```diff
@@ -52,10 +52,18 @@ global:
     containerdCertCopier:
       startupProbe: {}
     priorityClassName: ~
-  #Global flag to enable Operator based deployment
+  #Global flag to enable Operator based deployment (operator-awareness in Houston/Commander).
+  # Whether the airflow-operator SUBCHART is installed is governed separately by the top-level
+  # `airflow-operator.enabled` toggle (see Chart.yaml): leave it unset to track this flag, or set
+  # it false to keep awareness on while skipping the subchart (cluster already runs the operator).
```
danielhoherd: "leave it unset to track this flag, or set it false to keep awareness on while skipping the subchart" could be improved. What does "track this flag" mean? The wording here is very Claude-ish and not directed at the intended audience: the APC application operator.

tests/chart_tests/test_houston_configmap.py
```diff
@@ -1118,3 +1118,54 @@ def test_houston_configmap_certgenerator_custom_tag():
...
+
+def test_houston_configmap_operator_adoption_enabled():
+    """Setting global.operatorAdoption.enabled: true must propagate to production.yaml."""
+    docs = render_chart(
+        values={"global": {"operatorAdoption": {"enabled": True}}},
+        show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
+    )
+    prod = yaml.safe_load(docs[0]["data"]["production.yaml"])
+
+    assert prod["operatorAdoption"]["enabled"] is True
```
danielhoherd: These three could be one parametrized test. Or only the bottom two if we want to go that way.

values.yaml
karankhanchandani: So the intention when I was installing this in local setup was, operator was installed in k8s namespace before Astronomer namespace is installed. We also ship the operator, but I wanted to simulate "true" adoption flow, which is to NOT install the operator via helm as part of DP installation.

This PR was supposed to be merged way back, but the branch got out of sync, so I created a new branch.

I think claude got that context and added this comment here.

danielhoherd (approved):
Looks good

PR #3391, opened 2026-07-09
PINF-955: Scope worker stdout tee to Airflow 2 to stop AF3 task-log duplication in Elasticsearch

Vasu-Madaan:
## Description

The default `airflowLocalSettings` `pod_mutation_hook` (in `charts/astronomer/templates/houston/houston-configmap.yaml`) wraps KubernetesExecutor worker pods when the logging sidecar is enabled. The wrapper does two things:

1. Creates `/var/log/sidecar-log-consumer/finished` when the task subprocess exits, signalling the Vector sidecar to terminate so the K8s task pod can complete. Without it, task pods stay `1/2 NotReady`.
2. Tees stdout/stderr to `/var/log/sidecar-log-consumer/out.log`, which the sidecar ships to Elasticsearch.

The guard applied the `out.log` tee to **both** Airflow 2 (`airflow tasks run`) and Airflow 3 (`python -m airflow.sdk.execution_time.execute_workload`) worker commands.

## Problem

On Airflow 3 the tee is redundant. Airflow 3 writes structured task logs to `/usr/local/airflow/logs/dag_id=.../attempt=N.log`, which the sidecar already ships via the broad log path. With the tee also active, every AF3 task log line is indexed into Elasticsearch twice:

- once from the structured file (carries `run_id`/`task_id`/`dag_id`/`log_id` — this is what the Airflow UI reads), and
- once from `out.log` (no identifiers — invisible in the UI, but stored in ES).

Because the UI filters by `log_id`, the duplication is invisible in the UI and only shows up as doubled ingest/storage cost. It does not affect Airflow 2, where `out.log` is the only path task output takes to ES after the broad log path was version-gated out of the AF2 sidecar config in airflow-chart 1.17.36 (PINF-880).

## Fix

Split the guard into `is_af2` / `is_af3`, keep the `finished`-file termination shim for both, and route the tee (`log_cmd`) only for Airflow 2. On Airflow 3 the redirect becomes a plain terminator, so no `out.log` is written. Airflow 2 behavior is unchanged.

## Measured impact

AF3 KubernetesExecutor deployment (chart 1.17.36, runtime 3.2-5), one-minute test DAG, 71 task-log lines/run, DaemonSet disabled, sidecar enabled — Elasticsearch docs per task run:

| ALS | docs with `log_id` (structured) | docs without `log_id` (out.log) | total/run |
|---|---|---|---|
| Before (tee on AF3) | 71 | 182 | 253 |
| After (this PR) | 71 | 0 | 71 |

3.5x / -72% ES docs per AF3 task run. No change to what the Airflow UI displays. Pods still terminate cleanly (verified: `Completed` + GC, no `1/2 NotReady`).

## Testing

- Added `test_houston_configmap_loggingsidecar_scopes_stdout_tee_to_airflow2` asserting the `is_af2`/`is_af3` split, the AF2-only redirect, the removal of the unconditional `+ log_cmd`, and that the `finished`-file shim still applies.
- `uv run pytest tests/chart_tests/test_houston_configmap.py -k "loggingsidecar or scopes_stdout"` — 11 passed.

## Related

- Linear: PINF-955 (AF3 task-log duplication)
- Follow-up to the earlier chart fix PINF-880 (AF2 broad-path version gate, airflow-chart 1.17.36)


Shubham-astro (approved):

vishwas-astro:
@Shubham-astro and @rishkarajgi  - Are we going to backport this to 2.1 and to 1.1.8?

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -526,14 +526,22 @@ data:
...
+                      # would ingest every task log line a second time. Airflow 2 still needs the tee
+                      # because out.log is its only path to Elasticsearch after the broad log path was
+                      # version-gated out of the Airflow 2 sidecar config.
+                      redirect = log_cmd if is_af2 else " ; "
                       command_str = (
                           cmd_init
                           + " /entrypoint "
                           + shlex.join([str(arg) for arg in container.args])
-                          + log_cmd
+                          + redirect
```
pgvishnuram, line 551: L528 defines log_cmd but here its referenced as redirect

pgvishnuram:
@Shubham-astro this whole change will only work for new dataplane registration for existing dataplanes they have to update via clusterConfig

Vasu-Madaan:
> @Shubham-astro this whole change will only work for new dataplane registration for existing dataplanes they have to update via clusterConfig

Yeah, i have aligned this customer.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -526,14 +526,22 @@ data:
...
+                  is_af3 = int(version.split('.')[0]) >= 3 and container.args[0:3] == ["python", "-m", "airflow.sdk.execution_time.execute_workload"]
+                  if is_af2 or is_af3:
                       container.command = ["tini", "--"]
                       new_args = ["bash", "-c"]
+                      # Airflow 3 writes structured task logs to /usr/local/airflow/logs/**/*.log,
+                      # which the sidecar already ships to Elasticsearch. Teeing stdout to out.log
+                      # would ingest every task log line a second time. Airflow 2 still needs the tee
+                      # because out.log is its only path to Elasticsearch after the broad log path was
+                      # version-gated out of the Airflow 2 sidecar config.
+                      redirect = log_cmd if is_af2 else " ; "
```
Simpcyclassy, line 546: Correct fix, the finished-file termination shim in cmd_init stays shared for both branches and only the tee itself is scoped to is_af2, exactly matching the described AF2/AF3 split.

tests/chart_tests/test_houston_configmap.py
```diff
@@ -337,6 +337,46 @@ def test_houston_configmap_with_loggingsidecar_enabled():
     assert "vector" in prod_yaml["deployments"]["logging"]["loggingSidecar"]["image"]
 
 
+def test_houston_configmap_loggingsidecar_scopes_stdout_tee_to_airflow2():
```
Simpcyclassy, line 345: Good coverage of both the presence of the new tee scoping and the absence of the old unconditional log_cmd usage, consistent with this file's existing substring-assertion style for this configmap.

Simpcyclassy (commented):
Clean, narrowly scoped fix that traces correctly end to end: cmd_init's finished-file termination shim stays unconditional for both is_af2 and is_af3, only the redirect variable changes, so AF2 keeps its tee to out.log (its only path to Elasticsearch) while AF3 gets a no-op semicolon separator instead of double-ingesting its already-structured task logs. The measured before/after Elasticsearch doc counts in the description are a nice touch that make the fix easy to verify independent of the code reading. The new test asserts on exact substrings of the embedded Python string, which is brittle to incidental reformatting, but this matches the existing test style already used elsewhere in this file for the same configmap, so it's consistent rather than a new problem. Comfortable merging as-is.

rishkarajgi (commented):
Checked the one thing that actually matters here beyond the diff itself: whether AF3 task logs still reach Elasticsearch after the out.log tee is removed. They do -- `airflow-chart`'s `logging-sidecar-configmap.yaml` has `/usr/local/airflow/logs/**/*.log` as an unconditional Vector source whenever `airflowVersion >= 3.0.0`, independent of whether out.log ever gets written, so this isn't trading a real duplication fix for a silent AF3 logging gap. The is_af2/is_af3 split is also behavior-preserving for the gating condition itself (same two-clause boolean as before, just named) -- no pod loses the finished-file termination shim that didn't already have it.

pgvishnuram's comment on `redirect` vs `log_cmd` checks out as just the naming indirection, not a bug -- `log_cmd` is still the tee command, `redirect` is whichever value actually gets appended, conditional on `is_af2`. Traced it end to end and it's correct, consistent with Simpcyclassy's read.

Nothing here clears the bar for comment. Two other open PRs (#3499, #3496) touch the same two files, but at completely different line ranges (config line ~121 and test lines ~1084/1092 vs this PR's ~533-556 and ~345) -- checked directly, no real overlap.

PR #3292, opened 2026-06-11
PINF-713 k8s security

danielhoherd:
## Description

**PINF-713 — Kubernetes security hardening (Pod Security Standards "Restricted")**

1. **Container securityContext hardening across platform charts** — set `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `runAsNonRoot: true`, and an explicit non-root `runAsUser` on astronomer, airflow-operator, grafana, pgbouncer, nats, elasticsearch (exporter/nginx), kube-state, alertmanager, external-es-proxy, prometheus, prometheus-node-exporter, prometheus-postgres-exporter, and others.

2. **Pod-level `seccompProfile: RuntimeDefault`** added to `podSecurityContext` across the platform charts (including the astronomer registry `podSecurityContext`).

3. **Shared securityContext helpers (`templates/_helpers.tpl`)** — `platform.containerSecurityContext` layers a per-container override on top of the chart's `.Values.securityContext`, force-enables `readOnlyRootFilesystem: true`, and omits `runAsUser` on OpenShift (or when set to the `"auto"` sentinel) so the cluster SCC can assign a UID. Added the pod-level counterpart **`platform.podSecurityContext`** (omits `fsGroup`/`runAsGroup`/`runAsUser` on OpenShift). **Consolidated ~12 charts onto these two helpers**, removing the per-chart container/pod securityContext duplicates and inline `$required` merges (nginx, prometheus, elasticsearch, nats, airflow-operator, prometheus-node-exporter, alertmanager, grafana, kube-state, pgbouncer, prometheus-postgres-exporter, astronomer). Charts with a component-specific base (loggingSidecar, registry/exporter pod contexts, elasticsearch per-component blocks) were intentionally left and documented in-code.

4. **`securityContext` applied to all Jobs / CronJobs / Helm hooks** — houston cleanup & metrics cronjobs, DB-migration and upgrade-deployments hook jobs, commander jwks hooks, config-syncer cronjob, es-curator cronjob, nats jetstream job, etc.

5. **Per-component UID overrides** where the default non-root UID doesn't fit, e.g. astro-ui (`runAsUser: 101`), the global auth-sidecar (`runAsUser: 101`), grafana db-bootstrapper init container (`runAsUser: 1000`), and airflow-operator (`runAsUser: 65532`).

6. **New PSA opt-in config** `configs/enable-psa.yaml` to label namespaces with `pod-security.kubernetes.io/audit|warn=restricted` for validating conformance.

[...]

PR #3401, opened 2026-07-13
PINF-713 k8s security (#3292) [backport release-1.2]

pgvishnuram:
## Scope
Consolidates per-chart securityContext/seccomp helpers into two shared `platform.*` helpers (`templates/_helpers.tpl`), removes the old per-chart helpers, rewires deployments/statefulsets, and adds PSS-Restricted securityContext/podSecurityContext defaults across charts.

## release-1.2 adaptations
- **openshift flag:** the shared helpers use the flat `.Values.global.openshiftEnabled` (release-1.2 convention), not the nested `.Values.global.openshift.enabled` from the source commit.
- **Excluded** (absent on release-1.2): dp-link, navigator, pilot, external-secrets, and prometheus-node-exporter + their tests.
- **Tests:** kept release-1.2's `test_byo_sa.py` and `test_openshift.py` (the source commit's rewrites are coupled to master's template inventory and the nested openshift flag). New `test_security_context_override.py` and `test_pod_seccomp_profile.py` apply as-is.

## Verification
- `helm template` renders clean in default / OpenShift / unified / data-plane modes.
- Security-context test suite passes (`test_security_context_override`, `test_pod_seccomp_profile`, `test_openshift`); OpenShift path correctly omits `runAsUser`.


danielhoherd (approved):
LGTM. If tests pass then we should be good to go.

PR #3407, opened 2026-07-14
Local CP/DP k3d: self-healing, drift-free cross-cluster networking (no more manual reconcile after OrbStack restarts)

karankhanchandani:
## Summary
Makes the local CP↔DP k3d cross-cluster networking survive OrbStack/Mac restarts on its
own, so you no longer have to re-run `reconcile-k3d-orbstack-network.py` every time
connectivity "randomly" breaks after a restart. Local setup tooling only — no chart
templates, values, or runtime behavior are touched.

## Why
The reconcile helper pins the **current** k3d node / Service-LB IPs into each cluster's
CoreDNS `NodeHosts` and the DP node's `/etc/hosts`. Those targets are fragile:
- On an OrbStack restart, container/LB IPs **drift**, so the pins go stale.
- k3s's node-hosts controller **rewrites `NodeHosts`** and drops the reconcile's block.
- Docker **regenerates a node's `/etc/hosts`** on container restart, wiping appended pins.

Net effect: adopt/Commander, JWKS auth, image-pull registry auth, and even the browser all
broke after every restart until you manually reconciled again.

## What changed
Route CP↔DP through the **stable `astronomer-net` gateway → the existing host
`astro-cp-dp-proxy`** (SNI-passthrough on `:443`), instead of pinning drifting IPs.

**`bin/setup-cp-dp-k3d.py`**
- **Pod-level DNS** (`_ensure_cross_cluster_coredns_routing`): a `coredns-custom` ConfigMap on
  CP **and** DP resolves the peer plane's hostnames to the gateway. Lives in user config k3s
  never rewrites, and the gateway is fixed for the network's lifetime → persistent + drift-free.
  Covers Houston→Commander / `dp01/metadata` (pod-level) and Commander→Houston JWKS.
- **Node-level `/etc/hosts`** (`_ensure_node_hosts_daemonset`): self-healing DaemonSets re-apply
  the pin on every node/pod start (needed because Docker wipes `/etc/hosts` on restart and k3d
  can't set Docker `ExtraHosts`). Fixes node-level image-pull registry auth
  (`houston.<base>/v1/registry/authorization`, previously hit the DP's own cert → x509 mismatch).
  Two DaemonSets, each pinning the *other* plane's hosts: `astronomer-cp-hosts-pin` on the DP,
  `astronomer-dp-hosts-pin` on the CP.
- **Dual-stack proxy** (`_write_proxy_conf`): `listen [::]:443` in addition to IPv4, so browsers
  using Secure DNS/DoH (which resolve `localtest.me`'s public `::1`) can reach the clusters —
  previously `curl` worked but the browser got connection-refused on `::1`.
- `_ensure_dp_node_houston_hosts_pin` now targets the gateway instead of the drifting CP-nginx LB IP.

**`bin/k3d_setup_shared.py`** (reusable primitives, so any local setup script can pin node hosts
the same way):
- `_docker_network_gateway(network)`
- `NODE_HOSTS_DAEMONSET_TEMPLATE` + `_node_hosts_daemonset_manifest()` + `_apply_node_hosts_daemonset()`

**`configs/local-node-hosts-daemonset.yaml`** (new): the DaemonSet spec as a `str.format` template
(`{ds_name}`/`{hosts_line}`), following the existing `configs/mysql-manifest.yaml` convention —
reviewable/editable outside the Python.

## Testing

NA

## Merging

master


sudarshanrampuria (approved):
Tested Locally, And worked fine.

PR #3410, opened 2026-07-14
Dev updates 2026-07-14

danielhoherd:
## Description

Dev workflow updates:
- Fix k3d value precedence when using `main`
- Move `--version main`'s image tag floating (houston/astroUI to `main`, commander to `master`) out to an opt-in `configs/pull-main-master-tags.yaml`, applied via `--helm-values` on any `--version`
- Rename the local-checkout `--version` alias from `main` to `local` — it installs whatever chart is checked out on disk, which has nothing to do with the git branch `main`. Renamed `configs/local-main.yaml` to `configs/k3d-local-scenario.yaml` to match
- Broaden `configs/enable-git-sync-deploys.yaml` for git-sync dev testing (private-CA setup, generic tag placeholders instead of one-off branch names)
- Add a "one `render_chart()` call per test" rule to the chart-tests skill
- Bump the ruff pre-commit hook, refresh `uv.lock`

## Related Issues

N/A, KTLO type work

## Testing

No product code is affected by this change, no testing needed

## Merging

Merge everywhere


karankhanchandani (approved):
LGTM

PR #3411, opened 2026-07-14
APC-1539 - fix logging baseline policies

pgvishnuram:
## Description

elasticsearch curator, nginx, exporter not using the updated baseline policies this PR fixes the behaviour.

## Related Issues

https://linear.app/astronomer/issue/PINF-1022/qa-automation-for-the-k8s-policy-for-platfrom-level-validation-and

## Testing

general kyverno policy tests

## Merging

merge to master and release-2.1 , release-1.2


danielhoherd:
Adopting `platform.containerSecurityContext` here is the right call, and the OpenShift path is correct (verified: curator on `openshift.enabled` renders without `runAsUser`). Two things before merge:

**Tests are red — 10 failures.** The helper merges the component override over the chart-wide `.Values.securityContext` base, and `charts/elasticsearch/values.yaml:379` sets `runAsUser: 1000`. So default curator now renders `runAsUser: 1000`, which the updated assertions don't include:

```
Left contains 1 more item: {'runAsUser': 1000}
```

Only curator is affected — exporter/nginx set their own `runAsUser` (65534/101), so the override wins there.

**Decide whether `runAsUser: 1000` is what we want for curator, not just patch the test.** On master curator had no explicit `runAsUser` (image `USER` + `runAsNonRoot: true`); this pins it to UID 1000, which is the ES node UID and may not suit the `ap-curator` image. Render/CI can't prove it runs as 1000 — worth a quick cluster check. If 1000 is right, update the assertions; if not, give curator an explicit appropriate `runAsUser`.

**Add default + OpenShift test cases.** Both failing tests only exercise the override path. A no-override render and an `openshift.enabled` case (asserting `runAsUser` is omitted) would lock in the behavior that actually broke.

danielhoherd (approved):
Great catch @pgvishnuram! This should have been part of https://linear.app/astronomer/issue/PINF-713/ but fell through the cracks. I'm going back through the chart to ensure nothing else like this fell through.

PR #3422, opened 2026-07-17
PINF-1031 Add new scenario mechanism for running functional tests agasint arbitrary scenarios

danielhoherd:
### What's new

- Add a "scenario" mechanism to functional tests: a named test profile (topology, values overlays, k8s version, namespace labels), auto-discovered from `tests/functional/scenarios/*/test_profile.yaml` and rendered into its own CircleCI job by `bin/generate_circleci_config.py`.
- Add the first scenario, `auth-sidecar`: combines `global.authSidecar.enabled` with a PSS-Restricted-enforcing namespace. Closes the gap that let PINF-1033 reach QA instead of CI -- no test run had ever exercised the interaction between the two.
- Add `bin/run-scenario.py`: loads a scenario's `test_profile.yaml` and drives `reset-local-dev`/`helm-install.py` with the right flags.
- Add `configs/enable-auth-sidecar.yaml` values overlay.

### What changed

- Replace the `TEST_SCENARIO` env var with an explicit `--topology` CLI flag across `helm-install.py`, `setup-kind.py`, `reset-local-dev`, `conftest.py`, and the `Makefile`. The env var didn't reliably survive CI step/process boundaries; `--topology` does.
- `bin/helm-install.py`: add repeatable `--helm-values` and `--namespace-label` flags. Create and label the namespace before install when labels are given -- PSA enforcement isn't retroactive, so it has to land before `--create-namespace`.
- `bin/reset-local-dev`: export the tools dir onto `PATH` for the rest of the script (fixes a bare-`helm` shell-out failure downstream).
- Tolerate the `astronomer` namespace already existing when creating and labeling it.
- Update docs to match: `docs/local-development.md`, `tests/functional/README.md`, `.agents/skills/functional-tests/SKILL.md`, new `tests/functional/scenarios/README.md`.

The existing k8s version test mechanism is untouched -- scenarios are additive.

Screenshot 2026-07-17 at 11 51 08

## Related Issues

- <https://linear.app/astronomer/issue/PINF-1031/improve-astronomerastronomer-repo-ci>

## Testing

I have run this through CI several times during debugging, which is sufficient. All of these changes are CI related.

## Merging

For now this is only for 2.1.

### DIFF
--- a/.agents/skills/functional-tests/SKILL.md
+++ b/.agents/skills/functional-tests/SKILL.md
@@ -14,7 +14,7 @@ Functional tests run against a live Kubernetes cluster (kind) with the Helm char
 ## Critical Rules
 
 1. **Always run tests with `uv run`** — never `python3 -m pytest` or `python -m pytest`
-2. **Always set `TEST_SCENARIO`** before running or the kubeconfig path will be `None`
+2. **Pass `--topology` to `bin/reset-local-dev`** (`unified`/`control`/`data`) before running — pytest itself needs no env var, it infers topology from the test's own location on disk
 3. **Always use uppercase kubeconfig constants** from `tests.utils.k8s` — `KUBECONFIG_UNIFIED`, `KUBECONFIG_CONTROL`, `KUBECONFIG_DATA` (not lowercase variants)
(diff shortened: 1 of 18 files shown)


edmund-astronomer (approved):
LGTM

PR #3427, opened 2026-07-21
Change config to airflowOperator

sudarshanrampuria:
## Description

A cluster runs many operators, so a bare global.operator flag is very generic and can be miss leading. 

Renames the value everywhere it's consumed (values, Chart.yaml condition, templates, setup script, configs, docs, tests). 

## Testing

All test passing

## Merging

main and 2.1

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -99,7 +99,7 @@ global:
     containerdCertCopier:
       startupProbe: {}
     priorityClassName: ~
-  operator:
+  airflowOperator:
     # Turn on operator-based Airflow deployments. When enabled, this chart installs
     # the airflow-operator by default and grants Astronomer the permissions it needs
     # to manage Airflow deployments through the operator.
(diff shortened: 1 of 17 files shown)


danielhoherd (approved):
I love this. It resonates with the zen of python's "explicit is better than implicit." It is also much easier to `grep` for now. Good stuff!

values.yaml
```diff
@@ -99,7 +99,7 @@ global:
     containerdCertCopier:
       startupProbe: {}
     priorityClassName: ~
-  operator:
+  airflowOperator:
     # Turn on operator-based Airflow deployments. When enabled, this chart installs
     # the airflow-operator by default and grants Astronomer the permissions it needs
```
timo-astronomer, line 104: i'd replace `Astronomer` with something like `APC`

```suggestion
    # the airflow-operator by default and grants APC the permissions it needs
```

values.yaml
```diff
@@ -99,7 +99,7 @@ global:
...
     # to manage Airflow deployments through the operator.
```
danielhoherd, line 105: It looks like there are some comments with terminology that we could improve.

PR #3433, opened 2026-07-22
pinf-1071-helm-upgrade-deletes-crds-when-crdcreate-is-disabled

sudarshanrampuria:
## Description

Toggling crd.create=true to false on a Helm upgrade deletes the CRD, which deletes all deployments.

Expected behavior: Helm should stop owning the CRD when it is removed from chart ownership, but the CRD object itself should not be deleted. Existing Airflow CRs and the operator should continue functioning. This behavior should also be documented explicitly.

Proposed solution: add the helm.sh/resource-policy: keep annotation to all CRDs.

## Related Issues

https://linear.app/astronomer/issue/PINF-1071/helm-upgrade-deletes-crds-when-crdcreate-is-disabled

## Testing

Local K3d testing

## Merging

main and 2.1

### DIFF
--- a/charts/airflow-operator/templates/crds/airflow.yaml
+++ b/charts/airflow-operator/templates/crds/airflow.yaml
@@ -4,6 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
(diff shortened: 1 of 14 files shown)


karankhanchandani (approved):
LGTM

PR #3435, opened 2026-07-22
Upgrade airflow operator to 1.6.0 rc4

sudarshanrampuria:
## Description

* Upgrade Airflow operator to rc4

* Fix airflow-operator metrics: make Prometheus scraping work
    
    The manager bound metrics to 127.0.0.1, unreachable from other pods, and the Service advertised 8443 while the manager listened on 8080. Removed the loopback bind (:8080) and aligned the Service to 8080, since there's no rbac-proxy sidecar. Tests updated.

* Update missing labels in all operator related resources.
* Add test cases for platformNodePool in operator

## Merging

main and 2.1


karankhanchandani (approved):
Changes looks good

PR #3441, opened 2026-07-23
PINF-1081 - flight-deck-db-provisioning-fails-when-cluster-roles-are-disabled

pgvishnuram:
## Description

This PR adds supports for flight deck job with necessary role and rolebinding for no cluster role installation needed for  failover enablement

## Related Issues

https://linear.app/astronomer/issue/PINF-1081/flight-deck-db-provisioning-fails-when-cluster-roles-are-disabled

## Testing

Refer ticket

## Merging


charts/astronomer/templates/commander/commander-flightdeck-role.yaml
```diff
@@ -0,0 +1,32 @@
+##############################
+## Commander Flighdeck Role ##
+##############################
+{{- if .Values.global.rbac.enabled }}
+{{- if or .Values.flightDeck.enabled (and .Values.global.dataPlaneFailover.enabled (eq .Values.global.plane.mode "data")) }}
```
danielhoherd: Do we have a named template for this condition? I like that pattern, so if we don't have a template already maybe we can make one?

pgvishnuram: i have updated this one with the named template

code-shivy (dismissed):
LGTM

Shubham-astro (approved):

PR #3445, opened 2026-07-27
remove create_linear_milestone step

sreenusuuda:
## Description

Summary

Deletes bin/create-linear-beta-milestone.py, the script that created a Linear project milestone via GraphQL when a beta/RC Astronomer Helm chart tarball was published.
Removes the corresponding create_linear_milestone job definition and its step in the release-to-internal-triggered workflow from .circleci/config.yml.j2.
Regenerates .circleci/config.yml from the template (uv run python bin/generate_circleci_config.py) so the two files stay in sync — no unrelated changes.

**Why**

We no longer want CI to auto-create Linear milestones on beta/RC chart releases, so this removes the automation and its dead code entirely rather than leaving it disabled.

## Related Issues
https://linear.app/astronomer/issue/PLA-294/remove-linear-milestone-creation-from-release-workflow


danielhoherd (approved):
Removing this failing step looks good to me.

danielhoherd:
This only affected release-2.1 branch. I rebased master on to release-2.1 so this is complete.

PR #3446, opened 2026-07-27
Fix webhook namespace for vwebserverscale validating webhook

sudarshanrampuria:
## Description

The vwebserverscale validating webhook hardcoded namespace: airflow-operator-system in an inline clientConfig instead of using the shared webhooks.clientConfig helper. Outside that namespace it targeted a nonexistent service, blocking webserver scaling (failurePolicy: Fail). Now routed through the helper so the namespace tracks .Release.Namespace, matching the other 9 webhooks.

## Merging

Main and 2.1

### DIFF
--- a/charts/airflow-operator/templates/webhooks/validating-webhook-configuration.yaml
+++ b/charts/airflow-operator/templates/webhooks/validating-webhook-configuration.yaml
@@ -152,11 +152,7 @@ webhooks:
   sideEffects: None
 - admissionReviewVersions:
   - v1
-  clientConfig:
-    service:
-      name: {{ .Release.Name }}-airflow-operator-webhook-service
-      namespace: airflow-operator-system
-      path: /validate-airflow-apache-org-v1beta1-webserver-scale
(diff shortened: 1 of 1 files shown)


charts/airflow-operator/templates/webhooks/validating-webhook-configuration.yaml
```diff
@@ -152,11 +152,7 @@ webhooks:
   sideEffects: None
 - admissionReviewVersions:
   - v1
-  clientConfig:
-    service:
-      name: {{ .Release.Name }}-airflow-operator-webhook-service
-      namespace: airflow-operator-system
-      path: /validate-airflow-apache-org-v1beta1-webserver-scale
+  clientConfig: {{ include "webhooks.clientConfig" (dict "path" "/validate-airflow-apache-org-v1beta1-webserver-scale" "Values" .Values "Release" .Release) | indent 4 }}
```
ianbuss, line 155: Looks like this is already the case in airflow-operator.

ianbuss (approved):

PR #3451, opened 2026-07-28
PINF-1093: enable flightdeck for CP-HA data planes

adhipjos:
## Summary

Commander's **flightdeck store** powers the CP↔DP **version-guard CAS** that rejects stale/out-of-order deployment messages. It was gated on `dataPlaneFailover`, so a **CP-HA install without failover left the guard off** — the exact multi-CP reordering CP-HA introduces (each CP has its own NATS) could then silently regress deployment state.

Net rule: **CP-HA ⇒ flightdeck on the DP**, independent of `dataPlaneFailover`.

## Changes

- **`flightdeck.enabled` helper** (`_helpers.yaml`) now also fires on a data plane when `controlPlaneHA.enabled` is set, alongside the existing `flightDeck.enabled` / `dataPlaneFailover` conditions. `flightDeck.enabled` remains **mode-agnostic** (unchanged from master), preserving the long-standing tested path.
- **Consolidated the inline-duplicated gates** in `cluster-local-data.yaml` (`flightdeck_db_name`) and `commander-deployment.yaml` (commander DSN env) onto the helper. `cluster-local-data` keeps its `(data|unified)` plane guard.
- **Inject `COMMANDER_CONTROL_PLANE_HA_ENABLED`** (`commander_env`) so commander initializes the store under CP-HA.
- Pilot untouched (failover-only executor; never renders under CP-HA alone).

## Tests

- New `test_control_plane_ha_flightdeck.py`: data-mode provisioning (init containers + DSN, configmap, RBAC), the `COMMANDER_CONTROL_PLANE_HA_ENABLED` env, unified-mode negative (CP-HA needs data mode), both-flags-off regression.
- Reworked `test_flightdeck_enabled` / `test_flightdeck_enabled_with_no_cluster_role` (also dedented the latter, which was nested inside the former and never collected by pytest).

## Verified in-cluster (KIND, data plane, CP-HA on / failover off)

```
msg="initializing flightdeck store" controlPlaneHAEnabled=true dataplaneFailoverEnabled=false
msg="flightdeck database connected"
```
`deployment_version` table present (guard live); negative control (both flags off) logs "both disabled" and provisions nothing.

## Related

- Commander counterpart (initializes the store under CP-HA): astronomer/commander#563
- Follow-up cleanup: PINF-1112
- Linear: PINF-1093

## Rollout note

This injects the DSN + CP-HA flag, but the guard only activates once a commander image containing astronomer/commander#563 is deployed.

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -959,14 +959,18 @@ Returns the string "true" or "false" — compare with eq.
 
 
 {{/*
-Whether flightdeck resources should render. Flightdeck is a data-plane-only concept — it
-backs the CP↔DP version-guard CAS (multi-CP message reordering) and DR-failover flight
-processing (migration between DP clusters), neither of which applies on a unified cluster
-(single CP + DP). So the store renders only when plane.mode == data, and then when any of
-flightDeck.enabled, dataPlaneFailover, or controlPlaneHA is set.
+Whether flightdeck resources should render. The store backs the CP↔DP version-guard CAS
+(multi-CP message reordering, needed under controlPlaneHA) and DR-failover flight processing.
+It renders when flightDeck.enabled is set explicitly, OR on a data plane when either
+dataPlaneFailover or controlPlaneHA is enabled.
+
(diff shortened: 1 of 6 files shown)


piyush-shrivastava (approved):
LGTM

PR #3452, opened 2026-07-29
PLX-793: Surface global.dataPlaneFailover.enabled into houston's own config

Shubham-astro:
## Description

`global.dataPlaneFailover.enabled` already exists as the chart-wide DR-failover master switch, The one thing it wasn't doing was reaching houston's own config map , so houston had no way to tell the UI "failover is on for this install." This PR adds that missing piece.

## Related Issues

https://linear.app/astronomer/issue/PLX-793/regions-nav-route-hidden-when-dp-failover-is-enabled-without-control

## Testing
-  Added unittests.
- Tested that regions tab is now available. 
Screenshot 2026-07-30 at 1 17 23 PM

## Merging

Master, 2.1

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -78,6 +78,10 @@ data:
       cookieName: {{ . }}
       {{- end }}
       {{- end }}
+      {{- if .Values.global.dataPlaneFailover.enabled }}
+      dataPlaneFailover:
+        enabled: {{ .Values.global.dataPlaneFailover.enabled }}
+      {{- end }}
       {{- if .Values.registry.auth.secretName }}
       registryAuthSecret: {{ .Values.registry.auth.secretName }}
       {{- end }}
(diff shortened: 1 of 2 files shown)


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -78,6 +78,10 @@ data:
       cookieName: {{ . }}
       {{- end }}
       {{- end }}
+      {{- if .Values.global.dataPlaneFailover.enabled }}
+      dataPlaneFailover:
+        enabled: {{ .Values.global.dataPlaneFailover.enabled }}
+      {{- end }}
```
pgvishnuram, line 84: is it not loaded from commander metadata ??

pgvishnuram: https://github.com/astronomer/astronomer/blob/master/charts/astronomer/templates/_helpers.yaml#L341-L353

Shubham-astro: So this is for if control plane has this feature enabled or not.

rishkarajgi (approved):

PR #3453, opened 2026-07-29
PINF-1114 - Enable endore scans for all images

sudarshanrampuria:
## Description

Wires an Endor Labs (`endorctl`) vulnerability scan across every platform image into CircleCI

### Trigger a scan

On-demand: trigger the pipeline with scan-docker-images: true (UI or API).
Release: click the approve-endorctl-scan approval on a release-* branch.

### What the scan does
Runs endorctl container scan on each image and fails the build only when a finding is (1) ≥ high severity, (2) has a fix version, and (3) isn't in endorignore. No-fix findings and ignored CVEs are reported but don't fail. Exit 0/1/2 = clean / actionable / errored.

Download the artifact
Each scan stores scan-results/ as the endorctl-scan-results artifact — grab endorctl-findings.csv from the job's Artifacts tab in CircleCI.

## Related Issues

https://linear.app/astronomer/issue/PINF-1114/add-endor-scan-to-release-ci-in-astronomerastronomer

## Testing

Test pipeline: https://app.circleci.com/pipelines/github/astronomer/astronomer/20168/workflows/dfa71544-8ef8-49a8-981c-c95ae3a155bf

Artifact Sample:  [endorctl-findings.csv](https://github.com/user-attachments/files/30506240/endorctl-findings.csv)

image

## Merging

All active branches


Simpcyclassy (commented):
This is a CI-only change (no chart templates touched) that wires an endorctl vulnerability gate into CircleCI, with on-demand and release-branch triggers. The .circleci/config.yml and config.yml.j2 stay in sync and the workflow gating (scan-docker-images vs qa-release vs default release) looks mutually exclusive and correct. scan-endorctl.py is well documented with clear docstrings, concurrent scanning, and a sensible severity/ignorefile/fix-version gate. One soft spot: get_cvss() admits in its own comment that the JSON key path is unverified and falls back to a guess, which means the CSV's CVSS column may silently stay empty; worth confirming against a real scan sample before this becomes load-bearing for triage. Comfortable with this merging as-is given it is observability/gating tooling, not a runtime chart change.

danielhoherd:
Instead of using CircleCI, we should use the exiting GHA that is used by other repos. Example: https://github.com/astronomer/astro-agent/blob/cf5823ae019f8ea3ef354f6ca8f3ad0f951876b9/.github/workflows/scan-endorlabs.yaml

danielhoherd:
Superseding this with #3564, which uses the shared GHA composite action (`astronomer/actions/.github/actions/endorlabs-image-scan`) instead of a custom CircleCI scanner, per the discussion above. Closing this one — thanks for the initial groundwork here.

PR #3462, opened 2026-07-31
PINF-986 toleration mount propagation

danielhoherd:
## Description

- `test_no_unexpected_default_tolerations`: renders the whole chart and asserts only `prometheus-node-exporter` (which must run on every node, including tainted ones, for accurate host metrics) has a default toleration -- previously only node-exporter's own toleration was tested, with nothing tying it to the "nothing else has one" half of the claim.
- `test_no_containers_use_unsafe_mount_propagation`: same render-everything-and-report-offenders pattern, asserting no `volumeMount` sets `mountPropagation` to `HostToContainer` or `Bidirectional`. Verified compliant by hand previously (2026-07-08); this makes it a real regression test.
- Both slot directly into `TestAllPodSpecContainers`'s existing `pod_manager_docs` sweep -- no new render/fixture infrastructure needed, no code changes, only test coverage was missing.

Closes PINF-986's `DENYlistTolerations` and `MountPropagation` checklist rows for this chart.

## Related Issues

https://linear.app/astronomer/issue/PINF-986

## Testing

These are all test changes, no additional special testing is needed for these changes.

## Merging

release-2.1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -459,14 +459,61 @@ workflows:
       - scenario-unified-1-35-1:
           requires:
             - build-artifact
+      # Manual gate for the custom functional scenarios on non-release branches.
+      # On master and release-X.Y they run automatically (below) and gate the
+      # release; on every other branch they wait on this approval before running.
+      - approve-scenarios:
+          type: approval
+          filters:
+            branches:
+              ignore:
+                - master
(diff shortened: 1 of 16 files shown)


edmund-astronomer (approved):
lgtm

PR #3466, opened 2026-08-04
PINF-1139 - fix namespace pools deletion pilot config

pgvishnuram:
## Description

Pilot incorrectly passes namespace configuration for pools configuration - which deletes the underlying namespaces instead of skipping them

## Related Issues

https://linear.app/astronomer/issue/PINF-1139/pilot-deletes-namespaces-when-namespace-pools-are-enabled

## Testing

QA should not see namespace being deleted when nampace pools are configured when on failover testing

## Merging

merge to master , release-2.1

### DIFF
--- a/charts/astronomer/templates/_helpers.yaml
+++ b/charts/astronomer/templates/_helpers.yaml
@@ -331,10 +331,14 @@ gcp_cloud_logging:
   value: "{{ include "houston.authServiceURL" . }}"
 {{- end }}
 
-{{ define "commander_env" -}}
+{{ define "common_shared_env" -}}
 - name: COMMANDER_NAMESPACE_POOLS_ENABLED
   value: {{ ternary "true" "false" .Values.global.namespaceManagement.namespacePools.enabled | quote  }}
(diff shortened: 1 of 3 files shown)


Shubham-astro (approved):
Small change, looks good

charts/astronomer/templates/_helpers.yaml
```diff
@@ -331,10 +331,14 @@ gcp_cloud_logging:
   value: "{{ include "houston.authServiceURL" . }}"
 {{- end }}
 
-{{ define "commander_env" -}}
+{{ define "common_shared_env" -}}
```
Shubham-astro, line 334: @pgvishnuram  are you thinking of adding more env under this.? Just asking due to the name "common_shared_env"

pgvishnuram: right now the core require is to share the COMMANDER_NAMESPACE_POOLS_ENABLED env vars for the current issue , also instead of adding 2 sets i creates them and included them on both.

PR #3475, opened 2026-08-05
Fix airflow-operator broken webhook TLS default

sudarshanrampuria:
## Description

With shipped defaults (certManager.enabled: false, no customCertsSecretName), the manager rendered a custom-certs volume pointing at an empty secret name, so the operator install failed out of the box.

This flips certManager.enabled to true by default and adds a template-time validation that fails fast with a clear message when: both cert sources are disabled, or useCustomTlsCerts is on without customCertsSecretName/caBundle. The guard is gated on operator.enabled.

## Related Issues

NA

## Testing

Unit test

## Merging

main and 2.1


Shubham-astro (approved):
Discussed changes, LGTM

PR #3479, opened 2026-08-05
fix  APC docs link

pgvishnuram:
## Description

fix docs link

## Related Issues

NA

## Testing

NA

## Merging

merge to all release branches


danielhoherd (approved):
These days, this new link is much more appropriate than the old one for the people who will see it at the end of a `helm install`, since the APC application admin is probably not a DAG author.

PR #3483, opened 2026-08-06
PINF-1148 allow jwks hook job service account overrides

pgvishnuram:
## Description

This PR allows overriding jwks hook job to use custom service account name which was shared with commander service account before this change and this now rectifies and gives more control on the naming 

## Related Issues

https://linear.app/astronomer/issue/PINF-1148/separate-jwks-hook-job-service-account-config-from-commander

## Testing

QA should able to override customer service account name for jwks hook job

## Merging

merge to master, release-2.1, 2.0 and 1.0


Shubham-astro (approved):
LGTM

danielhoherd (approved):

PR #3487, opened 2026-08-08
PINF-1170 - disable cluster scoped CES and CPS changes

pgvishnuram:
## Description

Disable ClusterExternalSecret and ClusterPushSecret by default

## Related Issues

https://linear.app/astronomer/issue/PINF-1170/disable-cluster-secret-defaults-and-fix-serviceaccount-rbac-template

## Testing

No action required from QA

## Merging

merge to master and release-2.1.x


danielhoherd:
Test gap here: [`test_rbac_disabled`](https://github.com/astronomer/astronomer/blob/103be02170d84f8bd069a8bc190f5e9114106146/tests/chart_tests/test_external_secrets.py#L226-L240) confirms the Role/RoleBinding/ServiceAccount don't render with `rbac.create: false`, but nothing renders `deployment.yaml` to check what the operator pod actually runs as in that state.

I rendered both sides to check. Before this PR, `rbac.create: false` still creates the named ServiceAccount and the pod uses it. After this PR (the new `$rbacEnabled` gate in [`serviceaccount.yaml`](https://github.com/astronomer/astronomer/blob/103be02170d84f8bd069a8bc190f5e9114106146/charts/external-secrets/templates/serviceaccount.yaml#L4-L5)), no ServiceAccount is created and the pod falls back to `serviceAccountName: default`. Any `serviceAccount.annotations` (e.g. IRSA `eks.amazonaws.com/role-arn`) silently stop applying too, since the SA carrying them never renders.

If that's the intended contract, worth adding a case that renders `DEPLOYMENT_TEMPLATE` with `rbac.create: False` and asserts `serviceAccountName == "default"`, so the fallback is asserted instead of implied. If it's not intended, someone using `rbac.create: false` today to bring their own Role/RoleBinding while keeping a real annotated SA would lose that silently on upgrade.

Separate naming note: this is the chart-local `external-secrets.rbac.create`, not `global.rbac.enabled` (used elsewhere in `deployment.yaml` for the ClusterRole vs. Role choice). Worth being explicit about which one in the PR description since both exist in this chart.

charts/external-secrets/values.yaml
```diff
@@ -70,10 +70,10 @@ systemAuthDelegator: false
 # -- if true, the operator will process cluster external secret. Else, it will ignore them.
 # When enabled, this adds update/patch permissions on namespaces to handle finalizers for proper
 # cleanup during namespace deletion, preventing race conditions with ExternalSecrets.
-processClusterExternalSecret: true
+processClusterExternalSecret: false
```
rishkarajgi, line 73: Shipped default flips `true`→`false` here (and for `processClusterPushSecret` just below). This is a backward-compat event: on upgrade, an existing failover install relying on `ClusterExternalSecret`/`ClusterPushSecret` silently stops reconciling them. Direction is correct (least-privilege), but please confirm it's intended and release-note it.

charts/astronomer/templates/commander/commander-role.yaml
```diff
@@ -38,7 +38,7 @@ metadata:
 rules:
 {{- if $.Values.global.dataPlaneFailover.enabled  }}
 - apiGroups: ["external-secrets.io", "generators.external-secrets.io"]
-  resources: ["clusterexternalsecrets", "clusterpushsecrets", "clustersecretstores", "externalsecrets", "pushsecrets", "secretstores", "generatorstates"]
+  resources: ["clustersecretstores", "externalsecrets", "pushsecrets", "secretstores", "generatorstates"]
```
rishkarajgi, line 41: Removing the `clusterexternalsecrets`/`clusterpushsecrets` verbs from the failover role is a good narrowing — but it's safe only if commander's failover Go code never touches those cluster-scoped CRDs. That call graph isn't visible from the chart; if it does still use them this becomes a runtime `RBAC: forbidden`. Worth confirming with someone who knows the commander side.

rishkarajgi (commented):
**Verdict: PASS** — no blockers, 2 should-fix, 1 nit.

**Direction of change (credited):** this is a clean **least-privilege narrowing**, not a widening. Every change removes or gates a permission — the commander failover ClusterRole drops the `clusterexternalsecrets`/`clusterpushsecrets` verbs, the operator flips `processClusterExternalSecret`/`processClusterPushSecret` to `false` (rendering `--enable-cluster-*-reconciler=false` and dropping the matching operator-role grants), and the ServiceAccount now additionally gates on `rbac.create`. Namespaced ExternalSecret/PushSecret/SecretStore and cluster *stores* are untouched, so functionality is preserved and the operator/commander roles stay internally consistent. README + values + templates are in sync.

**Should-fix:**
1. **This flips a shipped default, so it's a backward-compat / upgrade event.** Existing failover installs that rely on `ClusterExternalSecret`/`ClusterPushSecret` will silently stop reconciling them *and* lose the RBAC verbs on upgrade. The direction is right, but please confirm it's intended, release-note it, and confirm no in-field failover install depends on the cluster-scoped variants.
2. **Pin the new default in a render test.** The tests cover the commander rule and the SA gate, but nothing positively asserts the *default* render now omits the cluster CES/CPS grants and emits the two `--enable-cluster-*-reconciler=false` args — add one so it can't silently regress.

**Nit:** `$rbacEnabled` is recomputed in both `_helpers.tpl` and `serviceaccount.yaml` (+ trailing double-space) — cosmetic.

**Needs SME / can't verify from the chart:** whether commander's failover **Go** code still calls the removed cluster-scoped verbs. If it does, this becomes a runtime `RBAC: forbidden` break — worth a quick confirm from someone who knows the commander call graph, since the chart alone can't prove the grant is unused.

charts/external-secrets/templates/_helpers.tpl
```diff
@@ -76,7 +76,8 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 Create the name of the service account to use
 */}}
 {{- define "external-secrets.serviceAccountName" -}}
-{{- if .Values.serviceAccount.create }}
+{{- $rbacEnabled := and .Values.rbac.create .Values.serviceAccount.create  }}
```
rishkarajgi: SA name now falls back to `default` when `rbac.create=false` — consistent with the new `serviceaccount.yaml` gate and the updated `test_rbac_disabled`. Good. Minor: this `$rbacEnabled` expression is duplicated in `serviceaccount.yaml`; a shared helper would avoid drift.

charts/external-secrets/values.yaml
Simpcyclassy, line 73: This default flip silently stops reconciliation of any ClusterExternalSecret objects a customer already has on upgrade, needs an explicit release-note callout, not just No action required from QA in the PR description.

charts/external-secrets/values.yaml
```diff
@@ -70,10 +70,10 @@ systemAuthDelegator: false
...
 
 # -- if true, the operator will process cluster push secret. Else, it will ignore them.
-processClusterPushSecret: true
+processClusterPushSecret: false
```
Simpcyclassy, line 76: Same concern as processClusterExternalSecret above, this also silently disables existing ClusterPushSecret reconciliation on upgrade.

charts/external-secrets/templates/_helpers.tpl
Simpcyclassy: This rbacEnabled expression is duplicated verbatim in serviceaccount.yaml, consider extracting a shared named template so the two can't drift.

charts/astronomer/templates/commander/commander-role.yaml
Simpcyclassy, line 41: Good, this correctly trims commander's RBAC to match the new default of not processing cluster-scoped secrets, and the matching tests were updated in the same PR.

Simpcyclassy (changes requested):
The template and RBAC changes correctly implement the stated goal and are well tested, but the values.yaml default flip needs a customer-facing callout before merge. processClusterExternalSecret and processClusterPushSecret flip from true to false by default, which silently stops the operator from reconciling any ClusterExternalSecret/ClusterPushSecret objects a customer already has deployed once they upgrade, with no explicit opt-back-in guidance in the PR description, and Testing says only No action required from QA which undersells the customer impact for anyone actively using these cluster-scoped resources today. The serviceaccount.yaml + _helpers.tpl coupling of ServiceAccount creation to rbac.create is a real behavior change too (previously serviceAccount.create alone controlled it) but it is deliberate per the linked issue title and is correctly covered by the updated test_rbac_disabled, so that part is fine as long as it's intentional. Minor: the `and .Values.rbac.create .Values.serviceAccount.create` expression is now duplicated verbatim in both _helpers.tpl and serviceaccount.yaml, worth extracting into one named template. Commander RBAC and its tests are updated consistently with the resource-list trim. Please add a release-note / upgrade-note callout for the CES/CPS default flip before this ships.

charts/external-secrets/values.yaml
pgvishnuram: ClusterExternalSecret/ClusterPushSecret this is not at all used in any of our internal flows - safe to remove them

charts/astronomer/templates/commander/commander-role.yaml
pgvishnuram: we dont have code path that creates clusterexternalsecrets/clusterpushsecrets

charts/external-secrets/templates/_helpers.tpl
pgvishnuram: this is a expected change mocking our existing design format - let me fix the shared helper

charts/external-secrets/values.yaml
pgvishnuram: we dont use this any where in our code - safe to remove

charts/external-secrets/values.yaml
pgvishnuram: we dont use this any where in our code - safe to remove

danielhoherd (approved):
Looks good. Removing cluster scoped permissions always makes customers happy :)

PR #3490, opened 2026-08-12
Local k3d setup: self-heal Docker-stubbed bind mounts, fix DP install ordering, de-duplicate registry helpers

karankhanchandani:
## Description

Three related fixes to the local k3d dev setup, all surfaced by one broken run.

### 1. Setup wedges permanently when Docker stubs a bind-mounted config file

Docker creates the source of a bind mount as an empty directory when that source does not
exist. Three helper configs are bind-mounted as single files:

| File | Container |
|---|---|
| `registry-configs/*.yml` | the 5 pull-through registry proxies |
| `cp-dp-dnsmasq.conf` | `astro-cp-dp-dnsmasq` |
| `cp-dp-proxy-nginx.conf` | `astro-cp-dp-proxy` |

If those files go missing while the containers still exist, the next Docker/OrbStack restart
re-runs the containers and the daemon recreates each path as a directory. That wedges the
setup in both directions, and neither side recovers on its own:

- the containers can no longer mount a directory onto a file and exit 127
- the setup script then fails with `[Errno 21] Is a directory: '.../astronomer-registry-proxy-quay.yml'`

Adds `_write_bind_mount_file()` in `k3d_setup_shared.py`, which clears a stub directory before
writing, and routes all three sites through it. `_ensure_registry()` also no longer treats
`docker start` as success without checking: a container that will not stay running is now
recreated (cached layers live in the `*-data` volumes, so nothing is lost). Previously the
script reported success and continued to a cluster create whose image pulls had no working
mirror.

### 2. The DP Helm install ran before the SNI proxy it depends on

The DP's `astronomer-commander-jwks-hook` is a `pre-install,pre-upgrade` hook that fetches the
JWT signing key from `https://houston.<baseDomain>`. From inside the DP, that name resolves via
the `coredns-custom` override to the `astronomer-net` gateway, so the hook reaches the CP only
through the host's `astro-cp-dp-proxy`. That proxy was created four milestones *after* the DP
install.

A first-ever install works — no `coredns-custom` yet, so the hook resolves via the NodeHosts
pin. But that override is written at the end of every run and persists in the cluster, so from
the second run onward the DP install silently depends on a container the script starts later.
With the proxy down, the hook fails with `BackoffLimitExceeded` and takes the whole DP install
with it.

Moves `_setup_local_networking()` ahead of the Helm section (it only needs `settings` — the
published ports are fixed at cluster-creation time). The end-to-end check is split into
`_verify_local_networking(settings)` and stays at the end of the run, where it is meaningful.

### 3. `setup-local-registry.py` was a near-complete copy of the shared registry code

407 → 160 lines. It duplicated `_print`/`_debug`/`_run`, the container inspection helpers, the
spec table, the config generation, the ensure path, four path constants, and a second
hand-written copy of the k3d mirror map — which is why fix (1) had to be written twice.

The file cannot be imported (dash in the name), so the direction is one-way: `k3d_setup_shared`
owns the logic and the CLI keeps only `--status`, `--destroy/--purge`, and argparse. The mirror
map is now derived from the spec table via a `mirror_hosts` field, so adding a proxy is a single
tuple entry instead of edits in five places. Also drops a parameter `_get_registry_config_path()`
never used, and centralizes the `astronomer-net` literal as `DEFAULT_DOCKER_NETWORK`.

Docs: adds a troubleshooting entry for the 127 / `Errno 21` failure to
`docs/cp-dp-k3d-setup-guide.md` and `docs/local-registry.md`.

## Related Issues

NA

## Testing

- Generated k3d mirror config asserted byte-identical to the previous hardcoded string,
  including the `docker.io` / `index.docker.io` double alias.
- `_write_bind_mount_file()` exercised against a simulated Docker stub directory and against a
  missing parent directory.
- `--help` verified on all three entry points; each does no work before argparse.
- `setup-local-registry.py --status` run against live containers to exercise the new import path.
- Fix (1) confirmed against the real failure: the five registry proxies were dead at
  `Exited (127)` with stub directories on disk, and came back up on the next run with no manual
  cleanup.

```
Milestones summary:

+----+------------------------------------------------------------------------+------------+----------+------------------------------------------------------------------------------------+
| #  | Milestone                                                              | Status     | Duration | Details                                                                            |
+----+------------------------------------------------------------------------+------------+----------+------------------------------------------------------------------------------------+
| 1  | Validate prerequisites (docker/k3d/kubectl/helm)                       | ✅ Success  | 0.1s     |                                                                                    |
| 2  | Ensure Docker network `astronomer-net` exists                          | ✅ Success  | 0.1s     |                                                                                    |
| 3  | Ensure local pull-through registry proxy containers are running        | ✅ Success  | 0.3s     | config=/Users/karankhanchandani/.local/share/astronomer-software/k3d-registry.yaml |
| 4  | Generate TLS certs (mkcert) — one per plane, scoped to its own host... | ✅ Success  | 0.4s     | 2 cert(s): k3d-cp01, k3d-dp01                                                      |
| 5  | Ensure k3d clusters exist (CP=cp01, DP=dp01)                           | ✅ Success  | 37.8s    |                                                                                    |
| 6  | Apply namespace + secrets in all clusters (ns=astronomer)              | ✅ Success  | 0.6s     | tlsSecret=astronomer-tls caSecret=mkcert-root-ca                                   |
| 7  | Configure local DNS + SNI proxy for `localtest.me` (no /etc/hosts e... | ✅ Success  | 2.4s     | resolver present; proxy CP:8443 DP:[8444]                                          |
| 8  | Write CP/DP Helm values files                                          | ✅ Success  | 0.0s     | dir=/var/folders/1r/cdc2t03565l2h31qnmrt1yhw0000gn/T/astro-k3d-f3l31dr3            |
| 9  | Ensure Helm repo `astronomer-internal` is up-to-date                   | ✅ Success  | 2.4s     | version=2.1.0                                                                      |
| 10 | Helm install/upgrade Control Plane (context=k3d-cp01)                  | ⏳ Running  | -        |                                                                                    |
| 11 | Install cert-manager on cp01 (operator webhooks)                       | ✅ Success  | 11.7s    | version=v1.19.4                                                                    |
| 12 | Install ServiceMonitor CRD on cp01 (operator apiserver watch)          | ✅ Success  | 273.3s   | prometheus-operator v0.76.0                                                        |
| 13 | Create DP astronomer-bootstrap secrets                                 | ⏭️ Skipped | 0.0s     | postgresql subchart creates it automatically                                       |
| 14 | Pre-DP: update dp01 CoreDNS NodeHosts for CP ingress                   | ✅ Success  | 5.0s     |                                                                                    |
| 15 | Install cert-manager on dp01 (operator webhooks)                       | ✅ Success  | 12.6s    | version=v1.19.4                                                                    |
| 16 | Install ServiceMonitor CRD on dp01 (operator apiserver watch)          | ✅ Success  | 0.1s     | prometheus-operator v0.76.0                                                        |
| 17 | Helm install/upgrade Data Plane (context=k3d-dp01)                     | ✅ Success  | 63.8s    |                                                                                    |
| 18 | Reconcile cross-cluster DNS + node-level hosts pins (dp01)             | ✅ Success  | 2.9s     |                                                                                    |
| 19 | DP node: pin houston.<baseDomain> to CP ingress (dp01 /etc/hosts)      | ✅ Success  | 0.1s     |                                                                                    |
| 20 | Node /etc/hosts self-healing DaemonSets (CP + DP, survive OrbStack ... | ✅ Success  | 0.1s     |                                                                                    |
| 21 | Drift-free cross-cluster pod DNS (coredns-custom -> gateway -> SNI ... | ✅ Success  | 3.2s     |                                                                                    |
| 22 | Verify local DNS + SNI proxy end-to-end (name -> dnsmasq -> :443 pr... | ✅ Success  | 1.1s     |                                                                                    |
+----+------------------------------------------------------------------------+------------+----------+------------------------------------------------------------------------------------+

✅ Completed.
```
No chart templates changed; chart unit tests are unaffected.

## Merging

master

### DIFF
--- a/bin/k3d_setup_shared.py
+++ b/bin/k3d_setup_shared.py
@@ -324,20 +324,22 @@ proxy:
 """
 
 
-def _write_registry_config(config_path: Path, content: str) -> None:
-    """Write a registry proxy config file, clearing a directory left behind at that path.
+def _write_bind_mount_file(path: Path, content: str) -> None:
+    """Write a file that is bind-mounted into a container, clearing a directory left at that path.
 
-    Docker creates the source of a bind mount as an empty directory when the source does not
-    exist. If the config file goes missing while the container is registered (a wiped
-    `~/.local/share`, a Docker/OrbStack restart that re-runs `--restart always` containers), the
-    daemon recreates the path as a directory. That wedges the setup permanently: the container
-    can no longer mount a directory onto a file and exits 127, and writing the config here fails
-    with `[Errno 21] Is a directory`. Removing the stub restores a plain file.
+    Use this for every host file mounted with `-v <host file>:<container file>`, never a plain
+    `write_text`. Docker creates the source of a bind mount as an empty directory when the source
+    does not exist. If the file goes missing while its container still exists (a wiped
+    `~/.local/share`, a Docker/OrbStack restart that re-runs `--restart` containers), the daemon
+    recreates the path as a directory. That wedges the setup permanently: the container can no
+    longer mount a directory onto a file and exits 127, and writing the file here fails with
+    `[Errno 21] Is a directory`. Removing the stub restores a plain file.
     """
-    if config_path.is_dir() and not config_path.is_symlink():
-        _print(f"  Removing stale directory left at {config_path} (expected a file)")
-        shutil.rmtree(config_path)
-    config_path.write_text(content)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    if path.is_dir() and not path.is_symlink():
+        _print(f"  Removing stale directory left at {path} (expected a file)")
+        shutil.rmtree(path)
+    path.write_text(content)
 
 
 def _container_state(name: str) -> str | None:
@@ -412,7 +414,7 @@ def _ensure_local_registries(docker_network: str) -> None:
     REGISTRY_CONFIG_DIR.mkdir(parents=True, exist_ok=True)
     for spec in _REGISTRY_SPECS:
         config_path = REGISTRY_CONFIG_DIR / f"{spec.name}.yml"
-        _write_registry_config(config_path, _registry_docker_config(spec))
+        _write_bind_mount_file(config_path, _registry_docker_config(spec))
         _ensure_registry(spec, docker_network)
 
 
--- a/bin/k3d_setup_shared.py
+++ b/bin/k3d_setup_shared.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import json
 import os
 import shlex
+import shutil
 import subprocess
 import time
 from dataclasses import dataclass
@@ -27,6 +28,8 @@ HELPER_BIN_DIR = HELPER_DIR / "bin"
 REGISTRY_CONFIG_DIR = HELPER_DIR / "registry-configs"
 K3D_REGISTRY_CONFIG_PATH = HELPER_DIR / "k3d-registry.yaml"
 REGISTRY_IMAGE = "registry:2"
+
+DEFAULT_DOCKER_NETWORK = "astronomer-net"
 
 HELM_REPO_NAME = "astronomer-internal"
 HELM_CHART = f"{HELM_REPO_NAME}/astronomer"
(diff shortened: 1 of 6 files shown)


Shubham-astro (approved):
LGTM.

PR #3491, opened 2026-08-12
PLX-817: Allow disabling the Grafana link via Helm chart values

karankhanchandani:
## Description

`deployments.metricsReporting.grafana.enabled` was hardcoded to `true` in the
Houston configmap template, so the Grafana link in the UI could not be
disabled even when Grafana isn't in use 

This adds `global.metricsReporting.grafana.enabled` (default `true`, so
existing installs are unaffected) as a chart value, following the same
pattern already used by the sibling `global.metricsReporting.taskUsageMetrics.enabled`
flag, and wires it into the Houston configmap instead of the hardcoded literal.

## Related Issues

Fixes https://linear.app/astronomer/issue/PLX-817/allow-disabling-grafana-link-via-helm-chart

## Testing

- Added `test_houston_configmap_features_grafana_disabled`, asserting
  `deployments.metricsReporting.grafana.enabled` renders `false` when
  `global.metricsReporting.grafana.enabled: false` is set.
- Existing `test_houston_configmap_features_grafana_defaults` still asserts
  the default remains `true`.

## Merging

main

### DIFF
--- a/.github/workflows/pr_checks.yml
+++ b/.github/workflows/pr_checks.yml
@@ -36,7 +36,7 @@ jobs:
             ((errors += 1))
           fi
 
-          if [[ "${KTLO}" != "true" ]] && [[ "${AUTOMATED_PR}" != "true" ]] && [[ "${PR_BODY}" != *"astronomer/issue"* && "${PR_BODY}" != *"PINF-"* && "${PR_BODY}" != *"APC-"* && "${PR_BODY}" != *"PLA-"* ]] ; then
+          if [[ "${KTLO}" != "true" ]] && [[ "${AUTOMATED_PR}" != "true" ]] && [[ "${PR_BODY}" != *"astronomer/issue"* && "${PR_BODY}" != *"PINF-"* && "${PR_BODY}" != *"APC-"* && "${PR_BODY}" != *"PLA-"* && "${PR_BODY}" != *"PLX-"* ]] ; then
             echo "PR description does not contain an issue link"
             ((errors += 1))
           fi
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -126,7 +126,7 @@ data:
           enabled: {{ .Values.global.metricsReporting.taskUsageMetrics.enabled }}
           reportNumberOfDays: {{ .Values.houston.cleanupTaskUsageData.olderThan}}
         grafana:
-          enabled: true
+          enabled: {{ .Values.global.metricsReporting.grafana.enabled }}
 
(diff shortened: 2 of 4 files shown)


.github/workflows/pr_checks.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             ((errors += 1))
           fi
 
-          if [[ "${KTLO}" != "true" ]] && [[ "${AUTOMATED_PR}" != "true" ]] && [[ "${PR_BODY}" != *"astronomer/issue"* && "${PR_BODY}" != *"PINF-"* && "${PR_BODY}" != *"APC-"* && "${PR_BODY}" != *"PLA-"* ]] ; then
+          if [[ "${KTLO}" != "true" ]] && [[ "${AUTOMATED_PR}" != "true" ]] && [[ "${PR_BODY}" != *"astronomer/issue"* && "${PR_BODY}" != *"PINF-"* && "${PR_BODY}" != *"APC-"* && "${PR_BODY}" != *"PLA-"* && "${PR_BODY}" != *"PLX-"* ]] ; then
             echo "PR description does not contain an issue link"
```
karankhanchandani, line 40: Required because PLX ticket fail the PR check which it should not.

values.yaml
```diff
@@ -171,6 +171,9 @@ global:
   metricsReporting:
     taskUsageMetrics:
       enabled: false
+    grafana:
+      # Whether the UI shows a link to Grafana. Disable if Grafana is not in use.
+      enabled: true
```
pgvishnuram: if you do this -> this will overlap with the tags and it wont respect tags.monitoring=false wont remove the grafana

pgvishnuram: thats why we dont expose them and hard set them in the backend config - to be overridable at the houston config level - adding configuration here - then there no more tags concept to remove the whole component based on that.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -126,7 +126,7 @@ data:
           enabled: {{ .Values.global.metricsReporting.taskUsageMetrics.enabled }}
           reportNumberOfDays: {{ .Values.houston.cleanupTaskUsageData.olderThan}}
         grafana:
-          enabled: true
+          enabled: {{ .Values.global.metricsReporting.grafana.enabled }}
```
pgvishnuram: instead  hasKey and flag is enabled -> always defaults to true

pgvishnuram:
I am approving this PR -> this still needs a houston change switch from gdc to config.get

pgvishnuram (approved):

Shubham-astro (approved):

PR #3493, opened 2026-08-13
PINF-1182 Fix eso pdb service configuration

pgvishnuram:
## Description

Fix eso pdb service configuration

Currently ESO Pod Disruption Budget does not respect global config and does require its own spec to disable and deviates from our existing charting model. This change fixes the underlying issue and correct the config

## Related Issues

https://linear.app/astronomer/issue/PINF-1182/make-eso-and-pgbouncer-poddisruptionbudgets-respect-global

## Testing

QA should able to validate below config no longer generates pdb for eso and pgbouncer component
```
global:
    podDisruptionBudgets:
       enabled: false 
```

## Merging


charts/pgbouncer/templates/pgbouncer-poddisruptionbudget.yaml
```diff
@@ -1,9 +1,10 @@
-###################################
+#####################################
 ## Pgbouncer Pod Disruption Budget ##
-###################################
+#####################################
+{{- if .Values.global.podDisruptionBudgets.enabled }}
 {{- if and .Values.global.pgbouncer.enabled .Values.podDisruptionBudget.enabled }}
 kind: PodDisruptionBudget
-apiVersion: policy/v1beta1
+apiVersion: {{ include "apiVersion.PodDisruptionBudget" . }}
```
rishkarajgi, line 7: The apiVersion fix here (`policy/v1beta1` → the `policy/v1` helper) is correct and load-bearing: `policy/v1beta1` was removed in k8s 1.25, so on the supported 1.31–1.35 range this PDB could never have applied when enabled. One gap: this pgbouncer change has no render test — the umbrella `TestPDB` glob (`pod-disruption-budget`) doesn't match this dashless filename, so nothing exercises it. Please add a render case mirroring the new external-secrets tests.

tests/chart_tests/test_external_secrets.py
```diff
@@ -367,3 +368,44 @@ def test_global_pod_labels_applied(self, kube_version):
             pod_labels = doc["spec"]["template"]["metadata"]["labels"]
             for key, value in custom_labels.items():
                 assert pod_labels.get(key) == value, f"Label {key}={value} missing from {doc['metadata']['name']}"
+
+
+@pytest.mark.parametrize("kube_version", supported_k8s_versions)
+class TestExternalSecretsPodDisruptionBudget:
```
rishkarajgi, line 374: Good coverage for the external-secrets PDB (default-off / opt-in / global-off). Since this PR also changes the **pgbouncer** PDB template (apiVersion + global gate) but nothing exercises it, consider adding an equivalent set of pgbouncer cases so the `v1beta1`→`v1` fix is actually tested.

rishkarajgi (commented):
**Verdict: PASS** — no blockers, 1 should-fix, a couple of nits.

**What this fixes (verified against the chart):** two PodDisruptionBudget templates (external-secrets, pgbouncer) were the only two of ~15 PDBs not gated on `global.podDisruptionBudgets.enabled` and not using the shared `apiVersion.PodDisruptionBudget` helper. This aligns both. Notably, pgbouncer's PDB was on `apiVersion: policy/v1beta1`, which was **removed in k8s 1.25** and exists on none of the supported versions (metadata pins 1.31–1.35) — so when enabled it could never apply. This is a real fix, not just cleanup. Selectors are untouched (no immutable-`spec.selector` upgrade conflict) and default-off behavior is preserved.

**Root cause of why this shipped unnoticed:** the umbrella `TestPDB` globs files whose path contains `pod-disruption-budget`, but both of these files use a dashless `poddisruptionbudget` name — so they were silently excluded from the `policy/v1` and global-gate assertions.

**Should-fix:** the pgbouncer change has **zero render coverage**. The new tests cover external-secrets, but the actual `v1beta1`→`v1` fix on pgbouncer rides an untested path. Add pgbouncer cases mirroring the new ESO ones (renders `policy/v1` when enabled; absent when `global.podDisruptionBudgets.enabled: false`).

**Nits:** renaming both templates to the repo's `*-pod-disruption-budget.yaml` convention would let the existing `TestPDB` auto-cover them and prevent recurrence (out of scope here). Pre-existing awareness note: ESO `minAvailable: 1` with a single replica blocks node drains when the PDB is opted on.

<sub>Generated by an automated reviewer and posted for the team's consideration; not a full substitute for human review.</sub>

charts/external-secrets/templates/poddisruptionbudget.yaml
```diff
@@ -2,8 +2,9 @@
 ##########################
 ## external-secrets PDB ##
 ##########################
+{{- if .Values.global.podDisruptionBudgets.enabled }}
```
Simpcyclassy, line 5: Good addition of the missing global.podDisruptionBudgets.enabled gate, confirmed this key already exists with a default of true in values.yaml so it is nil-safe.

charts/pgbouncer/templates/pgbouncer-poddisruptionbudget.yaml
Simpcyclassy, line 7: Nice catch fixing the hardcoded policy/v1beta1, which is removed in the k8s versions this chart supports, in favor of the shared apiVersion.PodDisruptionBudget helper.

tests/chart_tests/test_external_secrets.py
```diff
@@ -19,6 +19,7 @@
 ROLEBINDING_TEMPLATE = "charts/external-secrets/templates/rolebinding.yaml"
 CLUSTER_SECRET_STORE_TEMPLATE = "charts/external-secrets/templates/cluster-secret-store.yaml"
 SECRETS_BACKEND_CREDENTIALS_TEMPLATE = "charts/external-secrets/templates/secrets-backend-credentials.yaml"
+PDB_TEMPLATE = "charts/external-secrets/templates/poddisruptionbudget.yaml"
```
Simpcyclassy, line 22: This PR fixes the same global-gate bug in pgbouncer's PDB template too, but no equivalent test was added for it there, please add parametrized coverage matching this file's pattern.

Simpcyclassy (commented):
Correct fix, verified against a values.yaml snapshot close to this PR's timeframe that global.podDisruptionBudgets.enabled already exists as a nested key with default true, so both templates gain a nil-safe global gate consistent with the rest of the chart. Bonus fix: pgbouncer's PDB was hardcoded to apiVersion policy/v1beta1, which is removed in the 1.31-1.35 k8s versions this chart supports, and is now correctly resolved via the shared apiVersion.PodDisruptionBudget helper, that alone was likely already a broken render for pgbouncer PDBs before this PR. The gap is test coverage: the same global-gate fix is applied to both external-secrets and pgbouncer templates, but only external-secrets gets the new parametrized test cases, pgbouncer's fix relies solely on the manual QA step mentioned in the PR body. Would like a pgbouncer-equivalent test added before or shortly after merge, but the fix itself is sound so this does not need to block.

tests/chart_tests/test_pgbouncer.py
```diff
@@ -343,3 +343,19 @@ def test_pgbouncer_networkpolicy_not_created_when_pgbouncer_disabled(self, kube_
...
         assert len(pgbouncer_policies) == 0
+
+    def test_pgbouncer_pdb_overrides(self, kube_version):
+        """Test that the pgbouncer PodDisruptionBudget Not created when disabled globally with  global.podDisruptionBudget set to false."""
+        docs = render_chart(
+            kube_version=kube_version,
+            values={
+                "global": {
+                    "pgbouncer": {"enabled": True},
+                    "podDisruptionBudget": {
```
rishkarajgi: This sets the singular `global.podDisruptionBudget`, but the template gates on the plural `global.podDisruptionBudgets` — no template reads the singular key, so it's a no-op. The render returns 0 only because the pgbouncer subchart's inner `podDisruptionBudget.enabled` defaults to false, not because of any global toggle; flipping this to True would still render 0. To actually test the new gate, turn the subchart PDB on (`pgbouncer.podDisruptionBudget.enabled: True`) and set the plural `global.podDisruptionBudgets.enabled: False`, like the ESO `global-pdb-overrides` case does.

rishkarajgi (commented):
Following up on the earlier round — thanks for adding the pgbouncer coverage. One issue with the new test, and it's the kind that matters because it gives false confidence: `test_pgbouncer_pdb_overrides` doesn't actually exercise the gate it claims to.

The pgbouncer PDB template gates on the **plural** `.Values.global.podDisruptionBudgets.enabled`, but the test sets the **singular** `global.podDisruptionBudget.enabled: False`. I grepped the chart — no template reads the singular key anywhere, so the value the test sets is an inert no-op and the plural gate stays at its default (`true`). The render returns 0 docs only because the pgbouncer subchart's *inner* `podDisruptionBudget.enabled` defaults to `false` (`charts/pgbouncer/values.yaml`), which the test never turns on — so flipping the singular value to `True` would still yield 0 docs, and the assertion can't distinguish "gate works" from "gate absent." To actually test the new global gate, turn the subchart PDB on (`pgbouncer.podDisruptionBudget.enabled: True`) and set the plural `global.podDisruptionBudgets.enabled: False`, mirroring the ESO `global-pdb-overrides` case.

The ESO test is correct — all three of its cases use the plural key and genuinely exercise the gate. And the underlying template change is good and upgrade-safe: the new outer gate defaults on (an opt-out suppression toggle, so existing installs' PDB rendering doesn't regress), and the pgbouncer apiVersion fix `policy/v1beta1` -> the `policy/v1` helper is a real fix since v1beta1 was removed in k8s 1.25. So it's just the one test to make non-vacuous. Not a gate; needs CODEOWNERS approval.

danielhoherd (approved):
Fairly simple eso chart change. Good cleanup on the pgbouncer pdb apiVerison. Tests look good. Rishabh's comment about the plural PDB need to be addressed before merging.

PR #3494, opened 2026-08-13
PINF-1032 k3d nginx resolver ipv6 workaround

danielhoherd:
## Description

- **Disable IPv6 resolution in nginx's k3d resolver.** `bin/setup-cp-dp-k3d.py` now patches each plane's nginx ConfigMap (`server-snippet: resolver <kube-dns ClusterIP> valid=30s ipv6=off;`) right after each Helm install. On these single-stack-IPv4 clusters, CoreDNS's `kubernetes` plugin errors (SERVFAIL) instead of returning clean NODATA on AAAA lookups for cluster-local Service names, and nginx's own resolver — used for the `auth_request` subrequest behind every `auth-url`-annotated ingress — queries both A and AAAA, so it treated that error as a total resolution failure. That 500'd every ingress fronted by Houston auth, including Prometheus and Grafana. Added `--enable-nginx-resolver-ipv6` (default off) to opt back into AAAA if ever needed.
- **Bump `ap-kuiper-reloader` 0.2.12 → 0.3.0** (`charts/prometheus/values.yaml`). Picks up the PINF-1032 fix: `filesd-reloader` now retries its initial Postgres connection with backoff instead of exiting on the first blip, and no longer runs under `uvicorn --reload`, which was masking worker crashes from Kubernetes and left the pod permanently wedged after any transient DB outage.
- **`configs/enable-git-sync-private-ca.yaml`:** point `global.privateCaCerts` at the standard `mkcert-root-ca` secret name (was the stale `private-ca` name) and enable `gitSyncRelay` metrics.

## Related Issues

https://linear.app/astronomer/issue/PINF-1032

## Testing

I have been testing this in k3d for a few days while troubleshooting other issues and it's been working for me, whereas previously prometheus was throwing 500 and 502.

## Merging

master, release-2.1


Simpcyclassy (approved):
LGTM!

configs/enable-git-sync-private-ca.yaml
```diff
@@ -29,5 +34,5 @@ global:
   # is what config-time credential validation against the private-CA Forgejo host needs
   # (TC-CA-12) -- a distinct trust layer from the relay's deployment-namespace mount.
   privateCaCerts:
-    - private-ca
+    - mkcert-root-ca
```
sudarshanrampuria, line 37: @danielhoherd CI is broken because of this change. It's not able to find `mkcert-root-ca` secret.

https://app.circleci.com/pipelines/github/astronomer/astronomer/20401/workflows/e646402d-64dc-4a32-aefa-6d3444f756c8/jobs/300533

PR #3495, opened 2026-08-13
PINF-1163: Fix incorrect SSL config in Houston hook jobs

sudarshanrampuria:
### Problem
houston-upgrade-deployments-job.yaml and houston-cp-refresh-job.yaml gated SSLMODE on .Values.global.sslmodeEnabled — a key that is never set in any values.yaml, so the block never rendered. When it did, it hardcoded SSLMODE=require, ignoring the operator's chosen SSL mode.

### Fix
Both jobs now use the real, documented keys global.ssl.enabled + global.ssl.mode, matching the rest of the chart:

## Related Issues

https://linear.app/astronomer/issue/PINF-1163/fix-incorrect-ssl-config-in-houston-hook-jobs

## Testing

Add test cases for it.

## Merging

main


charts/astronomer/templates/houston/helm-hooks/houston-upgrade-deployments-job.yaml
```diff
@@ -104,9 +104,9 @@ spec:
             # we need to hardcode it to be able to upgrade from prisma1 to prisma2
             - name: SCHEMA_NAME
               value: "houston$default"
-            {{ if .Values.global.sslmodeEnabled }}
+            {{- if and .Values.global.ssl.enabled .Values.global.ssl.mode }}
             - name: SSLMODE
-              value: "require"
+              value: {{ .Values.global.ssl.mode }}
```
rishkarajgi, line 104: Behavior change on upgrade worth a release note: SSL-enabled installs now get `SSLMODE=<mode>` (default `prefer`) here, where the previous (broken) path hardcoded `require`. This is the **post-upgrade** hook, so it runs on every upgrade of existing installs. Blast radius looks contained — the env sits in the `houston-bootstrapper` init container and the deployments-hook fan-out logic is untouched — but please confirm the `prefer` default is intended vs the old `require`, and the PINF-1163 backport targets.

charts/astronomer/templates/houston/helm-hooks/houston-cp-refresh-job.yaml
```diff
@@ -123,9 +123,9 @@ spec:
             # we need to hardcode it to be able to upgrade from prisma1 to prisma2
             - name: SCHEMA_NAME
               value: "houston$default"
-            {{ if .Values.global.sslmodeEnabled }}
+            {{- if and .Values.global.ssl.enabled .Values.global.ssl.mode }}
             - name: SSLMODE
-              value: "require"
+              value: {{ .Values.global.ssl.mode }}
```
rishkarajgi, line 123: This now matches the standard `global.ssl.enabled` / `global.ssl.mode` pattern used by the rest of the Houston components (configmap, commander, db-migration) — good consistency fix. Same `require`→`mode` default change applies here as on the upgrade-deployments hook, so the release-note point covers both.

rishkarajgi (commented):
**Verdict: PASS** — no blockers, 1 should-fix, 2 nits.

**What this fixes (verified against the chart):** the `SSLMODE` env in `houston-cp-refresh-job.yaml` and `houston-upgrade-deployments-job.yaml` was gated on `global.sslmodeEnabled`, a key that exists nowhere in `values.yaml` — so on SSL-enabled installs these two hooks silently dropped `SSLMODE` while every other Houston component keys off `global.ssl.enabled`/`global.ssl.mode`. The fix aligns both to the standard pattern and is complete across all Houston hook jobs (db-migration was already correct). Good parametrized render-test coverage added.

**Should-fix:** behavior change on upgrade — SSL-enabled installs will now emit `SSLMODE=<mode>` (default `prefer`) where the old broken path hardcoded `require`. Blast radius looks contained (env is in the `houston-bootstrapper` init container; the upgrade-hook fan-out logic is untouched; sibling jobs already connect this way), but worth a **release note** and a check of the **PINF-1163 backport targets** (looks like an active `release-2.1` train).

**Nits:** `houston-db-migration-job.yaml` still uses the plainer `if .ssl.enabled` guard without the mode check — aligning it is a reasonable follow-up (the PR's `and` form is the more defensive one). The `{{ if`→`{{- if` whitespace change is cosmetic and correct.

charts/astronomer/templates/houston/helm-hooks/houston-cp-refresh-job.yaml
```diff
@@ -123,9 +123,9 @@ spec:
             # we need to hardcode it to be able to upgrade from prisma1 to prisma2
             - name: SCHEMA_NAME
               value: "houston$default"
-            {{ if .Values.global.sslmodeEnabled }}
+            {{- if and .Values.global.ssl.enabled .Values.global.ssl.mode }}
```
Simpcyclassy, line 121: Good fix, this now correctly gates on the real global.ssl.enabled/mode keys instead of the never-set sslmodeEnabled, confirmed both defaults exist in values.yaml so this is nil-safe.

charts/astronomer/templates/houston/helm-hooks/houston-upgrade-deployments-job.yaml
Simpcyclassy, line 104: Same fix applied consistently here; note for existing installs with ssl.enabled already true, SSLMODE will now actually get injected for the first time, worth a callout in the PR description as a behavior change on upgrade rather than a pure dead-code fix.

Simpcyclassy (commented):
Solid root-cause fix. Confirmed against the base values.yaml that global.ssl.enabled (default false) and global.ssl.mode (default "prefer") already exist and are used elsewhere, so the new `and .Values.global.ssl.enabled .Values.global.ssl.mode` guard is nil-safe and matches convention, unlike the old sslmodeEnabled key which genuinely never existed anywhere. For any customer who already has global.ssl.enabled set true, this fix changes runtime behavior on upgrade since SSLMODE was previously never injected due to the dead key; that is exactly the intended fix, but worth calling out explicitly in the PR description as a behavioral change on upgrade, not just a dead-code removal. Tests are added in the same parametrized class so they inherit the existing kube_version matrix, and cover default/enabled/disabled/empty-mode cases well. Comfortable with this merging as-is.

tests/chart_tests/test_astronomer_houston_hook_job.py
```diff
@@ -229,6 +229,31 @@ def test_upgrade_deployments_init_containers_disabled_with_custom_houston_secret
...
+        "ssl,sslmode_expected",
+        [
+            (None, None),
+            ({"enabled": True, "mode": "verify-full"}, "verify-full"),
+            ({"enabled": False, "mode": "require"}, None),
+            ({"enabled": True, "mode": ""}, None),
+        ],
+        ids=["default", "enabled-uses-mode", "disabled-mode-ignored", "enabled-empty-mode"],
+    )
+    def test_upgrade_deployments_job_sslmode(self, kube_version, ssl, sslmode_expected):
```
Simpcyclassy, line 242: Good coverage of the default/enabled/disabled/empty-mode matrix, and it correctly inherits the class-level kube_version parametrization rather than reinventing it.

charts/astronomer/templates/houston/helm-hooks/houston-upgrade-deployments-job.yaml
sudarshanrampuria: Mentioned in the task that we need to mention this change in release notes.

Shubham-astro (approved):

rishkarajgi (commented):
Re-reviewed against the earlier round. Both open items from before are resolved: `houston-db-migration-job.yaml` now uses the same defensive `and .Values.global.ssl.enabled .Values.global.ssl.mode` guard as the two originally-broken hook jobs, and the scope has grown to cover `houston-deployment.yaml` and `houston-worker-deployment.yaml` too -- those weren't using the wrong key, but they were missing the `and ...mode` check, so `ssl.enabled: true` with an empty `mode` would have rendered `SSLMODE=` instead of omitting it. Every Houston-adjacent component now gates the same way.

The new chart-wide test (`test_sslmode_follows_global_ssl_mode`) is a real step up -- it renders the whole chart with every feature on and asserts SSLMODE consistency across every container, not just the files this PR touches. Checked its one exemption (Grafana's `bootstrapper` container) against the actual chart: it really does read a separate `global.ssl.grafana.sslmode` key, so that's a legitimate carve-out, not a hack to make the test pass.

The one thing still open from the first round: the release-note / backport-targets ask. Sounds like that's being tracked in the PINF-1163 ticket itself rather than in this PR's description -- worth a final confirmation it's actually been done before this merges, since it won't block on this diff either way.

Also checked: #3000 (open, adds global labels support) touches all five of the same template files, but every one of its hunks sits near the top of the file (right after `apiVersion:`/`kind:`) while this PR's changes are 80-100+ lines deeper in the container spec -- no real overlap, safe to merge in either order.

Nothing new to add beyond confirming the above -- comfortable with this as-is.

PR #3496, opened 2026-08-14
PINF-1115: add git-sync-shared-volume live-cluster scenario

danielhoherd:
## Description

New live-cluster scenario (`git-sync-shared-volume`) validating the git-sync `repoShareMode=shared_volume` fix end to end — [PINF-1115](https://linear.app/astronomer/issue/PINF-1115) (Shared Volume "Stale file handle" on the `dags` mount) and [PINF-1190](https://linear.app/astronomer/issue/PINF-1190) (airflow-chart's pre-install/pre-upgrade init hook Job, [airflow-chart#592](https://github.com/astronomer/airflow-chart/pull/592)).

Auto-discovered by `bin/generate_circleci_config.py`, same approval-gating pattern as `auth-sidecar`/`deployment-lifecycle`/`git-sync-private-ca`.

**Tests:**
- `test_shared_volume_deployment_reaches_ready` / `test_shared_volume_dags_mount_has_no_stale_handle` — a fresh `git_sync` + `shared_volume` Airflow Deployment reaches ready, and its scheduler's `dags` mount is actually readable. Direct proof, not just inferred from readiness — this bug's whole symptom was `ls: cannot access 'dags': Stale file handle`.
- `test_shared_volume_init_hook_used_relay_image` — checks the init hook Job succeeded and used the `ap-git-sync-relay` image family, not upstream `git-sync` (the core PINF-1190 change — upstream `git-sync` has no equivalent of the relay's HTTPS+PAT credential helper). Checks the image *repository* only, deliberately not a specific tag: the tag gets bumped continuously for CVE fixes independently of this scenario, and no other scenario test asserts on an exact image tag.
- `test_pvc_created_on_transition_to_shared_volume` / `test_transition_deployment_dags_mount_has_no_stale_handle` — a deployment created plain and then **upgraded** (not created fresh) into `shared_volume` mode gets the PVC created via the pre-upgrade hook and reaches a working, non-stale `dags` mount. This is the actual PVC install-vs-upgrade lifecycle fix from the #592 critical-review round, using `snapshot_release_revisions`/`previous_revisions` so the wait doesn't short-circuit on the deployment's pre-transition ready state.

**Supporting changes:**
- Extended `upsert_deployment()` with `git_sync_repo_share_mode` → `dagDeployment.gitSyncRepoShareMode` (houston-api's actual GraphQL field/enum, confirmed against `src/schema/types.js` and the generated schema — lowercase `git_daemon`/`shared_volume` wire values, defaults to `git_daemon`).
- `configs/pin-git-sync-shared-volume-test-images.yaml` pins `astronomer.airflowChartVersion` to an airflow-chart#592 build (`1.18.8-build44928`) and `global.gitSyncRelay.images.gitSync.tag` to `0.5.0` (astronomer/git-sync-relay#71, includes `git-sync-once`/PINF-1188) — both verified published before pinning. **Note:** #592 has since picked up one more commit (`feba50ba`, restoring the PVC as a coexisting plain resource alongside its hook declaration) that this pin predates; not believed to affect this scenario's own assertions (that fix addresses a chart-version-upgrade prune hazard this scenario doesn't exercise), but the pin should be bumped before relying on this scenario to cover that specific case.
- `configs/git-sync-shared-volume-static-storage.yaml` points `shared_volume` mode's PVC at a `storageClassName` that deliberately doesn't correspond to a real `StorageClass` object — needed because this scenario's kind cluster has no ReadWriteMany-capable dynamic provisioner, but `shared_volume` mode's PVC hardcodes `accessModes: [ReadWriteMany]` (a real product requirement, not something to relax for the test). The test itself statically pre-binds a `hostPath` PersistentVolume with this same `storageClassName` before each release's PVC is created (`_create_static_shared_volume_pv`); static PV/PVC binding only needs matching `accessModes`/`capacity`/`storageClassName`, not a real `StorageClass` resource.
- Since Kubernetes doesn't apply `fsGroup`-based ownership fixups to `hostPath` volumes (unlike most real network/CSI-backed RWX volumes), that stand-in directory comes out root-owned and unwritable by git-sync-relay's non-root init container — `_fix_static_shared_volume_pv_hostpath_permissions` runs a throwaway root-privileged Job to `chmod` it first. Same failure class as PINF-156's real NFS/Filestore permission bug, but here it's an artifact of this test's hostPath stand-in, not a bug in #592 itself.

**Explicitly out of scope** for this scenario (documented in the test file's module docstring):
- The concurrent-sync race between the pre-upgrade hook Job and an already-running relay Deployment's own poll/webhook cycle — timing-dependent, not something this fixture shape can reliably force to interleave.
- `shared_volume` against a private-CA host — `git-sync-private-ca` already covers private-CA trust for `git_daemon` mode; a `shared_volume` + private-CA combination would need its own Forgejo-backed fixture, kept out to stay scoped to the shared_volume-specific fixes here.
- The SSH and HTTPS+PAT credential-helper wiring itself — both deployments here use `HTTPS_NONE` against a public repo, so the credential helper is render-tested (airflow-chart's own chart tests) but never exercised live by this scenario.

## Related Issues

- [PINF-1115](https://linear.app/astronomer/issue/PINF-1115)
- [PINF-1190](https://linear.app/astronomer/issue/PINF-1190)
- [airflow-chart#592](https://github.com/astronomer/airflow-chart/pull/592)
- [git-sync-relay#71](https://github.com/astronomer/git-sync-relay/pull/71) (merged)

## Testing

- New test file collects cleanly under `pytest --collect-only` (5 tests, no import errors)
- Regenerating `.circleci/config.yml` is idempotent (matches the `circle-config-yaml` pre-commit hook's own consistency check)
- Full repo pre-commit suite passes on all changed/new files
- Human review (rishkarajgi): PASS, with 1 should-fix + 1 nit — both addressed (`ff7bfddcc`): dropped an exact-tag assertion in favor of the repository-prefix check described above, and documented the HTTPS+PAT credential-helper gap explicitly as scope rather than silently expanding the PR.
- **Two live-cluster runs so far, both failed on scenario-infrastructure bugs (not the product fix under test), both fixed:**
  - Build 300592: a known JWKS cold-start race on the first `upsertDeployment` call (missing the same `only_rerun` retry guard `git-sync-private-ca` already carries), and the RWX-storage gap described above (fixed via the static hostPath PV route).
  - Build 302666: the hostPath `fsGroup` permission gap described above (fixed via the chmod Job, `ed6940b3b`).
- **Not yet confirmed green** — pushed and awaiting a fresh run of the approval-gated `scenario-git-sync-shared-volume` CI job.


rishkarajgi (commented):
Reviewed together with the fix it validates, [airflow-chart#592](https://github.com/astronomer/airflow-chart/pull/592).

**Verdict: PASS** — no blockers, 1 should-fix, a nit. This is a solid E2E scenario that genuinely exercises the [airflow-chart#592](https://github.com/astronomer/airflow-chart/pull/592) fix rather than just asserting setup.

**Credited (verified):**
- **The CircleCI change is a correct *regeneration*, not a hand-edit** — `.circleci/config.yml` is rendered from `config.yml.j2` by `generate_circleci_config.py`, which auto-globs `tests/functional/scenarios/*/test_profile.yaml`. This PR adds the new `test_profile.yaml`, the config delta matches the j2 loop output, and the `.j2` was correctly left untouched.
- **The scenario meaningfully proves the fix** — it execs into the scheduler pod and asserts the `dags` mount is readable and non-empty (direct proof the ESTALE / empty-dags race is gone), asserts the init Job succeeded on the pinned image, and exercises the pre-*upgrade* PVC-creation path via the transition fixture.
- **The cross-repo seam is self-guarding** — `test_shared_volume_init_hook_used_pinned_images` asserts the rendered Job image ends `:0.5.0` (matching [#592](https://github.com/astronomer/airflow-chart/pull/592)'s new chart default) and that the Job succeeded, so a wrong image-pin path fails loudly.

**Should-fix — live auth-coverage gap:** the scenario only runs the `https-none` / public-repo path. The SSH and HTTPS-PAT credential-helper wiring — which is the PINF-1190 reason the relay image is reused on the init Job — is render-tested only, never exercised end-to-end here. The docstring notes the private-CA gap but not this one; worth stating so the coverage boundary is explicit.

**Nit:** `EXPECTED_GIT_SYNC_RELAY_TAG = "0.5.0"` is duplicated across this test, the pin overlay, and the airflow-chart default — three copies to keep in lockstep on the next bump.

**Note (backport):** this PR bases on `release-2.1`; worth confirming [#592](https://github.com/astronomer/airflow-chart/pull/592) targets the matching train(s) and that the ticket stays open until the backports land.

tests/functional/scenarios/git-sync-shared-volume/test_git_sync_shared_volume.py
```diff
@@ -0,0 +1,275 @@
...
+# with no credentials and this scenario needs no private-CA/PAT fixture of its own.
+GIT_SYNC_REPOSITORY_URL = "https://github.com/astronomer/apc-test-dags-public"
+
+ADMIN_EMAIL = "pinf-1115-git-sync-shared-volume@astronomer.io"
+ADMIN_PASSWORD = "Astronomer%123"
+WORKSPACE_LABEL = "pinf-1115-git-sync-shared-volume"
+FRESH_DEPLOYMENT_LABEL = "pinf-1115-shared-volume-fresh"
+TRANSITION_DEPLOYMENT_LABEL = "pinf-1115-shared-volume-transition"
+
+EXPECTED_GIT_SYNC_RELAY_TAG = "0.5.0"
```
rishkarajgi: Nit: this tag is duplicated across the test, `configs/pin-git-sync-shared-volume-test-images.yaml`, and the airflow-chart `gitSync.tag` default — three copies to keep in lockstep on the next bump. Consider deriving it from the pin overlay so there's a single source of truth.

danielhoherd:
Addressed both in `ff7bfddcc`:

- **Nit (tag duplication):** fixed. `EXPECTED_GIT_SYNC_RELAY_TAG` is gone — the test now asserts the init Job's image starts with `quay.io/astronomer/ap-git-sync-relay:` (repository only), not a specific tag. Checked first: no other scenario test in this repo asserts on an exact image tag, so this now matches that convention instead of being the one exception that would've needed manual upkeep on every CVE-driven tag bump.
- **Should-fix (HTTPS-PAT/SSH coverage gap):** agreed this is real and worth being explicit about rather than fixed here — added it as a third documented scope boundary in the module docstring, alongside the concurrency and private-CA gaps. Both deployments in this scenario use `HTTPS_NONE` against a public repo, so the credential-helper wiring that's the actual reason the init Job reuses the relay image is render-tested (airflow-chart's own chart tests) but not exercised live here. Didn't expand this PR to add a live HTTPS-PAT deployment fixture — that's a real follow-up, not a quick fix, and felt like a separate scoping decision rather than something to fold in silently.

Re: the backport note — agreed, tracking separately; this PR bases on `release-2.1` matching #592's own target.

tests/functional/scenarios/git-sync-shared-volume/test_git_sync_shared_volume.py
danielhoherd: Fixed in ff7bfddcc — dropped the tag assertion entirely (and the constant) rather than deriving it from the overlay: no other scenario test in this repo asserts on an exact image tag, since it's expected to drift continuously for CVE fixes. Now asserts the image's repository prefix (quay.io/astronomer/ap-git-sync-relay:) instead, which still proves the init Job uses the relay image family (the actual thing worth checking) without needing to stay in lockstep with the tag.

tests/functional/scenarios/git-sync-shared-volume/test_git_sync_shared_volume.py
```diff
@@ -0,0 +1,282 @@
...
+    and no other scenario test asserts on an exact image tag."""
+    release_name = shared_volume_deployment["release_name"]
+    namespace = _deployment_namespace(release_name)
+    jobs = _k8s_batch_v1_client_module.list_namespaced_job(namespace, label_selector="component=git-sync-relay").items
+    init_jobs = [j for j in jobs if j.metadata.name.endswith("-git-sync-relay-init")]
+    assert init_jobs, f"No git-sync-relay-init Job found in {namespace}"
+    job = init_jobs[0]
+    assert job.status.succeeded == 1, f"init Job did not succeed: status={job.status}"
+    image = job.spec.template.spec.containers[0].image
+    assert image.startswith(GIT_SYNC_RELAY_IMAGE_PREFIX), (
```
Simpcyclassy, line 388: This only asserts the image repository prefix, not the pinned 0.5.0 tag, so a typo in configs/pin-git-sync-shared-volume-test-images.yaml's tag would not be caught here despite the PR description's claim that this test sanity-checks the pinned image.

tests/functional/scenarios/git-sync-shared-volume/test_git_sync_shared_volume.py
```diff
@@ -0,0 +1,282 @@
...
+            dag_deployment_type="git_sync",
+            repository_url=GIT_SYNC_REPOSITORY_URL,
+            auth_type="HTTPS_NONE",
+            git_sync_repo_share_mode="shared_volume",
+        )
+    except HoustonError:
+        dump_pod_logs(_k8s_core_v1_client_module, "component=houston")
+        dump_pod_logs(_k8s_core_v1_client_module, "component=commander")
+        raise
+    wait_for_release_ready(
```
Simpcyclassy, line 344: Good use of snapshot_release_revisions/previous_revisions here so wait_for_release_ready doesn't short-circuit on the deployment's pre-transition ready state, this is exactly what the install-vs-upgrade PVC lifecycle fix needs to actually exercise.

Simpcyclassy (commented):
High quality scenario addition, well documented and mirrors the existing auth-sidecar/deployment-lifecycle/git-sync-private-ca pattern including the CircleCI duplicate-job-name workaround comment. One accuracy gap between the PR description and the code: the body claims test_shared_volume_init_hook_used_pinned_images sanity-checks that the init hook ran the pinned ap-git-sync-relay:0.5.0 so a typo in the pin overlay would be caught, but the actual test (named test_shared_volume_init_hook_used_relay_image) only asserts the image starts with the ap-git-sync-relay repository prefix, not the specific pinned tag, and its own docstring says as much. That is a reasonable deliberate choice given tag churn from CVE bumps, but the PR description overstates what this test guarantees and should be corrected. The transition-deployment fixture correctly uses snapshot_release_revisions/previous_revisions to avoid short-circuiting on pre-transition ready state, which is the actual install-vs-upgrade lifecycle bug being tested, that part is well done. Comfortable merging once the description is aligned with what the test actually checks.

danielhoherd:
The first CI run (build 300592) failed all 5 tests. Two distinct root causes, both fixed in caa8d49ff:

**1. JWKS cold-start race** — `shared_volume_deployment`'s fixture hit the known `"13 INTERNAL: failed to validate token"` race on its first `upsertDeployment` call. `git-sync-private-ca` already guards its first-touching test against this; this scenario was missing the same `@pytest.mark.flaky(reruns=5, reruns_delay=5, only_rerun=[JWKS_COLD_START_ERROR])` decorator. Added to the first test touching each fixture.

**2. RWX storage gap (the real finding)** — `transition_deployment`'s init hook Job hit `DeadlineExceeded`, traced to `ProvisioningFailed: ... NodePath only supports ReadWriteOnce and ReadWriteOncePod`. `shared_volume` mode's PVC requires `ReadWriteMany` (a real product requirement, not something to relax — see `airflow-chart`'s own `values.yaml` comment on `gitSyncRelay.volumeSync.storageClassName`), but this scenario's kind cluster is single-node with no RWX-capable dynamic provisioner. No `activeDeadlineSeconds` value fixes this — the PVC can never bind on that StorageClass. Confirmed via grep this is the first scenario/chart test in either repo to need RWX at all.

Fix: each fixture now statically pre-binds a `hostPath` PersistentVolume for its own release (via `claimRef`) right after `upsertDeployment` returns and the release name is known, but before commander's helm install/upgrade reaches the init hook Job. Static PV/PVC binding only needs matching `accessModes`/`capacity`/`storageClassName` — the `storageClassName` doesn't need to resolve to a real `StorageClass` object — so this works without standing up a real network-filesystem provisioner in CI. Safe here because the cluster is single-node, so hostPath behaves like real shared storage would for every pod that mounts it.

New `configs/git-sync-shared-volume-static-storage.yaml` wires `astronomer.houston.config.deployments.deployMechanisms.gitSyncRelay.storageClassName` — the only way to set this at all, since it has no modeled Helm value in the `astronomer` chart itself, only houston's own raw `houston.config` passthrough.

Re-triggering the approval-gated scenario job now.

danielhoherd:
Correction to the last line above: I don't have CircleCI trigger/approval access from here. `scenario-git-sync-shared-volume-approved` is gated behind the `approve-scenarios` manual approval job on this branch's pipeline — needs a click in the CircleCI UI to actually run against a live cluster.

rishkarajgi:
please fix tests

danielhoherd:
Root cause for the CircleCI failure (build 302666, all 5 tests): Kubernetes does not apply `fsGroup`-based ownership fixups to `hostPath` volumes — a deliberate, long-standing k8s limitation, unlike most real network/CSI-backed RWX volumes. `_create_static_shared_volume_pv`'s stand-in hostPath directory (`DirectoryOrCreate`) therefore stays root:root 0755, and `gitSyncRelay.securityContext`'s `runAsUser: 50000` (non-root) init hook container can't write into it — `git-sync-once` exits 1 immediately (surfaced only as a bare `exit_code: 1`/`reason: Error`, no message).

Since the pre-install/pre-upgrade hook Job failing blocks the whole Helm release, both test timeouts traced back to this single cause:
- `shared_volume_deployment` (fresh install): "no Deployments/StatefulSets found yet" — nothing else in the release exists yet either, since Helm blocks *everything* behind a failing pre-install hook.
- `transition_deployment` (upgrade): showed the actual failed `git-sync-relay-init` pod directly (`exit_code: 1`), since its pre-transition resources were already up from the earlier plain-deployment create.

Same failure class as PINF-156's real NFS/Filestore permission bug, but here it's an artifact of this test's hostPath stand-in, not a real bug in #592 — a genuine RWX export doesn't enforce per-UID ownership the way a fresh hostPath dir does.

Fix (`ed6940b3b`): a throwaway root-privileged Job chmods the hostPath directory to `0777` before the static PV references it, mounting the identical path so it doesn't depend on docker/node-level access from the test process. Also fixed an idempotency gap in my own first pass — `transition_deployment`'s namespace already exists by this point (from its earlier plain-deployment create) while `shared_volume_deployment`'s doesn't yet, so the fix applies a `Namespace` doc in the same `kubectl apply` rather than `kubectl create namespace`, which would have failed on the former.

Verified: collects cleanly (5 tests), full pre-commit suite passes on the changed file. Not yet re-run live — re-triggering the approval-gated scenario job next.

tests/functional/scenarios/git-sync-shared-volume/test_git_sync_shared_volume.py
danielhoherd: Confirmed already fixed — the PR description now correctly names `test_shared_volume_init_hook_used_relay_image` and says it checks the image repository only, not the pinned tag. Looks like this fell through the cracks without a reply; nothing left to do here.

tests/functional/scenarios/git-sync-shared-volume/test_git_sync_shared_volume.py
danielhoherd: Thanks!

danielhoherd:
I've changed the merge base to master. I don't know why Claude ever set up the PR to merge into release-2.1.

danielhoherd:
Fixed the `scenario-auth-sidecar-approved` failure too (`9afb43ef1`) — unrelated to this PR's own git-sync-shared-volume work, but it was blocking CI on this branch and I'd already root-caused it while investigating.

Not a Bitnami-image issue (the actual image is `quay.io/astronomer/ap-postgresql`, Chainguard-based per PINF-26) and not new — `charts/postgresql`'s `init-postgresql-run` container (PINF-347) runs `runAsUser: 0` with `capabilities.add: [CHOWN, FOWNER]` by default, which always violates PSA `restricted`. It's only ever been caught here because `auth-sidecar` is the only scenario that applies `pod-security.kubernetes.io/enforce=restricted` to the platform namespace — every other scenario renders the identical violating pod spec and just never enforces the policy that would reject it. Confirmed against master's own history (4 of the last 8 commits also failed this same check, independent of this branch).

Fix: `postgresqlRunPermissions.enabled` already has a supported off switch for exactly this (PINF-347 built it for restricted-PodSecurity/OpenShift-style clusters — `global.openshift.enabled` already hard-disables it, relying on the image's own entrypoint to handle `/var/run/postgresql` without a k8s-side chown). Wired the same disable into `configs/psa-restricted.yaml`, the overlay this scenario already uses for every other PSA-incompatible component.

Verified via `helm template` with this scenario's exact overlay stack: main postgres container renders fully `restricted`-compliant (`runAsNonRoot`, no added capabilities, `seccompProfile` set) with zero init containers. `tests/chart_tests/test_postgres.py` (60 tests) unchanged/passing — no default behavior touched, this only applies inside the opt-in PSA-restricted config overlay.

Simpcyclassy (approved):
The follow-ups from my last review are resolved: the description now correctly names test_shared_volume_init_hook_used_relay_image and states it only checks the image repository, and @rishkarajgi's tag-duplication nit is gone too. One real but non-blocking gap: airflow-chart#592 is still unmerged a month later and has picked up a commit since this pin was cut, so it's confirmed stale. Approving; the transition test genuinely proves the fix and pytest collects clean.

PR #1511, opened 2022-05-09
Sync release-0.29 to master

andriisoldatenko:
## Description

Do not merge this PR until this description is replaced with relevant info.

## Related Issues

Do not merge this PR until this text is replaced with links to related issues.

## Testing

Do not merge this PR until this text is replaced with details about how these changes were tested.

## Merging

Do not merge this PR until it lists which release branches this PR should be merged / cherry-picked into.

PR #3497, opened 2026-08-17
PINF-347: Mount a writable /var/run/postgresql on the internal PostgreSQL statefulsets

Shubham-astro:
## Description

Upgrading a release with in-cluster PostgreSQL enabled (observed 0.37.7 → 2.0.1, including the 1.0.1 → 1.1.x path) fails on startup:
```
chmod: /var/run/postgresql: Read-only file system
PostgreSQL Database directory appears to contain a database; Skipping initialization
postgres: could not access the server configuration file "/bitnami/postgresql/data/postgresql.conf": No such file or directory
```
Root cause: the postgres container runs with readOnlyRootFilesystem: true (master statefulset.yaml, slave statefulset-slaves.yaml), but the only writable mounts are /tmp and the data PVC. On a fresh install the Chainguard image resolves unix_socket_directories to /tmp, so nothing touches /var/run/postgresql and it works by accident. On an upgrade, the PVC's persisted postgresql.conf (written by the old Bitnami image) still has unix_socket_directories = '/var/run/postgresql', so postgres tries to chmod that path and hits the read-only root filesystem.

There was also a half-wired leftover from a previous attempt at this fix (#3105, closed): a pg-conf emptyDir volume was declared in statefulset.yaml but never referenced by any volumeMounts entry, and the slave statefulset had no equivalent volume at all.

## Fix

Mount a dedicated writable emptyDir at /var/run/postgresql on the postgres container in both statefulsets:

- `charts/postgresql/templates/statefulset.yaml` : added the postgresql-run volume mount alongside the existing tmp mount, and renamed the previously-unreferenced pg-conf volume to postgresql-run (repurposing it rather than adding a second dangling volume).
- `charts/postgresql/templates/statefulset-slaves.yaml` : added both the postgresql-run volumeMount and the corresponding volume declaration, since the slave container carries the same readOnlyRootFilesystem: true and had no mount at all previously.

This works regardless of which value unix_socket_directories resolves to (stale /var/run/postgresql on upgraded PVCs, or /tmp on fresh installs) because it doesn't depend on reading or rewriting any customer-persisted config — it just makes the path postgres might try to chmod writable. An init-container sed rewrite of the persisted postgresql.conf was considered and rejected as more fragile and as mutating customer state.

## Related Issues

https://linear.app/astronomer/issue/PINF-347/add-readonlyrootfilesystem-volume-mounts-for-internal-postgresql

## Testing

- Updated test_postgresql_statefulset_defaults to expect the new postgresql-run mount on the master statefulset.
- Added test_postgresql_statefulset_slaves_volume_mounts to cover the same mount + volume on the slave statefulset.
- Tested upgrade on a cluster and confirmed that postgresql was Running:

Screenshot 2026-08-18 at 3 44 59 PM

Screenshot 2026-08-18 at 3 45 27 PM

## Merging

Main, 2.1


charts/postgresql/templates/statefulset.yaml
```diff
@@ -70,6 +70,32 @@ spec:
...
+        resources: {{- toYaml .Values.resources | nindent 10 }}
+        command:
+          - sh
+          - -c
+          - |
+            mkdir -p /var/run/postgresql
+            chown {{ .Values.securityContext.runAsUser }}:{{ .Values.securityContext.fsGroup }} /var/run/postgresql
+            chmod 0700 /var/run/postgresql
+        securityContext:
+          runAsUser: 0
```
Simpcyclassy: This init container is unconditional (no enable/disable flag) and hardcodes runAsUser: 0, unlike the existing init-chmod-data container which is gated behind volumePermissions.enabled and doesn't hardcode a UID. That forces every install, including OpenShift/restricted-PodSecurity clusters that currently avoid a root init container via that flag, to run one with no opt-out.

_Reviewed by Claude, verified by @Simpcyclassy_

charts/postgresql/templates/statefulset-slaves.yaml
```diff
@@ -62,8 +62,33 @@ spec:
...
+        resources: {{- toYaml .Values.resources | nindent 10 }}
+        command:
+          - sh
+          - -c
+          - |
+            mkdir -p /var/run/postgresql
+            chown {{ .Values.securityContext.runAsUser }}:{{ .Values.securityContext.fsGroup }} /var/run/postgresql
+            chmod 0700 /var/run/postgresql
+        securityContext:
+          runAsUser: 0
```
Simpcyclassy: Same concern as the master statefulset: this init container always renders and hardcodes runAsUser: 0 with no toggle to disable it for restricted-SCC/PodSecurity environments.

_Reviewed by Claude, verified by @Simpcyclassy_

charts/postgresql/templates/statefulset.yaml
```diff
@@ -277,7 +305,7 @@ spec:
       - emptyDir: {}
         name: tmp
       - emptyDir: {}
-        name: pg-conf
+        name: postgresql-run
```
Simpcyclassy, line 310: Good catch reusing the previously-dangling pg-conf volume rather than adding a second unused one; confirmed it wasn't referenced by any volumeMounts before this rename, so no upgrade impact.

_Reviewed by Claude, verified by @Simpcyclassy_

tests/chart_tests/test_postgres.py
```diff
@@ -32,10 +32,51 @@ def test_postgresql_statefulset_defaults(self, kube_version):
...
 
+        # The postgresql-run emptyDir is created root-owned by kubelet; fsGroup only
+        # changes its GID, not its UID, so the non-root postgres container can't chmod
+        # it itself. This init container hands ownership to the runAsUser before the
+        # main container starts..
+        init_containers = sts["spec"]["template"]["spec"]["initContainers"]
+        assert len(init_containers) == 1
+        init_container = init_containers[0]
+        assert init_container["name"] == "init-postgresql-run"
+        assert init_container["securityContext"]["runAsUser"] == 0
```
Simpcyclassy, line 73: No test renders this with global.openshift.enabled: true; worth adding one given the hardcoded runAsUser: 0 on the new init container bypasses the openshift.enabled guard used elsewhere in this template.

_Reviewed by Claude, verified by @Simpcyclassy_

Simpcyclassy (commented):
Solid root-cause fix with good test coverage for both statefulsets; filing as COMMENT since this PR is still a draft. The new init-postgresql-run container is unconditional and hardcodes runAsUser: 0 with no opt-out, unlike the existing init-chmod-data pattern gated by volumePermissions.enabled, which could force a root-running container onto OpenShift or restricted-PodSecurity customers who previously avoided one. Worth adding a toggle, or at least a rendering test with global.openshift.enabled: true, before this leaves draft.

_Reviewed by Claude, verified by @Simpcyclassy_

charts/postgresql/templates/statefulset.yaml
Shubham-astro: init-postgresql-run is now gated behind volumePermissions.enabled (same toggle init-chmod-data already uses), and runAsUser comes from volumePermissions.securityContext.runAsUser instead of being hardcoded to 0.

charts/postgresql/templates/statefulset-slaves.yaml
Shubham-astro: init-postgresql-run is now gated behind volumePermissions.enabled (same toggle init-chmod-data already uses), and runAsUser comes from volumePermissions.securityContext.runAsUser instead of being hardcoded to 0.

charts/postgresql/templates/statefulset.yaml
rishkarajgi: This init container forces `runAsUser: 0` unconditionally, but the rest of the chart (including the `fsGroup` block just above) gates UID-0 behind `not .Values.global.openshift.enabled`. On OpenShift's restricted SCC a UID-0 request is rejected at admission, so the postgres pod won't start there. If the internal DB is expected to run on OpenShift, this needs the same `not openshift` guard; if not, worth noting OpenShift is out of scope.

charts/postgresql/templates/statefulset-slaves.yaml
rishkarajgi: Same OpenShift concern on the replica init container — rendered unconditionally as `runAsUser: 0` with no `not .Values.global.openshift.enabled` guard, so it'll hit the restricted-SCC admission rejection there too. Keep this in lockstep with whatever the primary does.

charts/postgresql/templates/statefulset.yaml
```diff
@@ -70,6 +70,32 @@ spec:
...
+        image: {{ template "postgresql.image" . }}
+        imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+        resources: {{- toYaml .Values.resources | nindent 10 }}
+        command:
+          - sh
+          - -c
+          - |
+            mkdir -p /var/run/postgresql
+            chown {{ .Values.securityContext.runAsUser }}:{{ .Values.securityContext.fsGroup }} /var/run/postgresql
+            chmod 0700 /var/run/postgresql
```
rishkarajgi, line 85: `chmod 0700` plus a chown to `runAsUser:fsGroup` (1001 by default) means that on OpenShift, where the main container runs as an SCC-assigned arbitrary UID rather than 1001, it can't write this socket dir — reintroducing the exact original failure even if the UID-0 init container were somehow admitted. Group-ownership via `fsGroup` + group-writable perms would survive the UID reassignment.

rishkarajgi (commented):
Nice targeted fix — a writable emptyDir for the runtime dir is the right minimal approach rather than loosening `readOnlyRootFilesystem`, and I checked the data-safety angles carefully since this is a StatefulSet change on the internal DB. One thing to sort for OpenShift before merge, otherwise it looks good.

Data-safety checks (all pass):
- The new `postgresql-run` volume is an ephemeral `emptyDir: {}`, not a PVC, mounted at exactly `/var/run/postgresql` — it doesn't overlap `/bitnami/postgresql`, so no shadowing of the data dir.
- The primary reuses the pre-existing `pg-conf` emptyDir (which was declared but had no `volumeMount` referencing it anywhere — an orphan), so renaming it to `postgresql-run` removes no live mount.
- The `data` PVC volume/mount is untouched, so the `helm upgrade` pod re-roll doesn't touch persisted data. `readOnlyRootFilesystem: true` is preserved on the main container — the fix is a targeted writable mount, not a loosening of RO-root. 
- Both statefulsets have render tests for the new mount.

Upgrade behavior (stating, not blocking): the pod-template change rolls the postgres primary and replica pods on upgrade — brief DB downtime, expected since it's fixing a hard-broken upgrade, and the data PVC is untouched so no data loss. Harmless/idempotent for already-healthy non-OpenShift installs beyond that one restart.

**Should-fix — the new `init-postgresql-run` init container isn't OpenShift-guarded and will likely break there.** It renders unconditionally with `runAsUser: 0` / `runAsNonRoot: false`, then `chmod 0700` a dir chowned to `securityContext.runAsUser:fsGroup` (1001 by default). Everywhere else the chart gates UID-0 / `fsGroup` behind `not .Values.global.openshift.enabled` (the `fsGroup` block right above this one does exactly that), matching the hardening convention of not hardcoding a UID and letting the SCC assign one. On OpenShift's restricted SCC two things go wrong: the `runAsUser: 0` request is rejected at admission so the pod won't start; and even if it were admitted, the main container runs as an SCC-assigned arbitrary UID (not 1001), so a `chmod 0700` (owner-only) dir owned by 1001 isn't writable by it — which reintroduces the original failure. Note the pre-existing `init-chmod-data` is also UID 0 but gated behind `volumePermissions.enabled` (default false), whereas this one is always-on, so it changes default-install OpenShift behavior. If OpenShift is in scope for the internal DB: guard the UID-0/chown path behind `not openshift` (or use `fsGroup` group-ownership + group-writable perms instead of `chmod 0700`) and add the manual OpenShift-verification note. If it's out of scope, worth stating that in the PR so the ungated UID-0 init container is a recorded non-concern.

Nit: primary and replica reach the same end state by slightly different means — the primary renames the orphan `pg-conf` volume while the slaves add a fresh one, and the two `volumeMount` entries order their keys differently. Aligning them would make the primary/replica parity obvious at a glance.

Open question I couldn't answer from the chart alone: the PR shows three symptoms (chmod read-only, skipping init, `postgresql.conf` missing). My read is the bitnami entrypoint runs under `set -e`, so the early chmod failure aborts before it writes `postgresql.conf` — making all three one cascade that the writable mount resolves — but that entrypoint isn't in this repo, so worth confirming the conf-missing symptom actually disappears on a real `0.37.7 → 2.0.1` upgrade.

charts/postgresql/templates/statefulset-slaves.yaml
Shubham-astro: Already addressed — runAsUser is now wrapped in {{- if not .Values.global.openshift.enabled }} on both statefulsets, matching the guard the main container already uses in this template. The value itself is also no longer hardcoded: it comes from volumePermissions.securityContext.runAsUser (default 0), and runAsNonRoot is derived from that value instead of being a separate hardcoded flag, so the two can't drift out of sync

charts/postgresql/templates/statefulset.yaml
Shubham-astro: Already addressed — runAsUser is now wrapped in {{- if not .Values.global.openshift.enabled }} on both statefulsets, matching the guard the main container already uses in this template. The value itself is also no longer hardcoded: it comes from volumePermissions.securityContext.runAsUser (default 0), and runAsNonRoot is derived from that value instead of being a separate hardcoded flag, so the two can't drift out of sync

charts/postgresql/templates/statefulset.yaml
Shubham-astro: done

charts/postgresql/templates/statefulset-slaves.yaml
```diff
@@ -62,8 +62,47 @@ spec:
         fsGroup: {{ .Values.securityContext.fsGroup }}
       {{- end }}
       serviceAccountName: {{ template "postgresql.serviceAccountName" . }}
-      {{- if and .Values.volumePermissions.enabled .Values.persistence.enabled }}
       initContainers:
+      {{- if .Values.volumePermissions.enabled }}
```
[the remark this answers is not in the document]
Shubham-astro: added a new flag postgresqlRunPermissions which is true by default

charts/postgresql/templates/statefulset.yaml
```diff
@@ -70,6 +70,46 @@ spec:
         fsGroup: {{ .Values.securityContext.fsGroup }}
       {{- end }}
       serviceAccountName: {{ template "postgresql.serviceAccountName" . }}
+      initContainers:
+      {{- if .Values.volumePermissions.enabled }}
```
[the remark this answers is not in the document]
Shubham-astro: added a new flag postgresqlRunPermissions which is true by default

charts/postgresql/templates/_helpers.tpl
```diff
@@ -156,6 +156,22 @@ Also, we can't use a single if because lazy evaluation is not an option
...
+postgresqlRunPermissions.enabled is explicitly set to: this container needs
+capabilities.add [CHOWN, FOWNER], which OpenShift's default restricted SCC does not
+grant, so enabling it there risks the pod failing SCC admission entirely rather than
+just falling back to the original crash-loop bug. Clusters running a custom SCC that
+does permit this can't override the auto-disable via values - the real fix for
+OpenShift is the image-level entrypoint change (see the ap-vendor/postgresql
+companion fix), which needs no k8s-side chown at all. See PINF-347.
+*/}}
+{{- define "postgresql.postgresqlRunPermissions.enabled" -}}
+{{- and .Values.postgresqlRunPermissions.enabled (not .Values.global.openshift.enabled) -}}
```
rishkarajgi, line 172: This resolves the earlier finding cleanly: the init container is now hard-disabled on OpenShift (`not .Values.global.openshift.enabled`) — where the restricted SCC wouldn't grant CHOWN/FOWNER and the pod would fail admission — while staying opt-out elsewhere, and both statefulsets gate on it. Good fix.

rishkarajgi (approved):
Approving — the OpenShift concern from the earlier round is fully resolved, and I re-verified the data-safety at the current head.

The OpenShift fix (verified): the new `postgresql.postgresqlRunPermissions.enabled` helper is `and .Values.postgresqlRunPermissions.enabled (not .Values.global.openshift.enabled)`, and both the primary and slave StatefulSets gate the `init-postgresql-run` container behind it — so it's hard-disabled on OpenShift (whose restricted SCC won't grant CHOWN/FOWNER) while staying opt-out elsewhere. `runAsUser` is now configurable (default 0) with `runAsNonRoot` derived, and there are tests asserting non-render on OpenShift on both statefulsets. That's exactly the right shape.

Data-safety re-verified: `postgresql-run` is an ephemeral `emptyDir` mounted at exactly `/var/run/postgresql`, not a PVC, no overlap with `/bitnami/postgresql`; the primary's rename of the orphan `pg-conf` volume (declared but never mounted at base) leaves no dangling mount; slaves mirror it correctly. The `data` PVC is untouched, so the upgrade re-roll doesn't risk data. RO-root preserved on the main container.

Two things to note (not blocking the approval):
- **Default-on root-capable init container.** `postgresqlRunPermissions.enabled: true` renders a root-by-default init container on every non-OpenShift install. It's gated, OpenShift-safe, and a crash-loop remediation rather than a new feature, so default-on is defensible — but a **security review** on that posture is worth doing before merge (I'm flagging it, not gating on it).
- **Seam with ap-vendor #1511.** The `unix_socket_directories`/conf-repair half lives in the image (PR #1511, same ticket), and the chart still pins the `-2` image tag — so the full crash fix needs #1511 (and its image-tag bump) to ship in lockstep. #3497 on its own is a correct, self-contained change (mount the socket dir), so it's fine to approve independently. Still needs the usual CODEOWNERS approval.

charts/postgresql/templates/statefulset-slaves.yaml
```diff
@@ -62,8 +62,39 @@ spec:
         fsGroup: {{ .Values.securityContext.fsGroup }}
       {{- end }}
       serviceAccountName: {{ template "postgresql.serviceAccountName" . }}
-      {{- if and .Values.volumePermissions.enabled .Values.persistence.enabled }}
       initContainers:
+      {{/* Hard-disabled on OpenShift regardless of postgresqlRunPermissions.enabled - see
+           the "postgresql.postgresqlRunPermissions.enabled" helper in _helpers.tpl for why.
+           Because of that, everything below can assume global.openshift.enabled is false
+           whenever it actually renders, so it doesn't need its own OpenShift branching. */}}
```
danielhoherd: This comment is effectively a duplicate of what is in the _helpers.tpl file so we can get rid or it.
```suggestion
```

charts/postgresql/templates/statefulset.yaml
```diff
@@ -70,6 +70,38 @@ spec:
         fsGroup: {{ .Values.securityContext.fsGroup }}
       {{- end }}
       serviceAccountName: {{ template "postgresql.serviceAccountName" . }}
+      initContainers:
+      {{/* Hard-disabled on OpenShift regardless of postgresqlRunPermissions.enabled - see
+           the "postgresql.postgresqlRunPermissions.enabled" helper in _helpers.tpl for why.
+           Because of that, everything below can assume global.openshift.enabled is false
+           whenever it actually renders, so it doesn't need its own OpenShift branching. */}}
```
danielhoherd: Same duplicate comment
```suggestion
```

charts/postgresql/templates/statefulset.yaml
danielhoherd: Shouldn't 0700 be 0760 to add g+rw on /var/run/postgresql ? or 0720 if we really only want g+w

charts/postgresql/templates/statefulset-slaves.yaml
```diff
@@ -62,8 +62,39 @@ spec:
...
+        image: {{ template "postgresql.image" . }}
+        imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+        resources: {{- toYaml .Values.resources | nindent 10 }}
+        command:
+          - sh
+          - -c
+          - |
+            mkdir -p /var/run/postgresql
+            chown {{ .Values.securityContext.runAsUser }}:{{ .Values.securityContext.fsGroup }} /var/run/postgresql
+            chmod 0700 /var/run/postgresql
```
danielhoherd, line 77: Same mode concern here.

Shubham-astro:
Waiting for approval of https://github.com/astronomer/ap-vendor/pull/1511 approach of image release.

charts/postgresql/templates/statefulset.yaml
Shubham-astro: 0700 is intentional given both containers share the same UID.

PR #3498, opened 2026-08-17
PINF-1202 Unify mkcert-root-ca / private-ca secret naming across dev topologies

danielhoherd:
## Description

`topology=unified`/KIND (`bin/setup-kind.py`) and the k3d CP/DP topology (`bin/setup-cp-dp-k3d.py`) each independently built their own "mint a locally-trusted CA via mkcert and register it as a Secret" step, under two different names: `private-ca` (KIND) and `mkcert-root-ca` (k3d). No shared code, no shared convention.

That let PR #3494 accidentally swap `private-ca` for `mkcert-root-ca` in `configs/enable-git-sync-private-ca.yaml`, which doesn't exist as a secret under `topology=unified`. Every pod mounting `global.privateCaCerts` (`commander`, `houston`, `registry`, `alertmanager`, `prometheus`, `vector`) hit `FailedMount` and hung until the install timed out. Sudarshan reverted the name in `11c238ebd`.

Diagnosing that turned up two more bugs in the KIND-side fixture:
- `bin/setup-kind.py` created the `private-ca` secret via `--from-file={path}` with no `cert.pem=` key prefix, so the data key defaulted to the file's basename instead of the `cert.pem` key every consumer's `subPath` mount requires.
- The fixture's content was a leaf certificate for a fake hostname (`server.example.org`), not a CA at all, so it couldn't have been doing what `configs/local-dev.yaml`'s own comment claimed ("or the platform stops trusting its own astronomer-tls CA").

This PR collapses both topologies onto one name, one mechanism:

- `bin/setup-kind.py` now resolves the CA via `_mkcert_caroot()`/`_mkcert_path()` (`bin/k3d_setup_shared.py`), the same helper the k3d topology already used, instead of minting its own leaf-cert fixture.
- Removed `bin/certs.py`'s `create_astronomer_private_ca_certificates()` and the `generate-private-ca` subcommand (and its call in `bin/reset-local-dev`) as dead code now that KIND doesn't mint its own cert.
- Secret is `mkcert-root-ca` in both topologies now. Updated `configs/local-dev.yaml`, `configs/enable-git-sync-private-ca.yaml`, `configs/enable-git-sync-deploys.yaml` with comments on the history so this doesn't regrow a second name.

## Related Issues

https://linear.app/astronomer/issue/PINF-1202

## Testing

`bin/certs.py`, `bin/setup-kind.py`, `bin/k3d_setup_shared.py` parse; `bin/reset-local-dev` passes `bash -n`; the three edited YAML configs parse. `uv run pytest tests/chart_tests/test_privateca.py tests/chart_tests/test_containerd_privateca.py tests/chart_tests/test_security_context_override.py tests/chart_tests/test_default_chart.py` — 295 passed, 48 skipped, unaffected as expected since none of this touches chart templates. Haven't re-run the `git-sync-private-ca` CI scenario itself yet against this branch.

I have also tested this in my private-ca git-sync https+pat setup locally and it works.

### DIFF
--- a/bin/setup-kind.py
+++ b/bin/setup-kind.py
@@ -11,10 +11,10 @@ from pathlib import Path
 
 import yaml
 from certs import (
-    astronomer_private_ca_cert_file,
     astronomer_tls_cert_file,
     astronomer_tls_key_file,
 )
+from k3d_setup_shared import _mkcert_caroot, _mkcert_path
 from kubernetes import client, config
 from kubernetes.client.exceptions import ApiException
 
@@ -280,16 +280,22 @@ def setup_common_cluster_configs() -> None:
     """
     create_namespace()
     create_astronomer_tls_secret()
-    create_private_ca_secret()
+    create_mkcert_root_ca_secret()
 
 
-def create_private_ca_secret() -> None:
+def create_mkcert_root_ca_secret() -> None:
     """
-    Create the private-ca secret in the KIND cluster using self-signed certificates.
+    Create the mkcert-root-ca secret in the KIND cluster: mkcert's own root CA (the same
+    CA that signs astronomer-tls, via create_astronomer_tls_certificates()'s `mkcert -install`),
+    so the platform trusts its own dev TLS chain -- the identical role the k3d CP/DP topology's
+    `--mkcert-root-ca-secret-name` secret plays via _mkcert_caroot() in k3d_setup_shared.py.
+    Both topologies now share that one helper and one secret name/content; there is no separate
+    KIND-only "private CA" concept.
 
     Raises:
         RuntimeError: If secret creation fails.
     """
+    root_ca = _mkcert_caroot(_mkcert_path())
 
     cmd = [
         KUBECTL_EXE,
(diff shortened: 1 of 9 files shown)


bin/setup-kind.py
```diff
@@ -298,14 +304,19 @@ def create_private_ca_secret() -> None:
...
         "generic",
-        "private-ca",
-        f"--from-file={astronomer_private_ca_cert_file}",
+        "mkcert-root-ca",
+        # Every consumer of global.privateCaCerts (custom_ca_volume_mounts / custom_ca_volumes
+        # in charts/astronomer/templates/_helpers.yaml, plus the copies in the alertmanager,
+        # prometheus, and vector subcharts) mounts each secret with `subPath: cert.pem`, per the
+        # contract documented in configs/local-dev.yaml ("Each secret must have one data entry
+        # 'cert.pem'"), so the key= prefix here is required, not optional.
+        f"--from-file=cert.pem={root_ca}",
```
Simpcyclassy, line 313: Good fix, this restores the cert.pem key prefix that subPath mounts require, which is exactly the missing piece that broke the old private-ca fixture.

_Reviewed by Claude, verified by @Simpcyclassy_

Simpcyclassy (commented):
Clean, well-documented fix that collapses two dev-topology secret fixtures into one and removes now-dead code in bin/certs.py. None of this touches chart templates or values.yaml, so there is no in-place-upgrade or plane-mode-gating risk, only local dev/CI tooling. The PR body notes the git-sync-private-ca CI scenario, the exact one that broke before, hasn't been re-run against this branch yet, worth confirming before merge.

_Reviewed by Claude, verified by @Simpcyclassy_

Simpcyclassy (approved):
LGTM

PR #3499, opened 2026-08-18
Laminar POC script

karankhanchandani:
## Description

## Related Issues

## Testing

## Merging

### DIFF
--- a/.github/workflows/pr_checks.yml
+++ b/.github/workflows/pr_checks.yml
@@ -36,7 +36,7 @@ jobs:
             ((errors += 1))
           fi
 
-          if [[ "${KTLO}" != "true" ]] && [[ "${AUTOMATED_PR}" != "true" ]] && [[ "${PR_BODY}" != *"astronomer/issue"* && "${PR_BODY}" != *"PINF-"* && "${PR_BODY}" != *"APC-"* && "${PR_BODY}" != *"PLA-"* ]] ; then
+          if [[ "${KTLO}" != "true" ]] && [[ "${AUTOMATED_PR}" != "true" ]] && [[ "${PR_BODY}" != *"astronomer/issue"* && "${PR_BODY}" != *"PINF-"* && "${PR_BODY}" != *"APC-"* && "${PR_BODY}" != *"PLA-"* && "${PR_BODY}" != *"PLX-"* ]] ; then
             echo "PR description does not contain an issue link"
             ((errors += 1))
           fi
--- a/bin/setup-laminar-poc.py
+++ b/bin/setup-laminar-poc.py
@@ -1,0 +1,570 @@
+#!/usr/bin/env python3
+"""
+POC: install the Laminar hypervisor onto an APC data plane created by `bin/setup-cp-dp-k3d.py`.
+
+This is an investigation tool for the "Laminar Hypervisor Integration for APC" design work
+(Linear project "Laminar integration (L)"). It is NOT a supported install path, and nothing
+here is a proposal for how the astronomer chart should eventually ship Laminar — it exists to
+let us run the real thing against a real APC data plane and find out what breaks.
+
+What it does, against a data-plane context you name explicitly:
+
+  install-laminar
+    - installs KEDA (the astronomer chart does not ship it; worker-queue autoscaling needs it)
+    - builds the laminar image from a local checkout of github.com/astronomer/laminar and
+      imports it straight into the k3d cluster (no registry, no ACR token to expire)
+    - creates the `laminar` namespace
+    - creates Laminar's own database on the data plane's postgres, plus the `laminar-connection`
+      secret it reads via LAMINAR_DATABASE_URL
+    - renders the laminar manifests (base + hypervisor, optionally apiserver) and applies them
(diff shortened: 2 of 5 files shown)


charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -121,12 +121,16 @@ data:
         dagOnlyDeployment:
           enabled: {{ .Values.global.deployMechanisms.dagOnlyDeployment.enabled }}
 
+      {{- $grafanaLinkEnabled := true }}
+      {{- if and (hasKey .Values.global "grafana") (kindIs "map" .Values.global.grafana) (hasKey .Values.global.grafana "enabled") }}
```
Simpcyclassy, line 125: This makes metricsReporting.grafana.enabled follow a new global.grafana.enabled toggle, but that key is never added to values.yaml (charts/astronomer/values.yaml has no `global.grafana` block). The hasKey guard keeps existing installs safe (defaults to true, same as before), but the toggle is undiscoverable to chart consumers without reading this template. Please add `global.grafana.enabled: true` to values.yaml with a comment, or drop this if it's not actually needed for the Laminar POC.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -121,12 +121,16 @@ data:
...
+      {{- $grafanaLinkEnabled = .Values.global.grafana.enabled }}
+      {{- end }}
       metricsReporting:
         taskUsageMetrics:
           enabled: {{ .Values.global.metricsReporting.taskUsageMetrics.enabled }}
           reportNumberOfDays: {{ .Values.houston.cleanupTaskUsageData.olderThan}}
         grafana:
-          enabled: true
+          enabled: {{ $grafanaLinkEnabled }}
```
Simpcyclassy, line 133: Also worth noting: this grafana-toggle change has nothing to do with the Laminar POC script and isn't mentioned anywhere in the (placeholder) PR description. If it's intentional, it should ship as its own PR with a real description; if it's a rebase/merge artifact, it should be dropped.

tests/chart_tests/test_houston_configmap.py
```diff
@@ -1084,14 +1084,24 @@ def test_houston_configmap_features_elasticsearch_custom_logging():
 
 
 def test_houston_configmap_features_grafana_defaults():
-    """Validate that metricsReporting.grafana.enabled is always true."""
+    """Validate that metricsReporting.grafana.enabled defaults to true."""
     docs = render_chart(
         show_only=["charts/astronomer/templates/houston/houston-configmap.yaml"],
     )
     prod = yaml.safe_load(docs[0]["data"]["production.yaml"])
     assert prod["deployments"]["metricsReporting"]["grafana"]["enabled"] is True
```
Simpcyclassy, line 1094: Good coverage for the new false-path, and updating the sibling docstring (no longer 'always true') is a nice touch.

bin/setup-laminar-poc.py
```diff
@@ -0,0 +1,784 @@
+#!/usr/bin/env python3
```
Simpcyclassy, line 1: The script itself is clearly labeled as an investigation tool (not a supported install path), documents its assumptions and known gaps (e.g. the Astro-ID-labels workaround) well, and follows the existing bin/*-k3d.py convention using k3d_setup_shared. No concerns with the script's content, only with it being bundled into the same PR as the unrelated production template/CI changes above.

.github/workflows/pr_checks.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             ((errors += 1))
           fi
 
-          if [[ "${KTLO}" != "true" ]] && [[ "${AUTOMATED_PR}" != "true" ]] && [[ "${PR_BODY}" != *"astronomer/issue"* && "${PR_BODY}" != *"PINF-"* && "${PR_BODY}" != *"APC-"* && "${PR_BODY}" != *"PLA-"* ]] ; then
+          if [[ "${KTLO}" != "true" ]] && [[ "${AUTOMATED_PR}" != "true" ]] && [[ "${PR_BODY}" != *"astronomer/issue"* && "${PR_BODY}" != *"PINF-"* && "${PR_BODY}" != *"APC-"* && "${PR_BODY}" != *"PLA-"* && "${PR_BODY}" != *"PLX-"* ]] ; then
```
Simpcyclassy, line 39: Minor: adding the PLX- issue prefix here is also unrelated to "Laminar POC script" and isn't called out in the description. Low risk on its own, but same scope-bundling concern as the grafana change above.

Simpcyclassy (commented):
Draft PR whose description is 100% unfilled placeholder text (four "Do not merge" markers), so this cannot be un-drafted as-is regardless of the code itself. The diff also bundles three unrelated things under a "Laminar POC script" title: a 784-line investigation script, an undocumented grafana toggle in houston-configmap, and an unrelated CI issue-prefix change. Marking COMMENT only because of draft status; the real ask before this leaves draft is filling in the description and considering splitting the grafana/CI changes into their own PR.

PR #3507, opened 2026-08-19
PINF-898 enable global ingress flag

pgvishnuram:
## Description

Today APC renders Ingress objects unconditionally. There is no supported way to install the platform without default generated ingress objects them. This change introduces `global.ingress.enabled `  to control the creation of ingress object across the APC platform.

## Related Issues

https://linear.app/astronomer/issue/PINF-898/helm-chart-nginx-ingress-objects-rendered-when-nginxenabled-false

## Testing

**Unit tests**: Cover ingress objects not being created when `global.ingress.enabled` is set to `false`.

**Manual/QA verification**:
1. Deploy with `global.ingress.enabled: false`.
2. Confirm no ingress objects exist in the platform namespace (expect `0`):
   - With `jq`: `kubectl get ingress -n <platform-namespace> -o json | jq '.items | length'`
   - Without `jq`: `kubectl get ingress -n <platform-namespace> --no-headers | wc -l`
## Merging

merge to master and release-2.x


rishkarajgi (commented):
Verdict: PASS — this is a clean gated rollout and I checked the upgrade blast-radius concretely rather than by inspection. It adds a master kill-switch: all 12 platform Ingress templates are wrapped in `{{- if eq (include "global.ingress.enabled" .) "true" }}`, backed by a new helper and `global.ingress.enabled: true` (default-on) in values. Materializing base and head and diffing the rendered Ingress manifests at default values across `plane.mode = unified/control/data` gives byte-identical output in all three — no annotation/class/host/TLS change — so on an existing install `helm upgrade` is a no-op for ingress; the flag only does anything when explicitly set to false, which suppresses every Ingress (verified the parent helper is reachable from the subcharts). The helper uses `hasKey` rather than `default`, so a legitimate `enabled: false` isn't coerced back to true — back-compat handled correctly. Test added for the disabled case; existing test covers the default. Nit only: the helper's `else -> true` fallback (fires only if a user nulls `global.ingress`) isn't directly unit-tested. Not a gate; needs CODEOWNERS approval.

danielhoherd (approved):
Looks good. This is a straightforward k8s object toggle, nothing difficult about it. I'm happy to see the tests.

PR #3508, opened 2026-08-19
PINF-578

Shubham-astro:
## Description

Curator fixes

## Related Issues

## Testing

- TBD

## Merging

- TBD

### DIFF
--- a/charts/elasticsearch/templates/_helpers.tpl
+++ b/charts/elasticsearch/templates/_helpers.tpl
@@ -148,11 +148,112 @@ imagePullSecrets:
 {{- end -}}
 {{- end -}}
 
+{{/*
+The timestring curator uses for its age filter, unquoted.
+
+Two sources of truth, in priority order:
+  1. global.logging.loggingSidecar.indexPattern -- the customer-facing knob, which also
+     drives the index name the vector logging sidecar writes to (via Houston).
+  2. curator.age.timestring -- the subchart default, used when the sidecar is off.
+
+Kept separate from "curator.indexPattern" so the granularity helpers below can inspect
+the raw value without having to strip quotes.
+*/}}
+{{- define "curator.age.timestringRaw" -}}
+{{- if and .Values.global.logging.loggingSidecar.enabled .Values.global.logging.loggingSidecar.indexPattern -}}
+{{- .Values.global.logging.loggingSidecar.indexPattern -}}
+{{- else -}}
+{{- .Values.curator.age.timestring -}}
+{{- end -}}
+{{- end -}}
+
 {{- define "curator.indexPattern" -}}
-{{ if and .Values.global.logging.loggingSidecar.enabled .Values.global.logging.loggingSidecar.indexPattern }}
-{{- .Values.global.logging.loggingSidecar.indexPattern | squote }}
-{{ else }}
-{{- .Values.curator.age.timestring | squote}}
+{{- include "curator.age.timestringRaw" . | squote -}}
+{{- end -}}
+
+{{/*
+The finest time unit actually present in the effective timestring.
+
+Curator's age filter with "source: name" parses the index name through
+strptime(timestring) and treats the result as the index's age date. strptime anchors
+missing components to their lowest value, so strptime("2026.04", "%Y.%m") yields
+2026-04-01. Comparing that against a *day*-granularity retention means the current
+month's index looks older than "10 days" from the 11th onward and gets deleted, even
+though it is the index currently being written to. See PINF-578.
+
+Deriving the unit from the timestring keeps the two halves of the filter consistent:
+  %Y.%m.%d -> days   (chart default)
+  %Y.%m    -> months
+  %Y       -> years
+
+Ordered finest-first so the most specific component present wins. Every unit curator
+accepts is covered on purpose: "curator.age.validate" compares an operator's explicit
+unit against this result, so a missing directive here would turn a legitimate config
+(e.g. weekly %Y.%W indices with unit: weeks) into a spurious template failure.
+Falls back to days -- the historical default -- for a timestring with no recognised
+directive.
+*/}}
+{{- define "curator.age.derivedUnit" -}}
+{{- $timestring := include "curator.age.timestringRaw" . -}}
+{{- if regexMatch "%S" $timestring -}}seconds
+{{- else if regexMatch "%M" $timestring -}}minutes
+{{- else if regexMatch "%H" $timestring -}}hours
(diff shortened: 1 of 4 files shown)


charts/elasticsearch/templates/_helpers.tpl
```diff
@@ -148,11 +148,112 @@ imagePullSecrets:
...
+Falls back to days -- the historical default -- for a timestring with no recognised
+directive.
+*/}}
+{{- define "curator.age.derivedUnit" -}}
+{{- $timestring := include "curator.age.timestringRaw" . -}}
+{{- if regexMatch "%S" $timestring -}}seconds
+{{- else if regexMatch "%M" $timestring -}}minutes
+{{- else if regexMatch "%H" $timestring -}}hours
+{{- else if regexMatch "%[dj]" $timestring -}}days
+{{- else if regexMatch "%[WU]" $timestring -}}weeks
```
Simpcyclassy, line 202: `%V` belongs in this class alongside `W` and `U`. Curator's `date_regex` handles `V` (and compensates for week-of-year by appending a weekday before strptime), so `%G.%V` is a config it will happily run, but this helper won't match it here.

Simpcyclassy (commented):
`curator.age.derivedUnit` misses `%V` and `%G`, which curator's own date_regex accepts, so an ISO-week pattern falls through to the `days` default and reintroduces the exact PINF-578 deletion. The new hard `fail` also turns an existing `unit: days` plus `%Y.%m` install into an aborted `helm upgrade`, which needs saying out loud in a real title and description rather than TBD. The logic and the tests are otherwise strong, so both are worth resolving before this leaves draft.

charts/elasticsearch/templates/_helpers.tpl
```diff
@@ -148,11 +148,112 @@ imagePullSecrets:
...
+{{- else if regexMatch "%m" $timestring -}}months
+{{- else if regexMatch "%[Yy]" $timestring -}}years
+{{- else -}}days
```
Simpcyclassy, line 205: This fallback points the wrong way. `days` is near the finest end of the scale, so any timestring this chain doesn't recognise (`%G.%V` being the realistic one, since neither `%G` nor `%V` is matched above) silently derives day-granularity retention on a coarse index name, which is precisely the deletion PINF-578 is about. Fall back to `years`, or `fail` with the same message as `curator.age.validate`, so an unknown pattern errs toward keeping data.

charts/elasticsearch/templates/_helpers.tpl
```diff
@@ -148,11 +148,112 @@ imagePullSecrets:
...
+{{- $configured := .Values.curator.age.unit | toString -}}
+{{- $derived := include "curator.age.derivedUnit" . -}}
+{{- $timestring := include "curator.age.timestringRaw" . -}}
+{{- $ranks := dict "seconds" 1 "minutes" 2 "hours" 3 "days" 4 "weeks" 5 "months" 6 "years" 7 -}}
+{{- if not (hasKey $ranks $configured) -}}
+{{- fail (printf "elasticsearch.curator.age.unit is %q, which curator does not accept; valid units are seconds, minutes, hours, days, weeks, months, years (or leave it unset to derive it from the index timestring %q)" $configured $timestring) -}}
+{{- end -}}
+{{- if lt (int (index $ranks $configured)) (int (index $ranks $derived)) -}}
+{{- fail (printf "elasticsearch.curator.age.unit is %q but the curator index timestring is %q, whose finest component is %q. A retention unit finer than the index-name granularity makes curator delete the index it is still writing to, destroying every log written since the last curator run (PINF-578). Fix this by either unsetting elasticsearch.curator.age.unit so it is derived as %q, setting it to %q or coarser, or switching to a finer index pattern such as \"%%Y.%%m.%%d\"." $configured $timestring $derived $derived $derived) -}}
```
Simpcyclassy, line 255: Worth being explicit that this is an upgrade-blocking failure, not just a fresh-install guard: anyone today running `unit: days` with a `%Y.%m` indexPattern gets an aborted `helm upgrade` on the release that ships this. I think aborting is the right call over silent data loss, but it needs a release note and a line in the PR body.

charts/elasticsearch/templates/curator/es-curator-configmap.yaml
```diff
@@ -33,11 +34,19 @@ data:
           ignore_empty_list: True
           disable_action: False
         filters:
+        # `unit` must not be finer than the finest component of `timestring`, or curator
+        # deletes the index it is still writing to. See PINF-578 and the granularity rule
+        # documented on curator.age in the elasticsearch chart values.yaml.
+        {{- if .Values.curator.age.unit }}
```
Simpcyclassy, line 40: This branch exists only to pick which YAML comment to emit into the rendered ConfigMap. The three static lines above already say the important part, and the rendered `unit:` value tells you the rest. I'd drop the conditional and keep the output stable.

charts/elasticsearch/values.yaml
```diff
@@ -266,9 +266,15 @@ curator:
...
+  #
+  # Leave `unit` null (the default) and it is derived from the timestring:
+  #   %Y.%m.%d -> days    %Y.%m -> months    %Y -> years
+  # Set it explicitly only if you need a different unit, and note that a value finer
+  # than the timestring granularity is rejected at `helm template` time.
+
   age:
     timestring: "%Y.%m.%d"
-    unit: "days"
+    unit: ~
```
Simpcyclassy, line 277: Changing a shipped default from `"days"` to null is a behaviour change for anyone relying on the default with a non-daily indexPattern, so call it out in the body. Also there's a stray blank line between the new comment block and `age:` that detaches the docs from the key.

tests/chart_tests/test_elasticsearch.py
```diff
@@ -630,6 +635,101 @@ def test_elasticsearch_curator_with_indexPatterns_defaults(self, kube_version):
...
+        assert LS["actions"][1]["filters"][0]["unit"] == "days"
+        assert LS["actions"][1]["filters"][0]["unit_count"] == 10
+
+    @pytest.mark.parametrize(
+        "indexPattern,expected_unit",
+        [
+            ("%Y.%m.%d", "days"),
+            ("%Y.%m.%d.%H", "hours"),
+            ("%Y.%m", "months"),
+            ("%Y", "years"),
```
Simpcyclassy, line 648: Nice table, and good that these assert the rendered `unit` rather than just a successful render. Please add `%Y.%W` and `%G.%V` rows here; the second one is what surfaces the missing `%V`/`%G` handling in `derivedUnit`.

PR #3510, opened 2026-08-20
PINF-74: Remove duplicate DATABASE_URL env vars from Houston pods

Shubham-astro:
<html><head></head><body><h2>Description</h2>
<p>Houston containers have been rendering two <code>DATABASE_URL</code> entries since 2022 — one explicit, one from the <code>houston_environment</code> helper. This PR removes the duplicates.</p>
<p>Removing them naively is what makes this non-trivial. Helm diffs list items by merge key, and <code>env</code> uses <code>name</code>, so dropping one of two identical keys makes Helm emit <code>{"name":"DATABASE_URL","$patch":"delete"}</code> — and the apiserver applies that directive to <em>every</em> entry sharing the key, wiping the variable entirely. That is <a href="https://github.com/kubernetes/kubernetes/issues/58477">kubernetes#58477</a>, it caused an incident when #1463 first attempted this, and it was reverted in #1747 with the guard test skipped pending an upstream fix. The upstream fix is still unmerged, so this works around the bug rather than waiting on it.</p>
<p>Two groups, handled differently:</p>
<ul>
<li><strong>The six hook-job containers</strong> (<code>houston-db-migration-job</code>, <code>houston-upgrade-deployments-job</code>, <code>houston-cp-refresh-job</code>) carry <code>helm.sh/hook-delete-policy: before-hook-creation</code>. Helm recreates them every release, so they never go through the patch path — the duplicates are simply removed.</li>
<li><strong>The two long-lived Deployments</strong> are patched on upgrade, so their containers are renamed (below) to route around the bug.</li>
</ul>
<p>Note the removed entry was already dead: it pointed at <code>astronomer-bootstrap</code> and sat <em>before</em> the <code>houston_environment</code> include, which emits <code>DATABASE_URL</code> again from <code>houston.backendSecret</code>. Kubelet resolves duplicates last-wins, so the backend secret was already the effective value. No container's resolved environment changes in this PR.</p>
<p>This also unblocks Helm 4, which <a href="https://github.com/helm/helm/issues/31529">treats duplicate env vars as an error</a>.</p>
<h2>Container renames — user visible</h2>

Template | Before | After
-- | -- | --
houston-deployment.yaml (init container) | wait-for-db | houston-wait-for-db
houston-worker-deployment.yaml | houston | houston-worker

<p>Containers merge on <code>name</code> too, so renaming makes Helm delete the old container and add a new one wholesale, carrying a single <code>DATABASE_URL</code>. No env-level diff is computed, so the delete directive is never generated. The rename is load-bearing for upgrade safety, not cosmetic.</p>
<p>The worker rename also aligns with the vector sidecar's pre-existing <code>container_name: houston-worker</code> filter, fixing a latent log-shipping mismatch.</p>
<p><strong>Impact:</strong> anything referencing these container names breaks — <code>kubectl logs -c houston</code> against a worker pod, saved queries, dashboards, runbooks, or customer tooling. This needs a loud callout in the release notes for whichever release carries it.</p>
<p><code>wait-for-db</code> still exists in the hook jobs and in Grafana; only the one in <code>houston-deployment.yaml</code> is renamed. The API's own main container is still <code>houston</code>.</p>
<h2>Prometheus alert changes</h2>
<p>Both follow from the worker rename, in opposite directions:</p>
<ul>
<li><code>CriticalComponentPodCrashLooping</code> matches an explicit name list. The worker was covered because its container was named <code>houston</code>, so <code>houston-worker</code> is <strong>added</strong> to keep coverage.</li>
<li><code>AirflowDeploymentUnhealthy</code> matches <code>.*worker</code> as a pattern. <code>houston</code> never matched it, but <code>houston-worker</code> does — so <code>container!~"houston-.*"</code> is <strong>excluded</strong> to stop the platform worker firing an Airflow-tier alert. Since <code>kube_pod_container_status_running</code> exports <code>0</code> for a non-running container, a single worker restart would otherwise trip it.</li>
</ul>
<h2>Related Issues</h2>
<p>https://linear.app/astronomer/issue/PINF-74/put-duplicate-env-var-check-back-when-k8s-bug-is-fixed</p>

<h2>Testing</h2>
<ul>
<li><code>TestDuplicateEnvironment</code> is re-enabled. It renders the chart with all features enabled and asserts no container has duplicate env vars. It is the check that proves this change and prevents regression. Now parametrized over <code>unified</code> / <code>control</code> / <code>data</code> plane modes, so templates that only render in a single plane are covered too.</li>
<li>Upgraded a cluster from 2.0.1 to 2.1.0:
<ul>
<li>Confirmed exactly one <code>DATABASE_URL</code> per container on the live Deployment specs, sourced from <code>&lt;release&gt;-houston-backend</code>.</li>
<li>Tested deployment creation and tested logs were coming
<ul>
<li>Daemonset:
&lt;img width="1574" height="911" alt="Screenshot 2026-08-20 at 2 19 33 PM" src="https://github.com/user-attachments/assets/08762be1-b500-435c-9f12-7bfdbf40a739" /&gt;
&lt;img width="1519" height="526" alt="Screenshot 2026-08-20 at 2 20 02 PM" src="https://github.com/user-attachments/assets/58072af2-4ad6-42e2-b7e0-d4482ccb02df" /&gt;</li>
<li>Sidecar logging:
&lt;img width="1201" height="421" alt="Screenshot 2026-08-20 at 2 27 41 PM" src="https://github.com/user-attachments/assets/7d75c409-fbe7-4e93-8246-54e0420b31d8" /&gt;
&lt;img width="1556" height="553" alt="Screenshot 2026-08-20 at 2 28 47 PM" src="https://github.com/user-attachments/assets/c07f269b-0a89-4971-b1fc-7b5109cd491b" /&gt;</li>
</ul>
</li>
</ul>
</li>
</ul>
<h2>Merging</h2>
<p>Master.</p>
<p>Whichever release branch this is cherry-picked into needs a release-note callout for the two container renames, per the impact section above.</p></body></html>## Description

Houston containers have been rendering two `DATABASE_URL` entries since 2022 — one explicit, one from the `houston_environment` helper. This PR removes the duplicates.

Removing them naively is what makes this non-trivial. Helm diffs list items by merge key, and `env` uses `name`, so dropping one of two identical keys makes Helm emit `{"name":"DATABASE_URL","$patch":"delete"}` — and the apiserver applies that directive to *every* entry sharing the key, wiping the variable entirely. That is [[kubernetes#58477](https://github.com/kubernetes/kubernetes/issues/58477)](https://github.com/kubernetes/kubernetes/issues/58477), it caused an incident when #1463 first attempted this, and it was reverted in #1747 with the guard test skipped pending an upstream fix. The upstream fix is still unmerged, so this works around the bug rather than waiting on it.

Two groups, handled differently:

- **The six hook-job containers** (`houston-db-migration-job`, `houston-upgrade-deployments-job`, `houston-cp-refresh-job`) carry `helm.sh/hook-delete-policy: before-hook-creation`. Helm recreates them every release, so they never go through the patch path — the duplicates are simply removed.
- **The two long-lived Deployments** are patched on upgrade, so their containers are renamed (below) to route around the bug.

Note the removed entry was already dead: it pointed at `astronomer-bootstrap` and sat *before* the `houston_environment` include, which emits `DATABASE_URL` again from `houston.backendSecret`. Kubelet resolves duplicates last-wins, so the backend secret was already the effective value. No container's resolved environment changes in this PR.

This also unblocks Helm 4, which [[treats duplicate env vars as an error](https://github.com/helm/helm/issues/31529)](https://github.com/helm/helm/issues/31529).

## Container renames — user visible

| Template | Before | After |
|---|---|---|
| `houston-deployment.yaml` (init container) | `wait-for-db` | `houston-wait-for-db` |
| `houston-worker-deployment.yaml` | `houston` | `houston-worker` |

Containers merge on `name` too, so renaming makes Helm delete the old container and add a new one wholesale, carrying a single `DATABASE_URL`. No env-level diff is computed, so the delete directive is never generated. The rename is load-bearing for upgrade safety, not cosmetic.

The worker rename also aligns with the vector sidecar's pre-existing `container_name: houston-worker` filter, fixing a latent log-shipping mismatch.

**Impact:** anything referencing these container names breaks — `kubectl logs -c houston` against a worker pod, saved queries, dashboards, runbooks, or customer tooling. This needs a loud callout in the release notes for whichever release carries it.

`wait-for-db` still exists in the hook jobs and in Grafana; only the one in `houston-deployment.yaml` is renamed. The API's own main container is still `houston`.

## Prometheus alert changes

Both follow from the worker rename, in opposite directions:

- `CriticalComponentPodCrashLooping` matches an explicit name list. The worker was covered because its container was named `houston`, so `houston-worker` is **added** to keep coverage.
- `AirflowDeploymentUnhealthy` matches `.*worker` as a pattern. `houston` never matched it, but `houston-worker` does — so `container!~"houston-.*"` is **excluded** to stop the platform worker firing an Airflow-tier alert. Since `kube_pod_container_status_running` exports `0` for a non-running container, a single worker restart would otherwise trip it.

## Related Issues

https://linear.app/astronomer/issue/PINF-74/put-duplicate-env-var-check-back-when-k8s-bug-is-fixed

Also related: astronomer/issues#5227, original de-dup #1463, revert #1747.

## Testing

- `TestDuplicateEnvironment` is re-enabled. It renders the chart with all features enabled and asserts no container has duplicate env vars. It is the check that proves this change and prevents regression. Now parametrized over `unified` / `control` / `data` plane modes, so templates that only render in a single plane are covered too.
- Upgraded a cluster from 2.0.1 to 2.1.0:
  - Confirmed exactly one `DATABASE_URL` per container on the live Deployment specs, sourced from `<release>-houston-backend`.
  - Tested deployment creation and tested logs were coming
    - Daemonset:
      Screenshot 2026-08-20 at 2 19 33 PM
      Screenshot 2026-08-20 at 2 20 02 PM
    - Sidecar logging:
      Screenshot 2026-08-20 at 2 27 41 PM
      Screenshot 2026-08-20 at 2 28 47 PM

## Merging

Master.

Whichever release branch this is cherry-picked into needs a release-note callout for the two container renames, per the impact section above.


charts/astronomer/templates/houston/helm-hooks/houston-upgrade-deployments-job.yaml
```diff
@@ -81,11 +81,6 @@ spec:
             - "echo"
             - "DB ready"
           env:
-            - name: DATABASE_URL
-              valueFrom:
-                secretKeyRef:
-                  name: astronomer-bootstrap
-                  key: connection
             {{- include "houston_environment" . | indent 12 }}
```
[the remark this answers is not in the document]
Shubham-astro: The ordering here is the other way round. The explicit `astronomer-bootstrap` block
sat *before* `{{- include "houston_environment" }}`, and the helper emits its own
`DATABASE_URL` from `houston.backendSecret`. Kubelet resolves duplicate env names
last-wins, so this container was *already* using the backend secret — placeholder
and all — on fresh installs. That's why it only ever produced a `duplicate name`
warning. Removing the overwritten entry changes no container's resolved env. So the change is fine

charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -85,12 +85,7 @@ spec:
           livenessProbe: {{ tpl (toYaml .Values.houston.waitForDB.livenessProbe) . | nindent 12 }}
           {{- end }}
           env:
-            - name: DATABASE_URL
-              valueFrom:
-                secretKeyRef:
-                  name: astronomer-bootstrap
-                  key: connection
-           {{- include "houston_environment" . | indent 12 }}
+            {{- include "houston_environment" . | indent 12 }}
```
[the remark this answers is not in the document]
Shubham-astro: Same as the houston-deployment thread: the explicit `astronomer-bootstrap` entry
came *before* the `houston_environment` include, which emits `DATABASE_URL` again
from `houston.backendSecret`. Last-wins means this init container was already
resolving to the backend secret, so removing the dead entry doesn't repoint it.

Also worth noting this job has its own `houston-bootstrapper` init container right
after `wait-for-db`, with the same ordering as the API deployment — nothing about
that sequence changes in this PR.

charts/astronomer/templates/houston/helm-hooks/houston-db-migration-job.yaml
```diff
@@ -77,11 +77,6 @@ spec:
             - "echo"
             - "DB ready"
           env:
-            - name: DATABASE_URL
-              valueFrom:
-                secretKeyRef:
-                  name: astronomer-bootstrap
-                  key: connection
             {{- include "houston_environment" . | indent 12 }}
```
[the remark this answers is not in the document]
Shubham-astro: Same as the other two threads — the explicit block was ordered before the
`houston_environment` include and was being overwritten by it, so the effective
`DATABASE_URL` was already `<release>-houston-backend` here. No behavioural change
from removing it.

One extra note for this file: it's a `post-upgrade,post-install` hook with
`hook-delete-policy: before-hook-creation`, so it's recreated from scratch every
release rather than patched — which is also why it needed no container rename.

charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
```diff
@@ -137,7 +137,15 @@ spec:
...
-        - name: houston
+        # NOTE: renamed from "houston" in the same release that removed the
+        # duplicate DATABASE_URL env var. Renaming makes Helm delete the old
+        # container and add a new one, instead of emitting an ambiguous
+        # `{"name":"DATABASE_URL","$patch":"delete"}` directive that the apiserver
+        # would apply to *every* entry with that name, wiping the var entirely.
+        # See https://github.com/kubernetes/kubernetes/issues/58477 and PINF-74.
+        # The name now also matches the component/service labels this pod's vector
+        # sidecar already ships logs under.
+        - name: houston-worker
```
rishkarajgi, line 142: This rename (from `houston`) is load-bearing for upgrade safety: it makes `helm upgrade` replace the container with a fresh env (one `DATABASE_URL`) instead of diffing the existing container's env list and emitting the `{name:DATABASE_URL,$patch:delete}` that k8s #58477 would spread to all entries. It also aligns with the vector sidecar's pre-existing `container_name: houston-worker` filter (fixing a latent log-shipping mismatch). Verified head renders exactly one DATABASE_URL here.

charts/prometheus/templates/prometheus-alerts-configmap.yaml
```diff
@@ -619,7 +619,7 @@ data:
             runbook_url:
             summary: '{{`{{ $labels.pod }}`}} Pod is CrashLooping'
           expr: |
-            rate(kube_pod_container_status_restarts_total{container=~"astro-ui|commander|grafana|houston|prometheus|registry|sqlproxy"}[15m]) * 60 * 5 > 0
+            rate(kube_pod_container_status_restarts_total{container=~"astro-ui|commander|grafana|houston|houston-worker|prometheus|registry|sqlproxy"}[15m]) * 60 * 5 > 0
```
rishkarajgi, line 622: Correct and necessary consequence of the worker rename — the `=~` match is anchored, so without adding `houston-worker` here the renamed container's crashloops would go unalerted. `houston` is retained since it's still the API container.

rishkarajgi (approved):
Approving — I gave this one extra scrutiny because `DATABASE_URL` is how Houston reaches its DB, so a botched de-dup would be a platform outage, and there's a real k8s strategic-merge-patch wipe (#58477) in play. Verified the outage vectors directly and they're closed:

- **Exactly one `DATABASE_URL` on every Houston workload.** The explicit literal entries are removed from all five templates (deployment, worker, and the three helm-hook jobs), and each container still pulls in `houston_environment`, which emits a single `DATABASE_URL` from `astronomer-houston-backend`. So every DB-using container renders exactly one — no zero (wiped), no two (still duplicated). Where the removed entry differed (the `wait-for-db` init had an `astronomer-bootstrap` copy), it was already shadowed by Kubernetes last-wins, so the kept value is unchanged.
- **Source-level removal, not a `$patch:delete`.** The only `$patch:delete` in the diff is inside an explanatory comment — no actual delete directive anywhere, which is the safe shape (it avoids the very wipe the ticket describes).
- **Upgrade-safe by construction.** The two containers whose count changed 2→1 are renamed (`houston`→`houston-worker`, `wait-for-db`→`houston-wait-for-db`), so on `helm upgrade` the apiserver replaces the container with a fresh full env (one entry) rather than diffing its env list and emitting the `{name:DATABASE_URL,$patch:delete}` that triggers #58477. The count-unchanged containers stay un-renamed, and all three hook jobs recreate fresh (`before-hook-creation`) so there's no in-place merge. No window where a Houston pod comes up with no `DATABASE_URL`.
- The worker rename also **fixes** a pre-existing vector log-shipping mismatch (the sidecar already filtered on `container_name: houston-worker`), and the prometheus crashloop alert correctly adds `houston-worker` (anchored `=~` wouldn't have matched it otherwise). The `TestDuplicateEnvironment` guard is re-enabled — the realization of "put the duplicate-env check back."

One caveat worth stating given the severity of the failure mode (not blocking): the no-wipe guarantee is design-verified (renamed container ⇒ replace-not-merge; hook jobs recreate) and the templates render one entry, but no test performs an actual `helm upgrade` from a prior release to prove no pod transiently comes up with zero `DATABASE_URL`. That's a known upgrade-e2e CI gap rather than this PR's burden — but an upgrade smoke test would convert this from design-verified to regression-guarded, and it'd be worth it here. The re-enabled `TestDuplicateEnvironment` could also assert `DATABASE_URL` is *present* per workload to cover the 'not zero' side, not just 'not two'. Still needs CODEOWNERS approval.

charts/prometheus/templates/prometheus-alerts-configmap.yaml
pgvishnuram: do we still have sqlproxy ??

charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
danielhoherd: This solution of renaming the container is very creative, and a clever way to work around some of the challenges. I actually like the container rename too, because it's a better description of what is running in the pod.

My only concern with this is if any customer tooling relies on the old name, this will break it. We will need to make sure to communicate this change loudly in any release that gets this change.

charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
```diff
@@ -137,7 +137,15 @@ spec:
...
       containers:
-        - name: houston
+        # NOTE: renamed from "houston" in the same release that removed the
+        # duplicate DATABASE_URL env var. Renaming makes Helm delete the old
+        # container and add a new one, instead of emitting an ambiguous
+        # `{"name":"DATABASE_URL","$patch":"delete"}` directive that the apiserver
+        # would apply to *every* entry with that name, wiping the var entirely.
+        # See https://github.com/kubernetes/kubernetes/issues/58477 and PINF-74.
+        # The name now also matches the component/service labels this pod's vector
+        # sidecar already ships logs under.
```
pgvishnuram: @danielhoherd @pgvishnuram -> make sure you comments should be one-liner or make it small

charts/astronomer/templates/houston/api/houston-deployment.yaml
```diff
@@ -68,7 +68,7 @@ spec:
           imagePullPolicy: {{ .Values.images.houston.pullPolicy }}
           securityContext: {{- include "platform.containerSecurityContext" (list $ nil) | nindent 12 }}
           resources: {{- toYaml .Values.houston.resources | nindent 12 }}
-        - name: wait-for-db
+        - name: houston-wait-for-db
```
Simpcyclassy, line 73: This rename is load-bearing for exactly the same reason as the worker one, but it has no comment, and the worker pod's own `wait-for-db` init container keeps its old name because it never had the duplicate. Please add the same short NOTE here so the next person doesn't "clean up" the asymmetry and reintroduce the wipe.

tests/chart_tests/test_read_only_root_filesystem.py
```diff
@@ -86,7 +86,7 @@ def test_houston_pods_read_only_root_filesysystem_settings(self, pod_manager_doc
                     assert container["volumeMounts"] == [{"name": "etc-ssl-certs", "mountPath": "/etc/ssl/certs_copy"}], (
                         f"{pod_name}/{container['name']} etc-ssl-certs-copier mounts are wrong"
                     )
-                case "wait-for-db":
+                case "wait-for-db" | "houston-wait-for-db":
```
Simpcyclassy, line 90: Worth a one-line comment that both names are intentional: the api pod's init container is renamed, the worker pod's is not. Otherwise this reads like a leftover and someone will drop the old case.

tests/chart_tests/test_default_chart.py
```diff
@@ -266,7 +266,6 @@ def test_no_containers_use_unsafe_mount_propagation(self):
         )
 
 
-@pytest.mark.skip("See issue https://github.com/astronomer/issues/issues/5227 for details about when to reenabling this.")
 class TestDuplicateEnvironment:
```
Simpcyclassy, line 269: Good to have this guard back, and it does assert real rendered output rather than just "did not throw". One gap: it renders once with `get_all_features()`, so the control-only and data-only plane renders never get the duplicate check. Adding a plane-mode parametrize here would cover the templates that only exist in those modes.

Simpcyclassy (changes requested):
Renaming the worker container to `houston-worker` makes it match the `.*worker` matcher in the AirflowDeploymentUnhealthy expr further up this same file, so worker restarts will fire a bogus airflow-tier alert; exclude it there. The description also never mentions either container rename, which is the riskiest and most support-visible part of the change and belongs in the body next to the screenshots. The de-dup itself traces clean, so this is the alert exclusion plus a description update.

charts/prometheus/templates/prometheus-alerts-configmap.yaml
Simpcyclassy: Agree on this line, but the rename has a second effect in this same file that isn't handled. The AirflowDeploymentUnhealthy expr near the top matches `container=~".*(scheduler|...|worker|...)"`, and Prometheus anchors that regex, so `houston-worker` now matches where `houston` did not. `kube_pod_container_status_running` is still exported as `0` for a container that isn't running, so a worker restart lasting 15m gives sum=0, count=1 and the alert fires with `tier: airflow` against whatever the platform `release` relabel produces (empty in single-namespace mode, `<release>-houston` in namespace-pools mode). Adding `container!~"houston-.*"` to both matchers in that expr closes it.

Simpcyclassy: We do, just not from this chart. `sqlproxy` is the container from the separately installed `pg-sqlproxy-gcloud-sqlproxy` release that GCP Software installs run in the platform namespace; it's also referenced in `charts/astronomer/templates/istio/sidecars.yaml`. Nothing here renders it, so grep comes up empty, but it should stay in the matcher.

tests/chart_tests/test_default_chart.py
```diff
@@ -266,7 +266,6 @@ def test_no_containers_use_unsafe_mount_propagation(self):
         )
 
 
-@pytest.mark.skip("See issue https://github.com/astronomer/issues/issues/5227 for details about when to reenabling this.")
```
danielhoherd, line 269: I'm happy to see this `skip` removed! In its place we should put a comment that once we move to helm 4 we can delete this test, because helm 4 treats duplicate env vars as a blocking bug, which satisfies the goal of this test.

tests/chart_tests/test_read_only_root_filesystem.py
danielhoherd: Actually we should rename the worker pod's container too. They both do the same thing. It's effectively the same container.

charts/astronomer/templates/houston/api/houston-deployment.yaml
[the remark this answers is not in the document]
pgvishnuram: also can you check why cp-refresh needs db bootstrapper and wait for db

tests/chart_tests/test_default_chart.py
Shubham-astro: Added the docstring

tests/chart_tests/test_default_chart.py
Shubham-astro: done

tests/chart_tests/test_read_only_root_filesystem.py
Shubham-astro: agreed.

charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
Shubham-astro: done

Shubham-astro: done

charts/astronomer/templates/houston/api/houston-deployment.yaml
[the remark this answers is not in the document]
Shubham-astro: @pgvishnuram, I think these are redundant here and probably copy-paste here from other hook jobs. cp-refresh is `post-upgrade` only (no `post-install`), so it never runs on a fresh install. However i would rather not remove them in this PR as this only removes overwritten duplicate changes and no runtime behaviour, I will raise a separate ticket for this.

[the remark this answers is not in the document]
Shubham-astro: https://linear.app/astronomer/issue/PINF-1215/remove-redundant-wait-for-db-and-houston-bootstrapper-init-containers

charts/prometheus/templates/prometheus-alerts-configmap.yaml
```diff
@@ -34,7 +34,7 @@ data:
         {{- end }}
         {{- if .Values.defaultAlerts.airflow.enabled }}
         - alert: AirflowDeploymentUnhealthy
-          expr: sum by(release) (kube_pod_container_status_running{container=~".*(scheduler|scheduler-gc|webserver|api-server|worker|statsd|pgbouncer|metrics-exporter|redis|flower|triggerer)"}) - count by(release) (kube_pod_container_status_running{container=~".*(scheduler|scheduler-gc|webserver|worker|statsd|pgbouncer|metrics-exporter|redis|flower|triggerer)"}) < 0
+          expr: sum by(release) (kube_pod_container_status_running{container=~".*(scheduler|scheduler-gc|webserver|api-server|worker|statsd|pgbouncer|metrics-exporter|redis|flower|triggerer)", container!~"houston-.*"}) - count by(release) (kube_pod_container_status_running{container=~".*(scheduler|scheduler-gc|webserver|worker|statsd|pgbouncer|metrics-exporter|redis|flower|triggerer)", container!~"houston-.*"}) < 0
```
Simpcyclassy, line 37: Confirmed this closes the bogus AirflowDeploymentUnhealthy alert I flagged last review: houston-worker restarts would have matched the unanchored .*worker term and fired a tier:airflow alert without this exclusion.

Simpcyclassy (approved):
The AirflowDeploymentUnhealthy and container-crashloop alert exprs now handle the houston-worker rename exactly as I asked (container!~"houston-.*" added, houston-worker added to the crashloop matcher), so the bogus-alert risk from my last pass is closed. TestDuplicateEnvironment is genuinely re-enabled and now parametrized across unified/control/data plane modes, and I traced the removed entries to confirm the surviving houston_environment DATABASE_URL was already last-wins in every case, so no connection string changes. One gap remains: the description still never mentions either container rename (houston to houston-worker, wait-for-db to houston-wait-for-db), which is the part most likely to surprise anyone with tooling keyed on the old names; worth adding before merge but not blocking given the code is sound.

charts/prometheus/templates/prometheus-alerts-configmap.yaml
Simpcyclassy: Resolved. Verified in the diff: container!~"houston-.*" was added to both matchers in the AirflowDeploymentUnhealthy expr, and houston-worker was added to the crashloop rule's container list separately.

tests/chart_tests/test_default_chart.py
Simpcyclassy: Confirmed fixed: the test now sets global.plane.mode to unified/control/data and re-renders per mode, so control-only and data-only plane templates get the duplicate-env check too.

charts/astronomer/templates/houston/api/houston-deployment.yaml
Simpcyclassy: Resolved, the NOTE comment was added above houston-wait-for-db explaining the rename. Same treatment landed on the worker container rename too.

tests/chart_tests/test_read_only_root_filesystem.py
Simpcyclassy: Minor: the reply says agreed but I don't see the comment actually added next to the wait-for-db/houston-wait-for-db case in the current diff. Not blocking, just flagging so it doesn't fall through.

charts/astronomer/templates/houston/worker/houston-worker-deployment.yaml
Simpcyclassy: Agree, and this is still open: the PR description hasn't been updated to call out either rename. Given this is what would confuse a customer grepping container names post-upgrade, it belongs in the body next to the upgrade screenshots.

charts/astronomer/templates/houston/api/houston-deployment.yaml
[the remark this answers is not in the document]
Simpcyclassy: Agree with author's call to track this separately (PINF-1215) rather than widening this PR's scope. cp-refresh being post-upgrade-only means the redundant init containers there are a latent cleanup item, not a correctness issue introduced by this change.

PR #3512, opened 2026-08-24
PLX-930: remove the houston-cleanup-airflow-db CronJob

Simpcyclassy:
## Description

Companion chart change for [houston-api#2920](https://github.com/astronomer/houston-api/pull/2920) (adds `perClusterConfig` to the `cleanupAirflowDb` API and removes the `cleanup-airflow-db` script this CronJob invoked).

Related 
https://linear.app/astronomer/issue/PLX-930/evaluate-removing-cleanup-airflow-db-script-in-favor-of-apischeduled
https://linear.app/astronomer/issue/PLX-706/bound-concurrency-in-clusterdeployment-batch-jobs

Removes the CronJob template entirely rather than building a replacement scheduling mechanism inside this chart. Anyone who wants recurring airflow-db cleanup should call the existing, RBAC-gated `cleanupAirflowDb` API directly on their own schedule (`perClusterConfig: true` for the same per-cluster behavior this CronJob had), rather than relying on a Helm-managed CronJob.

## Changes

- Removes `charts/astronomer/templates/houston/cronjobs/houston-cleanup-airflow-db-cronjob.yaml`.
- Trims `houston.cleanupAirflowDb` in `values.yaml` down to just `enabled`. That key still gates `deployments.deploymentLifecycle.cleanupAirflowDb.enabled` (the API feature flag) via `houston-configmap.yaml` — unchanged. The `schedule`/`olderThan`/`outputPath`/`dropArchives`/`dryRun`/`provider`/`bucketName`/`providerEnvSecretName`/`tables`/probe values only ever existed to template into this CronJob's args and pod spec, so they're now dead and removed.
- Removes `tests/chart_tests/test_astronomer_houston_airflow_db_cleanup.py`. Its 3 CronJob-rendering tests no longer apply; its one configmap-flag test was already duplicated by `test_houston_configmap.py`'s `test_houston_configmap_with_cleanup_airflow_db_enabled`/`_disabled` (untouched, still pass — `enabled` still exists and still flows into the same configmap key).
- Removes the now-dead `houston-cleanup-airflow-db-cronjob.yaml` entry from `test_byo_sa.py`'s service-account-per-template map, and the `cleanupAirflowDb` probe block from `enable_all_probes.yaml` (nothing renders a CronJob for those probe values to apply to).

## Testing

**Could not run `make test-unit` in this sandbox** — no network access to fetch the `helm` binary this repo's chart tests shell out to (`get.helm.sh` is blocked). Verified manually instead:
- `values.yaml` still parses as valid YAML.
- No other file in the repo references the removed sub-keys (`cleanupAirflowDb.schedule`, `.olderThan`, etc.) or the deleted CronJob template's filename.
- No `values.schema.json` exists for this chart to update.
- `tests/enable_all_features.yaml` and `configs/enable-all-features-control.yaml` (which set `houston.cleanupAirflowDb.enabled: true`) don't need changes — `enabled` still exists and still means the same thing.

**Please run `make test-unit` (specifically `tests/chart_tests/test_houston_configmap.py` and `tests/chart_tests/test_byo_sa.py`) before merging** to confirm the chart renders correctly with this CronJob template gone.

## Manual Test Plan

1. `helm template` this chart with `houston.cleanupAirflowDb.enabled: true` and confirm no `houston-cleanup-airflow-db-data` CronJob resource is rendered (previously it would be).
2. Confirm the houston configmap's `production.yaml` still sets `deployments.deploymentLifecycle.cleanupAirflowDb.enabled: true` in that case (the API feature flag still works).


charts/astronomer/values.yaml
```diff
@@ -363,43 +363,19 @@ houston:
     livenessProbe: {}
     startupProbe: {}
 
-  # Cleanup airflow db data
-  # This runs as a CronJob
+  # Airflow db cleanup (PLX-930): no longer a built-in CronJob. `enabled` still
+  # gates the cleanupAirflowDb API/Config Governance feature flag
+  # (deployments.deploymentLifecycle.cleanupAirflowDb.enabled) -- call the
+  # cleanupAirflowDb GraphQL query directly (pass perClusterConfig: true for
```
rishkarajgi, line 369: This tells operators `perClusterConfig: true` gives "the same per-cluster behavior the old CronJob had," but as of houston-api#2920's current head that fan-out is a bare `Promise.all` over every cluster with no concurrency cap and no per-cluster failure isolation — the script it replaces bounded this at 10 in-flight and isolated per-cluster failures. Worth holding this recommendation until that's fixed, or at least caveating it for large fleets.

rishkarajgi (commented):
Verified this against the repo rather than just the description: rendered the chart before/after with `houston.cleanupAirflowDb.enabled: true` (the CronJob really disappears, the configmap flag still flows through), grepped the whole monorepo for the removed template and values keys (nothing else references them), and ran the full chart unit suite on this head (6293 passed, 0 failed) since I had helm/network access the author's sandbox didn't. Nothing in the removal itself needs to change.

One thing worth a look before this merges, noted inline: the values.yaml comment here (and the PR description) tells operators the `perClusterConfig: true` call is a drop-in replacement for the CronJob's per-cluster behavior, but the houston-api PR implementing that fan-out currently does it with an unbounded `Promise.all` and no per-cluster failure isolation, unlike the script it's replacing. That's really a houston-api#2920 concern, but this is the PR that will point customers at it.

No blocking asks from the chart side. This needs at least one CODEOWNERS approval as usual; no security review trigger here.

charts/astronomer/values.yaml
```diff
@@ -363,43 +363,19 @@ houston:
     livenessProbe: {}
     startupProbe: {}
 
-  # Cleanup airflow db data
-  # This runs as a CronJob
+  # Airflow db cleanup (PLX-930): no longer a built-in CronJob. `enabled` still
+  # gates the cleanupAirflowDb API/Config Governance feature flag
+  # (deployments.deploymentLifecycle.cleanupAirflowDb.enabled) -- call the
```
Simpcyclassy, line 368: Verified this exact path against houston-configmap.yaml (deploymentLifecycle.cleanupAirflowDb.enabled at line 141-142) and against houston-api's src/lib/deployments/config/index.js, which still reads deploymentLifecycle.cleanupAirflowDb.enabled today, so this claim holds independent of houston-api#2920.

Simpcyclassy (commented):
I traced houston.cleanupAirflowDb.enabled into deployments.deploymentLifecycle.cleanupAirflowDb.enabled and confirmed houston-api still reads that exact path today, and a repo-wide grep found nothing else referencing the deleted CronJob file or its sub-keys. Rishkarajgi's point about perClusterConfig's Promise.all concurrency is fair but is really a houston-api#2920 scoping question I can't verify from this sandbox. Nothing on the chart side blocks this merge.

charts/astronomer/values.yaml
Simpcyclassy: Fair catch, but that's houston-api#2920's fan-out implementation, not something this chart PR controls, and I can't fetch that branch in this sandbox to verify the Promise.all concern myself. If it's unbounded there, I'd rather soften this comment's wording than block the chart removal on it, happy to update once #2920 settles.

edmund-astronomer (commented):
LGTM just one nit - we have some dead `houston-cleanup-airflow-db` selectors in

```txt
- charts/astronomer/templates/houston/api/houston-networkpolicy.yaml:53
- charts/astronomer/templates/commander/commander-networkpolicy.yaml:118
- charts/astronomer/templates/registry/registry-networkpolicy.yaml:91
```

that can be cleaned up

Simpcyclassy:
Good catch @edmund-astronomer , fixed in 803f9618f -- removed the three dead `component: houston-cleanup-airflow-db` podSelector entries (houston, commander, registry NetworkPolicies). They never matched anything after this PR anyway since the CronJob that set that label is gone.

edmund-astronomer (approved):
LGTM just take a look at the pre-commit. I think its unrelated to your changes but broke because of new month

PR #3515, opened 2026-08-25
Fix `release_bom` on release-1.2: Python 3.12 CI image

Shubham-astro:
## Description

This PR fixes `release_bom` failure on this branch:
```
ERROR: Package 'astronomer-bom-generator' requires a different Python: 3.12.3 not in '>=3.13'
```

## Related Issues

## Merging

1.2,1.1

### DIFF
--- a/bin/generate_circleci_config.py
+++ b/bin/generate_circleci_config.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
-"""This script is used to create the circle config file so that we can stay
-DRY."""
+"""Generate the CircleCI config."""
 
+import datetime
 import subprocess
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


danielhoherd (approved):

rishkarajgi (approved):

bin/generate_circleci_config.py
```diff
@@ -12,7 +12,7 @@
 metadata = yaml.safe_load((git_root_dir / "metadata.yaml").read_text())
 kube_versions = metadata["test_k8s_versions"]
 
-ci_runner_version = "2026-05"  # This should be the current YYYY-MM
+ci_runner_version = (datetime.datetime.now()).strftime("%Y-%m")
```
Simpcyclassy, line 15: This makes config generation time-dependent. .pre-commit-config.yaml runs this hook against any yaml or tpl file, and the run_pre_commit job's pre-commit run --all-files --show-diff-on-failure will fail as soon as datetime.now() computes a month that no longer matches what's committed in config.yml, since pre-commit fails any hook that modifies files. Please keep this a hardcoded string bumped by hand like before.

Simpcyclassy (changes requested):
Danielhoherd and rishkarajgi already approved this, but I don't think either caught that ci_runner_version now computes datetime.datetime.now() instead of a static string. run_pre_commit runs pre-commit --all-files on every build, and pre-commit fails hooks that rewrite files, so this breaks CI the moment the month rolls over past merge. Please revert ci_runner_version to a hardcoded, manually-bumped string instead of a computed date.

bin/generate_circleci_config.py
Shubham-astro: @Simpcyclassy this is what caused the failure since we forgot to bump the release-bom and python version on this branch was very old, we already follow this same in other branches like: https://github.com/astronomer/astronomer/blob/master/bin/generate_circleci_config.py#L14

PR #3516, opened 2026-08-25
Fix `sign-released-image`: pyenv shim breaks after cosign orb install

Shubham-astro:
## Description

The "Install Python dependencies" step fails on every branch:
```
/home/circleci/.pyenv/libexec/pyenv-exec: line 24: pyenv-version-name: command not found
Exited with code exit status 127
```

## Related Issues

## Merging

Master, 1.1,1.2,2.0,2.1

### DIFF
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -36,7 +36,8 @@ parameters:
 
 jobs:
   sign-released-image:
-    executor: docker-executor
+    docker:
+      - image: quay.io/astronomer/ci-helm-release:{{ ci_runner_version }}
     steps:
       - checkout
       - setup_remote_docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,8 @@ parameters:
 
 jobs:
   sign-released-image:
-    executor: docker-executor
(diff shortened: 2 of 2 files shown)


danielhoherd (approved):

rishkarajgi (approved):

.circleci/config.yml.j2
```diff
@@ -36,7 +36,8 @@ parameters:
 
 jobs:
   sign-released-image:
-    executor: docker-executor
+    docker:
+      - image: quay.io/astronomer/ci-helm-release:{{ ci_runner_version }}
```
Simpcyclassy, line 40: Confirmed ci-helm-release is already used by release_bom and test-unit for their Python-based steps in this same file, so this should have Python available without the pyenv shim that cosign/install was breaking.

Simpcyclassy (approved):
Moving sign-released-image off the pyenv-shimmed docker-executor onto the same ci-helm-release image already used by release_bom and test-unit sidesteps the cosign-orb-breaks-pyenv issue cleanly. This matches an established pattern already in this same config file, so it should have Python available without the shim conflict. Approving.

PR #3520, opened 2026-08-27
PINF-1236: Fix `release_bom` and `sign-released-image`

Shubham-astro:
## Description

Two release-pipeline jobs are failing on every branch. Both trace back to the CI
image, and neither is fixed by changing the image tag.

### 1. `release_bom` — Python 3.12 vs bom-generator >= 3.13
```
ERROR: Package 'astronomer-bom-generator' requires a different Python: 3.12.3 not in '>=3.13'
```
`bin/release-bom` installs the *latest* `astronomer-bom-generator`, and v0.16.0 raised its
floor to `>=3.13`. Every `ci-helm-release` tag ships Python 3.12.3 — including `2026-08`, which
is why the 2.0 release still failed after the image bump. No image tag fixes this. So we let `uv` provision the interpreter

### 2. `sign-released-image` — `docker: command not found`

The job previously ran on `cimg/python`, where the cosign orb's `install.sh` writes a frozen
`PATH` snapshot into `$BASH_ENV`, breaking pyenv's shims and failing `pip` with
`pyenv-version-name: command not found`. Moving to `ci-helm-release` (#3516) fixed that but
exposed the next problem: that image has no docker CLI.

`bin/sign-images.py` never shells out to docker — only `cosign` (`which`, `verify`, `sign`). The
`docker login` existed purely to populate the credential store cosign reads, and `cosign login`
writes the same store with the same `--password-stdin` flag. So:

- `docker login` → `cosign login <registry> --username … --password-stdin`
- dropped `setup_remote_docker` — nothing in the job builds, loads, or runs a container

## Related Issues

## Testing

- N/A

## Merging

Master, 2.1, 1.2 and 1.1 2.0


rishkarajgi (approved):

rishkarajgi (commented):
Checked both root-cause claims directly rather than taking them on faith: `bin/sign-images.py` really does only ever shell out to `which cosign`/`cosign verify`/`cosign sign` -- confirmed by reading the whole script, no `docker` invocation anywhere, so dropping `setup_remote_docker` is safe. `uv venv -p 3.13 --seed` is a real, correct way to provision a specific interpreter regardless of the CI image's default Python, and installing bom-generator into that venv rather than the system Python sidesteps the >=3.13 floor cleanly.

Couldn't verify `cosign login`'s exact flag behavior empirically (no cosign binary in this environment), but it's documented cosign functionality built specifically for this credential-store-without-a-docker-daemon case, consistent with the reasoning given.

Overlaps with a few other open PRs on `.circleci/config.yml` (expected, it's a large shared file) -- checked the closest one (#3453) and its nearest hunk sits a safe distance from the lines touched here.

Nothing here clears the bar for comment.

danielhoherd (changes requested):
Before merging this PR, I'd like us to do a git commit to this branch that allows us to run cosign and bom-generator jobs from this branch to ensure that they both work.

Also

bin/release-bom
```diff
@@ -42,11 +42,14 @@ json_file_name="astronomer-${version}.json"
 max_age="300"
 
 # Install bom-generator (latest release)
+# The CI image ships Python 3.12, but bom-generator requires >=3.13
 bom_tag=$(latest_github_release "astronomer/astronomer-bom-generator")
-pip install "git+https://${GITHUB_TOKEN}@github.com/astronomer/astronomer-bom-generator.git@${bom_tag}"
+bom_venv="$(mktemp -d)/venv"
+uv venv -p 3.13 --seed "${bom_venv}"
+"${bom_venv}/bin/pip" install "git+https://${GITHUB_TOKEN}@github.com/astronomer/astronomer-bom-generator.git@${bom_tag}"
```
danielhoherd: Using `uv tool` will set this up in its own venv and make it available as a command. That's exactly how I've been running this tool on my laptop.

```suggestion
uv tool install "git+https://${GITHUB_TOKEN}@github.com/astronomer/astronomer-bom-generator.git@${bom_tag}"
```

The only potential problem with this `uv tool install` puts commands into `$HOME/.local/bin` which may not be in the `$PATH` in the container, so we would need to run `$HOME/.local/bin/bom-generator` below.

bin/release-bom
```diff
@@ -42,11 +42,14 @@ json_file_name="astronomer-${version}.json"
...
 
 # Generate bom file and indexes
-bom-generator --version "${version}"
+"${bom_venv}/bin/bom-generator" --version "${version}"
```
danielhoherd: This is tied to the `uv tool install` comment above
```suggestion
"$HOME/.local/bin/bom-generator" --version "${version}"
```

danielhoherd (approved):
@Shubham-astro has reported to me that the test code he committed has passed successfully, so this looks good to merge after we revert that test code, record those passing tests in this thread, and fix the merge conflict.

bin/release-bom
```diff
@@ -42,18 +48,24 @@ json_file_name="astronomer-${version}.json"
...
 
 # Install bom-generator (latest release)
+# The CI image ships Python 3.12, but bom-generator requires >=3.13
 bom_tag=$(latest_github_release "astronomer/astronomer-bom-generator")
-pip install "git+https://${GITHUB_TOKEN}@github.com/astronomer/astronomer-bom-generator.git@${bom_tag}"
+uv tool install --python 3.13 "git+https://${GITHUB_TOKEN}@github.com/astronomer/astronomer-bom-generator.git@${bom_tag}"
 
 # Generate bom file and indexes
-bom-generator --version "${version}"
+"$HOME/.local/bin/bom-generator" --version "${version}"
```
Simpcyclassy, line 56: This full path is what actually makes the uv tool install fix work, since ~/.local/bin isn't guaranteed to be on PATH in the container. Good that this picked up danielhoherd's suggestion from the earlier review round.

.circleci/config.yml
```diff
@@ -669,6 +680,32 @@ workflows:
               only:
                 - '/^release-\d+\.\d+$/'
 
+  # TEMPORARY -- remove before merging this PR.
```
Simpcyclassy: This whole test-release-jobs workflow, plus the matching image_tag elif branch above, is marked "TEMPORARY -- remove before merging this PR" but it's still in the diff (same block in config.yml.j2). Should come out before this lands, per the PR's own comment and danielhoherd's request.

Simpcyclassy (dismissed):
Dropping docker for cosign login and letting uv provision Python 3.13 both trace correctly to the root causes described, I confirmed sign-images.py never touches docker so setup_remote_docker really is dead weight. The TEMPORARY test-release-jobs workflow and its elif branch are still here despite the "remove before merging" comment, and the merge conflict danielhoherd flagged looks unresolved. Requesting changes to strip the temporary scaffolding and clear the conflict, the actual fix logic looks correct to me.

bin/release-bom
Simpcyclassy: Confirmed this made it into the current diff, uv tool install with the full $HOME/.local/bin/bom-generator path below. Looks correct to me.

PR #3521, opened 2026-08-27
PINF-1221: flag the PgBouncer auth_type behaviour change in the 2.x values migration

Shubham-astro:
## Description

Both 0.37.x and 1.x serve Airflow Deployment PgBouncer client auth with `auth_type = scram-sha-256` — the upstream Airflow chart default since chart 1.12.0, which Astronomer first bundled in platform 0.37.4. With a plain-text
password in the PgBouncer auth file, PgBouncer derives a SCRAM secret using PBKDF2 (4096 iterations) on every client login, and PgBouncer is single-threaded. Both its probes are `tcpSocket` checks, so saturation produces no log output — just probe failures, restarts during peak runs, and idle server connections piling up on the metadata database.

Two customers have hit this after an upgrade: Walmart (PINF-569, Sept 2025) and a second account this week — a 7-hour outage with missed SLAs.

**This setting carries into 2.x unchanged.** Helm is still the default deployment mode (`deployments.mode.helm.enabled: true`, with `mode.operator.enabled` driven by `global.airflowOperator.enabled`, default `false`), and an existing Deployment cannot be converted from Helm mode to operator mode. So every Deployment that survives an upgrade stays Helm-rendered and `deployments.helm.airflow.pgbouncer.*` — `auth_type` and `extraIni` included — remains fully effective. Operator-mode Deployments (2.1+, opt-in, new Deployments only) are the exception: `addPgBouncerComponentToSpec` in houston-api maps no `auth_type` onto the Airflow CR and the operator uses `md5`.

**This PR changes no values and pins nothing.** The migrated output is byte-identical to before. It adds a single advisory, printed on every run of either 2.x migration script, covering:

- that `auth_type` and `extraIni` carry over and stay effective,
- the per-login PBKDF2 cost on a single-threaded proxy, and the resulting sizing   implication,
- that upgrading a Deployment re-renders and restarts its PgBouncer, so it should  happen outside scheduled batch windows
- the operator-mode exception.

The advisory prints whether or not an `auth_type` key is present, because the common case is a customer on the chart default with no key of their own — a mutation-triggered warning would never fire for them.

Applied to both `037x-to-2x` and `1x-to-2x`; shared logic lives in `helm_chart_values_migration_shared.py`. Internal upgrade docs updated to match.

Whether md5 should become the platform default is a separate decision, and it belongs in the chart or Houston defaults rather than in a migration script. This PR is correct either way: if the default flips, explicit customer overrides still win and the advisory simply becomes less urgent.

## Related Issues

- PINF-1221

## Merging

- `master`, `release-2.1`, `release-2.0`


bin/migrate-helm-chart-values-1x-to-2x.py
```diff
@@ -135,6 +137,7 @@ def main(argv: list[str] | None = None) -> int:
 
     yml, data = load_yaml(input_path)
     changes = migrate_values(data)
+    notes = collect_pgbouncer_auth_type_notes(data)
 
     if args.dry_run:
```
rishkarajgi, line 142: None of the four `print_migration_notes` call sites this `main()` gained are covered by a test — `test_migrate_helm_chart_values.py` doesn't reference pgbouncer or either new helper anywhere. The 037x script at least gets the collector function unit-tested directly; this one's advisory wiring could stop firing silently and nothing here would catch it.

rishkarajgi (commented):
This is a low-risk change — it prints an advisory and touches no migrated values, and the preservation tests back up the byte-identical-output claim. Two things I'd want addressed first:

- The 1x-to-2x script's new advisory-printing wiring has no test coverage anywhere, direct or indirect (see inline comment).
- The description is missing the Testing section the PR template asks for — worth a line on how the unchanged-output claim was verified.

Once those are in, this looks good to merge. Needs the usual CODEOWNERS sign-off before it lands.

bin/helm_chart_values_migration_shared.py
```diff
@@ -952,6 +953,72 @@ def apply_nginx_csp_policy_migrations(root: CommentedMap) -> list[MigrationChang
...
+    Returns:
+        Advisory note strings to print for the operator running the migration.
+    """
+    if not isinstance(root, CommentedMap):
+        return []
+
+    return [PGBOUNCER_AUTH_TYPE_NOTE]
+
+
+def print_migration_notes(notes: list[str], stream: Any = None) -> None:
```
Simpcyclassy, line 1004: Nice that this fires on every main() exit path (dry-run, in-place, --output, stdout) in both scripts, not just the happy path - that's an easy branch to miss when wiring an advisory in after the fact.

Simpcyclassy (changes requested):
The advisory is a real pre-upgrade stderr notice wired into every exit path of both migration scripts, not a NOTES.txt post-hoc message or a dead comment, and it changes zero rendered values as claimed. The 1x-to-2x wiring has no test coverage anywhere though, same gap rishkarajgi already flagged, and that's the upgrade path most customers actually take. Good scoping otherwise - land that missing test before merge.

bin/migrate-helm-chart-values-1x-to-2x.py
Simpcyclassy: Confirmed - grepped tests/migration/test_migrate_helm_chart_values.py and neither collect_pgbouncer_auth_type_notes nor print_migration_notes shows up anywhere in it. This is also the more common upgrade path per the incidents cited in the description, so I'd treat this as blocking rather than a nice-to-have.

PR #3522, opened 2026-08-27
APC-1590 - add laminar  sub chart

pgvishnuram:
## Description

This PR adds laminar component as a sub chart

## Related Issues

https://linear.app/astronomer/issue/APC-1590/lm-04-build-the-laminar-sub-chart-hypervisor-only

## Testing

QA Testing notes updated in the linear ticket

## Merging

merge to master and release-2.2


charts/laminar/templates/apiserver/networkpolicy.yaml
```diff
@@ -0,0 +1,33 @@
+##################################
+## Laminar API Server NetworkPolicies ##
+##################################
+{{- if eq (include "laminar.enabled" .) "true" }}
+{{- if .Values.global.networkPolicy.enabled }}
+{{- $port := .Values.apiServer.port }}
+{{- range $name, $ns := dict "allow-istio-ingress" "istio-system" "allow-symphony-ingress" "symphony" "allow-keda-ingress" "keda" }}
```
karankhanchandani, line 7: Where is namespace where astronomer is installed? I think commander is going to communicate with apiserver right?

charts/laminar/templates/_helpers.yaml
```diff
@@ -0,0 +1,123 @@
...
+{{- define "laminar.image" -}}
+{{- if .Values.global.privateRegistry.enabled -}}
+{{ .Values.global.privateRegistry.repository }}/ap-laminar:{{ .Values.images.laminar.tag }}
+{{- else -}}
+{{ .Values.images.laminar.repository }}:{{ .Values.images.laminar.tag }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Image pull secrets.
```
ianbuss, line 47: Since we will require auth for laminar in all cases (unless the user mirrors it to their own repo without auth) I am not sure if this should be gated only behind `privateRegistry`

charts/laminar/templates/apiserver/networkpolicy.yaml
ianbuss, line 7: There are some astro-specific things here. In fact they probably all are?

charts/laminar/templates/apiserver/rbac.yaml
```diff
@@ -0,0 +1,56 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: laminar-apiserver-airflows
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "laminar.serviceAccountName" (list $ "api-server") }}
+    namespace: {{ .Release.Namespace }}
+---
+# Allow the laminar apiserver to use the (cluster-scoped) TokenReview API to validate
+# tokens issued to KEDA for worker scaling metrics. See ADR-0009.
```
ianbuss, line 40: I don't think we need this now actually, we should be able to just use the OIDC endpoint from the API Server without this permission.

charts/laminar/templates/hypervisor/hypervisor-rbac.yaml
```diff
@@ -0,0 +1,38 @@
...
+  name: laminar-hypervisor
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "laminar.labels" (list $ "hypervisor") | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["serviceaccounts/token"]
```
ianbuss, line 18: I don't think we need this one either (serviceaccounts/token)?

charts/laminar/templates/secret.yaml
```diff
@@ -0,0 +1,17 @@
+###############################
+## Laminar database connection ##
+###############################
+{{- if eq (include "laminar.enabled" .) "true" }}
+{{- if and .Values.database.create .Values.database.connectionString }}
```
ianbuss, line 5: Would db-bootstrapper create this DB?

charts/laminar/values.yaml
```diff
@@ -0,0 +1,144 @@
...
+    # explicitly true: Harmony uses server-side apply
+    - name: LAMINAR_HYPERVISOR__CONFIGMAP_METRICS_ENABLED
+      value: "true"
+    # explicitly empty: Harmony uses server-side apply
+    - name: LAMINAR_HYPERVISOR__DISABLED_METRICS_CSV
+      value: ""
+    - name: LAMINAR_HYPERVISOR__DRY_RUN_HEALERS_CSV
+      value: "CatatonicWorkerTerminator"
+    - name: LAMINAR_HYPERVISOR__ENABLE_HEALERS
+      value: "True"
```
ianbuss, line 130: We can probably tidy this up a bit and just have configuration in the configmap? But by default we want to disable a bunch of stuff in APC's Laminar:

`LAMINAR_HYPERVISOR__ENABLE_HEALERS=False`

`LAMINAR_HYPERVISOR__CONFIGMAP_METRICS_ENABLED=False`

We can go through all the settings and check for sensible defaults for APC. [https://github.com/astronomer/laminar/blob/main/src/laminar/settings.py](<https://github.com/astronomer/laminar/blob/main/src/laminar/settings.py>)

charts/laminar/templates/apiserver/networkpolicy.yaml
ianbuss: We would need the `nginx` ingress namespace and `astronomer` for Commander

Chart.yaml
```diff
@@ -46,6 +46,12 @@ dependencies:
     tags:
       - monitoring
 
+  # APC Laminar
+  - name: laminar
```
Simpcyclassy, line 50: New subchart, zero coverage under tests/chart_tests. When airflow-operator and external-secrets were added they each got a dedicated test file (test_airflow_operator.py, test_external_secrets.py); laminar should get one too before this merges, at minimum a render test with the chart enabled/disabled.

Simpcyclassy (commented):
Several concrete asks from karankhanchandani and ianbuss, missing commander/nginx namespaces in the apiserver NetworkPolicy, healer defaults still True, broad token-review RBAC, are still unresolved at this commit. Plane-gating is solid (laminar.enabled locks everything to data/unified, matching the repo's eq(include ...) == "true" convention), but there's no test file under tests/chart_tests, unlike airflow-operator or external-secrets. It's still a draft with a placeholder description, so I'm commenting rather than approving or requesting changes.

charts/laminar/templates/apiserver/networkpolicy.yaml
Simpcyclassy: Still just istio-system/symphony/keda in the current apiserver NetworkPolicy (charts/laminar/templates/apiserver/networkpolicy.yaml:7). nginx-ingress and the astronomer/Commander namespace aren't in the allow-list yet, so real traffic would still get dropped as-is.

charts/laminar/values.yaml
Simpcyclassy: LAMINAR_HYPERVISOR__ENABLE_HEALERS is still "True" in the current values.yaml (line 129), doesn't look like this got flipped to the safer False default you asked for yet.

charts/laminar/templates/apiserver/networkpolicy.yaml
pgvishnuram: This is not at all needed for us

charts/laminar/templates/secret.yaml
pgvishnuram: yes adding new changes in the helm chart

charts/laminar/values.yaml
pgvishnuram: This is now cleaned up and removed unneeded values for APC

charts/laminar/templates/_helpers.yaml
ianbuss: Why was this resolved? The point still stands?

charts/laminar/templates/_helpers.yaml
```diff
@@ -0,0 +1,146 @@
...
+Call with (list $ "<component>"), e.g. (list $ "api-server").
+*/}}
+{{- define "laminar.fullname" -}}
+{{- $ := index . 0 -}}
+{{- $component := index . 1 -}}
+{{- printf "%s-%s" $.Release.Name $component | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+ServiceAccount name for a component. Mirrors prometheus.serviceAccountName.
```
ianbuss: Do we need "`Mirrors prometheus.serviceAccountName`"?

charts/laminar/templates/_helpers.yaml
```diff
@@ -0,0 +1,146 @@
...
+{{- define "laminar.env" -}}
+{{- $ := index . 0 -}}
+{{- $component := index . 1 -}}
+- name: LAMINAR_API_URL
+  value: http://0.0.0.0:8000/laminar
+- name: LAMINAR_KUBE_NAMESPACE
+  valueFrom:
+    fieldRef:
+      fieldPath: metadata.namespace
+- name: LAMINAR_DATABASE_URL
```
ianbuss, line 95: As of astronomer/laminar#3057 we can now specify the DB URL in a file via `LAMINAR_DATABASE_URL_FILE`, which can be dynamically updated. Laminar will restart if it detects a change.

charts/laminar/templates/apiserver/apiserver-rbac.yaml
ianbuss, line 1: Nit on file naming, should it be `apiserver-cluster-role.yaml`?

charts/laminar/templates/apiserver/apiserver-serviceaccount.yaml
```diff
@@ -0,0 +1,19 @@
+##################################
+## Laminar API Server ServiceAccount ##
+##################################
+{{- if eq (include "laminar.enabled" .) "true" }}
+{{- if .Values.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
```
ianbuss, line 8: If this is false, can Laminar work? How will it make Kube API calls — will some other process inject the token?

charts/laminar/templates/helm-hooks/laminar-bootstrapper-job.yaml
```diff
@@ -0,0 +1,74 @@
...
+    heritage: {{ .Release.Service }}
+    plane: {{ .Values.global.plane.mode }}
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-1"
+    "helm.sh/hook-delete-policy": before-hook-creation
+  {{- if .Values.global.argoCD.annotation.enabled }}
+    argocd.argoproj.io/sync-wave: "0"
```
ianbuss, line 24: Out of interest, why do we need argo annotations?

charts/laminar/templates/hypervisor/rbac.yaml
ianbuss, line 1: nit: can we align the file naming scheme for both apiserver and hypervisors

charts/laminar/templates/hypervisor/rbac.yaml
```diff
@@ -0,0 +1,96 @@
...
+      - schedulers
+      - statsds
+      - triggerers
+      - redis
+      - webservers
+      - pgbouncers
+      - dagprocessors
+      - eventschedulers
+      - apiservers
+    verbs: ["get"]
```
ianbuss: I will change this to the following upstream, as I can see us wanting to start watches on all of the other things too

```
  - apiGroups: ["airflow.apache.org"]
    resources: ["airflows"]
      - airflows
      - schedulers
      - statsds
      - triggerers
      - redis
      - webservers
      - pgbouncers
      - dagprocessors
      - eventschedulers
      - apiservers
    verbs: ["get", "list", "watch"]
  - apiGroups: ["airflow.apache.org"]
    resources:
      - airflows/status
    verbs: ["get"]
```

Nothing for you to do but it does raise the question of how we should keep these manifests in sync with upstream Laminar?

charts/laminar/values.yaml
```diff
@@ -0,0 +1,140 @@
+# Default values for the Laminar.
+
+# Node scheduling. Defaults fall back to global.platformNodePool when unset.
+nodeSelector: {}
+affinity: {}
+tolerations: []
+
+images:
+  laminar:
+    repository: core.harbor.astro-qa.link/astrocloud/laminar
```
ianbuss, line 10: Let's get this updated to `quay.io/ap-laminar` next week

charts/laminar/values.yaml
```diff
@@ -0,0 +1,140 @@
...
+    annotations: {}
+    # -- The name of the service account to use. If not set and create is true, a name is generated using the fullname template
+    name: ""
+    # ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
+    automountServiceAccountToken: true
+
+# ---------------------------------------------------------------------------
+# laminar.env config file mounted at /laminar.env (subPath laminar.env).
+# When create=true, a ConfigMap named `name` is templated from `content`.
+# When create=false, an existing ConfigMap named `name` is referenced.
```
ianbuss, line 49: Is there a scenario when we would want to support that?

charts/laminar/values.yaml
```diff
@@ -0,0 +1,140 @@
...
+      value: NEVER
+    - name: LAMINAR_HYPERVISOR__QUEUED_TASK_SECOND_THRESHOLD
+      value: "480"
+    - name: LAMINAR_APPLY_CUSTOM_DDL
+      value: "true"
+    # explicitly empty: Harmony uses server-side apply
+    - name: LAMINAR_HYPERVISOR__DISABLED_METRICS_CSV
+      value: ""
+    - name: LAMINAR_HYPERVISOR__DRY_RUN_HEALERS_CSV
+      value: "CatatonicWorkerTerminator"
```
ianbuss: Let's move all these to the `env` file?

tests/chart_tests/test_laminar.py
```diff
@@ -0,0 +1,41 @@
...
+    for x in Path(f"{git_root_dir}/charts/laminar/templates/").glob("**/*.yaml")
+    if not x.name.startswith("_")
+)
+
+
+@pytest.mark.parametrize(
+    "kube_version",
+    supported_k8s_versions,
+)
+class TestLaminar:
```
ianbuss, line 38: Any more tests we want to do? Perhaps around some config variations?

charts/laminar/templates/_helpers.yaml
ianbuss: We can do this as a follow up when a version of Laminar is released with this change in. We should plan to release a first APC candidate Laminar next week.

rishkarajgi (commented):
The plane-mode gating and default-off wiring look solid, but the new apiserver/hypervisor services ship with no NetworkPolicy at all in the current diff, which is the one thing I'd want fixed before this merges. A couple of smaller things worth a look too: the default laminar image points at what looks like an internal QA registry, and the plane-mode test never actually exercises `data` mode, the plane this component is meant to run on.

- No NetworkPolicy for the new apiserver/hypervisor services (inline comment on apiserver-service.yaml)
- Default image registry looks QA-only, not production (values.yaml)
- Plane-gating test skips `data` mode (test_laminar.py)

Since this adds new RBAC and network-exposed services, it'll need the mandatory security review before merge, on top of the usual CODEOWNERS approval.

charts/laminar/templates/apiserver/apiserver-service.yaml
```diff
@@ -0,0 +1,22 @@
...
+kind: Service
+apiVersion: v1
+metadata:
+  name: {{ include "laminar.fullname" (list $ "api-server") }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "laminar.labels" (list $ "apiserver") | nindent 4 }}
+    app.kubernetes.io/name: apiserver
+spec:
+  type: ClusterIP
```
rishkarajgi, line 14: Neither this service nor hypervisor-service.yaml has a NetworkPolicy alongside it — every other exposed service in this chart (houston, pgbouncer, commander, elasticsearch, etc.) ships one, and an earlier revision of this PR actually had one for the apiserver. With default-deny network policy on, this traffic just gets dropped.

charts/laminar/values.yaml
rishkarajgi, line 10: This looks like an internal QA registry host rather than a production one — every other image default in this file (dbBootstrapper right below) points at quay.io/astronomer. Is astro-qa.link intended as the shipped default once this reaches release-2.2?

tests/chart_tests/test_laminar.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_laminar_disabled_by_default(self, kube_version):
+        """Test that laminar is disabled by default."""
+        docs = render_chart(kube_version=kube_version)
+
+        laminar_docs = [
+            doc for doc in docs if str(doc.get("metadata", {}).get("labels", {}).get("chart", "")).startswith("laminar-")
+        ]
+        assert laminar_docs == []
+
+    @pytest.mark.parametrize("plane_mode", ["control", "unified"])
```
rishkarajgi, line 48: This only covers control (excluded) and unified (included) — data is the plane laminar is actually meant to run on per the comment on laminar.enabled in _helpers.yaml, and it's never rendered here.

charts/laminar/templates/_helpers.yaml
pgvishnuram: This is cleaned up now - resolving the comment

charts/laminar/templates/apiserver/apiserver-serviceaccount.yaml
pgvishnuram: we are following the same pattern for all the component to allows customers to toggle them for the components which they dont need kube api token

charts/laminar/templates/helm-hooks/laminar-bootstrapper-job.yaml
pgvishnuram: We support argoCD based deployments and this annotation is needed for hook ordering

tests/chart_tests/test_laminar.py
pgvishnuram: We want to cover each and every changes to be added as test cases

tests/chart_tests/test_laminar.py
pgvishnuram: Yes correct - we are yet to add elaborated tests for the sub changs

charts/laminar/values.yaml
pgvishnuram: This is now cleaned up and moved to env based config

charts/laminar/templates/hypervisor/rbac.yaml
pgvishnuram: This is now fixed

charts/laminar/values.yaml
pgvishnuram: This is now fixed

charts/laminar/templates/apiserver/apiserver-serviceaccount.yaml
ianbuss: Yes, but this component definitely does need a kube api token :)

tests/chart_tests/test_laminar.py
pgvishnuram: This is now fixed we have added some elaborative tests

tests/chart_tests/test_laminar.py
pgvishnuram: We have some elaborative tests

charts/laminar/templates/apiserver/apiserver-serviceaccount.yaml
pgvishnuram: This is a optional feature even for commander as well we have the same feature spec - we are following same pattern across all the services - if this is disabled services will break.

ianbuss (approved):
Approving as a first pass but we do need follow ups for the small nits (file renames etc).

PR #3523, opened 2026-08-31
APC-1615: Expose worker queue feature flag and per-deployment cap <DO NOT MERGE>

Shubham-astro:
## Description

This PR Adds global.workerQueues and renders it into the Houston ConfigMap, so platform admins can turn the Worker Queues feature on at install/upgrade time.
```
global:
  workerQueues:
    enabled: false
    maxPerDeployment: 10
```
## Related Issues

https://linear.app/astronomer/issue/APC-1615/wq-04-add-worker-queue-feature-flag-and-per-deployment-cap-config-keys

## Testing

- Added Testcases

## Merging

Master

### DIFF
--- a/charts/astronomer/templates/houston/houston-configmap.yaml
+++ b/charts/astronomer/templates/houston/houston-configmap.yaml
@@ -121,6 +121,10 @@ data:
         dagOnlyDeployment:
           enabled: {{ .Values.global.deployMechanisms.dagOnlyDeployment.enabled }}
 
+      workerQueues:
+        enabled: {{ .Values.global.workerQueues.enabled }}
+        maxPerDeployment: {{ .Values.global.workerQueues.maxPerDeployment }}
+
       {{- $grafanaLinkEnabled := true }}
       {{- if and (hasKey .Values.global "grafana") (kindIs "map" .Values.global.grafana) (hasKey .Values.global.grafana "enabled") }}
       {{- $grafanaLinkEnabled = .Values.global.grafana.enabled }}
--- a/values.yaml
+++ b/values.yaml
@@ -179,6 +179,18 @@ global:
       dagTarballVersionValidation:
         enabled: true
 
+  # Worker queues: multiple named Celery worker groups per Airflow deployment.
+  workerQueues:
+    enabled: false
+    # Maximum worker queues per deployment. This is not resource protection --
+    # per-queue CPU, memory and replica bounds already come from
+    # resourceManagement.components[workers]. The cap exists because every queue
+    # adds a full podTemplateSpec to the Airflow CR's spec.workers[] array, and a
+    # Kubernetes object has a hard ceiling of roughly 1.5 MiB. Past it the apply
+    # is rejected outright, so the deployment stops reconciling. Raise at your
+    # own risk.
+    maxPerDeployment: 10
+
   # Deploy mechanisms
   deployMechanisms:
     dagOnlyDeployment:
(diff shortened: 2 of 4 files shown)


rishkarajgi (commented):
This is a clean, self-contained addition — a new global.workerQueues.{enabled, maxPerDeployment} value threaded into the Houston ConfigMap, off by default, with render tests covering both the default and an overridden cap. I checked it against the companion houston-api PR (APC-1615, houston-api#2929) and the key names/defaults match on both sides.

Nothing to flag before merge.

values.yaml
```diff
@@ -179,6 +179,18 @@ global:
       dagTarballVersionValidation:
         enabled: true
 
+  # Worker queues: multiple named Celery worker groups per Airflow deployment.
```
rishkarajgi, line 182: do we need to introduce this in values.yaml as we are now managing dataplane configs in the cluster config houston schema?

Shubham-astro: @rishkarajgi I am okay with controlling this. only from houston, but with no chart value, turning worker queues on becomes an updateCluster mutation after Houston is up, rather than a --set global.workerQueues.enabled=true during the upgrade.

- The flag exists to decouple the CR-builder switch from the upgrade that backfills each deployment's default queue. Making enablement a separate deliberate act after the upgrade is the behaviour we want. 
 
Let convert this PR to draft for now and not merge it and i can test with just cluster config only.

charts/astronomer/templates/houston/houston-configmap.yaml
```diff
@@ -121,6 +121,10 @@ data:
         dagOnlyDeployment:
           enabled: {{ .Values.global.deployMechanisms.dagOnlyDeployment.enabled }}
 
+      workerQueues:
+        enabled: {{ .Values.global.workerQueues.enabled }}
```
karankhanchandani, line 125: This requires operator mode to be enabled as well. Can we do this?

```
enabled: {{ .Values.global.airflowOperator.enabled && .Values.global.workerQueues.enabled }}
```

values.yaml
```diff
@@ -179,6 +179,18 @@ global:
...
+  workerQueues:
+    enabled: false
+    # Maximum worker queues per deployment. This is not resource protection --
+    # per-queue CPU, memory and replica bounds already come from
+    # resourceManagement.components[workers]. The cap exists because every queue
```
Simpcyclassy, line 187: Good call documenting why the cap exists (K8s object size ceiling on spec.workers[], not resource protection) - saves the next person from assuming this is an arbitrary limit and trying to raise it freely.

Simpcyclassy (commented):
Karankhanchandani's gating question is legit since spec.workers fanout only lives in houston-api's operator path, but the fix belongs in houston-api validation, not an && here. Defaults, plane-mode gating, and the two new render tests all look correct and this is purely additive for existing installs. Marked draft and DO NOT MERGE so I'm treating this as early feedback, not a blocker.

charts/astronomer/templates/houston/houston-configmap.yaml
Simpcyclassy: Traced this through houston-api - the spec.workers[] fanout only exists in the operator code path (src/lib/deployments/operator/index.js) and upsert-deployment's workerQueuesEnabled check never looks at operator mode, so the concern holds up. I'd gate it in houston-api validation rather than here though, since `&&` isn't valid Helm/Sprig syntax anyway and ANDing the flags in this configmap would hide each flag's real state from other consumers.

PR #3524, opened 2026-09-01
PLA-416: Thread split_software_automation_tests through the stage-cluster trigger

sreenusuuda:
## Description

1. Adds a split_software_automation_tests boolean param to the trigger-stage-cluster-build job and threads it through bin/trigger_stage_cluster_update.py into the workflow_extra_params_json payload sent to terraform-aws-astronomer's feature_stack/stage_cluster workflow generator.

2. The release-branch stage-cluster build (build-and-release-helm-chart → trigger-stage-cluster-build) now passes split_software_automation_tests: true, so the downstream automation run splits into p0 and p1/p2 suites instead of one combined run.

3. No behavior change for any other caller — the new param defaults to false everywhere else.

## Related Issues

https://linear.app/astronomer/issue/PLA-416/tag-the-priorities-to-the-all-test-cases-and-run-only-p0-as-part-of


bin/trigger_stage_cluster_update.py
```diff
@@ -123,11 +124,13 @@ def main(circleci_token: str, astro_path: str, branch: str):
     arg_parser.add_argument("--circleci_token", type=str, required=True)
     arg_parser.add_argument("--astro_path", type=str, required=True)
     arg_parser.add_argument("--branch", type=str, required=True)
+    arg_parser.add_argument("--split_software_automation_tests", type=str, required=False, default="false")
```
danielhoherd: This should be

```suggestion
    arg_parser.add_argument("--split_software_automation_tests", action="store_true")
```

https://docs.python.org/3/library/argparse.html#action

danielhoherd: Also these could all use `help="description of what the argument does"`

danielhoherd (changes requested):
We should improve the argparse logic before merging this. The `store_true` syntax is what we should be using for boolean args.

bin/trigger_stage_cluster_update.py
sreenusuuda: addressed

.circleci/config.yml.j2
```diff
@@ -197,10 +200,11 @@ jobs:
           name: Triggering update to stage cluster
           command: |
             set -e
-            bin/trigger_stage_cluster_update.py \
-              --astro_path=/tmp/workspace/<< parameters.artifact_subdir >> \
-              --circleci_token=$CIRCLECI_API_KEY_TEST \
-              --branch=$CIRCLE_BRANCH
+            args=(--astro_path=/tmp/workspace/<< parameters.artifact_subdir >> --circleci_token=$CIRCLECI_API_KEY_TEST --branch=$CIRCLE_BRANCH)
+            if [ "<< parameters.split_software_automation_tests >>" = "true" ]; then
```
danielhoherd, line 204: This uses `/bin/sh` syntax, but `args=` is a bash only feature, so we must be using bash. So let's make this also bash native syntax


```suggestion
            if [[ "<< parameters.split_software_automation_tests >>" == "true" ]] ; then
```

danielhoherd (approved):
Looks good. I left one more optional improvement that shouldn't block merging this.

rishkarajgi (commented):
Both of Daniel's earlier comments look addressed in the current head (`store_true` on the new arg, `help=` on all four). The one outstanding suggestion (bash-native `[[ ]]` vs the POSIX `[ = ]` test) is already marked optional/non-blocking on the thread, and it is: `args=(...)` on the line above is already a bash-only construct, so the block is bash regardless of which test syntax follows — a style-consistency point, not a correctness one.

Traced the actual payload rather than just the description's claim: `split_software_automation_tests` goes into the same top-level `parameters` dict as `workflow_gen`/`workflow_name` (not nested inside `workflow_extra_params_json`), which matters because terraform-aws-astronomer#224 declares it as a top-level pipeline parameter with the same name — a mismatch here (e.g. nesting it under the extra-params JSON) would have made the value silently unreachable on the other side. It lines up.

Nothing here blocking merge.

PR #3529, opened 2026-09-04
Replace software-upgrade-automation with apc-automation

danielhoherd:
- Rename software-upgrade-automation
- Replace deprecated prek `auto_update` key
- Update pre-commit hooks
- `prek run --all-files`


Simpcyclassy (approved):
Verified the `auto_update` to `update` key rename matches prek 0.5's actual config change, and the ruff/shellcheck-py rev bumps point at tags that exist. Grepped the repo for leftover software-upgrade-automation references and only found the one docstring this PR already fixes, no CI or image references to worry about. Small, low-risk maintenance change, approving.

PR #3531, opened 2026-09-07
PINF-1236: Fix sign-released-image version resolution

Shubham-astro:
## Description

PR Fixes `sign-released-image`: derive version from CIRCLE_TAG, run after release_bom.

## Changes
- Moved sign-released-image out of build-and-release-helm-chart (branch-filtered) into release-bom-workflow (tag-filtered), with requires: release_bom so the BOM exists before signing reads it.
- Version now comes from CIRCLE_TAG with the leading v stripped (--version "${CIRCLE_TAG#v}"), matching the BOM's chart-version key.
- Removed the Set image tag to sign step and its ineffective exit 0 guard. If CIRCLE_TAG is somehow unset the job now fails loudly instead of falling through to an empty --version.

## Related Issues

https://linear.app/astronomer/issue/PINF-1236/release-pipeline-release-bom-and-sign-released-image-failures-root

## Testing

- Not verifiable pre-merge: an actual signing run requires a real release tag, since the job is now tag-triggered by design.

## Merging

Main, 1.x, 2.x

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,28 +32,24 @@ jobs:
             echo "$DOCKER_PASSWORD" | cosign login docker.io --username $DOCKER_USERNAME --password-stdin
             echo "$QUAY_PASSWORD" | cosign login quay.io --username $QUAY_USERNAME --password-stdin
       - run:
-          name: Set image tag to sign
-          command: |
-            set -xe
-            # Only sign tagged releases or specific QA releases
-            # For regular release job
-            if [ -n "$NEXT_TAG" ]; then
-              echo "export IMAGE_TAG=$NEXT_TAG" >> $BASH_ENV
-            # For custom QA release job
-            elif [ -n "$IMG_TAG" ]; then
-              echo "export IMAGE_TAG=$IMG_TAG" >> $BASH_ENV
-            else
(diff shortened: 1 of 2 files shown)


rishkarajgi (commented):
This just moves sign-released-image to run after release_bom in the tag-triggered workflow and derives its version from CIRCLE_TAG instead of the old NEXT_TAG/IMG_TAG dance, with a hard failure instead of a silent skip. Checked the version-stripping and BOM-URL logic against bin/sign-images.py and bin/release-bom — they agree. One heads up: PR #3453 adds a new job right after this same spot in the file, so whichever of these merges second will need a small conflict resolution there.

Simpcyclassy (approved):
Checked the stripped CIRCLE_TAG against bin/release-bom and bin/sign-images.py directly: both land on the same astronomer-<version>.json name, with or without an rc/alpha suffix, so the version resolution is correct. Agree with rishkarajgi that this is a clean fix and the fail-loud change is the right call over the old silent exit 0; worth resolving the noted conflict with #3453 at merge time in the workflow block. Also confirmed config.yml and config.yml.j2 stay in lockstep, so the repo's own consistency check will pass.

PR #3532, opened 2026-09-07
APC-1771: Make the Houston database schema configurable instead of hardcoded

piyush-shrivastava:
## Description

Installing APC fails on a database that does not permit creating the `houston$default` schema, because the chart hardcodes that name in seven places. houston-api's half of this shipped in astronomer/houston-api#2955, which stripped the schema qualifier from five migrations and from the one schema-sensitive raw query. That unblocked customers who bring their own `backendSecretName` and so supply their own connection string. This is the chart half: it makes the schema a value instead of a literal.

New value `global.houston.database.schemaName`, defaulting to `"houston$default"`. It sits under `global` because the `astronomer`, `prometheus` and `prometheus-postgres-exporter` subcharts all need it, and Helm does not share one subchart's values with another. Threaded through:

| Location | What reads it |
| --- | --- |
| `houston-deployment.yaml` | `SCHEMA_NAME` |
| `houston-worker-deployment.yaml` | `SCHEMA_NAME` |
| `houston-db-migration-job.yaml` | `SCHEMA_NAME` |
| `houston-cp-refresh-job.yaml` | `SCHEMA_NAME` |
| `houston-upgrade-deployments-job.yaml` | `SCHEMA_NAME` |
| `prometheus-statefulset.yaml` | `DATABASE_SCHEMA_NAME` |
| `prometheus-postgres-exporter` queries | `SET search_path` |
| `houston-backend-secret.yaml` | `?schema=` in the connection URL |

After this, the only remaining occurrences of the literal in the chart are the default itself and the comment above it.

The five `SCHEMA_NAME` values feed the `houston-bootstrapper` init container, which is what runs `CREATE SCHEMA` and writes the schema into the connection secret Houston reads. I confirmed against `ap-db-bootstrapper:2.0.13` that it already honours arbitrary values: `SCHEMA_NAME` is a `click` option with `envvar="SCHEMA_NAME"`, and running the image against a scratch Postgres with `SCHEMA_NAME=my_custom` creates `my_custom` and produces `?schema=my_custom`. So no bootstrapper change is needed for this, and PINF-275 is not a prerequisite.

I also dropped five stale `# we need to hardcode it to be able to upgrade from prisma1 to prisma2` comments, which no longer described anything true.

## The backend-secret fix is a behaviour change, please look at it

> [!IMPORTANT]
> `houston-backend-secret.yaml` held the `?schema=houston%24default` this ticket is about, and it was both **unreachable and malformed**. Fixing it changes when the template renders, so review that guard rather than reading this as a find-and-replace.

What it rendered before:

```
postgresql://%!s(<nil>):%!s(<nil>)@%!s(<nil>):<nil>/%!s(<nil>)?schema=houston%!d(MISSING)efault
```

Two independent bugs:

- **`%24` was eaten by printf.** In `printf "...?schema=houston%24default"`, Go reads `%24d` as a width-24 `%d` verb, consumes a sixth argument that does not exist, and emits `%!d(MISSING)`. The string `houston%24default` was never once produced.
- **The clause was mutually exclusive with its own data.** Line 5 only rendered the secret when `houston.backendConnection` was *empty*, and the clause then read `.Values.houston.backendConnection.user/pass/host/port/db`. Every field was nil whenever the line ran.

Now:

```
postgresql://houston:s3cr3t@pg.example.com:5432/houston?schema=houston%24default
```

The guard became `or (not backendConnection) backendSecretConnection`, which is deliberately surgical: it opens only the new working path.

**All three pre-existing render paths are unchanged**, which is the part worth checking:

| Values | Before | After |
| --- | --- | --- |
| nothing set | random placeholder | random placeholder |
| `backendSecretName` set | template skipped | template skipped |
| `backendConnection` without `backendSecretConnection` | template skipped | template skipped |
| `backendConnection` + `backendSecretConnection` | **template skipped** | **real connection URL** |

Only the last row changes, and it previously produced no secret at all.

## Review feedback addressed

**Credentials are now percent-encoded** (`aa73d79`). Interpolating `user`, `pass` and `db` verbatim was pre-existing, but making the clause reachable made it live. It is worse than misparsing: feeding the old form to a URL parser raises `Port could not be cast to integer value as 'p'`, because the host:port split lands inside the password. All three components now go through a new `houston.uriComponent` helper, which is `urlquery` piped through `replace "+" "%20"` — `urlquery` alone renders a space as `+`, and a `+` in userinfo is a literal plus rather than a space. sprig has no `urlEncode` in this Helm version. The schema keeps plain `urlquery`, where `+` for space is correct.

**The postgres exporter now follows the value** (`aa73d79`). My original claim that it "degrades gracefully" was wrong. Postgres does tolerate a missing schema in `search_path`, but that is beside the point: with the tables in `my_custom` and `search_path` set to `houston$default, public`, the table is not found and the query errors. Only a `public` override would have survived, by coincidence. The exporter configmap now uses `tpl` instead of `printf` — the idiom this repo already uses in 83 other places — so the query can read the value. The identifier is also quoted now, and I confirmed against real Postgres that `SET search_path TO "houston$default", public` still resolves an unqualified table.

**The value moved under a `database` namespace** (`4c528e7`). Now `global.houston.database.schemaName`, so future database values have somewhere to go. It has to stay under `global`: I checked with a scratch parent chart that from inside the `prometheus` subchart, a `global.*` key resolves while an `astronomer.houston.*` key comes back empty. Nesting it under the astronomer subchart would have worked for the five `SCHEMA_NAME` sites and silently failed for prometheus and the exporter.

**Ticket references removed from comments** (`4c528e7`). All six, including the two in test docstrings.

## Related Issues

- Fixes [APC-859](https://linear.app/astronomer/issue/APC-859/remove-unnecessary-houstondollardefault-schema-from-database) / astronomer/issues#8024
- Parent: [APC-1771](https://linear.app/astronomer/issue/APC-1771/apc-installation-and-db-migration-fails-if-schema-is-not)
- houston-api half: astronomer/houston-api#2955 (`main`), astronomer/houston-api#2972 (`release-2.1`)
- [PINF-275](https://linear.app/astronomer/issue/PINF-275/need-way-to-configure-the-schema-name-in-db-bootstrapper) — **not a prerequisite**, see above. Worth retitling: the schema name is already configurable in db-bootstrapper; the real defect is that omitting `SCHEMA_NAME` creates a schema literally named `None`, because `CREATE SCHEMA IF NOT EXISTS "{schema_name}"` has no falsy guard.

## Known issue found during review, not fixed here

The `randAlphaNum 5` fallback in `houston-backend-secret.yaml` can clobber a working connection secret. `lookup` returns an empty dict whenever Helm cannot query the cluster — `helm template`, `--dry-run`, and any GitOps flow that applies rendered output — so `$existingConnection` is empty and the else branch writes a five-character random value as the entire connection string. A real `helm upgrade` preserves it correctly because `lookup` works, which is why this would only surface in the render-then-apply path. This template is already ArgoCD-aware (it carries the sync-wave annotation), so that path is supported.

Left alone deliberately: the guard change here does not alter which value combinations reach that branch, and it deserves its own change rather than a footnote in a schema PR. Being tracked separately.

## Testing

**Full chart suite: 6464 passed**, exit 0.

```bash
pytest tests/chart_tests -q -n 4
```

`helm lint` clean with the default and with `schemaName=my_custom` plus the exporter enabled. A full render under both settings produces 129 documents each, with zero `%!` verbs and zero `<nil>` values.

New `tests/chart_tests/test_houston_db_schema.py` covers the cross-cutting contract in one place, since it spans three subcharts: the default on every template, the override on every template, the HA-gated cp-refresh job separately, the exporter queries with the exporter enabled, and a whole-release consistency check that fails if any component disagrees. Extended `test_houston_backend_secret.py` with the connection URL, a custom schema, `$` encoding, credential encoding for `/ ? # @` and a space, and the without-opt-in skip.

Three things I checked rather than assumed:

- **The default did not move.** The pre-existing `env_vars["DATABASE_SCHEMA_NAME"] == "houston$default"` assertion in `test_prometheus_statefulset.py` passes untouched, so I added coverage instead of editing it. That test failing would have been the signal that upgrades break.
- **The encoded URLs round-trip.** Each asserts back through `urlparse` to the values that were configured, not just to an expected string.
- **Both new guards are non-vacuous.** Reintroducing the literal in one template fails the consistency check with `schema values disagree across the release: {'public', 'houston$default'}`; restoring the hardcoded `search_path` fails the exporter test. Reverting each restores green.

Not covered: a live install on a non-default schema. The values-generation half is covered above, and the admission half was verified on the parent ticket against real Postgres (fresh installs on `public`, `my_custom` and `houston$default`, plus the 2.0.x upgrade path and `pg_dump` schema equivalence).

## Merging

`master`. APC-859 carries the **2.1.1** label, so this wants cherry-picking to `release-2.1` as well, alongside astronomer/houston-api#2972.

Backwards compatible by design: the default preserves current behaviour exactly, so existing installs are unaffected and only a fresh install would set the new value. Note that *changing* it on an existing install points Houston at an empty schema, which is why `values.yaml` documents it as fresh-install-only rather than a tunable.

### DIFF
--- a/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
+++ b/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
@@ -2,7 +2,15 @@
 ## Houston Bootstrap Secrets ##
 ###############################
 {{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
-{{- if and (not .Values.houston.backendSecretName) (not .Values.houston.backendConnection) }}
+{{- /*
+  Rendered when the customer is not bringing their own secret. The
+  backendSecretConnection clause below builds the connection string from
+  backendConnection, so allow that combination through too -- it used to be
+  excluded by the (not .Values.houston.backendConnection) test, which made the
+  clause unreachable (APC-859). Supplying backendConnection without the opt-in
+  still skips this template, as before.
+*/}}
+{{- if and (not .Values.houston.backendSecretName) (or (not .Values.houston.backendConnection) .Values.houston.backendSecretConnection) }}
 {{- $existingSecret := (lookup "v1" "Secret" .Release.Namespace (include "houston.backendSecret" .)) | default dict }}
 {{- $existingConnection := dig "data" "connection" "" $existingSecret }}
 kind: Secret
@@ -22,8 +30,14 @@ metadata:
 type: Opaque
 data:
   {{- if .Values.houston.backendSecretConnection }}
-  # in prisma1 has hardcoded schema, so for now to migrate to prisma2 we need to hardcode it too.
-  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db) | b64enc | quote }}
+  {{- /*
+    The schema goes through urlquery because Prisma reads it from this query
+    parameter, and the default contains a "$" that has to arrive as %24. Passing
+    it as a %s argument rather than writing it into the format string also keeps
+    printf from reading the escape as a width specifier -- "%24default" was
+    parsed as a width-24 %d verb and rendered "%!d(MISSING)".
+  */}}
+  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=%s" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db (urlquery .Values.global.houston.schemaName)) | b64enc | quote }}
   {{- else if $existingConnection }}
   connection: {{ $existingConnection | quote }}
   {{- else }}
--- a/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
+++ b/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
@@ -31,13 +31,21 @@ type: Opaque
 data:
   {{- if .Values.houston.backendSecretConnection }}
   {{- /*
-    The schema goes through urlquery because Prisma reads it from this query
-    parameter, and the default contains a "$" that has to arrive as %24. Passing
-    it as a %s argument rather than writing it into the format string also keeps
-    printf from reading the escape as a width specifier -- "%24default" was
-    parsed as a width-24 %d verb and rendered "%!d(MISSING)".
+    Every interpolated component is percent-encoded, because a "/", "?", "#" or
+    "@" in a password would otherwise be read as URI structure and the
+    connection would fail. This is also what the bootstrapper expects of a
+    connection string it is handed -- it calls unquote() on the password.
+
+    userinfo and the path use houston.uriComponent rather than urlquery, since
+    urlquery renders a space as "+" and a "+" in userinfo is a literal plus. The
+    schema keeps urlquery because it really is a query parameter.
+
+    Passing the schema as a %s argument rather than writing it into the format
+    string also keeps printf away from the escape -- "%24default" was parsed as
+    a width-24 %d verb and rendered "%!d(MISSING)".
   */}}
-  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=%s" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db (urlquery .Values.global.houston.schemaName)) | b64enc | quote }}
+  {{- $conn := .Values.houston.backendConnection }}
+  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=%s" (include "houston.uriComponent" $conn.user) (include "houston.uriComponent" $conn.pass) $conn.host ($conn.port | toString) (include "houston.uriComponent" $conn.db) (urlquery .Values.global.houston.schemaName)) | b64enc | quote }}
   {{- else if $existingConnection }}
   connection: {{ $existingConnection | quote }}
   {{- else }}
--- a/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
+++ b/charts/astronomer/templates/houston/api/houston-backend-secret.yaml
@@ -7,7 +7,7 @@
   backendSecretConnection clause below builds the connection string from
   backendConnection, so allow that combination through too -- it used to be
   excluded by the (not .Values.houston.backendConnection) test, which made the
-  clause unreachable (APC-859). Supplying backendConnection without the opt-in
+  clause unreachable. Supplying backendConnection without the opt-in
   still skips this template, as before.
 */}}
 {{- if and (not .Values.houston.backendSecretName) (or (not .Values.houston.backendConnection) .Values.houston.backendSecretConnection) }}
@@ -45,7 +45,7 @@ data:
     a width-24 %d verb and rendered "%!d(MISSING)".
   */}}
   {{- $conn := .Values.houston.backendConnection }}
-  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=%s" (include "houston.uriComponent" $conn.user) (include "houston.uriComponent" $conn.pass) $conn.host ($conn.port | toString) (include "houston.uriComponent" $conn.db) (urlquery .Values.global.houston.schemaName)) | b64enc | quote }}
+  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=%s" (include "houston.uriComponent" $conn.user) (include "houston.uriComponent" $conn.pass) $conn.host ($conn.port | toString) (include "houston.uriComponent" $conn.db) (urlquery .Values.global.houston.database.schemaName)) | b64enc | quote }}
   {{- else if $existingConnection }}
   connection: {{ $existingConnection | quote }}
   {{- else }}
--- a/values.yaml
+++ b/values.yaml
@@ -15,6 +15,16 @@ global:
     mode: "unified"  # one of: control, data, unified
     domainPrefix: ""    # cluster-id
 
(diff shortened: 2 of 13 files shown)


charts/astronomer/templates/houston/api/houston-backend-secret.yaml
```diff
@@ -22,8 +30,14 @@ metadata:
...
-  # in prisma1 has hardcoded schema, so for now to migrate to prisma2 we need to hardcode it too.
-  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=houston%24default" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db) | b64enc | quote }}
+  {{- /*
+    The schema goes through urlquery because Prisma reads it from this query
+    parameter, and the default contains a "$" that has to arrive as %24. Passing
+    it as a %s argument rather than writing it into the format string also keeps
+    printf from reading the escape as a width specifier -- "%24default" was
+    parsed as a width-24 %d verb and rendered "%!d(MISSING)".
+  */}}
+  connection: {{ (printf "postgresql://%s:%s@%s:%s/%s?schema=%s" .Values.houston.backendConnection.user .Values.houston.backendConnection.pass .Values.houston.backendConnection.host (.Values.houston.backendConnection.port | toString) .Values.houston.backendConnection.db (urlquery .Values.global.houston.schemaName)) | b64enc | quote }}
```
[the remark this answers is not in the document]
piyush-shrivastava: Good catch. 
I fed the old unencoded form to a URL parser and it rejects it outright with "Port could not be cast to integer value as 'p'", because the host:port split lands inside the password.
Fair point that this only became live because I made the branch reachable, so I have fixed it here.

values.yaml
```diff
@@ -15,6 +15,16 @@ global:
...
 
+  # Postgres schema Houston's tables live in. Global because both the astronomer
+  # and prometheus subcharts need it.
+  #
+  # Existing installs must leave this alone: their tables are already in
+  # houston$default, and changing it points Houston at an empty schema.
+  # Set it only on a fresh install, e.g. to "public" when the database does not
+  # permit creating schemas (APC-859).
+  houston:
+    schemaName: "houston$default"
```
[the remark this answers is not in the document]
piyush-shrivastava: Fixed by switching the exporter configmap from printf to tpl, which is the idiom this repo already uses,so the query can read `global.houston.schemaName`. 

added tests that render the exporter enabled with both the default and my_custom.

charts/astronomer/templates/houston/api/houston-backend-secret.yaml
karankhanchandani, line 52: @danielhoherd Remember we used to see random character in connection string that was difficult to reproduce? I think this is that path.

values.yaml
karankhanchandani: Should we have it under database?

```
houston.database.schemaName
```

I think it would help having it under database because we can then configure any other values as well.

charts/astronomer/templates/houston/api/houston-backend-secret.yaml
```diff
@@ -2,7 +2,15 @@
 ## Houston Bootstrap Secrets ##
 ###############################
 {{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
-{{- if and (not .Values.houston.backendSecretName) (not .Values.houston.backendConnection) }}
+{{- /*
+  Rendered when the customer is not bringing their own secret. The
+  backendSecretConnection clause below builds the connection string from
+  backendConnection, so allow that combination through too -- it used to be
+  excluded by the (not .Values.houston.backendConnection) test, which made the
+  clause unreachable (APC-859). Supplying backendConnection without the opt-in
```
karankhanchandani: We should not have ticket references like APC-859 in the comments. The comments should be self contained.

values.yaml
piyush-shrivastava: Good idea on the database namespace, but `houston.database.schemaName` on its own would break two of
the consumers. 

That key lives under the astronomer subchart, and Helm does not share one subchart's values with another, so prometheus and prometheus-postgres-exporter cannot see it.

I went ahead with `global.houston.database.schemaName`. 
Global so that every subchart sees it, nested under database so that other database values have somewhere to go.

charts/astronomer/templates/houston/api/houston-backend-secret.yaml
piyush-shrivastava: Removed all six refs, including the two in test docstrings.

However, this is something that we'll need to clean up across other repos as well.
will create a separate ticket to remove such refs from comments as well as docstrings

rishkarajgi (commented):
Verified the schema-name plumbing across all five bootstrapper sites plus the exporter and backend-secret paths, and the guard/encoding fixes from the earlier review rounds hold up under the actual Go-template semantics (percent-encoding, tpl vs printf, global value propagation into the prometheus subcharts). Nothing here clears a blocking bar. One thing worth a second set of eyes given it touches a DB-credential secret: the non-default-schema path was validated against real Postgres on the parent ticket rather than in this diff's own tests, so that's worth confirming happened rather than taking on faith.

charts/astronomer/templates/houston/api/houston-backend-secret.yaml
danielhoherd: Great find! I think you're right! Lines 14,15 in the new branch show a race condition where if the secret has not been created yet, `$existingConnect` ends up empty, and would then fall all the way through to this line.

@karankhanchandani do you know if these changes fix that race condition? At first glance it seems like not, but hey, at least we would know where to do the work.

Simpcyclassy (approved):
Traced the backend-secret guard and the new houston.uriComponent helper directly against the base commit; the old branch really was unreachable and the %24-in-printf bug was real, both are fixed correctly and the randAlphaNum fallback that worried karankhanchandani is untouched pre-existing behavior, not something this PR makes worse. Agree with rishkarajgi's review that the schema plumbing holds up across all five bootstrapper sites plus the exporter and backend-secret paths, and the default staying "houston$default" everywhere keeps existing installs safe on upgrade. Test coverage (test_houston_db_schema.py) actually sweeps every consumer and checks whole-release consistency, approving.

PR #3533, opened 2026-09-08
Feature/no secrets as env vars

fhaynes:
## Description

## Related Issues

## Testing

## Merging

### DIFF
--- a/charts/astronomer/templates/houston/api/houston-vector-configmap.yaml
+++ b/charts/astronomer/templates/houston/api/houston-vector-configmap.yaml
@@ -4,6 +4,7 @@
 ##########################################
 {{- if .Values.houston.logging.loggingSidecar.enabled }}
 {{- include "houston.logging.loggingSidecar.validate" . }}
+{{- $vectorSecretsFromFiles := eq "true" (include "secretsFromFiles.enabled" (dict "ctx" $ "component" .Values.houston.logging.loggingSidecar)) }}
 {{- if or (eq .Values.global.plane.mode "control") (eq .Values.global.plane.mode "unified") }}
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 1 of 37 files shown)


Simpcyclassy (commented):
Traced every path this touches (houston family, commander, pilot, navigator, postgres, prometheus, es-proxy) and the secret material never lands in an env var, log, or ConfigMap once the flag is on; Vector's SECRET[] backend and the projected-volume mounts are the right call. Good catch pulling the dead FEDERATION_AUTH_TOKEN env var too, and the new test suite actually sweeps the whole rendered chart for leftover secretKeyRefs. Still a draft with the placeholder PR template, so commenting rather than approving until it's marked ready.

charts/astronomer/templates/houston/api/houston-vector-configmap.yaml
```diff
@@ -22,6 +23,27 @@ data:
...
       enabled: true
       address: "0.0.0.0:8686"
+{{- $cwSecret := and $vectorSecretsFromFiles .Values.houston.logging.loggingSidecar.cloudwatch.enabled (not .Values.houston.logging.loggingSidecar.cloudwatch.useIRSA) }}
+{{- $esSecret := and $vectorSecretsFromFiles .Values.houston.logging.loggingSidecar.elasticsearch.enabled (eq .Values.houston.logging.loggingSidecar.elasticsearch.auth.strategy "basic") }}
+{{- if or $cwSecret $esSecret }}
+
+    # Sink credentials are read from mounted files rather than the environment.
+    # remove_trailing_whitespace is required: without it the newline a Kubernetes
+    # Secret mount adds becomes a trailing space inside the quoted YAML scalar.
+    secret:
```
Simpcyclassy, line 33: Nice use of Vector's native `secret` directory backend + `SECRET[...]` interpolation here — the credential value itself never has to pass through the ConfigMap, only a reference to it.

PR #3534, opened 2026-09-08
PINF-1231 git sync relay

danielhoherd:
## Description

ap-git-sync-relay:0.6.0 which causes updates to be serialized, avoiding contention.

## Related Issues

https://linear.app/astronomer/issue/PINF-1231

## Testing

ap-git-sync-relay now has tests unit test that exercises the implementation logic by doing two successive commits, the first one slow, and asserts against the result. Further testing should do similar type of tests: make multiple quick commits, pushing after each one, and ensure that `ERROR` never shows up in the git-sync-relay logs.

## Merging

Master, release-2.1

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 ---
 exclude: '(venv|\.vscode|tests/k8s_schema|tests/chart_tests/test_data)' # regex
-auto_update:
+update:
   cooldown_days: 7
 repos:
   - repo: local
@@ -53,7 +53,7 @@ repos:
       - id: remove-tabs
         exclude_types: [makefile, binary]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.16.0"
+    rev: "v0.16.5"
     hooks:
       - id: ruff-check
         args:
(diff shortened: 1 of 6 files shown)


rishkarajgi (approved):

values.yaml
```diff
@@ -440,7 +440,7 @@ global:
         tag: 2026.8.24
       gitSync:
         repository: quay.io/astronomer/ap-git-sync-relay
-        tag: 0.5.2
+        tag: 0.6.0
```
Simpcyclassy, line 443: Manual pin of ap-git-sync-relay outside the apc-bot drift process. Reasonable here since it's tied to the PINF-1231 serialization fix, but flagging per our image-pin convention since it bypasses the normal automated bump.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 exclude: '(venv|\.vscode|tests/k8s_schema|tests/chart_tests/test_data)' # regex
-auto_update:
+update:
```
Simpcyclassy, line 3: This renames auto_update to update, but houston-api's .pre-commit-config.yaml still uses auto_update for the same cooldown_days block. If this key is read by shared update tooling, the rename may silently stop it from recognizing this repo's auto-update config — and it's unrelated to the git-sync-relay change this PR is about.

Simpcyclassy (approved):
The ap-git-sync-relay tag is manually pinned to 0.6.0 here instead of going through apc-bot, which is fine given it ties to the PINF-1231 serialization fix but is worth calling out as a process exception. The pre-commit-config.yaml rename of auto_update to update looks unrelated to git-sync-relay and diverges from houston-api, which still uses auto_update for the same cooldown_days block. Nothing here breaks chart rendering or upgrade safety, so approving with those two flagged for a second look.

PR #3535, opened 2026-09-09
PINF-955: Vector Daemonset improvements

Shubham-astro:
## Description

This PR:

- **Scopes the default Vector daemonset pipeline to a single writer per task log
  line.** On Airflow 3, KubernetesExecutor task pods run the containerised entrypoint
  with `subprocess_logs_to_stdout=True`, so every task line is emitted twice — once to
  `/usr/local/airflow/logs/**/attempt=N.log`, picked up by the `airflow_3_task_logs`
  file source, and once to container stdout, picked up by `airflow_k8s_logs`. Both
  pipelines feed the shared Elasticsearch sink with `action: create` and no
  deterministic `_id`, so every line is indexed twice.

- **Introduces `filter_k8s_task_logs`**, a filter transform situated between
  `transform_task_logs` and `transform_add_timestamp` on the stdout pipeline. It drops
  the stdout copy for KubernetesExecutor task output only, letting the file pipeline
  serve as the single authoritative source for task logs while system and component
  logs continue to flow through the stdout pipeline untouched:
  
 ```
  is_ke_task_pod = exists(.kubernetes.pod_labels.dag_id)
  is_task_output = exists(.dag_id) && exists(.task_id) && exists(.run_id)

  !(is_ke_task_pod && is_task_output)
  ```

### Behaviour

| stdout event | before | after |
|---|---|---|
| KE task pod, task line | indexed (duplicate) | dropped |
| KE task pod, supervisor line | indexed | indexed |
| Celery worker task line | indexed | indexed |
| scheduler / dag-processor line about a TI | indexed | indexed |

## Related
https://linear.app/astronomer/issue/PINF-955/af3-kubernetesexecutor-task-logs-are-ingested-twice

## Testing
- Unitests added
- Confirmed logs with the change: 
Screenshot 2026-09-09 at 6 15 24 PM

## Merge
Main, 2.1

### DIFF
--- a/charts/vector/templates/vector-configmap.yaml
+++ b/charts/vector/templates/vector-configmap.yaml
@@ -122,6 +122,8 @@ data:
           type: "vrl"
           source: |
             is_task_output = exists(.dag_id) && exists(.task_id) && exists(.run_id)
+            is_kubernetes_executor_pod = exists(.kubernetes.pod_labels.dag_id)
+
             !is_task_output || is_kubernetes_executor_pod
 
       transform_add_timestamp:
--- a/charts/vector/templates/vector-configmap.yaml
+++ b/charts/vector/templates/vector-configmap.yaml
@@ -116,14 +116,13 @@ data:
         type: filter
         inputs:
           - transform_task_logs
-        # Airflow 3 task processes echo their structured task log lines to stdout in
-        # addition to writing them to /usr/local/airflow/logs/**/attempt=N.log. The
-        # airflow_3_task_logs (file) source already ships those lines with full
-        # dag_id/run_id/task_id/attempt metadata parsed from the path, so shipping the
-        # stdout copy here as well double-ingests every task log line into
-        # Elasticsearch. Only pass through component/system logs, which have no
-        # file-based equivalent.
-        condition: '.log_type != "task"'
+        # AF3 tasks write each line to both stdout and attempt=N.log, so the file source
+        # and this one carry the same line to the sink. Drop the stdout copy.
+        condition:
+          type: "vrl"
+          source: |
+            is_task_output = exists(.dag_id) && exists(.task_id) && exists(.run_id)
+            !is_task_output || is_kubernetes_executor_pod
 
       transform_add_timestamp:
         type: remap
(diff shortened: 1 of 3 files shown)


rishkarajgi (commented):
Traced the full pipeline graph rather than reading the two transforms in isolation: `airflow_k8s_logs` (raw kubelet stdout) and `airflow_3_task_logs` (Airflow's own per-task files) converge at `transform_remove_fields` into one `elasticsearch` sink, so a task line appearing in both sources really does land twice today — confirmed `filter_k8s_task_logs` is correctly spliced in as the only consumer of `transform_task_logs`, nothing bypasses it.

The KubernetesExecutor exemption checks out against the actual chart, not just the PR's own claim: `airflow_3_task_logs`'s `include` already globs a bare `.../logs/**/*.log` path (no `-<release>` suffix) alongside `logs-*/**/*.log`, and `extract_release`'s regex only matches the dashed form — so a KubernetesExecutor pod's file copy is deterministically `release: unknown` and never queryable, which is exactly why the stdout copy has to survive for those pods while the file copy gets dropped by the paired `filter_task_logs_only` change. Also confirmed the drop condition is deliberately the tighter `dag_id && task_id && run_id`, not the existing looser `log_type == "task"` tag scheduler/dag-processor lines also carry — the new test pins that distinction.

Ran the suite against a reverted template to confirm the new tests actually discriminate: 20 fail without this change, 60/60 pass with it.

Worth knowing, not a concern: this is a genuinely separate duplication path from #3391/#3538 (today's houston-configmap pod_mutation_hook fix) — that one gates whether AF3 stdout gets teed into `out.log`/`err.log` for a per-pod logging sidecar, unrelated to either Vector source here. Both are legitimately needed; neither supersedes the other.

Since the PR body says this also targets `2.1`: checked, and `release-2.1`'s `vector-configmap.yaml` is identical to `master` in the touched region, so a cherry-pick should apply cleanly — unlike the surprises on the houston-configmap side today.

Nothing here blocking merge.

danielhoherd (approved):
Code changes look good.

@Shubham-astro please update the PR description to list changes modifying filter_task_logs_only and the KubernetesExecutor exemption. These are behavior changes worth noting.

Simpcyclassy (approved):
Agree with rishkarajgi's trace: filter_k8s_task_logs and the tightened filter_task_logs_only condition are a correctly matched pair, dropping the stdout duplicate while the is_kubernetes_executor_pod exemption keeps the only readable copy for ephemeral KubernetesExecutor task pods whose file copy indexes as release unknown. Daniel's ask to document the KubernetesExecutor exemption in the PR description is fair since it's a real behavior change, not just the headline dedup. Nothing in the template or new tests blocks merge.

PR #3543, opened 2026-09-10
APC-1767: astronomer chart MCP server component, ingress, and release wiring

Simpcyclassy:
## What this does
Adds a new component, the **MCP server**, to the Astronomer platform. It only runs on control-plane installs (not on data-plane-only installs), and is off by default until explicitly enabled.

This includes:
- The pieces needed to actually run it in Kubernetes (the workload itself, its own identity/permissions, and the network access rule that lets it receive traffic)
- A public web address so it can be reached from outside the cluster, protected by the same login/auth system as everything else
- Settings to turn it on/off and control which tool groups it exposes
- Test coverage for all of the above

As per conversation in the last apc demo call with @danielhoherd, I created the placeholder release to start from 1.0.0 instead of 0.1.0, because starting at 0 makes version numbers confusing, there's no clear rule for what counts as a small fix versus a big new feature. Starting at 1.0.0 keeps that clear from the beginning.

### Local K3d Testing
image

## What's intentionally not included
- Bumping the chart's version number — that happens separately at release time
- The MCP server's actual application code/image — that's built and published from its own repository

Related ticket: [APC-1767](https://linear.app/astronomer/issue/APC-1767/astronomer-chart-mcp-server-component-ingress-and-release-wiring)

2.2.0


charts/astronomer/templates/mcp-server/mcp-server-ingress.yaml
```diff
@@ -0,0 +1,84 @@
...
+silently producing that ingress.
+
+This "fail" only executes inside the already-true mcpServer.enabled branch above, and
+mcpServer.enabled defaults to false (see charts/astronomer/values.yaml) — so a BYO-ingress
+install that has not opted into the MCP server renders exactly as it did before this file
+existed. Flipping mcpServer.enabled on under authSidecar.enabled is the only way to hit this,
+and that combination has no supported answer yet (APC-1768), so failing loudly here is the
+correct behavior, not a bug to work around.
+*/}}
+{{- fail "mcpServer.enabled is not supported with global.authSidecar.enabled (BYO-ingress mode) in this release — see APC-1768. Disable mcpServer or the auth sidecar." }}
```
Simpcyclassy, line 24: This guard is correct in isolation, but it collides with this PR's own fixture changes: tests/enable_all_features.yaml and configs/enable-all-features-control.yaml both already set global.authSidecar.enabled: true, and this PR turns mcpServer.enabled on in the same two files. Every render_chart(get_all_features()) call (test_probes.py, test_pods_annotation.py, test_tags.py, test_byo_sa.py, and dozens more) will now hit this fail on a full-chart render instead of just the tests meant to exercise it.

tests/enable_all_features.yaml
```diff
@@ -14,6 +14,8 @@ astronomer:
         memory: 128Mi
   navigator:
     enabled: true
+  mcpServer:
+    enabled: true
```
Simpcyclassy, line 18: This file already sets global.authSidecar.enabled: true with a baseDomain and no plane override (defaults to unified), so mcpServer.enabled evaluates true here and trips the new fail() in mcp-server-ingress.yaml. get_all_features() backs ~40 other test files that render the full chart, so they'll all fail to render rather than just skip the mcp-server templates.

Simpcyclassy (commented):
Self review from Claude:
Enabling mcpServer.enabled: true in tests/enable_all_features.yaml and configs/enable-all-features-control.yaml trips the new BYO-ingress fail() guard, since both fixtures already set global.authSidecar.enabled: true. That aborts every render_chart(get_all_features()) call across roughly 40 existing test files, plus show-docker-images.py's control-plane run, not just the new mcp-server tests. Separately, mcp-server has no NetworkPolicy while houston/commander/registry/astro-ui all restrict their ingresses to nginx-only pod access, leaving the "no auth of its own" component reachable in-cluster with global.networkPolicy.enabled defaulting true; both need fixing before this leaves draft.

configs/enable-all-features-control.yaml
```diff
@@ -5,6 +5,8 @@ astronomer:
     enabled: true
   navigator:
     enabled: true
+  mcpServer:
+    enabled: true
```
Simpcyclassy, line 9: Same collision: this file already has global.authSidecar.enabled: true and plane.mode: control, so enabling mcpServer here trips the new BYO-ingress fail() and breaks show-docker-images.py's control-plane image discovery run.

charts/astronomer/templates/mcp-server/mcp-server-ingress.yaml
```diff
@@ -0,0 +1,84 @@
...
+##########################
+{{- if eq (include "mcpServer.enabled" .) "true" }}
+{{- if .Values.global.baseDomain }}
+{{- if .Values.global.authSidecar.enabled }}
+{{/*
+APC-1768: BYO-ingress (auth-sidecar) mode is explicitly out of scope for v1 of the MCP
+server. Every other ingress in this chart falls back to global.extraAnnotations only in
+this mode (see grafana-ingress.yaml), which is safe for them because there IS a platform
+nginx auth_request gate configured some other way, or the component tolerates being
+reachable ungated. Neither is true here: the MCP server holds no auth of its own by design
```
Simpcyclassy, line 12: Good design intent, but there's no NetworkPolicy backing this: houston, commander, registry, and astro-ui all restrict direct pod access to the ingress-controller only, mcp-server doesn't. With global.networkPolicy.enabled defaulting to true, any in-cluster pod can hit the Service directly and skip this auth-url gate entirely; worth adding a mirroring NetworkPolicy (see registry-networkpolicy.yaml) before merge.

charts/astronomer/templates/mcp-server/mcp-server-ingress.yaml
danielhoherd: Mentioning APC-1768 is not going to be helpful to customers.

tests/chart_tests/test_mcp_server.py
```diff
@@ -0,0 +1,302 @@
+"""Tests for the MCP server component (APC-1767): Deployment, Service, NetworkPolicy,
```
danielhoherd: If we are going to mention the linear ticket, we should mention the parent ticket, which is APC-1710.

configs/pull-main-master-tags.yaml
```diff
@@ -30,6 +30,13 @@ astronomer:
       pullPolicy: IfNotPresent
       repository: quay.io/astronomer/ap-houston-api
       tag: main
+    # APC-1767: manual entry -- this file is hand-maintained, not generated, so a new
+    # component's floating dev tag needs adding here explicitly. apc-mcp-server's default
+    # branch is main (confirmed via that repo's git remotes), matching astroUI/houston above.
+    mcpServer:
+      pullPolicy: IfNotPresent
+      repository: quay.io/astronomer/ap-mcp-server
```
danielhoherd, line 35: Does this exist? I'm unable to access it.

Simpcyclassy: Not yet. We have not started releasing

tests/chart_tests/test_mcp_server.py
Simpcyclassy: Dropping it from the comments

danielhoherd:
`mcpServer` is excluded from `tests/enable_all_features.yaml` (comment: "Covered in test_mcp_server.py"), but that's not quite true for two repo-wide guards that only run against that fixture: `test_probes.py::TestStartupProbes` (the Gatekeeper `allow-with-probes` 100%-coverage check) and `test_container_resources.py` (every container needs `resources`). Since mcp-server's Deployment never renders in `get_all_features()`, neither guard ever sees its container, so nothing currently protects the default `resources`/`startupProbe` from a future regression.

The values/template defaults look correct today, so this isn't blocking, but can you add explicit assertions to `test_deployment_default_values` in `test_mcp_server.py` — `resources` present and `startupProbe` non-empty — so the exclusion's claim actually holds?

charts/astronomer/templates/mcp-server/mcp-server-deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+            # public-hostname branch houston.internalauthurl carries for the
+            # data-plane case.
+            - name: HOUSTON_API_URL
+              value: "http://{{ .Release.Name }}-houston.{{ .Release.Namespace }}.svc.cluster.local:8871/v1"
+
+            # Chart toggle for the tool groups the catalog exposes (internal/config,
+            # internal/toolreg.Group in apc-mcp-server) — a CSV of group names.
+            - name: MCP_ENABLED_GROUPS
+              value: {{ include "mcpServer.enabledGroups" . | quote }}
+
```
adhipjos, line 105: need to add MCP_DISABLED_TOOLS to the env variables as well. This is to allow customers a way to disable individual tools

charts/astronomer/values.yaml
```diff
@@ -999,6 +1004,70 @@ navigator:
...
+
+  # -- Tool groups to enable, beyond the always-on core group (internal/toolreg.Group in
+  # apc-mcp-server). Joined into the MCP_ENABLED_GROUPS env var — see mcpServer.enabledGroups
+  # helper. Disabling a group here removes its tools from the catalog entirely, not just from
+  # what's visible to a given caller.
+  # @default -- ["platform_read", "platform_write"]
+  enabledGroups:
+    - platform_read
+    - platform_write
+
```
adhipjos, line 1039: add disabledTools as a value configurable by the user to disable individual tools

See config struct updates here: https://github.com/astronomer/apc-mcp-server/blob/71a6565129d26828068dd03ad54ffac423f97069/internal/config/config.go#L19

Consumed here: https://github.com/astronomer/apc-mcp-server/blob/71a6565129d26828068dd03ad54ffac423f97069/internal/config/config.go#L68

danielhoherd (approved):
The latest changes fix all the concerns I had, so this looks good to merge.

adhipjos:
@Simpcyclassy can you also update the houston side network policy for the api to allow apc-mcp-server using podselector labels. See here: https://github.com/astronomer/astronomer/blob/72388c2cb75987ee65c2a70fb98777ac2d6a7b5e/charts/astronomer/templates/houston/api/houston-networkpolicy.yaml#L70

Good idea to probably gate it as well on mcpServer.enabled

charts/astronomer/templates/mcp-server/mcp-server-serviceaccount.yaml
```diff
@@ -0,0 +1,19 @@
+###############################
+## MCP Server ServiceAccount ##
+###############################
+{{- if .Values.mcpServer.serviceAccount.create }}
+{{- if eq (include "mcpServer.enabled" .) "true" }}
```
adhipjos, line 5: no gating on global.rbac.enabled means service account rendered as 'default'

charts/astronomer/templates/mcp-server/mcp-server-ingress.yaml
```diff
@@ -0,0 +1,72 @@
+##########################
+## MCP Server Ingress   ##
+##########################
+{{- if eq (include "mcpServer.enabled" .) "true" }}
```
adhipjos, line 4: should be gated on global.ingress.enabled so that ingress doesn't render when ingress is disabled globally

adhipjos (commented):
Creating a new Linear issue for the changes

PR #3546, opened 2026-09-15
PLA-416: Thread split_software_automation_tests through the stage-clu…

sreenusuuda:
## Description

Cherry-picks #3524 (6477e5e2a) from main onto release-2.1.

Adds a split_software_automation_tests parameter that flows from the CircleCI trigger-stage-cluster-build job down to bin/trigger_stage_cluster_update.py, so the stage-cluster trigger can run p0 and p1/p2 software automation suites as separate sequential jobs instead of one combined job.

Changes:

bin/trigger_stage_cluster_update.py: added --split_software_automation_tests CLI flag (default false), passed through to the pipeline trigger payload; also added help text to the existing args.
.circleci/config.yml.j2 / .circleci/config.yml: added split_software_automation_tests job parameter, conditionally passes the new flag to the trigger script, and enables it for the qa artifact subdir workflow step.


karankhanchandani (approved):
LGTM

PR #3547, opened 2026-09-15
fix laminar env vars prefix

pgvishnuram:
## Description

This PR adds below changes for laminar enablement
- fixes the env vars injection variables
- asserts default vars tests
- adds commander laminar metadata config
-  supporting test cases 

## Related Issues

https://linear.app/astronomer/issue/APC-1813/fix-laminar-env-vars-prefix

## Testing

NA

## Merging

merge to master


sudarshanrampuria (approved):
LGTM. Not a blocker but to protect us from wrong future changes, we can add a test which will render the .env using `python-dotenv`, same what laminar does.

charts/laminar/templates/configmap.yaml
sudarshanrampuria: Should this have releaseName as prefix like all the other configmaps?

PR #3550, opened 2026-09-16
Pinf 1218 add test for sidecar custom config

sudarshanrampuria:
## Description

Adds a new functional scenario, sidecar-logging-custom-config, that verifies the Houston-injected log-shipping sidecar (global.logging.loggingSidecar) on real Airflow deployment pods when customConfig: true

The test does the following things: 

* Creates a real Airflow Deployment through Houston (as dag_deploy, then switched to git_sync) with global.logging.loggingSidecar.enabled + customConfig: true, and asserts:
* Sidecar injected on the right components:  the sidecar-log-consumer container lands on exactly the expected set of log-producing components.
* customConfig sources from a Secret, not a ConfigMap: every sidecar pod mounts its config from the fixed-name sidecar-config Secret via config-volume, and the default per-release <release>-sidecar-config ConfigMap is NOT created.
* DagDeployment-specific consumers — dag-server (dag_deploy) and git-sync-relay (git_sync) each carry the sidecar and mount the Secret.

## Related Issues

https://linear.app/astronomer/issue/PINF-1218/fix-dag-server-custom-sidecar-config-secret-mismatch#comment-83119ef7

## Testing

NA

## Merging

main


configs/enable-logging-sidecar-custom-config.yaml
```diff
@@ -0,0 +1,19 @@
+global:
+  logging:
+    loggingSidecar:
+      enabled: true
+
+      name: sidecar-log-consumer
+
+      # customConfig: true makes Houston mount the injected sidecar's config from a
+      # Secret named "sidecar-config" (fixed name) in each deployment's namespace,
```
danielhoherd:
```suggestion
      # Secret named "sidecar-config" in each deployment's namespace,
```

danielhoherd (approved):
Code looks good. I left one nit comment suggestion that should not block merging.

We should fix the merge conflict then squash merge this.

PR #3551, opened 2026-09-16
APC-1814: astronomer chart MCP server post-merge fixes

Simpcyclassy:
## What this does

Fixes a handful of bugs found shortly after the MCP server component was first added to the Astronomer Helm chart (in an earlier, already-merged PR). None of these are new features — they're cleanup so the component behaves correctly across different customer configurations.

Fixed:
- **A missing network rule** meant the MCP server's calls to Houston (the backend API) could get silently blocked when network policies are enabled.
- **A service account setting** wasn't respecting the customer's RBAC configuration, so it could render even when RBAC was turned off.
- **An ingress setting** wasn't respecting the global "disable all ingress" toggle.
- **A safety check** that's supposed to block an unsupported configuration (MCP server + "bring your own ingress" mode) only worked if the customer had also set a base domain — meaning it silently didn't fire for the exact case it was meant to catch.
- **Two hardcoded network ports** were fixed to actually use the customer's configured port instead of always assuming the default.
- **Added a chart setting** to disable individual tools by name (a more surgical version of the existing "disable a whole group of tools" setting), for when an operator needs to turn off just one tool during an incident.
- **Cleaned up the documentation comment** for the tool-group setting so it's clear it's a replace-the-list setting, not an add-to-the-list one.

Also looked into whether this component should be added to the chart's big "enable every feature" test fixture (something the original ticket assumed would be a simple follow-up). It turns out this component is permanently incompatible with the "bring your own ingress" feature by design, and that fixture enables both — so adding it there would break 15 unrelated tests for no real benefit. Left it out and added targeted test coverage instead.

## Related Issues
[APC-1814](https://linear.app/astronomer/issue/APC-1814/astronomer-chart-mcp-server-post-merge-fixes-houston-networkpolicy)

## How this was verified

- Manually rendered the Helm chart under each affected configuration to confirm each fix actually works and nothing else broke.
- Ran the full relevant chart test suite (301 tests) — all passing.


adhipjos (approved):
This looks good. It was just missing tests for the changes which I added. Good to merge

PR #3554, opened 2026-09-17
Fix executable detection: use shutil.which instead of bash -lc

pgvishnuram:
## Description

while running k3d with zsh with the current script it error out the terminal and resets the python environment.

_which now calls shutil.which(exe), which resolves against the PATH the process already inherited from the launching shell — correct for bash, zsh, CI, or IDE alike, with no login-shell sourcing.

## Related Issues

NA

## Testing

NA

## Merging

merge to release-1.x, 2.x, master


danielhoherd (approved):
Nice improvement. This will be much more reliable.

PR #3561, opened 2026-09-21
PINF-1260 fix missing nginx-metrics in prometheus dashboard

pgvishnuram:
## Description

Nginx metrics flag seems to be disabled while transition from old releases, this PR fixes the behaviour and adds the flag back to the default nginx args.Additionally this fixes the naming convention of ingress controller which changed from 1.x 

## Related Issues

https://linear.app/astronomer/issue/PINF-1260/fix-missing-nginx-metrics-in-prometheus-dashboard

## Testing

- base helm unit tests

## Merging

merge to master and all releases


charts/prometheus/templates/prometheus-config-configmap.yaml
```diff
@@ -136,7 +136,7 @@ data:
         relabel_configs:
           - source_labels: [__meta_kubernetes_service_name]
             action: keep
-            regex: "^{{ .Release.Name }}-nginx"
+            regex: "^{{ .Release.Name }}-{{ ternary "dp-nginx" "cp-nginx" (eq .Values.global.plane.mode "data")}}"
```
danielhoherd: since this is a regex, can we do `[cd]p-nginx` ?

danielhoherd (approved):
These changes all look good. I left one optional change for using a single regex vs a helm template to change the regex depending on cp or dp.

charts/prometheus/templates/prometheus-config-configmap.yaml
pgvishnuram: Yeah simplified the changes with the above ones

